PR #10, opened 2022-09-12
Repository annotation with JavaDoc outline

njr-11:
Creates the Repository annotation for #3 and experiments with a possible outline for the JavaDoc where we can put information about repository method naming patterns, keywords, return types and so forth as those aspects get decided later on.  The information that is initially there will not be correct or complete and hasn't even been discussed yet.

### DIFF
--- a/api/src/main/java/jakarta/data/Repository.java
+++ b/api/src/main/java/jakarta/data/Repository.java
@@ -1,0 +1,329 @@
+/*
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License, Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Designates a data repository interface that will be implemented by the container/runtime.<p>
+ *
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/data/Repository.java
```diff
@@ -0,0 +1,329 @@
...
+ * are discussed and defined.<p>
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface Repository {
+    /**
+     * Entity class. By default, detect automatically.
+     */
+    Class<?> value() default void.class;
```
otaviojava, line 328: :question: do we need this value's field?

njr-11: > ❓ do we need this value's field?

There needs to be some way of identifying the type of entity for a repository for cases where the only references to the entity type are parameterized.  The following, for example, won't work for solutions that process the annotations at run time because it has no way of knowing that it needs to look for the location in the database that stores Car entities,

```
@Repository
public interface CarRepository extends DataRepository<Car, Long>{
    List<Car> findByMakeAndModel(String brand, String vehicleModel);
}
```

One way to allow for that is to allow the entity type to be optionally specified here,

```
@Repository(Car.class)
public interface CarRepository extends ...
```

But there are plenty of other options that could be chosen instead, such as having a separate annotation for it, or requiring the user to compose at least one method that accepts or returns the entity type unparameterized, and so forth.  Do you have a preference?

otaviojava: We can do it with the annotation process with this way:

```java
    static List<String> findParameters(TypeMirror repository) {
        if (repository instanceof DeclaredType) {
            DeclaredType declaredType = (DeclaredType) repository;
            return declaredType.getTypeArguments().stream()
                    .map(TypeMirror::toString)
                    .collect(Collectors.toList());
        }
        return Collections.emptyList();
    }
```
Where the first parameter is the first element on the generic.
Anyway, we'll go further on it in the next discussion/issue about the ```CrudRepository```

njr-11: I don't see a way to use TypeMirror at run time, however, I looked into this further and it appears that ParameterizedType.getActualTypeArguments would be able to access the type parameters at run time, so I've updated the pull to remove the entity class from the annotation and its documentation.

otaviojava:
I'll cherry-pic this one.

otaviojava:
@njr-11 once I moved those commits to mine, I'll close this one.

PR #16, opened 2022-09-18
Add crud repository feature

otaviojava:
As we discussed last September 13, we started discussing repositories on the CrudRepository methods.

We used those methods to start the discussion:

Furthermore, this PR also provides the skeleton of the documentation to the repository type.

-  [CrudRepository](https://github.com/spring-projects/spring-data-commons/blob/main/src/main/java/org/springframework/data/repository/CrudRepository.java) by Spring
- [CrudRepository](https://github.com/micronaut-projects/micronaut-data/blob/master/data-model/src/main/java/io/micronaut/data/repository/CrudRepository.java) by Micronaut
- [Repository](https://github.com/eclipse-ee4j/nosql/blob/main/api/mapping/mapping-core/src/main/java/jakarta/nosql/mapping/Repository.java) by Jakarta NoSQL
- [EntityRepository](https://github.com/apache/deltaspike/blob/master/deltaspike/modules/data/api/src/main/java/org/apache/deltaspike/data/api/EntityRepository.java) by Delta Spike
- [PanacheRepository](https://github.com/quarkusio/quarkus/blob/main/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheRepository.java) by Quarkus

This PR includes:

* https://github.com/jakartaee/data/issues/14

:disappointed:  Sorry, I did not find a way to break this PR into smaller ones. 
:warning:  When it makes sense, please, explore the [suggestion mode on GitHub](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request).

Related: https://github.com/jakartaee/data/issues/8

https://github.com/jakartaee/data/issues/1

PR #20, opened 2022-09-21
OrderBy annotation for sorting

### LINKED ISSUE #18: Could OrderBy be better specified annotatively?
**Description**

As a:

...I need to be able to:

have a less lengthy and less error-prone way of defining custom repository methods

...which enables me to:

better understand the code that uses my repository methods and
reduce the opportunity for misspellings and mistakes in my repository methods.

Instead of writing queries such as the following that make the code almost unreadable,

```
List<Employee> findByDepartmentAndIdNotOrderByPositionDescLastNameAscFirstNameAsc(String deptNum, int empNum);

...
coworkers = employees.findByDepartmentAndIdNotOrderByPositionDescLastNameAscFirstNameAsc(deptNum, empNum);
```

suppose we allowed the sorting information to all be declared annotatively,

```
@OrderBy(value = "position", descending = true)
@OrderBy("lastName")
@OrderBy("firstName")
List<Employee> findByDepartmentAndIdNot(String deptNum, int empNum);

...
coworkers = employees.findByDepartmentAndIdNot(deptNum, empNum);
```

The only disadvantage I can think of is if someone wanted to duplicate the exact same query but with different ordering, but it's likely that special case would already be solved by the Sorts parameter that some of the implementations have.

njr-11:
fixes  #18

### DIFF
--- a/api/src/main/java/jakarta/data/repository/OrderBy.java
+++ b/api/src/main/java/jakarta/data/repository/OrderBy.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.repository;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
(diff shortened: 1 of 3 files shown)


otaviojava:
@njr-11 please, don't forget to add the documentation and this option on the [spec](https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/repository.asciidoc) as well.

otaviojava (dismissed):

otaviojava:
Anyway, IMHO: it is free to go.

njr-11:
> @njr-11 please, don't forget to add the documentation and this option on the [spec](https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/repository.asciidoc) as well.

@otaviojava thanks - with #16 now merged, I rebased onto that and added a [commit](https://github.com/jakartaee/data/pull/20/commits/fb8bf5a8b7775ec1944b659494dbe7ee3509ed24) that updates the repository spec document.  It also includes a couple of minor fixes that I spotted (making the indentation of the code examples consistent and a one-word correction) while making the change.

otaviojava:
@njr-11 about the documentation.
IMHO
We should not remove the method by the query that we support.
I would like to add this annotation as an option.
It is mainly on all vendors such as Spring, Micronaut, Delta Spike, JNoSQL. 
It is already a standard.

otaviojava (approved):

PR #33, opened 2022-10-03
Replacing assertion libraries: Hamcrest by AssertJ

eliasnogueira:
### Motivation

This PR replace Hamcrest assertion library by AssertJ to add more fluent asserts.

### Changed

- jakarta-data-api
    - Removed `hamcrest-all` in favour of `assertj-core`
    - Changed the assertions in `PageableTest` and `SortTest`
    - Removed redundant tests on `SortTest`
- jakarta-data-parent
    - Updated the following libraries
        - `mockito`.version` to `4.8.0`
        - `junit` to `5.9.0`
    - Fix typo on `mockito.version`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,20 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0.1] - 03-10-2022
+
+### Changed
+
+- jakarta-data-api
+    - Removed `hamcrest-all` in favour of `assertj-core`
+    - Changed the assertions in `PageableTest` and `SortTest`
+    - Removed redundant tests on `SortTest`
+- jakarta-data-parent
+    - Updated the following libraries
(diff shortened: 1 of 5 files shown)


otaviojava:
Thank you @eliasnogueira it sounds good to go.

@njr-11 What do you think?
could you review it?

njr-11:
> @njr-11 What do you think?
> could you review it?

This looks like it is primarily a change to how the unit tests are styled, which can be a matter of preference between developers.  I don't have a preference here and would lean toward whatever is the standard or convention for Jakarta specs, if there even is one.  I really have no idea.  The changes themselves look valid.  The intent of the tests has been preserved and I don't see any coverage lost or added.  It certainly makes sense at this point in the development cycle at the very beginning of the project to decide on what the standard will be for tests because now is when it's easiest to convert what is there.

otaviojava (approved):

njr-11 (approved):
Earlier this week I posted to the Jakarta Data mailing list to check if anyone was opposed to switching the style of test assertions and there were no objections, so I think we can merge this.

eliasnogueira:
@otaviojava @njr-11 I pulled the latest changes from `main` adding the new test (page size) and additional checks from the exception check test.

Could you please review it?

njr-11 (approved):

PR #34, opened 2022-10-04
Add test on Limit class

otaviojava:


njr-11 (approved):
Thanks for adding the unit test and spotting the missing `static`!

Issue #36, opened 2022-10-04
Return type of Page (and Slice) for repository methods


njr-11:
**Description**

As a:

- [ ] Application user/user of the configuration itself
- [x] API user (application developer)
- [ ] SPI user (container or runtime developer)
- [ ] Specification implementer

...I need to be able to:
define repository methods that return a Page of results

...which enables me to:
more efficiently work with paginated data.

Several of the existing implementations ([Spring Data](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Page.html), [Micronaut](https://micronaut-projects.github.io/micronaut-data/latest/api/io/micronaut/data/model/Page.html), [Jakarta NoSQL](http://www.jnosql.org/javadoc/jakarta/nosql/mapping/Page.html)) have a `Page<T>` interface for working with pages of data.  In the case of the first two, a distinction is made between a `Slice` and a `Page`, with the latter adding in a total count of available results and pages, which avoids collecting the expensive count information when unnecessary.  As pointed out in the comment from @odrotbohm, determination of the count cannot be deferred until the `getTotal*` methods are invoked because it cannot be assumed that resources continue to be available beyond the return of the repository method. 

In any case, I'm opening this issue to consider if Jakarta Data should also have `Page` and `Slice` interfaces to be used as a result type for repository methods, and if so, to consider what it will include from the existing implementations or otherwise.

Here is some initial analysis:

In Spring and Micronaut, a `Page` is [Iterable](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Iterable.html), which seems nice to be able to write code like `for (Product p : page) ...` without needing to convert the page to another type.

It looks like Spring and Jakarta NoSQL have a way to directly obtain a [Stream](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Stream.html) from a `Page`, although otherwise a stream could be obtained indirectly.

Spring and Micronaut have the ability to obtain the next and previous Pageable (page request information).
All 3 have a way to obtain the Pagination/Pageable information for the current page, which can be used to compute the next and previous page information.
Jakarta NoSQL has a method to directly obtain the next page, but this would require resources to continue to be available beyond the end of the repository method, which as stated earlier is not something that is possible for at least one of the existing implementations.

Spring and Micronaut have a `getNumberOfElements` method which seems useful for identifying how many results can be obtained from the page and for quickly identifying when a page is empty and so there are no more.
[...]

PR #37, opened 2022-10-05
Return type of Page for repository methods

otaviojava:
:bomb: Please don't merge it.

This PR aims to start the discussion around the ```Page``` and the ```PageableRepository``` new method.

https://github.com/jakartaee/data/issues/36


otaviojava:
I've update using this comments as base:
https://github.com/jakartaee/data/issues/36#issuecomment-1268074831

otaviojava:
@njr-11 @keilw @odrotbohm I created this PR based on the issue  https://github.com/jakartaee/data/issues/36#issuecomment-1268074831

Please, check if it makes sense based on those comments.

odrotbohm:
I guess whether `Streamable` should be part of this could be a separate discussion?

otaviojava:
Sure, I'll remove it.

otaviojava:
@odrotbohm removed.

api/src/main/java/jakarta/data/repository/Page.java
```diff
@@ -0,0 +1,39 @@
...
+public interface Page<T> extends Slice<T> {
+
+    /**
+     * Returns the total amount of elements.
+     * @return the total amount of elements
+     */
+    long getTotalElements();
+
+    /**
+     * Returns the number of total pages.
```
njr-11, line 35:
```suggestion
     * Returns the total number of pages.
```

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -129,14 +145,59 @@ public static Pageable page(long page) {
...
+    /**
+     * Creates a new Pageable at the given size, page and Sort
+     *
+     * @param page The page
+     * @param size The size
+     * @param sort the sort
+     * @return The pageable
+     * @throws IllegalArgumentException when page or size are negative
+     */
+    public static Pageable of(long page, long size, Sort sort) {
```
njr-11, line 163: Let's not tightly couple dynamic Sort with pagination.  It is best left decoupled as it was prior to these changes, where it is possible for the user to do both of:

```
page = products.findByNameLikeOrderByPriceDesc(pattern, pageable);
page = products.findByNameLike(pattern, pageable, Sort.desc("price"));
```

api/src/main/java/jakarta/data/repository/PageableRepository.java
```diff
@@ -28,6 +28,13 @@
  */
 public interface PageableRepository<T, K> extends CrudRepository<T, K> {
 
-
+    /**
+     * Returns a {@link Page} of entities meeting the paging restriction provided in the {@link Pageable} object.
+     *
+     * @param pageable the pageable to request a paged result , must not be null.
```
njr-11, line 34:
```suggestion
     * @param pageable the pageable to request a paginated result, must not be null.
```

njr-11 (changes requested):

api/src/main/java/jakarta/data/repository/Pageable.java
odrotbohm: I don't think that makes sense. Of course, `Sort` should be usable standalone, but as soon as you use a `Pageable`, it strictly *needs* a `Sort`, as otherwise the pagination order is not stable. Being able to provide a `Sort` instance in addition to a `Pageable` this only creates confusion:

- Callers wonder about two different ways to hand in the `Sort`? Via the `Pageable`? As additional parameter? Both?
- Implementors now have to decide how to interpret the additional `Sort`? Does that one need to get integrated into the `Pageable` or has the caller already done that?

That's why we decided that methods taking a `Pageable` would never take an additional `Sort`. The caller needs to prepare the `Pageable` instance completely, including sorting directives. The called code can work solely with the `Sort` instance contained in the `Pageable`.

odrotbohm:
> Sure, I'll remove it.

Thanks, Otavio! Just to be clear: I still think `Streamable` would be a useful addition. It's just that the type has been a constant topic of discussion for Spring Data, as some people don't like that it's defaulting the streaming on top of `Iterable`. On the other hand, the Spring Data flavor of it has really provided additional value due to all the additional methods on top that have been omitted here. In other words, I think the tension between "Should this be added at all?" and "Oh, let's provide plenty of cool helper methods!" warrants a dedicated discussion, as I can imagine that a standardized API comes to a different tradeoff than a standalone Spring project.

api/src/main/java/jakarta/data/repository/Pageable.java
njr-11: > * Callers wonder about two different ways to hand in the `Sort`? Via the `Pageable`? As additional parameter? Both?

There is nothing for callers to get confused about if it isn't possible to specify `Sort` via the `Pageable`.  Then the API itself doesn't allow you to attempt an erroneous combination of `Sort` via `Pageable` and `Sort` as method parameters.  This seems like a good reason for leaving it as is and not adding `Sort` to `Pageable`.

> * Implementors now have to decide how to interpret the additional `Sort`? Does that one need to get integrated into the `Pageable` or has the caller already done that?

`Pageable` is provided by the Jakarta Data API, not the implementation.  If the API doesn't allow associating a `Sort` with a `Pageable` then there isn't any decision or confusion for implementations on the matter.  So this also seems like a reason for leaving `Pageable` as is and not adding a `Sort` to it.

> Being able to provide a `Sort` instance in addition to a `Pageable` this only creates confusion

There is only confusion if `Pageable` also has its own `Sort`, which is what I'm aiming to avoid.

odrotbohm: A `Pageable` without a `Sort` doesn't make any sense for the reason described above. Pagination requires stable sorting.

otaviojava: @njr-11 about ```Pageable```without ```Sort```is like selling a car without an engine.


+1 for keep ```Pageable``` **with** ```Sort```

api/src/main/java/jakarta/data/repository/Slice.java
```diff
@@ -0,0 +1,61 @@
...
+     * @return the current Pageable
+     */
+    Pageable getPageable();
+
+    /**
+     * Returns the next {@link Pageable#next()}
+     *
+     * @return the next pageable
+     */
+    Pageable next();
```
otaviojava, line 60:
```suggestion
  Pageable nextPageable();
```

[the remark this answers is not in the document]
otaviojava: You're right fixed it.

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -129,14 +145,59 @@ public static Pageable page(long page) {
      * @param page The page
      * @param size The size
      * @return The pageable
+     * @throws IllegalArgumentException when page or size are negative
```
njr-11, line 148:
```suggestion
     * @throws IllegalArgumentException when page or size are negative or zero
```

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -129,14 +145,59 @@ public static Pageable page(long page) {
...
+    }
+
+    /**
+     * Creates a new Pageable at the given size, page and Sort
+     *
+     * @param page The page
+     * @param size The size
+     * @param sort the sort
+     * @return The pageable
+     * @throws IllegalArgumentException when page or size are negative
```
njr-11, line 161:
```suggestion
     * @throws IllegalArgumentException when page or size are negative or zero
```

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -129,14 +145,59 @@ public static Pageable page(long page) {
...
+    }
+
+    /**
+     * Creates a new Pageable at the given size, page and Sort
+     *
+     * @param page  The page
+     * @param size  The size
+     * @param sorts the sorts
+     * @return The pageable
+     * @throws IllegalArgumentException when page or size are negative
```
njr-11, line 175:
```suggestion
     * @throws IllegalArgumentException when page or size are negative or zero
```

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -129,14 +145,59 @@ public static Pageable page(long page) {
...
```
njr-11, line 188:
```suggestion
     * @throws IllegalArgumentException when page or size are negative or zero
```

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -79,13 +86,22 @@ public long getPage() {
         return page;
     }
 
+    /**
+     * Return the order collection
+     *
+     * @return the order collection
+     */
+    public List<Sort> getSorts() {
+        return Collections.unmodifiableList(sorts);
```
njr-11, line 95:
```suggestion
        return sorts;
```

In all cases, the list is already unmodifiable, so it doesn't make sense to wrap it again in another unmodifiable list.  Instead, just directly return the value here.

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -79,13 +86,22 @@ public long getPage() {
         return page;
     }
 
+    /**
+     * Return the order collection
```
njr-11, line 90:
```suggestion
     * Return the order collection if it was specified on this <code>Pageable</code>, otherwise an empty list.
```

api/src/main/java/jakarta/data/repository/Slice.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Returns the current {@link Pageable}
+     *
+     * @return the current Pageable
+     */
+    Pageable getPageable();
+
+    /**
+     * Returns the next {@link Pageable#next()}
```
njr-11, line 56:
```suggestion
     * Returns the next {@link Pageable#next()}, or <code>null</code> if it is known that there is no next page.
```

api/src/main/java/jakarta/data/repository/Pageable.java
njr-11: > @njr-11 about `Pageable`without `Sort`is like selling a car without an engine.

A better analogy would be selling a car without a built-in navigation system or map and someone to drive it. Many drivers can supply their own GPS or have other ways of knowing how to get around. The concept of a car still makes sense without these things sold and provided together.
The same is true of `Pageable`. It's the information for creating a page from a large sorted dataset. You can apply the page sizes and indices to different ways of sorting, with some provided by a `Sort` and others provided by a repository method's `OrderBy...`. There are different ways to tell it how to navigate which can be provided separately.

It's true that the pagination information isn't useful without sorting/ordering, but pagination information also isn't useful without a dataset, and everyone seems to agree on not coupling the dataset to `Pageable`. It should be a reasonable possibility for design that `Sort` not be coupled to `Pageable` either.

We should at least acknowledge that both proposals do make sense and this is a matter of preference for one style of design over another. This sort of thing ought to be a vote, but observing that existing implementations have already chosen the Sort-attached-to-Pageable approach, I can see that my preference for decoupling the two won't come out ahead, so I'll go ahead and review the portions of the pull that cover adding Sort to Pageable.

njr-11 (changes requested):

api/src/main/java/jakarta/data/repository/Pageable.java
otaviojava: It is a good point @njr-11

api/src/main/java/jakarta/data/repository/PageableRepository.java
```diff
@@ -28,6 +28,13 @@
...
+     * @return a page of entities
+     * @throws NullPointerException when pageable is null
+     */
+    Page<T> findAll(Pageable pageable);
```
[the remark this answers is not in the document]
njr-11: This comment only applied if decoupling `Sort` from `Pageable`, otherwise it is unnecessary, so I'll mark it resolved.

njr-11 (approved):
The only remaining comment is a minor typo that can be easily addressed, so I'll mark this approved now.

otaviojava:
@odrotbohm @njr-11 the ```Pageable```became a hot topic. So we'll create an issue only to talk about it, ok?

api/src/main/java/jakarta/data/repository/Pageable.java
graemerocher: I am in agreement with @odrotbohm as stated paging does require stable sorting so I don't see value in decoupling

PR #48, opened 2022-10-11
Consistently use AssertJ instead of JUnit 5 Assertions

timtebeek:
Applied through
```shell
mvn org.openrewrite.maven:rewrite-maven-plugin:4.35.1:init \
  -Ddependencies=org.openrewrite.recipe:rewrite-testing-frameworks:1.29.0 \
  -DactiveRecipes=org.openrewrite.java.testing.assertj.Assertj
mvn rewrite:run
```
@eliasnogueira as we discussed this afternoon. :) Continues on from #33 

Figured it's better to be consistent about which assertion library to use rather than use a mix of JUnit 5 & AssertJ.

### DIFF
--- a/api/src/test/java/jakarta/data/repository/LimitTest.java
+++ b/api/src/test/java/jakarta/data/repository/LimitTest.java
@@ -17,48 +17,51 @@
  */
 package jakarta.data.repository;
 
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+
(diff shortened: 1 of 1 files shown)


timtebeek:
I've signed all I could think of; even force pushed with a different email address in hopes of resolving this. If that's blocking my contribution feel free to recreate using the command shown above. :)

pzygielo:
revalidated ECA

njr-11 (approved):

Issue #39, opened 2022-10-05
Keyset pagination


njr-11:
@beikov originally added this comment under #35, but it's worth an issue of its own.  While the provider will implement `Page` and the `nextPageable` and `previousPageable` that are obtained from it, I'd prefer that we look into whether keyset pagination can be specified in a more standard way if possible rather than relying on particular vendors to possibly provide it.

> This is commonly referred to as "keyset pagination" which I implemented in Blaze-Persistence on top of JPA. Being able to extract `previousPageable` and `nextPageable` from a `Page` or `Slice` (like Spring Data allows it) would allow to nicely model this i.e.

```java
    page = employees.findAllByStatus(
                    "FULL TIME",
                    page == null ? PageRequest.of(100) : page.nextPageable()
    );
    ...
```

> With custom `Page` and `Pageable` implementations, the details can be left to the underlying implementation.

> You can find some details about this here: https://persistence.blazebit.com/documentation/1.6/core/manual/en_US/#anchor-keyset-pagination

_Originally posted by @beikov in https://github.com/jakartaee/data/issues/35#issuecomment-1268066903_

beikov:
I guess it might be worth adding something like a `KeysetAwarePage` and `KeysetPageable` to the API which provide additional details about the keyset parts. Another concept that could be interesting is a `KeysetCursor` which I called simply `Keyset` in Blaze-Persistence. A `KeysetCursor` would encapsulate values of an object, by which objects are sorted.

For example, when we query for `Cat` and `order by name, id desc`, we might get:

|name|id|
|------|--|
|A|2|
|A|1|
|B|3|

So the `KeysetCursor` of every object is the tuple `(name,id)` because that is what is ordered by.

A `KeysetAwarePage` contains at least 2 such cursors for the first and the last element, which serve as the basis for the previous and next `KeysetPageable` through `previousPageable` and `nextPageable`. For GraphQL Relay pagination it is necessary to attach a cursor to every object, so it might be worth adding optional support for accessing a list of cursors.

otaviojava:
I like the idea, but I am unsure if it will work on other data sources.
Maybe it specializes in JPA.

njr-11:
[...]

PR #52, opened 2022-10-20
keyset pagination API and JavaDoc

njr-11:
Keyset pagination API and JavaDoc for #39

### DIFF
--- a/api/src/main/java/jakarta/data/repository/KeysetPageable.java
+++ b/api/src/main/java/jakarta/data/repository/KeysetPageable.java
@@ -17,6 +17,8 @@
  */
 package jakarta.data.repository;
 
+import java.util.Objects;
+
 /**
  * <p>Keyset pagination is a form of pagination that aims to reduce the
  * possibility of missed or duplicate results by making the request for
@@ -158,6 +160,19 @@ public class KeysetPageable extends Pageable {
         public int size() {
             return keyset.length;
         }
+
+        /**
+         * String representation of the keyset cursor, including the number of
+         * key values in the cursor but not the values themselves.
+         *
+         * @return String representation of the keyset cursor.
+         */
+        @Override
+        public String toString() {
+            return new StringBuilder("Cursor@").append(Integer.toHexString(hashCode())) //
+                            .append(" with ").append(keyset.length).append(" keys") //
+                            .toString();
+        }
     }
 
     private final Cursor cursor;
@@ -171,6 +186,25 @@ public class KeysetPageable extends Pageable {
 
         this.cursor = new Cursor(keyset);
         this.mode = mode;
+    }
+
+    /**
+     * Indicates if this keyset pagination information is equal to other
+     * keyset pagination information. The comparison includes keyset cursor
+     * instances but does not do a deep comparison of keyset values within
+     * the cursor instance.
+     *
+     * @return true if equal. Otherwise false.
+     */
+    @Override
+    public boolean equals(Object o) {
+        KeysetPageable p;
+        return this == o || o != null
+                && o.getClass() == getClass()
+                && (p = (KeysetPageable) o).size == size
+                && p.page == page
+                && p.mode == mode
+                && p.cursor == cursor;
     }
 
     /**
@@ -190,5 +224,37 @@ public class KeysetPageable extends Pageable {
      */
     public Mode getMode() {
         return mode;
+    }
+
+    /**
+     * <p>Returns a hash code based upon:</p>
+     * <ul>
+     * <li>page number</li>
+     * <li>maximum page size</li>
+     * <li>mode</li>
+     * <li>keyset cursor</li>
+     * </ul>
+     *
+     * @return a hash code for the pagination information.
+     */
+    @Override
+    public int hashCode() {
+        return Objects.hash(size, page, mode, cursor);
+    }
+
+    /**
+     * Returns a string representation of the pagination information.
+     *
+     * @return a string representation of the pagination information.
+     */
+    @Override
+    public String toString() {
+        return new StringBuilder(100)
+                .append("KeysetPageable{size=")
+                .append(getSize()).append(", page=")
+                .append(getPage()).append(", mode=")
+                .append(mode).append(", cursor=")
+                .append(cursor).append("}")
+                .toString();
     }
 }
--- a/api/src/main/java/jakarta/data/repository/KeysetPageable.java
+++ b/api/src/main/java/jakarta/data/repository/KeysetPageable.java
@@ -63,8 +63,9 @@ import java.util.Objects;
  * <p>You can also construct a <code>KeysetPageable</code> directly, which
  * allows you to make it relative to a specific list of values. The number and
  * order of values must match that of the {@link OrderBy} annotations,
- * {@link Sort} parameters, or <code>OrderBy</code> name pattern of the
- * repository method. For example,</p>
+ * <code>Pageable</code> {@link Pageable#of(long, long, Sort...) Sort} parameters,
+ * or <code>OrderBy</code> name pattern of the repository method.
+ * For example,</p>
  *
  * <pre>
  * Employee emp = ...
@@ -90,7 +91,8 @@ import java.util.Objects;
  * Enclose the entire conditional expression of the <code>WHERE</code> clause
  * in parenthesis.
  * Sort criteria must be specified independently from the user-provided query,
- * either with the {@link OrderBy} annotation or {@link Sort} parameters.
+ * either with the {@link OrderBy} annotation or
+ * <code>Pageable</code> {@link Pageable#of(long, long, Sort...) Sort} parameters.
  * For example,</p>
  *
  * <pre>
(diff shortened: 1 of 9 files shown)


njr-11:
Rebased onto #37 for Slice, Page, and updates to Pageable.
Updated per above.
Additional specification documentation added.

otaviojava (approved):

njr-11:
Copying @beikov for additional review

api/src/main/java/jakarta/data/repository/KeysetPageable.java
```diff
@@ -0,0 +1,265 @@
...
+            return new StringBuilder("Cursor@").append(Integer.toHexString(hashCode())) //
+                            .append(" with ").append(keyset.length).append(" keys") //
+                            .toString();
+        }
+    }
+
+    private final Cursor cursor;
+    private final Mode mode;
+
+    KeysetPageable(Pageable copyFrom, Mode mode, Object... keyset) {
```
beikov, line 183: IMO we should accept a `Cursor` instead of `Object...` here for the keyset here and also add methods `Pageable#afterKeyset(Cursor)`/`Pageable#beforeKeyset(Cursor)`, as injecting a custom cursor might be desirable.

api/src/main/java/jakarta/data/repository/KeysetPageable.java
```diff
@@ -0,0 +1,265 @@
...
+         * or <code>OrderBy</code> name pattern of the repository method.
+         */
+        PREVIOUS
+    }
+
+    /**
+     * Represents keyset values, which can be a starting point for
+     * requesting a next or previous page.
+     */
+    public static class Cursor {
```
beikov, line 138: I think I would prefer to have an interface here and make the implementation private. If someone wants a different backing storage than an `Object[]` or want to provide some extensions to the cursor like the type of individual elements, allowing custom implementations is very useful.

api/src/main/java/jakarta/data/repository/KeysetPageable.java
```diff
@@ -0,0 +1,265 @@
...
+ * limitations under the License.
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.repository;
+
+import java.util.Objects;
+
+/**
+ * <p>Keyset pagination is a form of pagination that aims to reduce the
```
beikov, line 23: It's actually also helping with performance in case you want to mention this. Also see https://use-the-index-luke.com/no-offset

api/src/main/java/jakarta/data/repository/KeysetPageable.java
```diff
@@ -0,0 +1,265 @@
...
+ * Keyset pagination does not prevent misses and duplicates if the entity
+ * properties which are the sort criteria for existing entities are modified
+ * or if an entity is re-added with different sort criteria after having
+ * previously been removed.</p>
+ *
+ * <h2>Keyset Pagination with &#64;Query</h2>
+ *
+ * <p>Keyset pagination involves generating and appending to the query
+ * additional query conditions for the keyset properties. In order for that to
+ * be possible, the user-provided JPQL query must end with a
```
beikov, line 89: Reference to JPQL?

api/src/main/java/jakarta/data/repository/KeysetPageable.java
```diff
@@ -0,0 +1,265 @@
...
+     */
+    @Override
+    public boolean equals(Object o) {
+        KeysetPageable p;
+        return this == o || o != null
+                && o.getClass() == getClass()
+                && (p = (KeysetPageable) o).size == size
+                && p.page == page
+                && p.mode == mode
+                && p.cursor == cursor
```
beikov, line 209: Should use `equals`

api/src/main/java/jakarta/data/repository/KeysetPageable.java
```diff
@@ -0,0 +1,265 @@
...
+ *
+ * <p>To use keyset pagination, define a repository method with return value of
+ * {@link KeysetAwareSlice} or {@link KeysetAwarePage} and which accepts a
+ * special parameter (after the normal query parameters) that is a
+ * {@link Pageable}. For example,</p>
+ *
+ * <pre>
+ * &#64;OrderBy("lastName")
+ * &#64;OrderBy("firstName")
+ * &#64;OrderBy("id")
```
beikov, line 41: Not sure where to put that comment, but it might be worth pointing out that the fields by which the user orders by, should form a unique tuple for keyset pagination to work properly. Otherwise the database needs a notion of "ties" like in SQL the `... rows only with ties` clause to allow the server to fetch more rows, in case there are more with the same ordering.

Consider this:

```
users = [
{name: "A", id: 1},
{name: "B", id: 2},
{name: "B", id: 3},
{name: "C", id: 4}
]
```

When using `Pageable.size(2)` with just `@OrderBy("name")`, we will get either:

```
users = [
{name: "A", id: 1},
{name: "B", id: 2}
]
```

or:

```
users = [
{name: "A", id: 1},
{name: "B", id: 3}
]
```

and when paging to the next page:

```
users = [
{name: "C", id: 4}
]
```

because the data store uses `>`.

In Blaze-Persistence I validate the order by expressions and fail constructing a pagination query if I figure out during analysis that the expressions do not form a unique tuple.

njr-11:
@beikov thanks for the excellent review comments. I definitely missed mentioning the performance benefit and the requirements for the collection of keyset values to uniquely identify each entity.  I've made updates to accommodate before/after with a Cursor and have made the provided Cursor class extensible.  It might be possible to make it into an interface as well.  I'll look into that some more.

njr-11:
It ended up cleaner than I thought to switch Cursor to an interface. That is updated now as well.  All of the review comments should be addressed now.  Let me know if any additional changes are needed.

api/src/main/java/jakarta/data/repository/KeysetPageable.java
```diff
@@ -0,0 +1,328 @@
...
+                    && Arrays.equals(keyset, ((CursorImpl) o).keyset);
+        }
+
+        public Object getKeysetElement(int index) {
+            return keyset[index];
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(keyset);
```
beikov, line 227: Minor comment, should probably use `Arrays.hashCode` directly.

beikov (approved):
Looks great now, thanks for considering the suggested changes and great work! I added one more minor comment. Feel free to ignore that if you think it's not important.

api/src/main/java/jakarta/data/repository/KeysetPageable.java
njr-11: Thanks - I had to rebase onto latest anyways due to a merge conflict with a line in repository.asciidoc, after which I spotted a couple of minor grammar errors in my changes there as well as another place to state the requirement for sort criteria to uniquely identify entities, and I also added the update to Arrays.hashCode.  This should be all of the updates that I'm aware of.

mpredli01 (approved):
Everything looks good to me!

njr-11:
> otaviojava requested review from mpredli01 and keilw 5 days ago

@keilw all of the other reviewers have completed their reviews. Let me know if you had any comments on this or if you were not interested in reviewing. Thanks!

keilw (commented):
This Built-in implementation of `Cursor` seems a little odd. What is the purpose and is it intended to be overwritten by implementations?
If not, why is it an interface instead of just a class?

njr-11:
> This Built-in implementation of `Cursor` seems a little odd. What is the purpose and is it intended to be overwritten by implementations? If not, why is it an interface instead of just a class?

Originally I had only the built-in Cursor implementation as its own class (and no interface) as you are suggesting, with its purpose being to shield applications from having to deal with the Cursor at all. In many cases, the application only needs to supply the keyset values they are interested in and doesn't need to worry about getting a Cursor from somewhere just to hand the keyset values to the API.

It was brought up separately by one of the other reviewers that some Jakata Data providers will want to create their own custom Cursor implementations, so the interface was introduced, allowing for both patterns.

Another possibility could be having a Cursor class only and making it extensible, but @beikov would need to agree to that.

beikov:
>This Built-in implementation of Cursor seems a little odd. What is the purpose and is it intended to be overwritten by implementations?
>If not, why is it an interface instead of just a class?

My intention is to allow jakarta-data implementers to provide their own `Cursor` implementations to e.g. provide extensions like exposing the order by item expression or simply provide a custom serialization implementation.

To me, having an interface here is the most flexible. Since users most probably won't create `Cursor` instances themselves, I also don't think that the interface here is a problem. Can you elaborate what you think is odd in particular?

>Another possibility could be having a Cursor class only and making it extensible, but @beikov would need to agree to that.

We could surely do that, but then the object layout would be fixed i.e. every subtype has to use an `Object[]` as storage or ignore that. Can you explain what kind of problems you think the interface approach could lead to?

njr-11:
@keilw Would it be alright to merge this as it is now (with the interface) and then I'll open a second pull that would switch to an extensible class which can be discussed and decided separately?

api/src/main/java/jakarta/data/repository/KeysetPageable.java
```diff
@@ -0,0 +1,328 @@
...
+ * KeysetAwareSlice&lt;Customer&gt; getTopBuyers(int minOrders, float minSpent, Pageable pagination);
+ * </pre>
+ *
+ * <h2>Page Numbers and Totals</h2>
+ *
+ * <p>Page numbers, total numbers of elements across all pages, and total
+ * count of pages are not accurate when keyset pagination is used and should
+ * not be relied upon.</p>
+ */
+public class KeysetPageable extends Pageable {
```
graemerocher, line 119: As per my question in https://github.com/jakartaee/data/pull/64

Why is this not an interface?

api/src/main/java/jakarta/data/repository/KeysetPageable.java
```diff
@@ -0,0 +1,328 @@
...
+         * @return String representation of the keyset cursor.
+         */
+        @Override
+        public String toString();
+    }
+
+    /**
+     * Built-in implementation of Cursor.
+     */
+    static class CursorImpl implements Cursor {
```
graemerocher, line 196: should be `final`

api/src/main/java/jakarta/data/repository/KeysetPageable.java
njr-11: > As per my question in #64
> 
> Why is this not an interface?

It's a valid topic to discuss whether `Pageable` (and similarly `KeysetPageable`) ought to be made into an interface, but within this pull it is right for `KeysetPageable` to follow after `Pageable`.  Please open a separate issue to discussing changing to interface for Pageable.

api/src/main/java/jakarta/data/repository/KeysetPageable.java
njr-11: Good catch. I corrected this in https://github.com/jakartaee/data/pull/52/commits/24b8247df31b01ec682495b338621885a3774204 . Please review it.

api/src/main/java/jakarta/data/repository/KeysetPageable.java
njr-11: This topic is being addressed separately in #65

njr-11:
Rebased onto latest and resolved soft merge conflict with #64

PR #57, opened 2022-10-27
Introduce TCK project

### LINKED ISSUE #26: Create a blueprint of the TCK
Now that Jakarta EE specs have the ability to develop TCKs outside of the main TCK repository I think it would be a good idea to create a blueprint for how we want the TCK to operate. 
Having a blueprint in place early and having builds set up to create snapshot artifacts in the Jakarta stagging repository will allow implementors quick and easy access to test their implementations. 

Things to consider: 
- Do tests need to run inside of a Web or EJB container, or can they be written as standalone tests? 
- What test platform should we use? JUnit4/JUnit5/TestNG?
- What database(s) will users need for testing?  Or do we need to somehow create mock in-memory databases?
- How will we implement Signature Testing and how can we automate the creation of signature test files?

I think getting this set up as soon as possible will help the development of the spec since we will be able to write tests alongside updates to the API. I'm willing to help set this up 😄

### LINKED ISSUE #51: Licensing for TCK and TCK Distribution Zip
I have been working on #26
and have noticed that part of the current maven build is a check to ensure all files contain the Apache license: 
 https://github.com/jakartaee/data/blob/cd566ce3232d74b0e2c29683ed88822e26178d79/pom.xml#L286-L307

The problem is that the TCK needs to include signature testing, which means that we need to copy and modify code from the platform TCK to ensure consistent testing/reporting:
https://github.com/eclipse-ee4j/jakartaee-tck/tree/master/src/com/sun/ts/tests/signaturetest

Which are licensed under `EPL-2.0` and `GPL-2.0 WITH Classpath-exception-2.0`

Also from working on the Concurrency TCK I know that Eclipse requires the following license structure for TCKs:

- TCK Artifact (jar): `EPL-2.0` and `GPL-2.0 WITH Classpath-exception-2.0`
- TCK Distribution (zip which includes tck-jar + doc + samples): `EFTL` - a special license for [Eclipse TCKs](https://www.eclipse.org/legal/tck.php)

Possible solutions: 
- Switch jakarta/data to use `EPL-2.0` throughout
- Keep the API under `Apache 2.0` and have different licenses for the TCK and TCK dist.

I'm also open to other discussions/ideas

KyleAure:
Fixes #26
Introduces sample test classes for running against standalone, core, and web profiles.
- Core Profile `ee.jakarta.tck.data.core.example`: Runs a test using a CDI bean
- Standalone  `ee.jakarta.tck.data.standalone.example`: Runs a test on the client JVM when in standalone mode, otherwise packages and deploys the test to the application server when not in standalone mode 
- Signature Tests `ee.jakarta.tck.data.standalone.signature`: Runs on all profiles and in standalone mode 
- Web Profile `ee.jakarta.tck.data.web.example`: Runs custom servlet apps, simple servlet apps, and EJB apps 

Introduces signature testing for the Jakarta Data API.

Extends Arquillian to automatically include TCK framework packages into archives that are deployed to the server: 
- All archives `ee.jakarta.tck.data.framework.junit.anno`: include custom annotations that tag and proxy the Arquillian extension 
- All archives `ee.jakarta.tck.data.framework.junit.extensions`: include JUnit extensions that bypass Arquillian when running in standalone mode, and log failing assertions to help implementors fix issues 
- Core profile: has no additional packages
- Web Profile `ee.jakarta.tck.data.framework.servlet`: include custom servlet runner class 
- Full profile: same as web profile
- Signature tests `ee.jakarta.tck.data.framework.signature`: include the Jakarta TCK signature test runner classes

Fixes #51 
- Skips the apache rat plugin so source files can use EPL & GPL license headers
- TCK pom uses EPL & GPL


tck/src/main/java/ee/jakarta/tck/data/framework/junit/anno/Standalone.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Tags;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import ee.jakarta.tck.data.framework.junit.extensions.AssertionExtension;
+import ee.jakarta.tck.data.framework.junit.extensions.StandaloneExtension;
+
+/**
+ * These are test classes that can DO NOT depend on any Jakarta EE server
```
njr-11, line 31:
```suggestion
 * These are test classes that DO NOT depend on any Jakarta EE server
```

tck/src/main/java/ee/jakarta/tck/data/framework/junit/anno/Standalone.java
```diff
@@ -0,0 +1,51 @@
...
+ * technologies.
+ * 
+ * However, when running the TCK against a core/web/full profile server these
+ * tests will be deployed and run on the server.
+ * 
+ * The dynamic method in which we achieve the ability to run these tests either
+ * on the client JVM or server JVM depends on both the JUnit5 tag AND a system
+ * property: jakarta.standalone.test
+ * 
+ * If jakarta.standalone.test=true the Arquillian extension will be ignore,
```
njr-11, line 41:
```suggestion
 * If jakarta.standalone.test=true the Arquillian extension will be ignored,
```

tck/src/main/java/ee/jakarta/tck/data/framework/servlet/URLBuilder.java
```diff
@@ -0,0 +1,210 @@
...
+     * returns: http://localhost:80/servlet/endpoint1/
+     * </pre>
+     * @param path - The path to append
+     * @return This URL builder object
+     */
+    public URLBuilder withPath(String path) {
+        if (this.paths == null) {
+            this.paths = new ArrayList<>(Arrays.asList(path));
+        } else {
+            this.paths.addAll(Arrays.asList(path));
```
njr-11, line 115:
```suggestion
            this.paths.add(path);
```

tck/src/main/java/ee/jakarta/tck/data/framework/signature/DataSignatureTestRunner.java
```diff
@@ -0,0 +1,387 @@
...
+
+        // unlisted optional packages are technology packages for those optional
+        // technologies (e.g. jsr-88) that might not have been specified by the
+        // user.
+        // We want to ensure there are no full or partial implementations of an
+        // optional technology which were not declared
+        ArrayList<String> unlistedTechnologyPkgs = getUnlistedOptionalPackages();
+
+        // If testing with Java 9+, extract the JDK's modules so they can be used
+        // on the testcase's classpath.
```
njr-11, line 239: Jakarta Data will not ever be tested with Java 8 and earlier if that helps to optimize the subsequent code

tck/src/main/resources/ee/jakarta/tck/data/framework/signature/sig-test.map
```diff
@@ -0,0 +1,20 @@
+#
+# Copyright (c) Contributors to the Eclipse Foundation
```
njr-11, line 2:
```suggestion
# Copyright (c) 2022 Contributors to the Eclipse Foundation
```

dstepanov:
What is the purpose of this TCK? My understanding is that is for external projects like Micronaut Data to validate the integration. If yes, then I don't understand direct dependencies on EJB, Servlet API and Maven.

Also, the signature test should be a part of the internal test suite not the external one.

njr-11:
> What is the purpose of this TCK? My understanding is that is for external projects like Micronaut Data to validate the integration. If yes, then I don't understand direct dependencies on EJB, Servlet API and Maven.

Because the Jakarta Data specification defines requirements for standalone Jakarta Data providers as well as Jakarta EE product providers that integrate with Jakarta Data providers, the TCK needs to have multiple purposes, including the one you stated. There will be different modes of running the TCK for testing each:

- A standalone mode for testing compatibility of a Jakarta Data provider, such as Micronaut.  Running in this mode does not require other Jakarta EE technologies like Servlet and EJB.
- A web profile mode for testing a Jakarta EE web profile product's integration with Jakarta Data.  This covers both requirements that Jakarta Data places directly on Jakarta EE product providers as well as that a Jakarta Data providers that is shipped by a Jakarta EE web profile product complies with the Jakarta Data specification.
- A core profile mode, similar to above, although also not requiring Servlet and EJB. We haven't yet determined which Jakarta EE profile Jakarta Data will go into, and allowing for either at this point will be most flexible to whatever decision gets made.

It seems likely a secondary mode will be needed as well to designate whether a Jakarta Data provider is compatible with Jakarta Persistence (JPQL queries) or which types of Jakarta NoSQL databases for tests that cannot be written generically.

The TCK will need maven and some other Jakarta EE technologies as dependencies for building the entire TCK which includes all modes, but running  the TCK should only require the dependencies that are applicable to the mode that you are running.

Information on the available modes and when each mode is applicable definitely belongs in the TCK user guide and will need to get there eventually, but this is only the very first pull to get the TCK started out.

KyleAure:
> Also, the signature test should be a part of the internal test suite, not the external one.

The signature test is to verify that the API used to write your implementation has the same class/method signatures as the official API when it is released.  This is a test strategy that was first used by Java EE before it was Open Source and it was possible that implementations were using snapshot or beta versions of an API that had signature changes in the official release.  

Now that development is happening in the Open and artifact management is pretty universal this has less of a chance of happening, but something that is still required by Jakarta EE specifications. IMO I'd like to see an updated method of testing signatures that wasn't so clunky but for now, this is what we need to have.

dstepanov:
> > What is the purpose of this TCK? My understanding is that is for external projects like Micronaut Data to validate the integration. If yes, then I don't understand direct dependencies on EJB, Servlet API and Maven.
> 
> Because the Jakarta Data specification defines requirements for standalone Jakarta Data providers as well as Jakarta EE product providers that integrate with Jakarta Data providers, the TCK needs to have multiple purposes, including the one you stated. There will be different modes of running the TCK for testing each:
> 
>     * A standalone mode for testing compatibility of a Jakarta Data provider, such as Micronaut.  Running in this mode does not require other Jakarta EE technologies like Servlet and EJB.
> 
>     * A web profile mode for testing a Jakarta EE web profile product's integration with Jakarta Data.  This covers both requirements that Jakarta Data places directly on Jakarta EE product providers as well as that a Jakarta Data providers that is shipped by a Jakarta EE web profile product complies with the Jakarta Data specification.
> 
>     * A core profile mode, similar to above, although also not requiring Servlet and EJB. We haven't yet determined which Jakarta EE profile Jakarta Data will go into, and allowing for either at this point will be most flexible to whatever decision gets made.
> 
> 
> It seems likely a secondary mode will be needed as well to designate whether a Jakarta Data provider is compatible with Jakarta Persistence (JPQL queries) or which types of Jakarta NoSQL databases for tests that cannot be written generically.
> 
> The TCK will need maven and some other Jakarta EE technologies as dependencies for building the entire TCK which includes all modes, but running the TCK should only require the dependencies that are applicable to the mode that you are running.
> 
> Information on the available modes and when each mode is applicable definitely belongs in the TCK user guide and will need to get there eventually, but this is only the very first pull to get the TCK started out.

Why can't it be profile/container agnostic? I understand it's hard to write the TCK that would test all those scenarios.

@KyleAure @njr-11 The signature check looks like overkill. You can store the API version in the META-INF and have a TCK check that matches its version of the API ones. By that, you validate that the TCK matches the API release.

Please have the TCK building tool agnostic, as Micronaut doesn't use maven.

njr-11:
> Why can't it be profile/container agnostic? I understand it's hard to write the TCK that would test all those scenarios.

Restricting the TCK to profile agnostic would limit coverage to the least common denominator, excluding coverage of requirements that the Jakarta Data specification makes regarding integration with Jakarta EE Web and Core profile products.  I agree with you that this makes it more difficult, but I don't see how the added difficulty permits us to exclude this coverage.  With the work Kyle has done, it should hopefully be possible to write the vast majority of tests to apply to all profiles (and standalone), and then have a limited number of tests for the profile-specific integration.  That is what we should be aiming for.

pom.xml
```diff
@@ -73,22 +73,27 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
         <jakarta.annotation.version>1.3.5</jakarta.annotation.version>
-        <jakarta.enterprise.cdi.version>2.0.1</jakarta.enterprise.cdi.version>
+        <jakarta.ejb.version>4.0.1</jakarta.ejb.version>
```
otaviojava, line 76: If we agree to work on Jakata-Core, we should not include EJB on the project: 
https://jakarta.ee/release/10/images/jakarta-ee-10-platform.svg
@njr-11 what do you think?

njr-11: > If we agree to work on Jakata-Core, we should not include EJB on the project: https://jakarta.ee/release/10/images/jakarta-ee-10-platform.svg @njr-11 what do you think?

I agree, let's omit the EJB dependency given that the specification does not currently make any requirements with respect to EJBs that would need to be tested in the TCK.  I expect that at some point we will want to define how data repositories behave with respect to container managed transactions when the underlying resource is transaction capable, but that could be tested with the CDI-based `@Transactional` rather than with EJBs.

KyleAure: @otaviojava @njr-11 EJBs have been removed.  Thanks for the feedback.  Just put them to make sure the project was set up correctly for the Arquillian EJB enricher.  I agree that sticking with CDI's `@Transactional` is the best move 👍🏼

njr-11 (approved):

KyleAure:
@otaviojava Is there anything else holding up this PR? Or can it be merged?

otaviojava (approved):

PR #59, opened 2022-10-30
Add Streamable interface

otaviojava:
As @odrotbohm recommendation, I've created this PR for this discussion:

https://github.com/jakartaee/data/pull/37#issuecomment-1286980442

The main reason for this PR is that the Stream makes sense to handle with Data at Jakarta Data. However, the structure should consider the resource available for it.

> Spring Data's Page implements Streamable, an extension of Iterable that exposes a stream() method defaulting to non-parallel streaming.

https://github.com/jakartaee/data/issues/36#issuecomment-1268074831

So, I created this PR to start the discussion on the ```Streamable``` 

The biggest inspiration comes from [Spring Data Commons](https://github.com/spring-projects/spring-data-commons/blob/main/src/main/java/org/springframework/data/util/Streamable.java), where I took a feel method to start the Topic.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Streamable.java
+++ b/api/src/main/java/jakarta/data/repository/Streamable.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.repository;
+
+import java.util.List;
+import java.util.Set;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * This is interface and can therefore be used as the assignment target for a lambda expression or method reference.
+ */
+@FunctionalInterface
+public interface Streamable<T> extends Iterable<T>, Supplier<Stream<T>> {
+
+    /**
+     * Creates a non-parallel {@link Stream} of the underlying {@link Iterable}.
+     *
+     * @return a non-parallel Stream
+     */
+    default Stream<T> stream() {
+        return StreamSupport.stream(spliterator(), false);
+    }
+
+    /**
+     * Creates a new, unmodifiable {@link Set}.
+     *
+     * @return a Set
+     */
+    default Set<T> toSet() {
+        return stream().collect(Collectors.toUnmodifiableSet());
+    }
+
+    @Override
+    default Stream<T> get() {
+        return stream();
+    }
+
+    /**
+     * Returns whether the current {@link Streamable} is empty.
+     *
+     * @return whether the current {@link Streamable} is empty.
+     */
+    default boolean isEmpty() {
+        return !iterator().hasNext();
+    }
+
+    /**
+     * Creates a new, unmodifiable {@link List}.
+     *
+     * @return will never be {@literal null}.
+     */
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/data/repository/Streamable.java
```diff
@@ -0,0 +1,74 @@
...
+    default boolean isEmpty() {
+        return !iterator().hasNext();
+    }
+
+    /**
+     * Creates a new, unmodifiable {@link List}.
+     *
+     * @return will never be {@literal null}.
+     */
+    default List<T> toList() {
```
njr-11, line 71: We should remove `Slice.getContent()`, which is made redundant by this method.

api/src/main/java/jakarta/data/repository/Streamable.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * Simple interface to ease streamability of {@link Iterable}s.
+ * This is interface and can therefore be used as the assignment target for a lambda expression or method reference.
+ */
+@FunctionalInterface
+public interface Streamable<T> extends Iterable<T>, Supplier<Stream<T>> {
```
njr-11, line 32: Is there an intent to use this beyond pagination and as its own return type for repository methods?
`Streamable<Person> streamable = people.findByFirstName(name)`
If so, it seems like each call to streamable.get() would need to fetch the data again if allowing it to be backed by a JPA provider's `query.getResultStream()`, and then we should also be clear about requiring the user to ensure the stream is closed when done with it.  This would be fine if that's the case, but we need to spell out the requirements then so that the user can know what to rely upon from the implementation.
If it's only for pagination and restricted to operating an already-loaded list of entities for a page then we don't need to be concerned with these things.

njr-11 (changes requested):

api/src/main/java/jakarta/data/repository/Streamable.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Creates a non-parallel {@link Stream} of the underlying {@link Iterable}.
+     *
+     * @return a non-parallel Stream
+     */
+    default Stream<T> stream() {
```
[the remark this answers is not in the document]
odrotbohm: I wouldn't actually do that, as the presence of `get` is a necessary side effect of implementing `Supplier`. The `stream()` method is much more explicit and also in alignment with `Collection.stream()`, which `Streamable` is usually more of a replacement for. If in doubt, I'd even rather go ahead and remove the `Supplier` implementation. It was merely added in Spring Data to be able to hand a `Streamable` into methods that expect a `Supplier<Stream>` but that's definitely not the primary use case of it.

api/src/main/java/jakarta/data/repository/Streamable.java
odrotbohm: In Spring Data, we don't assume such indirection. Executing a query method results in the query result computed unless explicitly specified (e.g. when using `Stream`) directly. That's more of a specification question for query methods rather than a design question of the type here. Theoretically, you could also implement a query method returning an `Iterable` in a way that wraps some database cursor, but I've just never seen anyone do that because of all the nasty side effects that keeping resources open beyond the invocation of a method brings.

keilw: How about other frameworks, e.g. Micronaut, @graemerocher ?

graemerocher: We don't currently have a `Streamable` interface as of today. Could be nice though. We do allow returning `Stream` instances from finder methods though.

graemerocher: Just to say from my perspective I don't see any issue with supporting this though, it seems useful

api/src/main/java/jakarta/data/repository/Streamable.java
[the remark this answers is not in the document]
njr-11: I'd be fine with keeping `stream()` and getting rid of `Supplier` / `get()`.  My intent was just to avoid the redundancy and what you've suggested will achieve that.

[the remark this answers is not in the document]
graemerocher: yeah removing `Supplier<Stream>` makes sense to me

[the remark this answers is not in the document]
otaviojava: Done.

njr-11 (approved):

PR #64, opened 2022-11-03
Reorder the page and size parameters on the Pageable.java constructor…

mpredli01:
… to match those of the overload of() methods

Pursuant to our Jakarta Data meeting on 11/2, this PR addresses my concern over the order of the page and size variables in the constructor and overloaded of() methords.

Everything was successfully tested and compiled in my local branch.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Pageable.java
+++ b/api/src/main/java/jakarta/data/repository/Pageable.java
@@ -56,25 +56,16 @@ public class Pageable {
 
     private static final long DEFAULT_SIZE = 10;
 
-    private final long size;
-
     private final long page;
+
+    private final long size;
 
     private final List<Sort> sorts;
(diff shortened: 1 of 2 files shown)


graemerocher:
I'm wondering why `Pageable` is not an interface since by making it a class we limit evolving the implementation later (for example with new defaults methods)?

njr-11:
> I'm wondering why `Pageable` is not an interface since by making it a class we limit evolving the implementation later (for example with new defaults methods)?

It could be nice to have an interface. I don't understand the specific concern about default methods though. A class allows for adding new methods, implemented with a default behavior.

njr-11 (approved):
The comment about whether an interface should be used for `Pageable` is being addressed as a separate issue, #65, so I think we can merge this.

PR #66, opened 2022-11-04
switch to Cursor class

njr-11:
This pull is opened in response to a [review comment](https://github.com/jakartaee/data/pull/52#pullrequestreview-1165337102) from @keilw about switching the `Cursor` interface to a class.  It needs to be agreed upon with @beikov whether to cancel this pull without changes because the interface is preferable or whether to merge the pull and switch to a class.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/KeysetPageable.java
+++ b/api/src/main/java/jakarta/data/repository/KeysetPageable.java
@@ -144,7 +144,24 @@ public class KeysetPageable extends Pageable {
      * Represents keyset values, which can be a starting point for
      * requesting a next or previous page.
      */
-    public interface Cursor {
+    public static class Cursor {
+        /**
+         * Keyset values.
+         */
+        protected final Object[] keyset;
+
+        /**
+         * Constructs a keyset cursor with the specified values.
+         *
+         * @param keyset keyset values.
+         * @throws IllegalArgumentException if no keyset values are provided.
+         */
+        public Cursor(Object... keyset) {
+            this.keyset = keyset;
+            if (keyset == null || keyset.length == 0)
+                throw new IllegalArgumentException("No keyset values were provided.");
+        }
+
         /**
          * Returns whether or not the keyset values of this instance
          * are equal to those of the supplied instance.
@@ -153,7 +170,11 @@ public class KeysetPageable extends Pageable {
          * @return true or false.
          */
         @Override
-        public boolean equals(Object o);
+        public boolean equals(Object o) {
+            return this == o || o != null
+                    && o.getClass() == getClass()
+                    && Arrays.equals(keyset, ((Cursor) o).keyset);
+        }
 
         /**
          * Returns the keyset value at the specified position.
@@ -163,7 +184,9 @@ public class KeysetPageable extends Pageable {
          * @throws IndexOutOfBoundsException if the index is negative
          *         or greater than or equal to the {@link #size}.
          */
-        public Object getKeysetElement(int index);
+        public Object getKeysetElement(int index) {
+            return keyset[index];
+        }
 
         /**
          * Returns a hash code based on the keyset values.
@@ -171,14 +194,18 @@ public class KeysetPageable extends Pageable {
          * @return a hash code based on the keyset values.
          */
         @Override
-        public int hashCode();
+        public int hashCode() {
+            return Arrays.hashCode(keyset);
+        }
 
         /**
          * Returns the number of values in the keyset.
          *
          * @return the number of values in the keyset.
          */
-        public int size();
+        public int size() {
+            return keyset.length;
+        }
 
         /**
          * String representation of the keyset cursor, including the number of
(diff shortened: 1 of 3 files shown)


beikov:
I'll repost my questions for @keilw here

----

This Built-in implementation of Cursor seems a little odd. What is the purpose and is it intended to be overwritten by implementations?
If not, why is it an interface instead of just a class?

My intention is to allow jakarta-data implementers to provide their own Cursor implementations to e.g. provide extensions like exposing the order by item expression or simply provide a custom serialization implementation.

To me, having an interface here is the most flexible. Since users most probably won't create Cursor instances themselves, I also don't think that the interface here is a problem. Can you elaborate what you think is odd in particular?

Another possibility could be having a Cursor class only and making it extensible, but @beikov would need to agree to that.

We could surely do that, but then the object layout would be fixed i.e. every subtype has to use an Object[] as storage or ignore that. Can you explain what kind of problems you think the interface approach could lead to?

keilw:
@beikov 

Not sure, how using a class instead of an interface would change the way things are stored?
It would be Cursor, or can you explain here in concrete examples what difference the interface vs. class would make?

As this is defined in the public API of Jakarta Data, an interface makes it likely that implementors must also implement it and the TCK needs to take it into consideration.

beikov:
>Not sure, how using a class instead of an interface would change the way things are stored?
>It would be Cursor, or can you explain here in concrete examples what difference the interface vs. class would make?

The point is that implementers are free to implement the `Cursor` interface however they like if they want.

If an implementation wants to provide a custom `Cursor`, it can do so by returning that from `KeysetAwareSlice#getCursor()` or implicitly through `KeysetAwareSlice#nextPageable()`/`KeysetAwareSlice#previousPageable()`.

Reasons for `Cursor` subtypes could be:

* Provider specific API e.g. provide access to the query expression for a cursor element
* Custom serialization logic for cursor values
* Custom storage layout (not `Object[]`) for the cursor

keilw:
Not sure, if that "if they want" works for the spec, they may have to implement it anyway, that is the point I want to clarify with platform, Spec Committee or others.
There are "Optional Features" e.g. in [Enterprise Beans](https://jakarta.ee/specifications/enterprise-beans/4.0/jakarta-enterprise-beans-spec-opt-4.0.html) but that's usually an entire spec that is no longer mandatory.

I think similar to Jakarta NoSQL we already have several possible areas of "optionality", remember using GraphQL or not, but currently `KeysetPageable` and its `Cursor` are baked into the abstract `Pageable` class, making them inseperable, so at the moment it is not possible to factor them out into a separate (optional) module, while e.g. Jakarta NoSQL has ben structured into modules already, see [mapping](https://github.com/eclipse-ee4j/nosql/tree/main/api/mapping). 
It allows in theory (although for existing specs I am not aware that works, it's "all or nothing") to have a vendor like MongoDB only implement the [mapping-document](https://github.com/eclipse-ee4j/nosql/tree/main/api/mapping/mapping-document) module but ignore column or key-value.
Even if we may not be forced to modularize it, there is not just a "if they want" because the TCK (via some kind of "profile") must cover all cases, especially the one where vendors implement their own `Cursor`.

That's why I'm raising this question, I am OK to merge the other PR but unless we change it to a class (in that case I guess we could call it an implementation detail) we will have to deal with it in the TCK.

beikov:
I don't know what the TCK needs to deal with here exactly, but unless an implementer wants to provide extensions or customize things, it shouldn't be necessary for an implementer to provide a custom `Cursor` implementation. An implementer should be able to use the default implementation through the means of `Pageable#afterKeyset`/`Pageable#beforeKeyset`.

That being said, I want to clarify that the keyset aspect does not have anything to do with GraphQL and that this feature is not something that can IMO be separated out into a sub-module or anything like that. It is IMO important for usability to make this part of the core API.

keilw:
No but the point is the aspect (challenge) of **Optionality and Modularity** (in the JCP only Java ME Embedded really accomplished that) whether it's different types of databases, GraphQL or anything else some vendors just don't want to care about.

In Jakarta Persistence there is no optionality either, which is why certain special features come with either Hibernate or EclipseLink. If they were in the spec both of them had to comply with it.
If Pageable/KeysetPageable are part of the API then the TCK must test for it and currently I'm not even sure, if a vendor has a choice, but if they do then they must run the TCK in a special profile or mode, e.g.  "default" vs. "customized" or similar.

With JSR 385 we were one of the only JSRs (beside the now unused ME Embedded) to offer a flexible set of 7 profiles in the [TCK](https://github.com/unitsofmeasurement/unit-tck). Inspired by the CDI TCK but there only differences between e.g. the Web and Full Profile of Jakarta EE are taken into consideration.
Having a much smaller CDI is the challenge of "CDI Lite" and wherever there are possible customizations we may have to offer some kind of "profile" in the TCK as well.

otaviojava:
@njr-11 do we need this PR?
If yes, can you rebase it, please?

njr-11:
> @njr-11 do we need this PR? If yes, can you rebase it, please?

Good point.  We might need it depending on the outcome of a discussion that @keilw and @beikov were having, but don't know yet.

Looking over the discussion, one item that seems inaccurate is:

> allowing "some vendors" to override means, all of them would have to

Let's consider another spec interface, `jakarta.data.repository.CrudRepository`.  There is a possibility a vendor could override it with their own, such as `com.vendor.repository.VendorCrudRepository`, maybe to add some vendor-specific built-in convenience methods or for whatever reason.  In no way does that imply that other vendors must do likewise, nor does it imply that a Jakarta Data TCK must test for the possibility of such a scenario.  Capabilities that vendors add outside of the specification remain outside of the specification.

beikov:
IMO we don't need it, but it appears @keilw doesn't like the interface approach, though I don't fully understand why.

keilw:
Yes, the Jakarta Data TCK must test for the possibility of such a scenario, and as soon as there is customization, it barely matters, if `Cursor` was an interface or (abstract) class.

njr-11:
I'll reiterate that it is out of scope for a TCK to cover vendor-specific extensions to the specification.  The only way I could see it being valid for TCK to require a vendor to supply their own `Cursor` would be if the specification had a method like the following on an interface (maybe Page) that the vendor must supply, `<C extends Cursor> C getCursor(Class<C> vendorCursorExtension)`, in which case it would be correct for the TCK to test the method and its vendor extension `Cursor` result.  But since we don't do that and aren't going to do that, then the TCK must not test for Jakarta Data vendors providing extensions to `Cursor`.

It would be valid for the TCK to test application-provided extensions to `Cursor`, with the TCK application providing the extension and supplying it to `pageable.beforeKeysetCursor` and `pageable.afterKeysetCursor` methods, verifying that page contents are as expected when used with a database that is capable of keyset cursor pagination, and when used with a database that is incapable of keyset cursor pagination then the repository method raises `UnsupportedOperationException` or whatever exception we define to be the specification required behavior in that case.  I'll get a pull open to define the exception for the error condition because I don't think the spec has defined it yet.

keilw:
Of course the TCK must test for implementations of the `Pageable.Cursor` interface, only any possible addition and new methods to a class that implements it would be out of scope for our TCK, but the behavior of the methods the API itself declares must be tested.

keilw (commented):
I am relatively unbiased between interface and static class. If it should be customizable, then the TCK must take the cursor into consideration, regardless of class or interface.

njr-11:
This pull is out of date and it is unclear that it is needed anymore, so I'm canceling it.

Issue #67, opened 2022-11-04
Inconsistent Pageable factory methods


keilw:
**Description**

There is an inconsistency among static factory methods in `Pageable`

- of() with various arguments
- page() which is just a convenience wrapper around of()
- size() ditto

IMO better consistency would be reached by calling those:
- ofPage()
- ofSize()

Along the lines of `OptionalofNullable()` or methods like `ofPrefix()`, `ofRational()`, etc. in JSR 385 [MultiplyConverter](https://github.com/unitsofmeasurement/indriya/blob/master/src/main/java/tech/units/indriya/function/MultiplyConverter.java). Mixing the two conventions, even worse because there are `getPage()` and `getSize()` methods easily confused with `page()` or `size()` seems quite bad.

njr-11:
`ofPage` and `ofSize` are being added in #68
Also, in that pull, we had some discussion about the possibility of switching to record style accessor methods for `Pageable`, which is being voted on in #70.  I'm copying the example from that discussion here because this is likely where it would end up being implemented:

```
p = Pageable.ofSize(50)
            .withPage(5)
            .withSorts(Sort.asc("lastName"), Sort.asc("firstName), Sort.asc("id"))
            .afterKeyset(lname, fname, id);
page = p.page();
size = p.size();
sorts = p.sorts();
mode = p.mode();
cursor = p.cursor();
```

otaviojava:
@njr-11, what do you think about removing the "of"?

```java
p = Pageable.size(50)
            .page(5)
            .sort(Sort.asc("lastName"), Sort.asc("firstName), Sort.asc("id"))
            .afterKeyset(lname, fname, id);
```

keilw:
Still makes it easy to confuse `size()` and `page()` with `getSize()` and `getPage()`.

njr-11:
> @njr-11, what do you think about removing the "of"?
> 
> ```java
> p = Pageable.size(50)
>             .page(5)
>             .sort(Sort.asc("lastName"), Sort.asc("firstName), Sort.asc("id"))
>             .afterKeyset(lname, fname, id);
> ```

We don't know the outcome of the vote under #70 yet, but if the decision is made to have record-style accessors, then methods named like `size(s)` and `page(p)` will be needed for that purpose instead.

> If it's `withPage()` or `withSorts()` then why not also call it `withSize()` that would be more consistent.
[...]

Issue #65, opened 2022-11-03
Pageable and friends should be interfaces


graemerocher:
**Description**

The current design has `Pageable` etc. as classes.  This limits the ability the evolve the API in the future IMO and I am wondering why this is the case.

njr-11:
Some things to consider:
- If `Pageable` becomes an interface, who will provide the implementation? We need to allow for the possibility that there are multiple active Jakarta Data providers (maybe one for NoSQL and one for relational databases). It would be unfortunate if we needed to make the API more complex for users to ensure they are getting the `Pageable` instance from the right one.  It seems like we would still want the Jakarta Data spec to provide an implementation to use by default even if it becomes an interface.
- How does the use of a class for `Pageable` versus an interface limit the ability to evolve the API in the future?  I'm not arguing against having an interface, it just isn't clear to me specifically where it is currently limiting which we will want to ensure an interface-based solution addresses.  In the comments to another pull, I pointed out that methods (along with impl that is default) can be added to classes just as methods with default impl can be added to interfaces.

njr-11:
If we switch to an interface with the spec also providing a basic implementation, one way to do it could be something like,
pageable = `Pageable.ofSize(50).page(1).sortBy(...).withKeyset(...)`
where we have a couple of static `of` methods to construct the instances that are provided by the spec,
and keep the `get` methods for accessing values,
and let the other methods be optionally used to configure values in a fluent pattern.
Different providers that wanted to enhance `Pageable` with capability beyond the spec could define their own extensions to the `Pageable` interface along with a provider-specific way to obtain it, which applications would only need to be concerned with if they want the value-add function that goes beyond the spec.
Would this sort of approach work?

graemerocher:
A default implementation can still be provided and created from static methods of Pageable, I don't see that as an issue as you have done in your example above.

And indeed implementations could have provider-specific extensions that go beyond the spec.

PR #68, opened 2022-11-04
Pageable as an interface

njr-11:
This pull converts Pageable and KeysetPageable into interfaces, with static ofSize and ofPage methods for obtaining the implementation from the Jakarta Data spec, following the pattern discussed with @graemerocher in #65 and possibly also addressing some concerns on naming that were independently raised by @keilw in #67.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/KeysetPageable.java
+++ b/api/src/main/java/jakarta/data/repository/KeysetPageable.java
@@ -17,9 +17,6 @@
  */
 package jakarta.data.repository;
 
-import java.util.Arrays;
-import java.util.Objects;
-
 /**
  * <p>Keyset pagination is a form of pagination that aims to reduce the
  * possibility of missed or duplicate results by making the request for
@@ -50,7 +47,7 @@ import java.util.Objects;
  * <p>You can use a normal {@link Pageable} to request an initial page,</p>
  *
  * <pre>
- * page = employees.findByHoursWorkedGreaterThan(1500, Pageable.size(50));
+ * page = employees.findByHoursWorkedGreaterThan(1500, Pageable.ofSize(50));
  * </pre>
  *
  * <p>For subsequent pages, you can request pagination relative to the
@@ -68,13 +65,13 @@ import java.util.Objects;
  * <p>You can also construct a <code>KeysetPageable</code> directly, which
  * allows you to make it relative to a specific list of values. The number and
  * order of values must match that of the {@link OrderBy} annotations,
- * <code>Pageable</code> {@link Pageable#of(long, long, Sort...) Sort} parameters,
+ * {@link Pageable#sortBy(Sort...)} or {@link Pageable#sortBy(Iterable)} parameters,
  * or <code>OrderBy</code> name pattern of the repository method.
  * For example,</p>
  *
  * <pre>
  * Employee emp = ...
- * KeysetPageable pagination = Pageable.size(50).afterKeyset(emp.lastName, emp.firstName, emp.id);
+ * KeysetPageable pagination = Pageable.ofSize(50).afterKeyset(emp.lastName, emp.firstName, emp.id);
  * page = employees.findByHoursWorkedGreaterThan(1500, pagination);
  * </pre>
  *
@@ -99,7 +96,7 @@ import java.util.Objects;
  * in parenthesis.
  * Sort criteria must be specified independently from the user-provided query,
  * either with the {@link OrderBy} annotation or
- * <code>Pageable</code> {@link Pageable#of(long, long, Sort...) Sort} parameters.
+ * {@link Pageable#sortBy(Sort...)} or {@link Pageable#sortBy(Iterable)} parameters.
  * For example,</p>
  *
  * <pre>
@@ -116,23 +113,25 @@ import java.util.Objects;
  * count of pages are not accurate when keyset pagination is used and should
  * not be relied upon.</p>
  */
-public class KeysetPageable extends Pageable {
+public interface KeysetPageable extends Pageable {
     /**
      * Direction of keyset pagination.
      */
     public static enum Mode {
         /**
          * Indicates forward keyset pagination, which follows the
-         * direction of the {@link OrderBy} annotations, <code>Pageable</code>
-         * {@link Pageable#of(long, long, Sort...) Sort} parameters,
+         * direction of the {@link OrderBy} annotations,
+         * {@link Pageable#sortBy(Sort...)} or {@link Pageable#sortBy(Iterable)}
+         * parameters, repository method {@link Sort} parameters,
          * or <code>OrderBy</code> name pattern of the repository method.
          */
         NEXT,
 
         /**
          * Indicates a request for a page in the reverse direction of
-         * the {@link OrderBy} annotations, <code>Pageable</code>
-         * {@link Pageable#of(long, long, Sort...) Sort} parameters,
+         * the {@link OrderBy} annotations,
+         * {@link Pageable#sortBy(Sort...)} or {@link Pageable#sortBy(Iterable)}
+         * parameters, repository method {@link Sort} parameters,
          * or <code>OrderBy</code> name pattern of the repository method.
          * The order of results on each page follows the sort criteria
          * and is not reversed.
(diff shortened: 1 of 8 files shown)


api/src/main/java/jakarta/data/repository/KeysetPagination.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+/**
+ * Built-in implementation of KeysetPageable.
+ */
+class KeysetPagination extends Pagination implements KeysetPageable {
```
graemerocher, line 30: can be `final`

api/src/main/java/jakarta/data/repository/KeysetPageable.java
```diff
@@ -191,138 +190,95 @@ public interface Cursor {
...
-        this.mode = mode;
-    }
+    public Cursor getCursor();
 
+    /**
+     * Returns the direction of keyset pagination.
+     *
+     * @return the direction of keyset pagination.
+     */
+    public Mode getMode();
```
graemerocher, line 205: for future planning do we want to use `getMode()` or `mode()`? When upgrading to Java 17 these could be records

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -54,20 +51,29 @@
...
-
-    final List<Sort> sorts;
+    /**
+     * Creates a new <code>Pageable</code> with the given page number and with a default size of 10.
+     *
+     * @param pageNumber The page number.
+     * @return a new instance of <code>Pageable</code>.
+     * @throws IllegalArgumentException when the page number is negative or zero.
+     */
+    public static Pageable ofPage(long pageNumber) {
```
graemerocher, line 63: clarify that the return value is never null

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -54,20 +51,29 @@
...
-    Pageable(long page, long size, List<Sort> sorts) {
-        this.page = page;
-        this.size = size;
-        this.sorts = sorts;
+    /**
+     * Creates a new <code>Pageable</code> for requesting pages of the
+     * specified size, starting with the first page number, which is 1.
+     *
+     * @param maxPageSize The number of query results in a full page.
+     * @return a new instance of <code>Pageable</code>.
```
graemerocher, line 72: clarify that the return value is never null

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -93,12 +97,10 @@ public KeysetPageable afterKeyset(Object... keyset) {
...
      *        <code>OrderBy</code> name pattern of the repository method to which
      *        this pagination will be supplied.
-     * @return reverse keyset pagination.
+     * @return a new instance of <code>KeysetPageable</code> with reverse keyset pagination.
      * @throws IllegalArgumentException if no keyset values are provided.
      */
-    public KeysetPageable beforeKeyset(Object... keyset) {
-        return new KeysetPageable(this, KeysetPageable.Mode.PREVIOUS, new KeysetPageable.CursorImpl(keyset));
-    }
+    public KeysetPageable beforeKeyset(Object... keyset);
```
graemerocher, line 103: clarify that the return value is never null

api/src/main/java/jakarta/data/repository/Pagination.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import jakarta.data.repository.KeysetPageable.Cursor;
+
+/**
+ * Built-in implementation of Pageable.
+ */
+class Pagination implements Pageable {
```
graemerocher, line 31: can be final

graemerocher (changes requested):
Looks good, just a few comments. I am wondering if we should consider the future that Java 17 has records and many implementations can be records.

api/src/main/java/jakarta/data/repository/KeysetPageable.java
njr-11: > When upgrading to Java 17 these could be records

Do you mean that the implementations of the Pageable/KeysetPageable interfaces could be records?  Or that Pageable/KeysetPageable themselves could become records?  I'm assuming you mean the former because the latter would prevent the possibility of provider-specific extensions that #65 aims to allow for.

I think it would be a good idea for the implementations to be records.  To accommodate that, we will need to do much more than change `getMode()` into `mode()`.  We also gave `getSize()` and `getPage()` and `getSorts()` and `getCursor()` that would need to be switched, and then some of those would collide with fluent-style `.size()` and `.page()`.  I used the fluent pattern to make it easier to add more properties in future spec versions.  We could either keep that and use different (longer) names for the fluent API, for example,

```
p = Pageable.ofSize(50)
            .withPage(5)
            .withSorts(Sort.asc("lastName"), Sort.asc("firstName), Sort.asc("id"))
            .afterKeyset(lname, fname, id);
page = p.page();
size = p.size();
sorts = p.sorts();
mode = p.mode();
cursor = p.cursor();
```

Or we could do away with the fluent pattern for the core properties and have something like,

```
p = Pageable.ofSize(50, 5, Sort.asc("lastName"), Sort.asc("firstName), Sort.asc("id"))
            .afterKeyset(lname, fname, id);
page = p.page();
size = p.size();
sorts = p.sorts();
mode = p.mode();
cursor = p.cursor();
```

The unfortunate part about the latter is the possibility of someone confusing the page size and the page number because both are just numbers in the parameter list, although I tried to make it more clear which is the size by including `Size` in the method name and placing the size parameter first.

api/src/main/java/jakarta/data/repository/Pagination.java
njr-11: It cannot be because `KeysetPagination` inherits from it.

api/src/main/java/jakarta/data/repository/Pageable.java
njr-11: I update this occurrence and all of the similar places in commit https://github.com/jakartaee/data/pull/68/commits/b636074d94404592644014b165acd07a2827bafc along with the update to add `final`.  I haven't done anything yet about the record idea, but I think we should, but wanted to give it some more thought first.

otaviojava: I guess  "a new instance of <code>Pageable</code>." it is enough.
I mean, how a new instance can be null? 
IMHO it is redundant.

api/src/main/java/jakarta/data/repository/KeysetPageable.java
graemerocher: i mean the implementations could be records

graemerocher (approved):

api/src/main/java/jakarta/data/repository/KeysetPageable.java
keilw: I was also puzzled why KeySetPageable should also be an interface and not a class implementing Pageable or (if Java 17+ was the minimum JDK) a record if you like?

njr-11: > I was also puzzled why KeySetPageable should also be an interface and not a class implementing Pageable or (if Java 17+ was the minimum JDK) a record if you like?

It looks like you opened #72 for a separate discussion of this question, so I'll let this be resolved there.

And I'll leave renaming the fluent methods to #67.

api/src/main/java/jakarta/data/repository/Pageable.java
njr-11: > I guess "a new instance of `Pageable`." it is enough. I mean, how a new instance can be null? IMHO it is redundant.

It seemed redundant to me as well, but I didn't see any harm including the additional statement in the javadoc, so I agreed to do so.  @otaviojava did you see any other issues with this during your review, or are we okay to merge?

otaviojava: IMHO @njr-11 merge it!

PR #69, opened 2022-11-05
Clarify null, cleanup javadoc, declare final where possible

graemerocher:


otaviojava (approved):

api/src/main/java/jakarta/data/repository/Slice.java
```diff
@@ -48,14 +50,14 @@
...
-     * @return the current Pageable
+     * @return the current Pageable; will never be {@literal null}.
      */
     Pageable getPageable();
 
     /**
      * Returns the next {@link Pageable#next()}, or <code>null</code> if it is known that there is no next page.
      *
-     * @return the next pageable
+     * @return the next pageable; will never be {@literal null}.
```
njr-11, line 60: This is incorrect and contradicts two lines up where it explicitly states,
"or <code>null</code> if it is known that there is no next page."

api/src/main/java/jakarta/data/repository/CrudRepository.java
```diff
@@ -84,7 +84,7 @@
      *
      * @param ids must not be {@literal null} nor contain any {@literal null} values.
      * @return guaranteed to be not {@literal null}. The size can be equal or less than the number of given
-     * {@literal ids}.
+     * {@literal ids}.
```
njr-11, line 87: Minor: unnecessary addition of blank space at end of line.

njr-11 (changes requested):
Thanks for tidying up the JavaDoc in all these places!  I noticed that you made the same changes to Pageable JavaDoc about non-null return types that you asked me to make in your review of #68, and this will cause merge conflicts.  Would it make sense to omit those changes from this pull to avoid the conflict?

graemerocher:
thanks for the feedback, will cleanup

otaviojava (approved):

njr-11 (approved):

Issue #72, opened 2022-11-07
Why is Pageable just CursorPageable? [VOTE]


keilw:
**Description**

Looking at this [Pageable](https://github.com/jakartaee/data/blob/main/api/src/main/java/jakarta/data/repository/Pageable.java)
compared to [Micronaut](https://micronaut-projects.github.io/micronaut-data/1.1.2/api/index.html?io/micronaut/data/model/Pageable.html)
and [Spring Data](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Pageable.html)

There is no point in separating `Pageable` and `KeySetPageable`, because the "abstract" class/interface does no abstraction at all, it just deals with `KeySetPageable` and its `Cursor` anyway.
All this API supports is [cursor-based pagination](https://engage.so/blog/a-deep-dive-into-offset-and-cursor-based-pagination-in-mongodb/#what-is-cursor-based-pagination)
While Spring Data and Micronaut (without 3rd party extensions) both rely on
[offset-based pagination](https://engage.so/blog/a-deep-dive-into-offset-and-cursor-based-pagination-in-mongodb/#what-is-offset-based-pagination).

@graemerocher, @odrotbohm Was the offset-based pagination so far practiced in Micronaut or Spring Data a mistake and you would like to force its current users to switch all apps to cursor now, or why is there no offset `Pageable` here yet?
If both can be supported I would have imagined an abstract base `Pageable` with just the page number, size and (unless #54 gets them separated) sorts, but keeping all the `KeySet` stuff in `KeySetPageable`.
Then another (frankly the names would be a bit confusing, hence `CursorPageable` would sound clearer?) `OffsetPageable` class or interface (or record ;-)

That's how `Collection` works, it is inherited by specialized sub-interfaces like `List`, `Queue` or `Set`, but any factory methods like `of()` are in the subsequent subtypes.

Why was this concept broken here?

graemerocher:
offset based pagination is just more commonly used and understood and more easily made into a generic API, we looked at keyset pagination but were unsure how to make it generic so left it up to users to implement

njr-11:
[...]

PR #73, opened 2022-11-07
combine KeysetPageable into single Pageable interface

njr-11:
After reading #72, I realized we could possibly consolidate `KeysetPageable` and `Pageable` into a single `Pageable` interface, with the presence of a `Cursor` distinguishing whether keyset pagination is to be used.  This pull shows what that could look like.  It does seem to be a good way of simplifying the API.  The user no longer needs to think about there being two different pageables or decide which to put into their repository method signature or worry about type casting from one to the other.


api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -142,6 +140,21 @@ public static Pageable ofSize(int maxPageSize) {
     @Override
     public boolean equals(Object o);
 
+    /**
+     * Returns the keyset values which are the starting point for
+     * keyset pagination.
+     *
+     * @return the keyset values; <code>null</code> if using offset pagination.
+     */
+    public Cursor getCursor();
```
otaviojava, line 149: I'm not a big fan of API that returns null. 
We have other options, such as an `Optional` or checking if it makes sense to put Pageable with this responsibility.

otaviojava:
I don't like this approach, especially, because we put a tight coupling between both and there is a risk of Pageable doing too much.

api/src/main/java/jakarta/data/repository/Pageable.java
graemerocher: but all the other parts of the API return `null` and `Optional` hasn't been used anywhere. If we move to `Optional` we should be consistent

graemerocher:
My feeling is the opposite, seems a nice simplification of the API rather than having an explosion of types which can be more confusing

api/src/main/java/jakarta/data/repository/Pageable.java
otaviojava: I agree, if a value might return null it should return an Optional IMHO.

keilw:
Same for me, it goes against the ideas of abstracting the most common aspects I raised in #72.

otaviojava:
I like this point @keilw 
Btw: I saw the methods at `Pageable`

* KeysetPageable afterKeyset(Object... keyset);
* KeysetPageable beforeKeyset(Object... keyset);
* KeysetPageable afterKeysetCursor(Cursor keysetCursor);
* KeysetPageable beforeKeysetCursor(Cursor keysetCursor);

Should we remove it?

njr-11:
> Btw: I saw the methods at `Pageable`
    * KeysetPageable afterKeyset(Object... keyset);
    * KeysetPageable beforeKeyset(Object... keyset);
    * KeysetPageable afterKeysetCursor(Cursor keysetCursor);
    * KeysetPageable beforeKeysetCursor(Cursor keysetCursor);
   Should we remove it?

No, removing those would lose out on core scenarios for keyset pagination: requesting keyset pagination from a particular position.
Instead, it might be possible that refactoring could be done to present the capability in a different way, such as in this pull where they are all switched to return `Pageable`, making them consistent in terms of return type with the other fluent-style methods of the interface.

api/src/main/java/jakarta/data/repository/Pageable.java
njr-11: I'm willing to switch this to `Optional` if that's what everyone wants. Personally I didn't see the point because this method will always return a valid cursor when in the CURSOR_NEXT or CURSOR_PREVIOUS mode, and always `null` when in the OFFSET mode, so no one should ever use the `Optional` for anything besides `Optional.get()` which is just extra code path because you already know what you are getting from the mode.

njr-11: @otaviojava @graemerocher I added `Optional` as requested under https://github.com/jakartaee/data/pull/73/commits/356b656622d775a65ac308e1979560849e4e2127
Please review.
@keilw you indicated after discussion on today's Jakarta Data call that you are okay with this approach now. If you have any other review comments let me know.

njr-11: Sorry for the inconvenience - just after posting the above comment I found this pull had significant merge conflicts with #74 and I needed to rebase onto latest to resolve, which required updating the commit link above.

otaviojava:
@njr-11 could you rebase it, please?

njr-11:
> @njr-11 could you rebase it, please?

Yes, I rebased it onto latest again to resolve merge conflicts with #75

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -230,4 +256,88 @@ static Pageable ofSize(int maxPageSize) {
      * @return a new instance of <code>Pageable</code>. This method never returns <code>null</code>.
      */
     Pageable sortBy(Sort... sorts);
+
+    /**
+     * The type of pagination, which can be offset pagination or
+     * keyset cursor pagination which includes a direction.
+     */
+    enum Mode {
```
otaviojava, line 264: :question: why not move to be a class?
Just curious, because use inner-class here

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -230,4 +256,88 @@ static Pageable ofSize(int maxPageSize) {
...
+         * Offset pagination is used when a cursor is not supplied.
+         */
+        OFFSET
+    }
+
+    /**
+     * Represents keyset values, which can be a starting point for
+     * requesting a next or previous page.
+     */
+    interface Cursor {
```
otaviojava, line 298: The same here.

api/src/main/java/jakarta/data/repository/Pagination.java
```diff
@@ -17,55 +17,65 @@
...
 
-import jakarta.data.repository.KeysetPageable.Cursor;
-
 /**
  * Built-in implementation of Pageable.
  */
-class Pagination implements Pageable {
+final class Pagination implements Pageable {
+
+    private final Optional<Cursor> cursor;
```
otaviojava, line 33:
```suggestion
    private final Cursor cursor;
´´´

https://rules.sonarsource.com/java/RSPEC-3553

File: api/src/main/java/jakarta/data/repository/Pagination.java [hunk 5]
```diff
@@ -118,24 +151,73 @@ public List<Sort> sorts() {
...
     @Override
     public Pageable sortBy(Sort... sorts) {
-        return new Pagination(page, size, sorts == null ? Collections.emptyList() : List.of(sorts));
+        return new Pagination(page, size, sorts == null ? Collections.emptyList() : List.of(sorts), mode, cursor);
+    }
+
+    /**
+     * Built-in implementation of Cursor.
+     */
+    static final class CursorImpl implements Cursor {
```
otaviojava, line 178: It is not meaningful, once it is a default implementation.

```suggestion
static final class DefaultCursor implements Cursor {
```

otaviojava, line 178: The same question to the inner-class.
At this point, we should also have a unit-class for it.

api/src/main/java/jakarta/data/repository/Pageable.java
njr-11: > why not move to be a class?
Just curious, because use inner-class here

Basic common usage of keyset pagination will not require the user to directly interact with `Mode` and `Cursor`, so it's nice to keep these more advanced concepts less prominent as inner classes where they won't overwhelm the user when browsing through the list of API.  A typical user will only need to do:

```
Pageable p1 = Pageable.ofSize(50).sortBy(Sort.asc("id"));
KeysetAwarePage<Product> page1 = products.findBySomething(something, p1);
...
Pageable p2 = p1.afterKeyset(productId); // or page1.nextPageable()
KeysetAwarePage<Product> page2 = products.findBySomething(something, p2);
...
```

api/src/main/java/jakarta/data/repository/Pageable.java
njr-11: same reason as above

api/src/main/java/jakarta/data/repository/Pagination.java
njr-11: > https://rules.sonarsource.com/java/RSPEC-3553

The link that you cited about limiting usage of Optional to return types states,

> using Optional on the input side increases the work you have to do in the method without really increasing the value

which is abundantly clear from the changes I needed to make to get the `Cursor` field switched to an `Optional<Cursor>`. Nonetheless, that is the only way of achieving what you asked for in making `cursor()` return `Optional<Cursor>` on an API that we decided will be made compatible with Java records. With records, the constructor-like record declaration needs to accept the same type as the accessor method returns.  `Optional` just doesn't fit well with records. I never wanted to use `Optional` on `Cursor` in the first place. As stated in my earlier comment on the matter, it makes no sense in terms of usage patterns. And as you've seen here, it makes the API more difficult to write, implement, and even use. Switching to `Optional` has nothing to do with whether or not `KeysetPageable` and `Pageable` are made into a single interface, so the issue should never have been combined into this pull in the first place. I will remove the two commits from this pull that switch ever to `Optional` so that the matter can be considered separately.

api/src/main/java/jakarta/data/repository/Pagination.java
njr-11: Good input, I'll get it renamed, but will need to do so independently of the Commit suggestion due additional usage by the unit test class.

njr-11: I added commit https://github.com/jakartaee/data/pull/73/commits/dc7a6e61ba422de3253ef68cab2cd93eb02297fb to update it as you suggested.  Please review.

api/src/main/java/jakarta/data/repository/Pageable.java
otaviojava: Thanks :)

api/src/main/java/jakarta/data/repository/KeysetCursor.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        return this == o || o != null
+                && o.getClass() == getClass()
+                && Arrays.equals(keyset, ((KeysetCursor) o).keyset);
+    }
+
+    public Object getKeysetElement(int index) {
```
otaviojava, line 50: Should we handle an exception here?
E.g.: When the value is negative or ArrayIndexOutOfBoundsException?

otaviojava (approved):

otaviojava:
I'll merge it, for more fix we can create a new PR.

PR #75, opened 2022-11-09
Use record-style accessors for API

### LINKED ISSUE #70: API consistency and Java 17 future
**Description**

So in general I think we should be planning for a future where Java 17 records are more broadly adopted. 

That means considering accessors for the API (getters `getContent()` vs `content()`). The current API is inconsistent in this aspect. For example the `Limit` class uses record style accessors while `Page`, `Pageable` etc, use getters.

Can we have a vote on what we should use? It seems to me with records we should be considering record style accessors.

graemerocher:
Alters the API to use record-style accessors. To avoid confusion the `page(int)` and `size(int)` methods are renamed `newPage(int)` and `newSize(int)` respectively. Fixes #70

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Pageable.java
+++ b/api/src/main/java/jakarta/data/repository/Pageable.java
@@ -147,21 +147,21 @@ public interface Pageable {
      *
      * @return the page to be returned.
      */
-    long getPage();
+    long page();
 
     /**
      * Returns the requested size of each page
      *
      * @return the requested size of each page
      */
-    int getSize();
+    int size();
 
     /**
      * Return the order collection if it was specified on this <code>Pageable</code>, otherwise an empty list.
      *
-     * @return the order collection
+     * @return the order collection; will never be {@literal null}.
      */
-    List<Sort> getSorts();
+    List<Sort> sorts();
 
     /**
      * Returns the <code>Pageable</code> requesting the next page.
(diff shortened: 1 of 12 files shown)


api/src/main/java/jakarta/data/repository/Sort.java
```diff
@@ -63,7 +63,7 @@ private Sort(String property, Direction direction) {
     /**
```
njr-11, line 63: Should we also replace `Direction direction` with `boolean isAscending` to align with Java records?
```
public record Sort(String property, boolean isAscending) {
    public boolean isDescending() {
        return !isAscending();
    }

    public static Sort asc(String property) {
        return new Sort(property, true);
    }

    public static Sort desc(String property) {
        return new Sort(property, false);
    }
}
```

api/src/main/java/jakarta/data/repository/Pageable.java
```diff
@@ -177,7 +177,7 @@ static Pageable ofSize(int maxPageSize) {
      * @param pageNumber The page number
      * @return a new instance of <code>Pageable</code>. This method never returns <code>null</code>.
      */
-    Pageable page(long pageNumber);
+    Pageable newPage(long pageNumber);
```
njr-11, line 180: Another option that I thought of for the fluent-style configuration could be `.withSize(max)`, `.withPage(num)`, `.withSorts(sorts)` but `.newSize(max)`, `.newPage(num)`, ... that you came up with is shorter.

otaviojava (approved):

otaviojava:
I'll merge it.
If need change we can create another PR.

api/src/main/java/jakarta/data/repository/Pageable.java
otaviojava: It is a good point.
I don't want to spend a lot of time on it, but I like the `with` a prefix.

Issue #71, opened 2022-11-05
Standard Exception Hierarchy?


graemerocher:
**Description**

Currently the exception hierarchy is limited to `DataException`. One of the big value items of Spring/Micronaut Data is a standard access hierarchy for data access failures. In Spring this is:

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/dao/package-frame.html

In Micronaut this is less comprehensive but found here:

https://micronaut-projects.github.io/micronaut-data/latest/api/io/micronaut/data/exceptions/package-summary.html

The value here is that as a user regardless of the underlying database you get a consistent set of exceptions to catch and recover from.

This standard may be an opportunity to standardize this hierarchy.

//cc @odrotbohm

otaviojava:
We put some exceptions to start, if we need more we can open this issue again.

PR #77, opened 2022-11-11
Standard Exception Hierarchy

otaviojava:
# Major Changes

* Add new Exceptions

Ref: https://github.com/jakartaee/data/issues/71

### DIFF
--- a/api/src/main/java/jakarta/data/exceptions/DataAccessResourceFailureException.java
+++ b/api/src/main/java/jakarta/data/exceptions/DataAccessResourceFailureException.java
@@ -15,7 +15,7 @@
  *
  *  SPDX-License-Identifier: Apache-2.0
  */
-package jakarta.data;
+package jakarta.data.exceptions;
 
 /**
  * Exception thrown when the underlying resource fails to connect.
--- a/api/src/main/java/jakarta/data/exceptions/DataAccessResourceFailureException.java
+++ b/api/src/main/java/jakarta/data/exceptions/DataAccessResourceFailureException.java
@@ -18,7 +18,7 @@
 package jakarta.data.exceptions;
 
 /**
- * Exception thrown when the underlying resource fails to connect.
+ * Indicates a connection-related failure that might be intermittent in nature. The operation might succeed if reattempted by the application. If the data access resource is enlisted in a transaction, the application must ensure that the transaction is rolled back before retrying the operation under a new transaction.
  */
 public class DataAccessResourceFailureException extends DataException {
 
--- a/api/src/main/java/jakarta/data/exceptions/NonUniqueResultException.java
+++ b/api/src/main/java/jakarta/data/exceptions/NonUniqueResultException.java
@@ -17,6 +17,12 @@
  */
 package jakarta.data.exceptions;
 
+/**
+ * This exception is raised when execution of a repository method with a singular
+ * return type finds multiple results. To avoid raising an error for this, you can
+ * use the <code>findFirst...</code> method name pattern or supply
+ * <code>Limit.of(1)</code> as a parameter to explicitly request only the first result.
+ */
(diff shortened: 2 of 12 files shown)


graemerocher:
can we group the exceptions in a `exceptions` subpackage?

otaviojava:
@graemerocher done.

api/src/main/java/jakarta/data/exceptions/DataAccessResourceFailureException.java
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.exceptions;
+
+/**
+ * Exception thrown when the underlying resource fails to connect.
+ */
+public class DataAccessResourceFailureException extends DataException {
```
njr-11, line 23: A shorter name could be `DataConnectionException`

api/src/main/java/jakarta/data/exceptions/DataAccessResourceFailureException.java
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.exceptions;
+
+/**
+ * Exception thrown when the underlying resource fails to connect.
```
njr-11, line 21:
```suggestion
 * Indicates a connection-related failure that might be intermittent in nature. The operation might succeed if reattempted by the application. If the data access resource is enlisted in a transaction, the application must ensure that the transaction is rolled back before retrying the operation under a new transaction.
```
We should explain how applications can use this exception, which will also help guide how Jakarta Data providers know when to raise it.  My suggestion above assumes that the intent here is to notify the application of potentially intermittent connection errors that might succeed if retried.

api/src/main/java/jakarta/data/exceptions/NonUniqueResultException.java
```diff
@@ -0,0 +1,48 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.exceptions;
+
+public class NonUniqueResultException extends DataException {
```
njr-11, line 20:
```suggestion
/**
 * This exception is raised when execution of a repository method with a singular
 * return type finds multiple results. To avoid raising an error for this, you can
 * use the <code>findFirst...</code> method name pattern or supply
 * <code>Limit.of(1)</code> as a parameter to explicitly request only the first result.
 */
public class NonUniqueResultException extends DataException {
```

njr-11 (changes requested):

api/src/main/java/jakarta/data/exceptions/DataAccessResourceFailureException.java
graemerocher: this is obviously a use case but the main use case is to avoid the need to catch a database specific exception (`SQLException`, `HibernateException`, `MongoException`) thus coupling your application to the database you are using if you need to introduce exception handling

api/src/main/java/jakarta/data/exceptions/DataAccessResourceFailureException.java
otaviojava: Done

njr-11 (approved):

PR #78, opened 2022-11-14
Add alias method at Pageable

otaviojava:
I'm implementing the API and sort requires to create a ```Sort```

Does it make sense to have an alias?
It would reduce the cognitive load to use the API.

So: 

```java
Pageable pageable = Pageable.ofSize(55).asc("name").desc("age");;
```

Instead of: 
```java
Pageable pageable = Pageable.ofSize(55).sortBy(Sort.asc("name"), Sort.desc("age"));
```

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Pageable.java
+++ b/api/src/main/java/jakarta/data/repository/Pageable.java
@@ -19,6 +19,7 @@ package jakarta.data.repository;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * <p>This class represents pagination information.</p>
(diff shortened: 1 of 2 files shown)


njr-11 (changes requested):
I like how simple this is when there is a single sort criterion.  It gets confusing when there are multiple.  For example, the implementation of the pull differs from the example that is written up in the pull.  The former follows the pattern used by the other fluent style methods of `Pageable` that copy from an existing instance and replace the designated property.  The latter shows a pattern of copying from an existing instance and adding onto the sorts list,

> So:
> 
> ```java
> Pageable pageable = Pageable.ofSize(55).asc("name").desc("age");;
> ```
> 
> Instead of:
> 
> ```java
> Pageable pageable = Pageable.ofSize(55).sortBy(Sort.asc("name"), Sort.desc("age"));
> ```

As implemented the two are not equivalent, and you end up with:
maximum page size: 55
sort criteria: age descending

If we add this, we need to be very explicit in JavaDoc so the user does not make the same mistake. Something like this might work: "The sort criteria of the existing instance are not copied over to the new instance, with the latter only including the single sort criterion that is specified by this method."  But even then, it seems likely many users will not read the JavaDoc and some of those will assume the wrong behavior, causing applications to behave incorrectly, so I'd caution against adding this at all.

otaviojava:
I'll close this PR as a non-go.
Let's see in the future.

PR #79, opened 2022-11-14
Define error for when database cannot do keyset pagination

njr-11:
Specification should define that UnsupportedOperationException be raised when a repository method is requested to perform keyset pagination but the underlying database is incapable of it.


otaviojava (approved):

otaviojava:
It makes sense to me.

Issue #80, opened 2022-11-15
Provider model for plugging in third-party Jakarta Data providers


njr-11:
**Description**

As a:

- [x] Application user/user of the configuration itself
- [ ] API user (application developer)
- [ ] SPI user (container or runtime developer)
- [ ] Specification implementer

...I need to be able to:

use a Jakarta EE product with one or more third-party Jakarta Data providers

...which enables me to:

connect to both Relational databases as well as Column, Document, Graph, and Key-Value (NoSQL) databases in my application(s) via the repository pattern that is defined by Jakarta Data.

This issue is for standardizing a provider model that a Jakarta EE product (such as web or core profile) can use to integrate with third-party providers of Jakarta Data.  This issue will not attempt to deal with configuration, which is also important but is not being designed under this issue.

otaviojava:
IMHO: we should keep this option for each vendor.

E.g., there are more than four types that Jakarta NoSQL support, such as time series. Furthermore, we have solutions such as Microstream or rest applications.

njr-11:
This was done with CDI extensions instead.

PR #81, opened 2022-11-15
data provider model

njr-11:
proposal for #80


otaviojava:
:warning:  Use a String instead of an enum

graemerocher:
Hmm this kind of assumes runtime creation of repositories which is not the case for Micronaut

njr-11:
> Hmm this kind of assumes runtime creation of repositories which is not the case for Micronaut

The intent was only to provide of way of getting the repository instances from the Jakarta Data provider.  I can rename the `createRepository` method to `getRepository` and check to ensure JavaDoc uses terms like obtain instead of create. Would that help?  Or is more needed?

njr-11:
> From a top-level perspective, a few bits are missing regarding serviceability. How should a repository factory choose which data source/driver configuration to use? A repository interface can oftentimes be used with different backing technologies as the domain model can be represented in a relational/document/column-database.
> 
> A design for a `DataProvider` makes sense, but it should have intermediate steps to obtain a `RepositoryFactory` for a specific connection or database technology.

These are very good points which I agree with although was hesitant to include config at this point when config hasn't yet been defined for Jakarta Data.  But I can certainly update it to add `RepositoryFactory` along with at least some marker to represent that we will need to include config to identify the correct database and so forth.

graemerocher:
> > Hmm this kind of assumes runtime creation of repositories which is not the case for Micronaut
> 
> The intent was only to provide of way of getting the repository instances from the Jakarta Data provider. I can rename the `createRepository` method to `getRepository` and check to ensure JavaDoc uses terms like obtain instead of create. Would that help? Or is more needed?

Using `getRepository` makes more sense yes

graemerocher:
I'm wondering myself why this interface is needed at all. Can we not just specify that the DI provider will provide the implementation (CDI, Spring, Micronaut etc.)

njr-11:
> I'm wondering myself why this interface is needed at all. Can we not just specify that the DI provider will provide the implementation (CDI, Spring, Micronaut etc.)

Forms of dependency injection outside of Jakarta EE should not be impacted by this.
The intent of the provider mechanism is to allow for the possibility of multiple different Jakarta Data providers supplying repositories for an application. For example, Micronaut providing repositories for relational databases and a Jakarta NoSQL implementation providing repositories for NoSQL databases.  It might be possible with additional annotations for disambiguation to push the responsibility for CDI dependency injection onto the Jakarta Data providers, but I suspect that would have the side effect of requiring the additional annotations even when there is only one Jakarta Data provider present, which would make the simplest path more complicated for end users.

graemerocher:
sure but I am unsure how this is supposed to work in practice since in the case of Micronaut and Spring you need a bean container (`ApplicationContext`) in order to retrieve implementations. So would the provider implementation in this case bootstrap the whole bean context? Normally this is initialized separately. 

And if so this provider interface would need a way to shutdown the whole thing as well (implement `AutoCloseable` etc.)

njr-11:
I added a [commit](https://github.com/jakartaee/data/pull/81/commits/78eb388f30eecd421ccca27059d264b2f935f88e) to address all but one of the review comments. The suggested `RepositoryFactory` class from one review comment was added with the name of `RepositoryProvider` to address another review comment to avoid giving the impression that `Repository` instances must be dynamically created at run time. The following review comment is not addressed:

> sure but I am unsure how this is supposed to work in practice since in the case of Micronaut and Spring you need a bean container (`ApplicationContext`) in order to retrieve implementations. So would the provider implementation in this case bootstrap the whole bean context? Normally this is initialized separately.
> 
> And if so this provider interface would need a way to shutdown the whole thing as well (implement `AutoCloseable` etc.)

As an outsider without any existing repository-based product of my own and not knowing what are requirements of existing implementations vs how these products chose to externalize their API, there is only so much I'll be able to figure out with a design for this.  It might be better if Micronaut or Spring Data makes a proposal here instead that would fit with existing implementations.

graemerocher:
@mp911de what do you think? Personally this interface doesn't make much sense to me. How would an interface loaded via serviceloader obtain instance to the Spring context to resolve repository implementations for example? Using static state?

mp911de:
I'm also torn on how usable this is for the time being. If things are specified, then we need to stick to them. If we leave intentional gaps, we can figure out later who needs this factory and whether it is a problem to be solved at all.

We already have mechanisms to spin up repositories within the individual frameworks (Micrometer, Spring, Quarkus), so that isn't something to be solved for now.

graemerocher:
Agreed, which brings me back to my original position is that the specification should simply say that Jakarta Data should be paired with a dependency injection framework (CDI/Spring/Micronaut) for obtaining repository instances.

otaviojava:
Agreed, maybe hold or close this PR for a while.

njr-11:
> If we leave intentional gaps, we can figure out later who needs this factory and whether it is a problem to be solved at all.

It's already needed. I starting investigating this path because there has been discussion at the last 2 Jakarta Data meetings of putting out a Milestone 1 of Jakarta Data against which implementations could put out Alpha versions.  I don't work on a Jakarta Data provider product, but I do work on a Jakarta EE app server and I'm trying to prepare it to integrate these Jakarta Data providers. As soon as there is more than one Jakarta Data provider, the current model with CDI dependency injection is broken/incomplete.  You can't have 2 different providers handling the same CDI injection points.  I can solve this problem by having the Jakarta EE product be the single place to handle the CDI dependency injection and delegate to the appropriate Jakarta Data provider and had opened this pull to proposed standardizing that model.  If others don't like this approach, that's fine - we can get rid of this pull.  However, I do not agree with deferring a solution to the problem or choosing to not solve it at all.  I already need a solution to it in order to begin integrating with the alpha versions of Jakarta Data providers that all of you will be creating.

graemerocher:
a solution needs to be considered carefully in that case, the current proposal via service loader requires the use of static state to be stored somewhere if the repositories have to be resolved from a bean container. That would be an unfortunate requirement.

njr-11:
> the current proposal via service loader requires the use of static state to be stored somewhere if the repositories have to be resolved from a bean container. That would be an unfortunate requirement.

Are there any additional parameters that could be supplied to a `getRepository` request that would give you the state that you need?  Looking around at what's available during CDI dependency injection, it looks like there would be `jakarta.enterprise.context.spi.CreationalContext` and `jakarta.enterprise.inject.spi.BeanManager` / `jakarta.enterprise.inject.spi.BeanContainer` and also awareness of which application and even the `jakarta.enterprise.inject.spi.AnnotatedType` for the specific injection point for which the repository is being requested.  Would any of that help you to avoid static state?

graemerocher:
Those are specific to CDI

njr-11:
> Those are specific to CDI

Yes, they are. It should be fine to have a `getRepository` method signature that is specifically for the CDI dependency injection path if that would help give provider implementations the state that they need. (I still don't know if it will though)
Regarding vendor-specific dependency injection, the owners of those dependency injection models would need to decide if they even want to allow other Jakarta Data providers in their vendor-specific dependency injection models or only their own, and if so, assess whether the same issues that come up with multiple providers in CDI would apply to those models as well.

graemerocher:
as long as the spec doesn't directly depend on or require CDI and pull it in transitively sure. Spring and Micronaut don't implement CDI currently. A partial implementation of CDI exists for Micronaut that is not complete yet but I doubt Spring will ever implement CDI @mp911de can correct me if I am wrong.

njr-11:
> Spring and Micronaut don't implement CDI currently. A partial implementation of CDI exists for Micronaut that is not complete yet but I doubt Spring will ever implement CDI @mp911de can correct me if I am wrong.

This is completely fine, and actually provides another good reason behind having this ability to obtain repository instances directly from the provider -- because it enables Jakarta EE products to handle the CDI integration for you so that you won't ever need to.

From this, I assume that my suggestion of supplying additional parameters to help you identify the state will be useless.  But if we restrict the use of the `getRepository` method to only during CDI dependency injection when the other dependency injection models aren't involved, do the issues you raised with needing to know the state still apply?  Could the approach work as written if limited to CDI paths?

graemerocher:
maybe @tomas-langer can chip in

otaviojava:
> Forms of dependency injection outside of Jakarta EE should not be impacted by this.
The intent of the provider mechanism is to allow for the possibility of multiple different Jakarta Data providers supplying repositories for an application. For example, Micronaut providing repositories for relational databases and a Jakarta NoSQL implementation providing repositories for NoSQL databases. It might be possible with additional annotations for disambiguation to push the responsibility for CDI dependency injection onto the Jakarta Data providers, but I suspect that would have the side effect of requiring the additional annotations even when there is only one Jakarta Data provider present, which would make the simplest path more complicated for end users.

IMHO: It should not belong to the API, but to the implementation itself. E.g.: Jakarta NoSQL can use CDI extension, Quarkus, Micronaut and Spring to different way.

njr-11:
Rebased onto latest and added a commit to clarify that the DataProvider is only needed for CDI dependency injection and not for vendor-specific dependency injection.

otaviojava:
@njr-11 as we discussed, on CDI, we can do a similar thing that Spring, Micronaut does with **CDI Qualifier**.

At Eclipse JNoSQL, we provide the `Database` annotation where the user can define and configure the injection and where this data comes from:

```java
@Inject
@Database(value = DatabaseType.KEY_VALUE, provider = "databaseA")
private KeyValueTemplate template;
````

The producer:

```java
// producers methods
@Produces
@Database(value = DatabaseType.KEY_VALUE, provider = "databaseA")
public BucketManager getManagerA() {
    BucketManager manager = // instance;
    return manager;
}

```

I like this approach primarily because it keeps the scope of the specification and doesn't break this responsibility. 
IMHO: Jakarta Data should not be responsible for creating and destroying beans; it should go to the DI engine.

njr-11:
We talked about this on the call as well. We shouldn't burden the user with a requirement for extra qualifier information regardless of whether there are multiple providers or just one. The current pull needs to be completely redone. Some of the current providers don't want to be CDI based and at this point we can't assume anything about whether Jakarta Config will be available.

otaviojava:
Everything that we will do has trade-offs.

Given there is no perfect solution which one has less cognitive load?

@njr-11 let's enumerate the trade-off here:

## An annotation outside the spec

Include an annotation even outside the spec and be ready for use, e.g.:

```java
@VendorQualifier
@Inject 
private PersonRepository repository;
```
### Trade-offs

**Pro**
* Simplicity
* We keep the focus on the spec
* We can increase it later with more maturity


**Const** 
* It is not on the spec
* Each spec needs to document as it

## RepositoryProvider

We need to inject a provider, then put the parameters to use finally:

```java
@Inject
RepositoryProvider provider;

PersonRepository repository = provider.getRepository(PersonRepository.class, Person.class, "provider");
....


### Trade-offs

**Pro**
* It belongs to the API
* The user has an option inside the spec


**Const** 

* Complex when it is required
* We get out of the focus on the Spec:  We're creating and destroying bean which belongs to CDI on the Jakarta EE world
* We're not sure how often the user needs this for now

njr-11:
No one has the suggested the kind of RepositoryProvider pattern you are comparing with. RepositoryProvider, which I just stated needs to be reworked in light of discussion from the previous call, has not ever been proposed as something the user would interact with, nor does the spec have any proposal to prevent vendors from introducing their own qualifiers and using them in whatever ways they see fit. Lists of Pros and Cons will have a time and place, but the idea in need of comparison has not been sufficiently put forward for that to happen yet.

njr-11:
Closing this pull to avoid confusion. When this is reattempted per more recent discussion, I'll do so in a separate pull.

otaviojava:
My point is also if we need to the first release. How can we create something more straightforward than a single Qualifier annotation by a vendor?

IMHO: I believe we can live without it.

PR #84, opened 2022-11-17
Inconsistent Pageable factory methods

otaviojava:
Signed-off-by: Otavio Santana <otaviopolianasantana@gmail.com>

Update API to use:

```java
Pageable pageable = Pageable.size(50)
            .page(5)
            .sort(Sort.asc("lastName"), Sort.asc("firstName), Sort.asc("id"))
            .afterKeyset(lname, fname, id);
```

I checked in `Pageable` and `Limit` both use the `of` as a prefix on the method factory on the majority.

Ref: https://github.com/jakartaee/data/issues/67

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Pageable.java
+++ b/api/src/main/java/jakarta/data/repository/Pageable.java
@@ -199,7 +199,7 @@ public interface Pageable {
      * @param pageNumber The page number
      * @return a new instance of <code>Pageable</code>. This method never returns <code>null</code>.
      */
-    Pageable newPage(long pageNumber);
+    Pageable page(long pageNumber);
 
     /**
      * <p>Creates a new <code>Pageable</code> instance representing the same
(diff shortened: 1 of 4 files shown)


mp911de:
That's an actual improvement as the method names read more nicely. The only issue is that both methods make sense as `static` and instance methods, but that's a decision that has to be made.

graemerocher (approved):

njr-11 (approved):
To be clear, this pull does not actually achieve the following, which is impossible due to the incompatibility between the static and fluent style methods having otherwise identical signatures,

> Update API to use:
> 
> ```java
> Pageable pageable = Pageable.size(50)
>             .page(5)
>             .sort(Sort.asc("lastName"), Sort.asc("firstName), Sort.asc("id"))
>             .afterKeyset(lname, fname, id);
> ```

Instead, it does this:

```java
Pageable pageable = Pageable.ofSize(50)
            .page(5)
            .sortBy(Sort.asc("lastName"), Sort.asc("firstName), Sort.asc("id"))
            .afterKeyset(lname, fname, id);
```

which seems like a nice improvement to the naming, so I'm adding my approval of the changes.

otaviojava:
Thanks @njr-11

PR #93, opened 2022-11-22
name-pattern keywords for distinct, aggregates

njr-11:
Looking over Micronaut documentation for repository methods, you can use patterns with aggregates and distinct such as: findMaxPriceBy..., findMinPriceBy..., findSumAgeBy..., findAvgPriceBy,  countDistinctBy... which correspond to functions available in JPQL.

This pull gets these things added to our repository keywords documentation along with examples.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Repository.java
+++ b/api/src/main/java/jakarta/data/repository/Repository.java
@@ -96,7 +96,7 @@ import java.lang.annotation.Target;
  * <td style="vertical-align: top"><b>Example</b></td>
  * </tr>
  *
- * <tr style="vertical-align: top"><td><code>countBy</code></td>
+ * <tr style="vertical-align: top"><td><code>count...By</code></td>
  * <td>counts the number of entities</td>
  * <td><code>countByAgeGreaterThanEqual(ageLimit)</code></td></tr>
  *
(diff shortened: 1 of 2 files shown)


mp911de:
I'm a bit hesitant as aggregations are not in line with entities or CRUD operations. While this seems simple with JPQL, other stores require either an entirely different approach (e.g. MongoDB with aggregations) that results in potentially entirely different underlying behavior or stores that do not support aggregations at all (e.g. Cassandra, Redis).

From a user perspective, it is difficult to know whether a particular feature/keyword is supported or not as the generic repository API fails to explain this aspect.

I'm also not sure that `DISTINCT` was the best thing we introduced for Spring Data repositories.

otaviojava:
I agree with @mp911de

graemerocher:
This is one of the reasons we use `@Repository` primarily as a stereotype annotation and then have specific annotations per datastore like `@MongoRepository`, `@JdbcRepository` etc. since we can fail compilation for unsupported features in those datastores.

njr-11:
This pull has been updated to remove case insensitive comparisons, which is being moved to another pull so it can be considered separately, leaving only distinct and aggregates within this current pull.

njr-11:
Based on above comments, this pull is unwanted, so I'm closing it out without merging it.

PR #96, opened 2022-11-29
repository keywords for boolean conditions and Null

njr-11:
[Micronaut](https://micronaut-projects.github.io/micronaut-data/latest/guide/) and [Spring Data](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/) have repository keywords for boolean true/false conditions and checking for null.  Jakarta Data doesn't have these documented yet.  This pull adds them.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Repository.java
+++ b/api/src/main/java/jakarta/data/repository/Repository.java
@@ -172,7 +172,7 @@ import java.lang.annotation.Target;
  * <tr style="vertical-align: top"><td><code>False</code></td>
  * <td>boolean</td>
  * <td>Requires that the entity's attribute value has a boolean value of false.</td>
- * <td><code>findByIsCanceledFalseAndDurationLessThan(max)</code></td></tr>
+ * <td><code>findByIsCanceledFalse()</code></td></tr>
  *
  * <tr style="vertical-align: top"><td><code>First</code></td>
  * <td>find...By</td>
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/data/repository/Repository.java
```diff
@@ -169,6 +169,11 @@
  * match the parameter value, which can be a pattern.</td>
  * <td><code>findByNameEndsWith(surname)</code></td></tr>
  *
+ * <tr style="vertical-align: top"><td><code>False</code></td>
+ * <td>boolean</td>
+ * <td>Requires that the entity's attribute value has a boolean value of false.</td>
+ * <td><code>findByIsCanceledFalseAndDurationLessThan(max)</code></td></tr>
```
mp911de, line 175: The sample style is confusing. Let's have only `findByCanceledFalse`. `AndDurationLessThan` creates quite some noise. Also, `IsCanceled` would apply if the property is named `isCanceled` (`private boolean isCanceled`).

Same comment applies for the `True` and `null` samples.

njr-11: I missed the rename from `IsCanceled` to `Canceled`, so there's also this commit https://github.com/jakartaee/data/pull/96/commits/aad44dfa048f1b889ddf709134689b1485a9ab93 to cover that.

otaviojava (approved):

PR #99, opened 2022-12-06
Case insensitive comparisons

njr-11:
Existing repository implementations have repository keywords ([Micronaut](https://micronaut-projects.github.io/micronaut-data/latest/guide/), [Spring](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/)) and other [methods](https://docs.spring.io/spring-data/commons/docs/1.9.1.RELEASE/api/org/springframework/data/domain/Sort.Order.html#ignoreCase--) to indicate case insensitive comparisons for string properties.  These are useful so this pull adds to Jakarta Data.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/OrderBy.java
+++ b/api/src/main/java/jakarta/data/repository/OrderBy.java
@@ -71,6 +71,18 @@ public @interface OrderBy {
     boolean descending() default false;
 
     /**
+     * <p>Indicates whether or not to request case insensitive ordering
+     * from a database with case sensitive collation.
+     * A database with case insensitive collation performs case insensitive
+     * ordering regardless of the requested <code>ignoreCase</code> value.</p>
+     *
+     * <p>The default value is <code>false</code>.</p>
+     *
+     * @return whether or not to request case insensitive sorting for the property.
+     */
+    boolean ignoreCase() default false;
+
+    /**
      * <p>Entity attribute name to sort by.</p>
      *
      * <p>For example,</p>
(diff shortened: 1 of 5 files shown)


graemerocher (approved):

njr-11:
> Do we have a guarantee about this query beyond the SQL databases?

@otaviojava - No, but the section under which I listed it already covers this by stating, "some keywords listed here might not be supported in a particular store" (from #16). At some point we should go back and be more explicit about which keywords need to work on every database and which will vary by database. If we add an extension repository as you point out, that would be a natural place to document that.

I do think it's important to enable core capability beyond what is common to every database so that we do not restrict users so much that they are stuck with existing vendor-specific API and don't adopt Jakarta Data. By standardizing these keywords now, we also reserve them for future use by any NoSQL database that later adds the function, so NoSQL databases can benefit, too.

otaviojava:
@njr-11 maybe we should adjust a methodology around what we should or should not put as a standard API.

Such as at least three persistence types.


E.g.:
**LT**, **LTE**, **GT**, and **GTE** all work with `SQL`, `Wide-column`, `Graph`, and `Document`, but it does not work with Key-value.

If a query only works with a type, it does not make sense to include it, such as an **edge property** that makes sense only on the Graph database.

I am afraid that if we start to put specific keywords to each type, we'll have a confusing API.


https://db-engines.com/en/ranking

That is why specialization might work better than all the possibilities types here.

njr-11:
@otaviojava let's open a separate issue to cover that.

njr-11:
I've opened #100 to cover the general issue of differences in availability of repository keywords by database type, referencing back to Otavio's comments in this pull.  @otaviojava are you okay with merging this now?

otaviojava:
Hey @njr-11 
IMHO: No, we need to wait for a Yes or not on this Issue: 

https://github.com/jakartaee/data/issues/100

If we solve with "Yes" we merged if "no" we should not merge it.

otaviojava (approved):

PR #101, opened 2022-12-09
Switch to int for Limit.maxResults

njr-11:
When Limit was added, usage of the `long` data type was copied from Pageable, which has since been switched to use `int`.  Limit.maxResults should similarly be switched.  Result sizes of that magnitude aren't even supported by JPA, wouldn't be compatible or work properly for some return types, and would defeat the purpose of using pagination and limits in the first place.  When users expect that many results, they should instead be reading results in pages (preferably with keysets).  I wasn't sure about startAt/endAt - it seems like those could arguably go either way.  If `long`, then the difference between them could imply a maxResults that exceeds Integer.MAX_VALUE, so I went ahead and switched those as well.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Limit.java
+++ b/api/src/main/java/jakarta/data/repository/Limit.java
@@ -47,11 +47,11 @@ package jakarta.data.repository;
  * </ul>
  */
 public final class Limit {
-    private static final long DEFAULT_START_AT = 1L;
-    private final long maxResults;
-    private final long startAt;
+    private static final int DEFAULT_START_AT = 1;
+    private final int maxResults;
+    private final int startAt;
 
-    private Limit(long maxResults, long startAt) {
+    private Limit(int maxResults, int startAt) {
         this.maxResults = maxResults;
         this.startAt = startAt;
     }
@@ -62,7 +62,7 @@ public final class Limit {
      *
      * @return maximum number of results for a query.
      */
-    public long maxResults() {
+    public int maxResults() {
         return maxResults;
     }
 
@@ -72,7 +72,7 @@ public final class Limit {
      *
      * @return offset of the first result.
      */
-    public long startAt() {
+    public int startAt() {
         return startAt;
     }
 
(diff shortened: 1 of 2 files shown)


mp911de:
Fine for `maxResults` as we cannot represent arrays with a greater size than `Integer.MAX_VALUE`. That would work for Streams but still take a while. `startAt` should remain `long`. While some users choose to use JPA, there are many other stores out there that support larger offsets.

njr-11:
> Fine for `maxResults` as we cannot represent arrays with a greater size than `Integer.MAX_VALUE`. That would work for Streams but still take a while. `startAt` should remain `long`. While some users choose to use JPA, there are many other stores out there that support larger offsets.

@mp911de Thanks - that makes sense. I added a commit https://github.com/jakartaee/data/pull/101/commits/83e623400281a400ddf79e44cccd6e2e24ab4e6b to put `startAt` and `endAt` back to `long`.

api/src/main/java/jakarta/data/repository/Limit.java
```diff
@@ -113,6 +115,9 @@ public static Limit range(long startAt, long endAt) {
         if (endAt < startAt)
             throw new IllegalArgumentException("startAt: " + startAt + ", endAt: " + endAt);
 
-        return new Limit(endAt - startAt + 1, startAt);
+        if (endAt - startAt >= Integer.MAX_VALUE)
```
otaviojava, line 118: Does make sense to use a long and have a MAX value on Integer?

njr-11: > Does it make sense to use a long and have a MAX value on an Integer?

The integer maximum is on the range specified between the startAt/endAt, not on the startAt and endAt values themselves.  It does make sense to constrain the maximum range for the reasons discussed earlier in this pull -- anyone needing to return that many results ought to be using pagination, not getting them all at once with a Limit.  Allowing startAt and endAt to be long values might be useful for large data sets, although [JPA](https://jakarta.ee/specifications/persistence/3.0/apidocs/jakarta.persistence/jakarta/persistence/query#setFirstResult(int)) will only allow int.

otaviojava (approved):

Issue #100, opened 2022-12-09
Ability to use some repository keywords will vary based on support from the database


njr-11:
This discussion started under #99 (see comments from @otaviojava) but is a much broader topic that isn't specific to that single pull.  Several repository keywords, including GreaterThan, GreaterThanEqual, LessThan, LessThanEqual, IgnoreCase, Between, Or? ... will be available from some databases but not others.  We should improve the specification and API in ways that are most helpful in conveying this information so that it is clear to users which keywords to rely on for the databases or types of databases used.

otaviojava:
Let's take a step back on the Jakarta Data proposal. The main goal is to be data agnostic instead of looking only at a specific database type, such as a relational database.

Said that the Repository interface should have a more considerable cohesion around database types instead of going to a database type.

We have almost 400 database solutions in the market, and it would take a lot of work to keep and add support to those keyword databases.

I suggest exploring OOP in our favor. Therefore, the extensions like the market do with Spring and Jakarta NoSQL. Thus, CassandraRepository, MongoRepository, and so on.

So, allowing it to JPA creates its repository to put that specific behavior is a good solution to the API moving those to this specialization:

* https://github.com/jakartaee/data/pull/96
* https://github.com/jakartaee/data/pull/97
* https://github.com/jakartaee/data/pull/93
* https://github.com/jakartaee/data/issues/100

PR #102, opened 2022-12-12
specialization of repository types for supported reserved keywords

njr-11:
@otaviojava Here's my attempt at the specialization approach that I think you're asking for under #100.
I didn't go as far as enumerating the exact lists of reserved keywords for the core set and the additional keywords that are specific to SQL/JPQL/relational under this pull because for some I would be guessing, but I did include a place to put each, which can be done in a follow-on pull once the approach looks good.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/DataRepository.java
+++ b/api/src/main/java/jakarta/data/repository/DataRepository.java
@@ -17,6 +17,8 @@
  */
 package jakarta.data.repository;
 
+import java.util.Set;
+
 /**
  * Parent repository interface for all repositories.
  * Central repository marker interface. Captures the domain type to manage and the domain type's id type.
@@ -25,5 +27,27 @@ package jakarta.data.repository;
  */
 public interface DataRepository<T, K> {
 
-
+    /**
+     * <p>Returns the set of reserved keywords that are supported by the
+     * database to which the repository connects. Reserved keywords begin
+     * with a capital letter as they appear within a method name
+     * and as defined by the Jakarta Data specification in its list of
+     * {@link Repository Reserved Keywords}. This set also includes the
+     * Jakarta Data provider's vendor-specific reserved keywords (if any)
+     * that are supported by the database.
+     * This set does not include reserved method name prefixes,
+     * such as <code>find...By</code> and <code>deleteBy</code>.</p>
+     *
+     * <p>The following reserved keywords must be supported for all
+     * databases:</p>
+     *
+     * <ul>
+     * <li><code>And</code></li>
+     * <li>TODO specify the remainder of this list in a separate pull once the exact list is determined</li>
+     * </ul>
+     *
+     * @return reserved keywords for repository method names that are
+     *         supported by the database.
+     */
+    Set<String> supportedKeywords();
 }
--- a/api/src/main/java/jakarta/data/repository/SQLRepository.java
+++ b/api/src/main/java/jakarta/data/repository/SQLRepository.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.repository;
+
+/**
+ * <p>A repository for entities that are stored in a relational database
+ * with capability that is consistent with SQL and JPQL.</p>
+ *
+ * <p>The {@link #supportedKeywords() supportedKeywords} method return value
+ * of a repository that implements {@code SQLRepository} must include the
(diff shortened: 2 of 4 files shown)


api/src/main/java/jakarta/data/repository/DataRepository.java
```diff
@@ -25,5 +27,27 @@
...
+     *
+     * <ul>
+     * <li><code>And</code></li>
+     * <li>TODO specify the remainder of this list in a separate pull once the exact list is determined</li>
+     * </ul>
+     *
+     * @return reserved keywords for repository method names that are
+     *         supported by the database.
+     */
+    Set<String> supportedKeywords();
```
otaviojava, line 52: When we talk about Repositories, we primarily put the leading reference on DDD and the repository pattern.

DDD, a repository, is an object that participates in the domain but abstracts storage and infrastructure details away.

When we put a method like this, I don't believe we're putting the tech details away.

api/src/main/java/jakarta/data/repository/SQLRepository.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * <ul>
+ * <li><code>IgnoreCase</code></li>
+ * <li>TODO specify the remainder of this list in a separate pull once the exact list is determined</li>
+ * </ul>
+ *
+ * @param <T> type of entity managed by the repository.
+ * @param <K> type of key that is the unique identifier for the entity.
+ */
+public interface SQLRepository<T, K> extends DataRepository<T, K> {
```
otaviojava, line 36: I like the idea, but once it is not an agnostic data solution, it does not make sense to stay here.
I believe there are two solutions for it:

- Create a repository call "data-extensions" where we put ideas and particular behavior
- Let's the vendor handle that, so Jakarta NoSQL we'll have an extension, Spring Data and JPA team as well.

njr-11: > I like the idea, but once it is not an agnostic data solution, it does not make sense to stay here. I believe there are two solutions for it:
> 
>     * Create a repository call "data-extensions" where we put ideas and particular behavior
>     * Let's the vendor handle that, so Jakarta NoSQL we'll have an extension, Spring Data and JPA team as well.

Although I disagree with your view that reserved keywords shouldn't apply to some types of databases and not others, I have still been trying to figure out approaches to accommodate that viewpoint.  However, creating a new "data-extensions" git repository to separate out everything but the least common denominator is going way too far.

We need a vote on this, and I can volunteer to get one started, but for everyone to understand what they are actually voting on, we need to first collect the information on the exact set of keywords that are common across every database type versus the set of keywords for basic capability users will expect that isn't common and would be moved to a data-extensions git repo.

Based on prior discussions I understand that the following are not common:
Between, Empty, False, GreaterThan, GreaterThanEqual, IgnoreCase, LessThan, LessThanEqual, Null, Or, True

What about the remaining keywords? Which of these are common to all databases vs not?
And, Asc, Contains, Desc, EndsWith, First, Like, In, Not, OrderBy, StartsWith

otaviojava: Those that already are on the docs are the common ones:
https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/repository.asciidoc

otaviojava: I'm talking about the new features coming:

- https://github.com/jakartaee/data/pull/93
- https://github.com/jakartaee/data/pull/96
- https://github.com/jakartaee/data/pull/97
- https://github.com/jakartaee/data/pull/99

 what we have on the repository is fine!

njr-11: > Those that already are on the docs are the common ones: https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/repository.asciidoc

I'm confused.  Previous comments have stated that some of these are not common to all databases.

From #99:
> E.g.: **LT**, **LTE**, **GT**, and **GTE** all work with `SQL`, `Wide-column`, `Graph`, and `Document`, but it does not work with Key-value.

From #39:
> e.g., Cassandra does not have the `OR` indicator: https://cassandra.apache.org/doc/latest/cassandra/cql/dml.html#select-statement

otaviojava: Sorry, miscommunication on my side.

The goal is not to have support for all databases. 
It is impossible!!
However, to have at least a minimum amount.

Such as, if three databases are provided, it would be acceptable to put them as standard.


My point is to find this number.

One is too small and insufficient to a standard, but all of those are impossible.


We can start with a number magic: three types and see what is happening.

njr-11: Deciding based on a certain threshold of database types seems arbitrary. It doesn't change that users must still cope with the situation that some of our existing keywords apply only to certain databases. We need to address that regardless. If we have established that it is okay to include keywords like `GreaterThan` and `Or` in the specification despite some databases not having the capability, we should give the same treatment to additional keywords if any are brought in from Spring and Micronaut (or elsewhere) that apply to a subset of database types.  I don't say that we should bring them in without question. For each individually, we should consider if it is indeed useful for end users and valuable to standardize, and if so, then treat it as we would the other keywords that do not apply to all database types.

njr-11: Yes, let's discuss at the next meeting.
I'm well aware that keywords we don't standardize are not forbidden for Jakarta Data providers to add on their own.  However, the value of a specification is in standardization and alignment across multiple vendors for the benefit of the end user.  It is valuable to have the standard that enforces that when additional vendors add these capabilities, they will do so with the exact same keywords and patterns that are defined by the standard and compatible with other vendors rather than creating something completely different.  Standardization is also valuable for end users with requirements imposed by their employer to only code to standards.

PR #103, opened 2022-12-26
Release version 1.0.0-b1

otaviojava:
This PR creates a release version to 1.0.0-b1.

As we discussed, the first release primarily aims to receive feedback from the community and see some implementations/vendor feedback.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -6,16 +6,18 @@ All notable changes to this project will be documented in this file.
 The format is based on https://keepachangelog.com/en/1.0.0/[Keep a Changelog]
 and this project adheres to https://semver.org/spec/v2.0.0.html[Semantic Versioning].
 
-== 1.0.0
+== [Unreleased]
 
-=== Changed
+== [1.0.0-b1] - 2022-12-27
+
+
+=== Added
 
 * jakarta-data-api
-    * Removed `hamcrest-all` in favour of `assertj-core`
-    * Changed the assertions in `PageableTest` and `SortTest`
-    * Removed redundant tests on `SortTest`
+** Removed `hamcrest-all` in favour of `assertj-core`
+** Changed the assertions in `PageableTest` and `SortTest`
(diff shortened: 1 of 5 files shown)


mpredli01:
Hi @otaviojava:

Everything looks good to me!  Let's see what the Java community has to say.

Mike.

otaviojava:
@mpredli01 could you approve then?

pzygielo:
So - there is a plan to merge this to master and keep the fixed version there instead of next SNAPSHOT? Shouldn't this version be rather set by build job instead?
It looks that https://ci.eclipse.org/data/job/jakarta-data-deploy/ is at least timer-based and it deploys current `main`. It's 1.0.0-SNAPSHOT so far. If this PR gets merged as-is, next deployments will happen for 1.0.0-b1 version.

It doesn't seem right to me.

mpredli01 (approved):
Approved!

njr-11 (commented):
In git email notifications, I see that @pzygielo raised a concern about versioning in this pull, but I can't see it anywhere on the pull itself.  Other than that, having a beta/alpha at this point seems like a good idea.

otaviojava:
I agree with @njr-11 and @pzygielo 
We'll move it forward, but it would be fantastic if we fix it to the following versions.

pzygielo:
Is there a plan to deal with failures https://ci.eclipse.org/data/job/jakarta-data-deploy/32/console?

otaviojava:
Hey @pzygielo yes.

We cannot publish twice a non-snapshot version on Maven.

We have this issue with releasing and then returning to snapshot after the release:

https://github.com/jakartaee/data/issues/104

pzygielo:
> We cannot publish twice a non-snapshot version on Maven.

I know, I've already told that here 5 days ago.

> We have this issue with releasing and then returning to snapshot after the release:
> - #104

I believe setting version back to 1.0.0-SNAPSHOT could be done manually, couldn't it?

Issue #104, opened 2022-12-29
Create a an automation process release on Jakarta Data


otaviojava:
Today, we released our first alfa version. However, we did it manually.

Could we improve this process using any kind of automation process, such as Jenkins, or Github action, where we can update a version, deploy it, and then increase it?

otaviojava:
We the link we have: https://wiki.eclipse.org/JakartaEE_New_Infra_Release_Job

KyleAure:
@otaviojava I believe there is typically a common set of workflows for Jakarta Specs in [Jenkins](https://ci.eclipse.org): 

- Build and stage: 
  - Builds and stages the parent POM, API jar, and TCK jar to the Jakarta stagging repo: https://jakarta.oss.sonatype.org/content/repositories/staging/.  
  - Builds and stages the TCK distribution zip to the Eclipse downloads page: https://download.eclipse.org
- Stage to release: Releases a staged POM, API jar, and TCK jar to maven central: https://repo1.maven.org/maven2/jakarta/
- Stage to promotion: Releases a staged TCK distribution zip to the Eclipse downloads page: https://download.eclipse.org

Each of these follow the general template you linked to in the previous comment.

otaviojava:
@KyleAure thanks.
Yeap, we have a Jenkins instance, but we did it manually; we need to move if forward automatically.

KyleAure:
These builds have been created!
Link to Jenkins: https://ci.eclipse.org/data/
Link to mailing list with more information on each build: https://www.eclipse.org/lists/data-dev/msg00114.html

PR #105, opened 2023-01-02
Return version to SNAPSHOT

otaviojava:
It is not the perfect solution, but it is a good MVP.

We have an issue to solve this on this issue: https://github.com/jakartaee/data/issues/104
Signed-off-by: Otavio Santana <otaviopolianasantana@gmail.com>


mpredli01:
+ 1 from me as well!

PR #110, opened 2023-01-20
Document precedence of logical operators

njr-11:
The specification should document the precedence for `And` and `Or` so that users can be assured what behavior they can rely on getting when writing a repository method like,

`findByNumDependantsGreaterThanOrFilingStatusAndIncomeBetween`

Users will likely expect to get the precedence [defined by JPQL](https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1.html#operators-and-operator-precedence) or by Java, which both have `And` taking precedence and being evaluated first.

Also, I spotted a typo in the same file where a sentence is missing the verb, so I've corrected that as well in this pull.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -278,7 +278,7 @@ Jakarta Data implementations often support the following list of predicate keywo
 
 ====== Logical Operator Precedence
 
-For relational databases, the logical operator ```And``` takes precedence over ```Or```, meaning that ```And``` is evaluated on conditions before ```Or``` when both are specified on the same method. For other database types, the precedence is limited by what the database is capable of. For example, some graph databases are limited to precedence in traversal order.
+For relational databases, the logical operator `And` takes precedence over `Or`, meaning that `And` is evaluated on conditions before `Or` when both are specified on the same method. For other database types, the precedence is limited to the capabilities of the database. For example, some graph databases are limited to precedence in traversal order.
 
 === Special Parameter Handling
 
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -276,6 +276,10 @@ Jakarta Data implementations often support the following list of predicate keywo
 
 |===
 
+====== Logical Operator Precedence
+
+For relational databases, the logical operator ```And``` takes precedence over ```Or```, meaning that ```And``` is evaluated on conditions before ```Or``` when both are specified on the same method. For other database types, the precedence is limited by what the database is capable of. For example, some graph databases are limited to precedence in traversal order.
+
 === Special Parameter Handling
 
 Jakarta Data also supports particular parameters to define pagination and sorting.
--- a/api/src/main/java/jakarta/data/repository/Repository.java
+++ b/api/src/main/java/jakarta/data/repository/Repository.java
@@ -297,7 +297,7 @@ import java.lang.annotation.Target;
  * For relational databases, the logical operator <code>And</code>
  * is evaluated on conditions before <code>Or</code> when both are specified
  * on the same method. Precedence for other database types is limited to
- * what the database is capable of.
+ * the capabilities of the database.
  * </p>
  *
  * <table style="width: 100%">
--- a/api/src/main/java/jakarta/data/repository/Repository.java
+++ b/api/src/main/java/jakarta/data/repository/Repository.java
@@ -27,7 +27,7 @@ import java.lang.annotation.Target;
  * <p>Annotates a data repository interface that will be implemented by the container/runtime.</p>
  *
  * <p>This class is a CDI bean-defining annotation when CDI is available,
- * enabling the container/runtime to the implementation available via the
+ * enabling the container/runtime to make the implementation available via the
  * <code>jakarta.inject.Inject</code> annotation.</p>
  *
  * <p>For example,</p>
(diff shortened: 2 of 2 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -276,6 +276,8 @@ Jakarta Data implementations often support the following list of predicate keywo
 
 |===
 
+When both are specified on the same method, the logical operator ```And``` takes precedence over ```Or```, meaning that ```And``` is evaluated on conditions before ```Or```.
```
otaviojava, line 279: It belongs to JPQL and SQL databases; thus, it belongs to vendor documentation.
IMHO: We should not expect all vendors and databases to behave similarly.

njr-11: We are doing a disservice to users every time we force them to read a particular vendor's documentation (if they even document it - I can't find it) for something that could have just been standardized in the specification.  There is a subset of customers that will only code to standards and are unable to use vendor specific function.  They won't be able to use this at all if we don't say what the precedence is.  Does NoSQL have the opposite precedence? Or equal precedence deferring to order of appearance?  JPQL and SQL can do any precedence based on where we decide to put the parenthesis, although the practice to have `And` take precedence over `Or` seemed the most preferable due to widespread adoption of that approach.

otaviojava: I agree that being forced to read specific vendors is the only reason sometimes.


Graph database with Apache TinkerPop will follow the traversal flow; thus, it will follow the sequence you put; accordingly, if "Or" is first, it will execute `Or` first.

My question is, do we need to put this?
I mean, Why do we need to put the full minimum details of JPQL on the spec?
I checked: 

* Spring Data JPA: https://docs.spring.io/spring-data/jpa/docs/current/reference/html/
* Apache Delta Spike JPA: https://deltaspike.apache.org/documentation/jpa.html
* Micronaut Data: https://micronaut-projects.github.io/micronaut-data/latest/guide/

They did not even put an implementation specific to JPA; why do we, as a standard to the agnostic database, need to put it?

njr-11: Because undocumented behavior is useless behavior. No customer should be expected to use trial and error to discover how spec-defined function actually behaves or be expected to write product code that relies on experimental findings rather than documented behavior. If users are forced to do this, vendors could change it underneath them at a whim and break their implementations. Customer implementations that use it will be non-portable. Why make our users vulnerable to that?

> Why do we need to put the full minimum details of JPQL on the spec?

No one has proposed doing so. Unlike query by method name, JPQL is flexible and allows the use of parentheses to define whatever precedence you like. Query by method name from Jakarta Data is by its nature limited to an inflexible pattern with no way for the user to indicate their desired precedence. Jakarta Data defines an API allowing for `And` and `Or` to be used in query by method. By defining the API, the specification takes on the responsibiilty of defining its behavior. Where we neglect to do so, specification becomes untestable, and vendors with clearly broken implementations (such as treating all `Or` as `And` when both are present) will be able to pass the TCK and call themselves valid implementations of Jakarta Data. The right course of action is to fully specify the behavior of what we define.

Thanks for pointing out the precedence limitation for Apache Tinkerpop.  Is that Graph database limitation or specific to that vendor?  Do any of the other NoSQL databases or database types require a specific precedence? The ideal situation would be for the specification to define a single precedence for all. But lacking that, it would be valid for the specification to standardize the precedence per database type. Or if NoSQL databases are all over the place and not even common by type, then maybe the best we can do is to state the requirement for relational databases. Can you or anyone else from NoSQL provide that information? Then we can write the spec accordingly.

otaviojava: Apache Tinkerpop works as JDBC to Graph database; thus, we're talking about over 30 graphs databases: 

https://tinkerpop.apache.org/providers.html
Fm6I84ZWAAAX1Au

otaviojava: > Because undocumented behavior is useless behavior. 
I agree, but who does implement the behavior?

njr-11: The ideal would have been to clearly document for all, but it looks like we're limited in what the specification can state for non-relational database types.  Let's just do the best we can from the specification side. I replace the previous commit with this one https://github.com/jakartaee/data/pull/110/commits/39f8e072499635bef31d85a0b1fbb0aeb5ea915e to address the concerns here.

otaviojava: Yeap. It is our challenge and nightmare.

otaviojava (approved):

mpredli01 (changes requested):
This looks good to me with the following suggestions:
[a] please change "...limited to what the database is capable of" to "...limited to the capabilities of the database." (I'm a stickler for sentences ending in a preposition.
[b] I believe we can safely change ```{ variable }``` to '{ variable }`.  I believe the extra back ticks add depth to the highlighting.

njr-11:
@mpredli01 thanks for reviewing. I included both updates in commit https://github.com/jakartaee/data/pull/110/commits/05e303d9e34e3e85aa070437f608344564b84d6e

njr-11:
@mpredli01 could you mark your approval on the commit to address your review comments https://github.com/jakartaee/data/commit/05e303d9e34e3e85aa070437f608344564b84d6e or let me know if further changes are needed so that we can merge this?  Thanks!

mpredli01 (approved):

Issue #94, opened 2022-11-23
Should we have the OrderBy annotation?


otaviojava:
To sort any data using the Repository feature at Jakarta Data, we have three ways:

1.  Using the `Pageable` or `Sort`to makes it possible dynamically
2. The method by query through the `order by  in the method statically
3. Using the `Sort` annotation statically.

My question: Do we need Sort annotation? I guess one to each, dynamically and statically, is enough.

njr-11:
If we need to get rid of one (and I'm not sure that we do), I would vote for getting rid of the `...OrderBySomethingAscSomethingElseAscAnotherThingAsc` in the method name.  It makes for methods that are difficult to read and consume in code, whereas the other mechanisms, including the annotation are much cleaner and more readable.

mp911de:
No, you should not have a sort annotation.

Consider the following:

```java

@Sort(by="baz")
List findByFooOrderBySomething()
```

An annotation creates easily ambiguity because it stands in front of the method.

While a `Sort` argument indicates sort appending:

```java
List findByFooOrderBySomething(Sort sort)
```

njr-11:
> An annotation creates easily ambiguity because it stands in front of the method.
> 
> While a `Sort` argument indicates sort appending:
> 
> ```java
> List findByFooOrderBySomething(Sort sort)
> ```

Allowing the same capability to be specified in 2 different places on the same method is a cause of ambiguity regardless of whether it comes from an annotation, a method name, or wherever.  The order/position of each does not imply an intuitive behavior and we should raise an error for the conflict, which is what the Jakarta Data spec currently defines:

In `Sort`,
https://github.com/jakartaee/data/blob/f0888ead734aca15ef442c979185318845f0a722/api/src/main/java/jakarta/data/repository/Sort.java#L44-L50

and `@OrderBy`,
https://github.com/jakartaee/data/blob/main/api/src/main/java/jakarta/data/repository/OrderBy.java#L50-L56

and `Pageable`,
https://github.com/jakartaee/data/blob/main/api/src/main/java/jakarta/data/repository/Pageable.java#L240-L248

The first 2 above are lacking mention of the conflict with `Sort` that comes from a `Pageable` which was missed when `Sort` was added to `Pageable`.  I've added pull #95 to correct that.

amoscatelli:
[...]

PR #115, opened 2023-01-27
Remove Sortby annotation

otaviojava:
We had an issue discussing this annotation, but it is not clear, even with more documentation, if this annotation is proper.

So, 

- I propose removing it.
- Incentive to a vendor implement a feature like that, then move again to this discussion with a case. 

Ref: https://github.com/jakartaee/data/issues/94


keilw:
Let's see, what vendor/implementor reps like @graemerocher (for Micronaut) say, if they don't see a value at this point, then let'S remove it.

otaviojava:
@keilw, you have the discussion here:
https://github.com/jakartaee/data/issues/94

njr-11 (changes requested):
This matter was discussed in #94 and agreement was never reached on removing it, so there should not be a pull to do so.  As I pointed out there, if anything is to be removed for sorting, this is the wrong thing to remove.  In the various tests that I have been writing for Jakarta Data thus far, I have found this annotation to be extremely useful and much preferred over the lengthy method names with the `OrderBy`/`Asc`/`Desc` keywords such as `repo.findBySomethingGreaterThanEqualAndSomethngElseLessThanOrderBySomethingAscSomethingElseAscAnotherThingAsc(something, somethingElse);` so if anything is to be removed, my vote is for the latter, not the annotation.   I'm also fine with leaving things as is and keeping both for the sake of compatibility with existing solutions.  Additional reasons are discussed in #94 - I can't copy all of that here.

It also needs to be pointed out that even if this change were to be agreed upon, it is not correct as written and introduces errors into the spec in at least one place.

otaviojava:
@njr-11 I closed.

Sorry for my misunderstanding

Issue #119, opened 2023-02-22
TCK: How to test Entity


KyleAure:
This issue builds off of https://github.com/jakartaee/data/issues/112

## Description

Since we've decided that the API for Entity will remain in the Persistence and NoSQL specs where they will be developed separately. 
We still need to figure out a way to write TCK tests that use entities. 

This is tricky since we want to be able to test `jakarta.nosql.Entity` or `jakarta.persistence.Entity` or both during a single run of the TCK.  I've been trying to think of a few ways to do this with the least amount of duplication possible: 

## Main roadblock

The main roadblock here is that Entity is an annotation that needs to exist on an object at compile time. 

## Possible Solutions

Here are some possible solutions that I have thought of, but don't have a complete solution and need help choosing/refining one. 

### Dual Annotated Entities
Add both Entity annotations for example:

```java
@jakarta.nosql.Entity
@jakarta.persistence.Entity
public class Student {
  
  @jakarta.nosql.Id
  @jakarta.persistence.Id
  Long id;
}

@Repository
public interface StudentDirectory extends DataRepository<Student, Long> {
    void save(Student student);
    int countByAgeGreaterThanEqual(Integer age);
}
``` 

For a standalone repository provider or a Jakarta EE profile that only has a single repository provider, this will work.
However, what happens when `studentDirectory.save()` is run on a Jakarta EE profile that has both providers?
How can we run the same test against both providers? 

### Duplicate code
This is a messy but straightforward solution by just duplicating code. For example: 

```java
public class Student {
  @jakarta.persistence.Entity
  public static class Persistence { ...  }

  @jakarta.nosql.Entity
  public static class NoSQL { ... }
}

public interface StudentDirectory {
    @Repository
    public interface Persistence extends DataRepository<Student.Persistence, Long> { ... }

	@Repository
	public interface NoSQL extends DataRepository<Student.NoSQL, Long> { ... }
```

With this solution, we would have to be told by the runner of the TCK which repository(ies) are supported and run based on that input.  This will likely require some more complex Arquillan/Junit logic but should be doable.

### Generated source code
Use a pre-compile tool that will automatically generate source code, compile it, and include it in the TCK.
This will require some sort of parser to search for a mock Entity/Id annotations and use that class as a template, for example: 

```java
@ee.jakarta.tck.data.anno.Entity
public class Student {
  
  @@ee.jakarta.tck.data.anno.Id
  Long id;
}
```

Once parsed two files are created: 
```java
[...]

PR #121, opened 2023-02-27
Example of duplicate entities

KyleAure:
Related to #119


KyleAure:
Closing in favor of #120

PR #122, opened 2023-02-28
Document interoperability with Jakarta Transactions

njr-11:
The spec needs to state what must happen if there is a JTA transaction already on the thread when a repository method is invoked so that users can know what to expect, both for resources types that are enlistable (these participate in the transaction) and for those which are not (these do not participate in the transaction).  The spec should also acknowledge that a failure running a repository method might cause a transaction to be marked as rollback-only (JPA does this in some cases) so that users are not surprised by that behavior.


njr-11:
Rebased to resolve merge conflict.

jmstephensgit (approved):

PR #124, opened 2023-03-01
TCK needs to be able to put both entity annotations on same entity

KyleAure:
Related to #119 

- Fixes spec documentation: `jakarta.nosql.mapping.*` -> `jakarta.nosql.*`
- Discourages applications from using both Entity annotations on the same class
- Document that the Repository is only responsible for using the entity it supports.


njr-11 (approved):

njr-11:
This has been open for 4 weeks without any issues or objections raised against it and is blocking progress of the TCK, so I'm merging it.

PR #125, opened 2023-03-03
Compile against Java 17

njr-11:
[Jakarta EE 11](https://eclipse-ee4j.github.io/jakartaee-platform/jakartaee11/JakartaEE11ReleasePlan) is looking to make Java 21 the minimum supported runtime level.  Java 21 isn't available yet, but a step toward getting there is to update from Java 11 to Java 17, the most recent long term support release of Java that is currently available.  This enables us to update some classes to utilize Java records as we previously voted.  At the time, those classes were updated to look like Java records in the way that methods were formed, but with compilation still at Java 11, they still included implementation of accessor methods/equals/toString/hashCode and so forth simulating what Java records would provide.

This can be looked at after Otavio gets beta 2 out.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Limit.java
+++ b/api/src/main/java/jakarta/data/repository/Limit.java
@@ -45,17 +45,29 @@ package jakarta.data.repository;
  * <li>a <code>Limit</code> parameter is supplied in combination
  *     with the <code>First</code> keyword.</li>
  * </ul>
+ *
+ * @param maxResults maximum number of results for a query.
+ * @param startAt    starting position for query results (1 is the first result).
  */
-public final class Limit {
-    private static final long DEFAULT_START_AT = 1L;
-    private final int maxResults;
-    private final long startAt;
+public record Limit(int maxResults, long startAt) {
 
-    private Limit(int maxResults, long startAt) {
-        this.maxResults = maxResults;
-        this.startAt = startAt;
+    private static final long DEFAULT_START_AT = 1L;
+
+    /**
+     * Constructs a limit on the amount of query results.
+     *
+     * @param maxResults maximum number of results for a query.
+     * @param startAt    starting position for query results (1 is the first result).
+     */
+    public Limit {
+        if (startAt < 1)
+            throw new IllegalArgumentException("startAt: " + startAt);
+
+        if (maxResults < 1)
+            throw new IllegalArgumentException("maxResults: " + maxResults);
     }
 
+    // Override to provide method documentation:
     /**
      * <p>Maximum number of results that can be returned for a
      * single invocation of the repository method.</p>
@@ -66,9 +78,10 @@ public final class Limit {
         return maxResults;
     }
 
+    // Override to provide method documentation:
     /**
      * <p>Offset at which to start when returning query results.
-     * The first query result is position <code>1</code>.<p>
+     * The first query result is position <code>1</code>.</p>
      *
      * @return offset of the first result.
      */
@@ -86,9 +99,6 @@ public final class Limit {
      * @throws IllegalArgumentException if maxResults is less than 1.
      */
     public static Limit of(int maxResults) {
-        if (maxResults < 1)
-            throw new IllegalArgumentException("maxResults: " + maxResults);
-
         return new Limit(maxResults, DEFAULT_START_AT);
     }
 
@@ -109,9 +119,6 @@ public final class Limit {
      *         exceeds {@link Integer#MAX_VALUE}.
      */
     public static Limit range(long startAt, long endAt) {
-        if (startAt < 1)
-            throw new IllegalArgumentException("startAt: " + startAt);
-
         if (endAt < startAt)
             throw new IllegalArgumentException("startAt: " + startAt + ", endAt: " + endAt);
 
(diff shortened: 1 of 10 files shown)


KyleAure (approved):
Looks great!

otaviojava:
@njr-11 I agree that we should do it, but I'm unsure if we need to do it now!

How do you start creating a CI for testing on Java 17?

njr-11:
Sorry that I caused confusion with the CI acronym.  In the commit, it was referring to updating the Continuous Integration that runs with each pull, not referring to Compatible Implementation.  It's a good point that compatible implementations will need to support Java 17 as well, and by the time of Jakarta EE 11, will need to support Java 21 as well.  The [JNoSQL github project page](https://github.com/eclipse/jnosql) states that it supports Java 11 or higher.  Do you see any issues with it implementing a Jakarta Data spec that requires Java 17?  I don't think this should stop you from still supporting Java 11 for Jakarta NoSQL (although eventually for Jakarta EE 11, that should be moving up as well).

otaviojava:
Eclipse JNoSQL has support for both: Java 11 and Java 17:

https://github.com/eclipse/jnosql/tree/main/.github/workflows

I suggest doing the same.

MIcronaut has the plan to use Java 17 as a minimum requirement, but it is not there yet:

https://micronaut.io/2023/02/16/micronaut-framework-4-0-with-java-17-baseline/

We could define Appril as a baseline for using only Java 17.

njr-11:
Yes, April sounds like a good plan for this.  If we move too quickly, implementations will not be ready.  If we wait too long, it could become more difficult to get off of Java 11.

otaviojava:
@njr-11 could you fix this one?

njr-11:
> @njr-11 could you fix this one?

Thanks for pointing out it needed an update.  I rebased onto the latest and resolved the merge conflict, which turned out just to be on adding year 2023 to a copyright.

api/src/main/java/jakarta/data/repository/Sort.java
```diff
@@ -57,28 +57,35 @@
...
 
+    // Override to provide method documentation:
     /**
+     * Name of the property to order by.
+     *
      * @return The property name to order by; will never be {@literal null}.
      */
     public String property() {
-        return this.property;
+        return property;
```
otaviojava, line 85: Do we need this method?
Once you don't overwrite, you can remove it.

api/src/main/java/jakarta/data/repository/Sort.java
```diff
@@ -91,46 +98,23 @@ public boolean ignoreCase() {
...
 
+    // Override to provide method documentation:
     /**
+     * Indicates whether to sort the property in ascending order (true) or descending order (false).
+     *
      * @return Returns whether sorting for this property shall be ascending.
      */
     public boolean isAscending() {
-        return Direction.ASC.equals(direction);
+        return isAscending;
```
otaviojava, line 108: We can remove this as well.

api/src/main/java/jakarta/data/repository/Sort.java
njr-11: We will get the same implementation from record if we remove it, but the documentation will be lost.  I have a comment just above the method clarifying why it is there, `Override to provide method documentation:`

api/src/main/java/jakarta/data/repository/Sort.java
njr-11: We will get the same implementation from record if we remove it, but the documentation will be lost.  I have a comment just above the method clarifying why it is there, `Override to provide method documentation:`

njr-11:
@otaviojava @KyleAure @mpredli01 
To follow up on the decision we made at the meeting regarding the record constructor, I looked into where we can suppress the documentation of it in favor of the static methods.  It turns out that Java records do such a good job of covering all aspects of implementation (and JavaDoc) for you that if you attempt to omit the JavaDoc or even blank it out as emtpy, it simply generates JavaDoc for the constructor and uses that instead.  The best we can do then is to write the JavaDoc stating to use the static methods instead for better readability of code.  I have done this under commit https://github.com/jakartaee/data/pull/125/commits/a20e94ff3964d6f8160a2ab71740545aeb25db65
Please have a look at that and let me know if we can go with that approach. Otherwise, I can remove the use of record for Limit and Sort.  Either way, let's get a decision made so we can update to Java 17.

api/src/main/java/jakarta/data/repository/Limit.java
```diff
@@ -54,7 +54,11 @@ public record Limit(int maxResults, long startAt) {
     private static final long DEFAULT_START_AT = 1L;
 
     /**
-     * Constructs a limit on the amount of query results.
+     * <p>Limits query results. For more descriptive code, use:</p>
+     * <ul>
+     * <li>{@link #of(int) Limit.of(maxResults)} to cap the number of results.</li>
+     * <li>{@link #range(long, long) Limit.range(startAt, endAt)} to limit to a range.</li>
+     * </ul>
```
KyleAure, line 61: 👍🏼 I think this is the best compromise.

KyleAure (approved):

njr-11:
@otaviojava if this solution isn't acceptable, let me know and I'll switch it. I don't want it to keep holding up progress of the TCK.

PR #130, opened 2023-03-16
Documentation of Standalone and Jakarta EE modes for using Jakarta Data providers

njr-11:
This pull adds documentation of the ways of using Jakarta Data, which can be as Standalone providers, within Jakarta EE products, or both.
It explicitly documents that injection can be accomplished either by custom dependency injection or by CDI and addresses discovery of providers for the latter, giving Jakarta Data providers the flexibility to choose the approach that best suits them without imposing requirements on providers that already have their own dependency injection.

### DIFF
--- a/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
+++ b/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.spi.provider;
+
+import java.lang.annotation.Annotation;
+import java.util.Set;
+
+import jakarta.data.exceptions.MappingException;
+
+/**
+ * <p>Jakarta Data providers can make themselves discoverable by
+ * Jakarta EE products via the Jakarta Data Provider SPI. This enables
+ * Jakarta EE products with CDI support to provide CDI dependency injection
+ * and other integration with CDI on behalf of Jakarta Data providers.
+ * Jakarta Data providers with their own custom dependency injection
+ * are not required to implement or provide the Jakarta Data Provider SPI.</p>
+ *
+ * <p>Jakarta Data providers that implement the Jakarta Data Provider SPI
+ * make an implementation of this interface available to the
+ * {@link java.util.ServiceLoader ServiceLoader} as follows.</p>
+ *
+ * <p>The {@code DataProvider} implementation and related classes are packaged
+ * within the Jakarta Data provider's JAR file, also including a file with the
+ * following name and location,</p>
+ *
+ * <pre>META-INF/services/jakarta.data.spi.provider.DataProvider</pre>
+ * 
+ * <p>The content of the above file must be one or more lines, each specifying
+ * the fully qualified name of a {@code DataProvider} implementation that is
+ * provided within the JAR file.</p>
+ *
+ * <p>Jakarta EE applications do not use the Jakarta Data Provider SPI directly.
+ * Jakarta EE products are not required to make this class available to
+ * applications.</p>
+ */
+public interface DataProvider {
+    /**
+     * <p>Provides implementation of the specified repository interface.</p>
+     *
+     * @param <R>                 interface class that defines the data repository.
+     * @param repositoryInterface the repository interface.
+     * @param entityClass         type of entity that the repository persists.
+     * @return repository instance. Never {@code null}.
+     * @throws MappingException for inconsistencies between the repository or
+     *         entity class and the database.
+     */
+    <R> R getRepository(Class<R> repositoryInterface, Class<?> entityClass) throws MappingException;
+
+
+    /**
+     * <p>Returns the name of the Jakarta Data provider. This name can be used
+     * in configuration when referring to this provider and when logging or
+     * displaying messages to the user.</p>
+     *
+     * @return name that identifies the Jakarta Data provider. Never {@code null}.
+     */
+    String name();
+
+    /**
+     * <p>Invoked by the Jakarta EE product to notify the the Jakarta Data
+     * provider that the CDI managed bean for a previously-obtained repository
+     * instance has reached the end of its life cycle. If multiple invocations of
+     * {@link #getRepository(Class, Class) getRepository} return the same
+     * repository instance, multiple dispose notifications will be sent for the
+     * instance, each corresponding to the disposal of a different CDI managed bean.</p>
+     *
+     * @param repository instance that was previously returned by
+     *        {@link #getRepository(Class, Class) getRepository}.
+     */
+    void repositoryBeanDisposed(Object repository);
+
+    /**
+     * <p>Returns the type of entity annotations that are supported by this provider.
+     * For example, <code>jakarta.persistence.Entity</code> or <code>jakarta.nosql.Entity</code>.
+     * This tells the Jakarta EE product that it can request repositories for entities
+     * of this type from this provider.</p>
+     *
+     * @return type of entity annotations that are supported by this provider.
+     */
+    Set<Class<? extends Annotation>> supportedEntityAnnotations();
+}
--- a/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
+++ b/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
@@ -52,14 +52,13 @@ public interface DataProvider {
     /**
      * <p>Provides an instance that implements the specified repository interface.</p>
      *
-     * @param <R>                 interface class that defines the data repository.
-     * @param repositoryInterface the repository interface.
-     * @param entityClass         type of entity that the repository persists.
+     * @param <R>        interface class that defines the data repository.
+     * @param repository the repository interface.
      * @return repository instance. Never {@code null}.
      * @throws MappingException for inconsistencies between the repository or
      *         entity class and the database.
      */
-<R> R getRepository(Class<R> repository) throws MappingException;
+    <R> R getRepository(Class<R> repository) throws MappingException;
 
 
     /**
@@ -75,12 +74,12 @@ public interface DataProvider {
      * <p>Invoked by the Jakarta EE product to notify the Jakarta Data
      * provider that the CDI managed bean for a previously-obtained repository
      * instance has reached the end of its life cycle. If multiple invocations of
-     * {@link #getRepository(Class, Class) getRepository} return the same
+     * {@link #getRepository(Class) getRepository} return the same
      * repository instance, multiple dispose notifications will be sent for the
      * instance, each corresponding to the disposal of a different CDI managed bean.</p>
      *
      * @param repository instance that was previously returned by
-     *        {@link #getRepository(Class, Class) getRepository}.
+     *        {@link #getRepository(Class) getRepository}.
      */
     void repositoryBeanDisposed(Object repository);
 
--- a/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
+++ b/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
@@ -23,13 +23,10 @@ import java.util.Set;
 import jakarta.data.exceptions.MappingException;
 
 /**
- * <p>Jakarta Data providers can make themselves discoverable by
- * Jakarta EE products via the Jakarta Data Provider SPI.
- * Jakarta EE products with CDI support are thus able to automatically provide
- * a CDI extension on behalf of the discovered Jakarta Data providers
- * without those Jakarta Data providers themselves needing to depend on CDI.
- * Jakarta Data providers with their own custom dependency injection
- * are not required to implement or provide the Jakarta Data Provider SPI.</p>
+ * <p>In a Jakarta EE profile, Jakarta Data providers can use the
+ * Jakarta Data Provider SPI to register to provide repository instances
+ * to the Jakarta EE product's CDI extension that handles
+ * {@link jakarta.data.repository.Repository Repository} injection points.</p>
  *
  * <p>Jakarta Data providers that implement the Jakarta Data Provider SPI
  * make an implementation of this interface available to the
--- a/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
+++ b/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
@@ -24,9 +24,10 @@ import jakarta.data.exceptions.MappingException;
 
 /**
  * <p>Jakarta Data providers can make themselves discoverable by
- * Jakarta EE products via the Jakarta Data Provider SPI. This enables
- * Jakarta EE products with CDI support to provide CDI dependency injection
- * and other integration with CDI on behalf of Jakarta Data providers.
+ * Jakarta EE products via the Jakarta Data Provider SPI.
+ * Jakarta EE products with CDI support are thus able to automatically provide
+ * a CDI extension on behalf of the discovered Jakarta Data providers
+ * without those Jakarta Data providers themselves needing to depend on CDI.
  * Jakarta Data providers with their own custom dependency injection
  * are not required to implement or provide the Jakarta Data Provider SPI.</p>
  *
--- a/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
+++ b/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
@@ -59,7 +59,7 @@ public interface DataProvider {
      * @throws MappingException for inconsistencies between the repository or
      *         entity class and the database.
      */
-    <R> R getRepository(Class<R> repositoryInterface, Class<?> entityClass) throws MappingException;
+<R> R getRepository(Class<R> repository) throws MappingException;
 
 
     /**
--- a/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
+++ b/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
@@ -44,7 +44,7 @@ import jakarta.data.exceptions.MappingException;
  * the fully qualified name of a {@code DataProvider} implementation that is
  * provided within the JAR file.</p>
  *
- * <p>Jakarta EE applications do not use the Jakarta Data Provider SPI directly.
+ * <p>Jakarta EE applications do not use the Jakarta Data Provider SPI.
  * Jakarta EE products are not required to make this class available to
  * applications.</p>
  */
--- a/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
+++ b/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
@@ -50,7 +50,7 @@ import jakarta.data.exceptions.MappingException;
  */
 public interface DataProvider {
     /**
-     * <p>Provides implementation of the specified repository interface.</p>
+     * <p>Provides an instance that implements the specified repository interface.</p>
      *
      * @param <R>                 interface class that defines the data repository.
      * @param repositoryInterface the repository interface.
--- a/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
+++ b/api/src/main/java/jakarta/data/spi/provider/DataProvider.java
@@ -72,7 +72,7 @@ public interface DataProvider {
     String name();
 
     /**
-     * <p>Invoked by the Jakarta EE product to notify the the Jakarta Data
+     * <p>Invoked by the Jakarta EE product to notify the Jakarta Data
      * provider that the CDI managed bean for a previously-obtained repository
      * instance has reached the end of its life cycle. If multiple invocations of
      * {@link #getRepository(Class, Class) getRepository} return the same
--- a/api/src/main/java/jakarta/data/spi/provider/package-info.java
+++ b/api/src/main/java/jakarta/data/spi/provider/package-info.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * <p>This package defines the Service Provider Interface (SPI) by which
+ * Jakarta Data providers can make themselves discoverable by Jakarta EE products
+ * for CDI dependency injection.</p>
+ *
+ * <p>Jakarta Data providers that provide their own custom dependency injection
+ * or other vendor-specific mechanisms can ignore this SPI and do not need
+ * to provide or implement it.</p>
+ *
+ * <p>Jakarta EE applications do not use the Jakarta Data Provider SPI directly.
+ * Jakarta EE products are not required to make this package available to
+ * applications.</p>
+ */
+package jakarta.data.spi.provider;
--- a/api/src/main/java/jakarta/data/spi/provider/package-info.java
+++ b/api/src/main/java/jakarta/data/spi/provider/package-info.java
@@ -25,7 +25,7 @@
  * or other vendor-specific mechanisms can ignore this SPI and do not need
  * to provide or implement it.</p>
  *
- * <p>Jakarta EE applications do not use the Jakarta Data Provider SPI directly.
+ * <p>Jakarta EE applications do not use the Jakarta Data Provider SPI.
  * Jakarta EE products are not required to make this package available to
  * applications.</p>
  */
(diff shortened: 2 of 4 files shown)


api/src/main/java/jakarta/data/spi/provider/DataProvider.java
```diff
@@ -0,0 +1,96 @@
...
+     * <p>Returns the name of the Jakarta Data provider. This name can be used
+     * in configuration when referring to this provider and when logging or
+     * displaying messages to the user.</p>
+     *
+     * @return name that identifies the Jakarta Data provider. Never {@code null}.
+     */
+    String name();
+
+    /**
+     * <p>Invoked by the Jakarta EE product to notify the the Jakarta Data
```
frowe, line 75:
```suggestion
     * <p>Invoked by the Jakarta EE product to notify the Jakarta Data
```

api/src/main/java/jakarta/data/spi/provider/DataProvider.java
```diff
@@ -0,0 +1,96 @@
...
+ * the fully qualified name of a {@code DataProvider} implementation that is
+ * provided within the JAR file.</p>
+ *
+ * <p>Jakarta EE applications do not use the Jakarta Data Provider SPI directly.
+ * Jakarta EE products are not required to make this class available to
+ * applications.</p>
+ */
+public interface DataProvider {
+    /**
+     * <p>Provides implementation of the specified repository interface.</p>
```
frowe, line 53:
```suggestion
     * <p>Provides an instance of an implementation of the specified repository interface.</p>
```

api/src/main/java/jakarta/data/spi/provider/package-info.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * <p>This package defines the Service Provider Interface (SPI) by which
+ * Jakarta Data providers can make themselves discoverable by Jakarta EE products
+ * for CDI dependency injection.</p>
+ *
+ * <p>Jakarta Data providers that provide their own custom dependency injection
+ * or other vendor-specific mechanisms can ignore this SPI and do not need
+ * to provide or implement it.</p>
+ *
+ * <p>Jakarta EE applications do not use the Jakarta Data Provider SPI directly.
```
frowe, line 28:
```suggestion
 * <p>Jakarta EE applications do not use the Jakarta Data Provider SPI.
```

frowe (changes requested):

api/src/main/java/jakarta/data/spi/provider/DataProvider.java
njr-11: Corrected in https://github.com/jakartaee/data/pull/130/commits/670dd4f12fffd78fa33361474f772aabc64ffc28

otaviojava:
Could you explain when you need to use it?

I mean, CDI has the `@Inject` that allows it.

There is the [CDI Instance](https://docs.oracle.com/javaee/7/api/javax/enterprise/inject/Instance.html) to allow and see multiple implementations) where you can see those.

```java
@Inject 
@Any
Library repository;
```

Sorry, but it is still unclear why we need it if a CDI engine can handle that for us.

njr-11:
> Sorry, but it is still unclear why we need it if a CDI engine can handle that for us.

@otaviojava you are assuming that all of the Jakarta Data providers want to provide their own CDI extensions, but we have already heard from them that isn't the case. Some want to use their own dependency injection, and mostly others probably don't want to do either, and they shouldn't need to.  This will give the Jakarta EE products a way to bring providers into CDI without the providers needing to have any dependency on CDI, and accommodates custom dependency injection as well.

njr-11:
This is a very difficult balancing act because we have 3 groups with completely different objectives around dependency injection

 - Existing providers with pre-existing custom dependency injection models that want to keep them
 - JNoSQL that wants to use its own CDI-based approach
 - The Jakarta EE platform, which places a great deal of importance on being CDI-centric

The third category above hasn't really had representation in the spec group, but we should not be ignoring it and what the consequences might be if we are deemed inadequate. If we don't do anything at all and leave the spec as is, I find it extremely unlikely that Jakarta EE accepts our specification into a Jakarta EE profile. I really don't want to risk that.

I've tried to put together this compromise as something that all of us will dislike parts of, but that we can hopefully all find workable, because it should leave ways for everyone to achieve their objectives.

- It allows Spring and Micronaut to keep their custom dependency injection models. They would need to tolerate the existence of the provider SPI in the spec, but it's their choice whether to tie into it, and they will be able to pass the Standalone TCK (which is how non-EE profiles certify) even if they choose not to provide the SPI, just using java.inject and their custom dependency injection models.

- It asks JNoSQL to tie into the provider SPI for CDI integration when running in Jakarta EE profiles, but for Standalone, JNoSQL could do whatever it wants, with its own standalone-mode CDI extension, or whatever it chooses.

- For Jakarta EE, it offers a CDI-centric solution in Jakarta EE profiles (that we know some providers will choose not to adopt), while outside of Jakarta EE profiles it neglects to enforce a CDI-centric solution and allows custom dependency injection for Standalone and when running with combinations of Jakarta EE specs that aren't officially declared to be Jakarta EE profiles. There are statements in this pull that Jakarta EE as a whole I'm sure will be unhappy with, but I think we can at least defend it as being necessary for the compromise that needs to be made.

If we all argue for getting everything we want, then we will never reach any agreement as was the case in the previous attempt. I'm hoping we can come to a solution that is at least workable for everyone. Not our favorite solution, but a workable solution, where everyone preserves what is most important to them.

otaviojava:
> The Jakarta EE platform, which places a great deal of importance on being CDI-centric

Being CDI-centric at the same time that they don't want to use CDI resources such as CDI extension, `Instance`, and `Disposes`, this point does not make sense to me.


The method  ```void repositoryBeanDisposed(Object repository);```

CDI provides the `Disposes` annotation, which you can use as well: 
https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#declaring_disposer_method


After reviewing the specifications, creating a Provider for the Repository instance would not align with the single responsibility principle of the specification. Additionally, the current stage of the specification allows for any dependency injection, which CDI can handle on Jakarta EE with resources such as Instance, Dispose annotation, and qualifier.

While a DataProvider may be helpful for those who choose not to use dependency injection, covering this point in the specification may not be necessary.

njr-11:
> Being CDI-centric at the same time that they don't want to use CDI resources such as CDI extension, Instance, and Disposes, this point does not make sense to me.

It wouldn't make sense if it were the same group having positions that are in conflict. However, as pointed out there are multiple groups with different objectives here, one wanting to be CDI-centric, another wanting no part whatsoever in CDI. I'm sure that each believes their own position makes sense. Our specification needs to cope with the differing positions and bring everyone together with an agreeable compromise to be successful.

> After reviewing the specifications, creating a Provider for the Repository instance would not align with the single responsibility principle

Your argument here doesn't involve a correct understanding of single responsibility principle. Single responsibility does not mean that one module takes responsibility for everything that is needed to achieve the specification. It means that each module should have just one responsibility. If you apply the principle of single responsibility here, one could argue that your idea to have the Jakarta Data provider be responsible for everything - both supplying repositories and also responsible to integrate repositories with CDI - is contradictory to it. A solution where one module, a Jakarta Data provider, only has the responsible of providing repositories, and another module, a component of the Jakarta EE product, only has the responsibility of integrating repositories with CDI, seems a much better fit for the principle of single responsibility.

api/src/main/java/jakarta/data/spi/provider/DataProvider.java
```diff
@@ -0,0 +1,96 @@
...
+     * <p>Provides an instance that implements the specified repository interface.</p>
+     *
+     * @param <R>                 interface class that defines the data repository.
+     * @param repositoryInterface the repository interface.
+     * @param entityClass         type of entity that the repository persists.
+     * @return repository instance. Never {@code null}.
+     * @throws MappingException for inconsistencies between the repository or
+     *         entity class and the database.
+     */
+    <R> R getRepository(Class<R> repositoryInterface, Class<?> entityClass) throws MappingException;
```
otaviojava, line 62:
```suggestion
<R> R getRepository(Class<R> repository) throws MappingException;
```

You don't need the second one, you can get it from the class:


```java
        Class<?> repository = PersonRepository.class;
        System.out.println("repository: " + repository);
        ParameterizedType param = (ParameterizedType) repository.getGenericInterfaces()[0];
        Type[] arguments = param.getActualTypeArguments();
        System.out.println("the " + param);
        System.out.println("the entity: " + arguments[0]);
        System.out.println("the id: " + arguments[1]);
```

api/src/main/java/jakarta/data/spi/provider/DataProvider.java
```diff
@@ -0,0 +1,96 @@
...
+
+import java.lang.annotation.Annotation;
+import java.util.Set;
+
+import jakarta.data.exceptions.MappingException;
+
+/**
+ * <p>Jakarta Data providers can make themselves discoverable by
+ * Jakarta EE products via the Jakarta Data Provider SPI. This enables
+ * Jakarta EE products with CDI support to provide CDI dependency injection
```
otaviojava, line 28: That point is not true; once the CDI extension is part of the spec, you can do it without it:

https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#spi_full

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -464,3 +464,28 @@ for (Pageable p = Pageable.ofSize(25).sortBy(Sort.desc("yearBorn"), Sort.asc("na
...
+
+==== Jakarta EE Mode
+
+Jakarta EE mode means that the Jakarta Data provider is able to run within a Jakarta EE product, which is a product that provides a Jakarta EE profile or the full Jakarta EE platform, or a product that is comprised of a subset of Jakarta EE specifications. If the Jakarta EE product is running in a mode that provides Jakarta Contexts and Dependency Injection (CDI), then the Jakarta EE product coordinates CDI dependency injection for all Jakarta Data providers that it discovers via the Jakarta Data Provider SPI. Jakarta Data providers do not need to support CDI. Jakarta Data providers can continue using their own custom dependency injection models without CDI present. The Jakarta Data TCK includes a Jakarta EE mode for validating Jakarta Data Providers that leverages the `@Inject` annotation that is common to both CDI and custom dependency injection models.
```
otaviojava, line 479: > then the Jakarta EE product coordinates CDI dependency injection for all Jakarta Data providers that it discovers via the Jakarta Data Provider SPI. Jakarta Data providers do not need to support CDI.

Can we make it optional?
A CDI user can do it using only CDI. So, it should not be mandatory for any vendor.

api/src/main/java/jakarta/data/spi/provider/DataProvider.java
njr-11: That's a very good point, the entityClass parameter is redundant because it can be obtained from the repository class,  so let's go with your change to remove it.

api/src/main/java/jakarta/data/spi/provider/DataProvider.java
njr-11: Thanks for pointing this out.  I can see where this statement can be misunderstood. I've corrected it in commit https://github.com/jakartaee/data/pull/130/commits/3b3e7260e7eb69230b84027297d820c123585d45

spec/src/main/asciidoc/repository.asciidoc
njr-11: Pushing the responsibility onto the user would be a poor design. It's a natural fit for the Jakarta EE product to provide the CDI etxension and coordinate among the registered providers, automatically determining the correct one by the entity annotation so that the correct repository is created and injected without extra work from the user.  This will offer the best usability because it requires the customer to do the least amount of work, avoids duplication (the entity annotation alone is enough to know which provider they want), and is highly intuitive.  With the following we have a great model that provides superior usability.  Let's keep it that way.

```
@Entity
public class Product {
...
}

@Repository
public interface Products extends CrudRepository<Product, Long> {
...
}

@Inject
Products products;
```

njr-11:
> Another point: could you define "CDI-centric" and why not use CDI spec resources?

I'm sorry to have written things so unclearly as to suggest CDI spec resources would not be used by Jakarta EE products.  Hopefully the corrections I made for preceding review comments have cleared this up.  Jakarta EE products will be providing a CDI extension on behalf of the providers that register with the SPI.  So the CDI spec is being used just as well as in the proposal to make providers supply a CDI extension, except that we get even broader coverage because we can cover additional providers this way and we also get a simplified, superior user experience out of it.

CDI-centric isn't my own term to define, but I think we should be able to satisfy expectations of Jakarta EE with this enough to allow the Jakarta Data spec into a Jakarta EE profile.  We are requiring CDI support to come from the Jakarta EE product, which is the only place where all vendors will happily provide it (all profiles already include CDI).  Standalone providers and running with individual Jakarta EE specs outside of an official profile don't require CDI support, but could opt to do something on their own.  Jakarta EE should be fine with this.

api/src/main/java/jakarta/data/spi/provider/DataProvider.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Set;
+
+import jakarta.data.exceptions.MappingException;
+
+/**
+ * <p>Jakarta Data providers can make themselves discoverable by
+ * Jakarta EE products via the Jakarta Data Provider SPI.
+ * Jakarta EE products with CDI support are thus able to automatically provide
+ * a CDI extension on behalf of the discovered Jakarta Data providers
+ * without those Jakarta Data providers themselves needing to depend on CDI.
```
otaviojava, line 30: It did not understand this point, who use CDI extension and don't depend on CDI at the same time?

Emily-Jiang: IIUC, I think what @njr-11 meant is if a Jakarta EE runtime supports CDI, they can use CDI extension and they should be discovered by the Jakarta Data Provider. I think this might be a nice loosely coupled model. Jakarta Data itself does not depend on CDI but honours CDI. Maybe the sentence should be rephrased to make it clearer.

api/src/main/java/jakarta/data/spi/provider/DataProvider.java
```diff
@@ -0,0 +1,96 @@
...
+ * Jakarta Data providers with their own custom dependency injection
```
otaviojava, line 31:
```suggestion
 * Jakarta Data providers with dependency injection
```
IMHO: it only makes sense if you don't have any DI at all.
But, I may be missing something.

api/src/main/java/jakarta/data/spi/provider/package-info.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * <p>This package defines the Service Provider Interface (SPI) by which
+ * Jakarta Data providers can make themselves discoverable by Jakarta EE products
+ * for CDI dependency injection.</p>
```
otaviojava, line 22:
```suggestion
 * for CDI dependency injection, if you don't want to use CDI extension.</p>
```

spec/src/main/asciidoc/repository.asciidoc
otaviojava: I agree with you that it should be this way.

What I don't understand is the relation between CDI and this Data provider.

Could you show some code to express why this is so crucial? and why not use what CDI already provides to us?

api/src/main/java/jakarta/data/spi/provider/DataProvider.java
njr-11: Correct, it's the Jakarta EE product/runtime that supports CDI and can easily provide the CDI extension, and this gets us past the problem where some Jakarta Data providers don't want to depend on CDI themselves or directly interact with it at all.  I agree, I'll rephrase this to take out the confusing wording and try to state more directly, addressing both this review comment and the one before it and the one after at.  See commit https://github.com/jakartaee/data/pull/130/commits/3f27958aa294658ac1f9d96be110143da68b0558 which now states the purpose of this SPI more directly, and specifically for Jakarta EE profiles, and leaves the discussion about different options in Standalone and so forth to the spec doc where it is better written.

api/src/main/java/jakarta/data/spi/provider/package-info.java
njr-11: That wouldn't be accurate because it _does_ result in CDI extension being used (the one provided by the Jakarta EE product/runtime). I need to update it to say that. Thanks for spotting this. It's corrected in https://github.com/jakartaee/data/pull/130/commits/9cecee168fdcb791c8442dd7a82db2d72ecbacc5

otaviojava:
> CDI-centric literally meaning the approach is centered around the concept and usage of CDI. To your question, I think you're asking why use "@Inject" and not "https://github.com/resource"?

Yeap, why not use CDI extension, Inject, Dispose of CDI, and create this Provider?

spec/src/main/asciidoc/repository.asciidoc
njr-11: I think we keep getting tripped up and talking past each other because there are two modes for running - Standalone vs in a Jakarta EE profile and we each tend to be talking about the opposite one than the other is thinking of.  Hopefully additional language that I've been adding when addressing review comments will help.

If the Jakarta Data provider is running Standalone and so is the only provider in the system, then having a module that provides your own CDI extension that handles all `@Repository` injection points is just fine.  The Jakarta Data provider won't collide with any others on those injection points because there aren't any others. This pull should be consistent with allowing this.

If the Jakarta Data provider is running in a Jakarta EE profile where there can be multiple Jakarta Data providers and the Jakarta EE product runtime itself, then we have a collision if each provider and the Jakarta EE product runtime are all supplying a CDI extension to take over handling of all the injections points for `@Repository`.  Two meetings ago, we discussed two different idea for avoiding the collisions.  One idea was having the Jakarta Data providers only handle injection points annotated with their own special qualifier, which would resolve the collision but at the price of adding unnecessary burden onto the end user.  The other idea was that Jakarta Data providers could just register via a provider SPI with the Jakarta EE product and the Jakarta EE product runtime would coordinate them all from its single CDI extension, which is the approach covered under this pull.  This approach _does_ use what CDI provides to us, it's just the Jakarta EE product runtime that provides the CDI extension in this case, not the Jakarta Data providers all providing a bunch of competing CDI extensions for the same injection point.  We've already tried out the solution and it works quite well.  I know it's an inconvenience if JNoSQL ends up implementing the Jakarta Data provider SPI for running in Jakarta EE profile, and for running in Standalone has a separate module with its own CDI extension.  But it shouldn't be that much trouble.  The SPI is only a single interface with a few methods on it, some of which are one-liners like the provider name "JNoSQL" and supported entity annotation class `jakarta.nosql.Entity.class` and hopefully the others like creating a repository instance can just delegate to what you already have for your CDI extension.

api/src/main/java/jakarta/data/spi/provider/package-info.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * <p>This package defines the Service Provider Interface (SPI) by which
+ * Jakarta Data providers can register to provide repository instances
+ * to the Jakarta EE product's CDI extension that handles
```
otaviojava, line 22: @njr-11 @Emily-Jiang @frowe 
That is what I'm not following.

The reason to need to register to Jakarta Data and CDI.

We agree that should be CDI-centric; my point is we don't need this SPI; the CDI extension is complete.

E.g.:

You can find the repositories in the `BeforeBeanDiscovery`, and using the event `AfterBeanDiscovery ` you can register the repositories to CDI.

This interface has an [add bean](https://jakarta.ee/specifications/cdi/3.0/apidocs/) method to do without this SPI.

Why not use the CDI capabilities to make it happen?

Why should I make this mandatory if I can make it easier only using CDI?


If you take a look at Bean from CDI is exactly what you're looking for: 
https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#bean

My point is, CDI provides this discovery and it works like a charm.

njr-11: I'm pretty sure that Spring and Micronaut have not ever agreed to have their Jakarta Data providers be CDI-centric.  Even if they did, what you suggest would cause collisions in a Jakarta EE profile where multiple providers all try to handle the injection point with different CDI extensions.  What you are suggesting only works like a charm in a Standalone mode that only has one Jakarta Data provider.  The provider SPI here coordinates amongst all Jakarta Data providers and makes it work with no collisions on a single CDI extension, supplied by the Jakarta EE product/runtime.  The Jakarta EE product/runtime provides the CDI-centric solution when all of the Jakarta Data providers themselves cnanot.  It uses CDI events like `AfterBeanDiscovery` just like you are saying.  It's just the Jakarta EE product/runtime handling the CDI-centric aspect.  I don't see why this is being disputed so much.  Hopefully it is just a misunderstanding.

spec/src/main/asciidoc/repository.asciidoc
otaviojava: I got it; the primary goal of this provider is to avoid Ambiguous Exception CDI when it has multiple vendors.

```java
@Inject
Products products;
```

Given an imaginary scenario with multiple JPA providers, such as Hibernate and Eclipse Link JPA, how can this provider help you?

I mean, those providers will return the same entity on the method `supportedEntityAnnotations()`

njr-11: A single Jakarta Data provider backed by JPA should be able to cover all relational databases, so a scenario where you need to have multiple Jakarta Data providers all backed by JPA does not seem very likely/realistic.  That scenario is more of an edge case that could be covered by configuration (associating a repository type with a specific provider name) if the need for it ever arises.  The SPI does include a method to collect the provider name for this.

otaviojava: We can add if it makes sense, but please, let me understand and try to help.

The scenario here is to work with Jakarta NoSQL and Jakarta Data on the same server.

Can we assume that an entity will run only in a single database?

If yes, given a scenario where I have two entities, one to Jakarta NoSQL and one to Jakarta Data:

```java
@jakarta.nosql.Entity
class AirPlane{
....
}
```


```java
@jakarta.persistence.Entity
class Car{
....
}
```

Where we'll have two repositories:

```java
@Repository
public interface Airport extends CrudRepository<AirPlane, Long> {
}

@Repository
public interface Garage extends CrudRepository<Car, Long> {
}

```

In this case, when I inject both repositories, IMHO: there is no ambiguous exception once `Airport` is provided to the Jakarta NoSQL vendor, and `Garage` should be provided to the Jakarta Data vendor.


```java
@Inject//visible only to NoSQL
Airport airport;
@Inject//visible only to Relational database
Garage garage;
```

We need to document it somehow, such as: As the provider discovers and provides a repository that can support it; otherwise, it will ignore it.

On the code side, we can return the entity from the Repository with this code:


```java
public boolean isNoSQLEntity(Class<?> type) {
    Optional<Class<?>> entity = getEntity(type);
    return entity.map(c -> c.getAnnotation(jakarta.nosql.Entity.class) != null)
            .orElse(false);
}

public boolean isRelationalEntity(Class<?> type) {
    Optional<Class<?>> entity = getEntity(type);
    return entity.map(c -> c.getAnnotation(jakarta.persistence.Entity.class) != null)
            .orElse(false);
}

private Optional<Class<?>> getEntity(Class<?> repository) {
    Type[] interfaces = repository.getGenericInterfaces();
    if (interfaces.length == 0) {
        return Optional.empty();
    }
    ParameterizedType param = (ParameterizedType) interfaces[0];
    Type[] arguments = param.getActualTypeArguments();
    if (arguments.length == 0) {
        return Optional.empty();
    }
    Type argument = arguments[0];
    if (argument instanceof Class<?> entity) {
        return Optional.of(entity);
    }
    return Optional.empty();
}
```

So, you can check the repository type and register only where it makes sense:

```java
isNoSQLEntity(Airport.class); //true
isRelationalEntity(Airport.class);//false
isNoSQLEntity(Garage.class); //false
isRelationalEntity(Garage.class);//true
```

Hope this one can help. Please, if there is any concern.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -464,3 +464,28 @@ for (Pageable p = Pageable.ofSize(25).sortBy(Sort.desc("yearBorn"), Sort.asc("na
...
 
+== Jakarta Data Providers
+
+=== Modes
+
+Jakarta Data Providers implement the Jakarta Data API and specification requirements for one or more of the following modes.
+
+==== Standalone Mode
+
+Standalone mode means that the Jakarta Data provider is able to run on its own outside of a Jakarta EE platform or profile. Standalone Jakarta Data providers can choose whether or not to provide their own custom dependency injection leveraging `@Inject`, whether or not to provide extensions to tie into dependency injection via Jakarta Contexts and Dependency Injection (CDI) SE, whether or not to tie into the Jakarta Data Provider SPI, and whether or not to define other vendor specific ways of accessing Jakarta Data repositories. The Jakarta Data TCK includes a Standalone mode for validating Jakarta Data Providers that requires at least one of the first three of the aforementioned approaches.
```
KyleAure, line 475: This section very throughly explores the different options for a Jakarta Data Provider for Standalone implementations.  When it comes to the TCK I think we'll need some more documentation on the order in which we try to discover the provider and what happens if none is found. 

For now, I'll tag issue #119 but will likely open a new issue for this once this is finalized and merged. 
Thank you.

spec/src/main/asciidoc/repository.asciidoc
njr-11: Nice - I didn't realize they are now open to CDI support and pursuing it.  That really changes things a lot.  Here's a writeup in a separate pull #131 of some rules for CDI extensions that we could use to avoid conflicts is most cases and in edge cases cope with them.  The approach does work when I tried it out with multiple providers and no longer requires the central abstraction to coordinate.

otaviojava: @njr-11 are you running Jakarta NoSQL and Data at Open Liberty?

njr-11: > @njr-11 are you running Jakarta NoSQL and Data at Open Liberty?

The goal is to do so once the spec chooses a solution for multiple providers (could be either this pull or #131) and JNoSQL has a version that aligns with it.  In the mean time we are using other providers/mock providers to try things out.  Both approaches have shown to be workable.  We just need to choose one and more forward with it.

otaviojava: Nice, I like the other pull solution.

I'll give some suggestions.

njr-11:
Closing this because the solution from #131 was chosen instead.

PR #131, opened 2023-03-20
Avoiding conflicts between CDI extensions

njr-11:
Come up with a set of rules for CDI extensions to reduce/avoid colliding on the same repositories when there are multiple providers active in the system.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -470,13 +470,13 @@ To run in environments with Jakarta Contexts and Dependency Injection (CDI), Jak
 
 === Entity Annotation Class
 
-The `jakarta.persistence.Entity` annotation from the Jakarta Persistence specification can be used by repository entity classes for Jakarta Data providers that are backed by a Jakarta Persistence provider. Jakarta Data providers that are not backed by a Jakarta Persistence provider must not support the `jakarta.persistence.Entity` annotation.
+The `jakarta.persistence.Entity` annotation from the Jakarta Persistence specification can be used by repository entity classes for Jakarta Data providers that are backed by a Jakarta Persistence provider. Other Jakarta Data providers must not support the `jakarta.persistence.Entity` annotation.
 
-The `jakarta.nosql.Entity` annotation from the Jakarta NoSQL specification can be used by repository entity classes for Jakarta Data providers that are backed by NoSQL databases. Jakarta Data providers that are not backed by NoSQL databases must not support the `jakarta.nosql.Entity` annotation.
+The `jakarta.nosql.Entity` annotation from the Jakarta NoSQL specification can be used by repository entity classes for Jakarta Data providers that are backed by NoSQL databases. Other Jakarta Data providers must not support the `jakarta.nosql.Entity` annotation.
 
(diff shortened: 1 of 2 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -464,3 +464,22 @@ for (Pageable p = Pageable.ofSize(25).sortBy(Sort.desc("yearBorn"), Sort.asc("na
 }
 ----
 
+== Jakarta Data CDI Extension
+
+To run in environments with Jakarta Contexts and Dependency Injection (CDI), Jakarta Data providers each register their own CDI extension (`jakarta.enterprise.inject.spi.Extension`) to produce the bean instances that are defined via the `Repository` annotation and injected via the `Inject` annotation. The Jakarta Data specification employs several strategies for reducing and avoiding conflicts between Jakarta Data providers. The entity annotation class is used as the primary strategy. The Jakarta Data provider name serves as a secondary strategy.
+
+=== Entity Annotation Class
+
+The `jakarta.persistence.Entity` annotation from the Jakarta Persistence specification can be used by repository entity classes for Jakarta Data providers that are backed by a Jakarta Persistence provider. Jakarta Data providers that are not backed by a Jakarta Persistence provider must not support the `jakarta.persistence.Entity` annotation.
```
otaviojava, line 473: Maybe it is an issue for a non-native speaker, but It has a cognitive load to understand this point:

> Jakarta Data providers not backed by a Jakarta Persistence provider must not support the `jakarta.persistence.Entity` annotation.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -464,3 +464,22 @@ for (Pageable p = Pageable.ofSize(25).sortBy(Sort.desc("yearBorn"), Sort.asc("na
...
+
+The `jakarta.nosql.Entity` annotation from the Jakarta NoSQL specification can be used by repository entity classes for Jakarta Data providers that are backed by NoSQL databases. Jakarta Data providers that are not backed by NoSQL databases must not support the `jakarta.nosql.Entity` annotation.
+
+Jakarta Data providers can introduce and support custom entity annotations, but the class name of all supported entity annotation types must end with `Entity`. This enables Jakarta Data providers to know if a repository entity class contains entity annotations that belong to a different Jakarta Data provider without knowing in advance the full list of entity annotation types that exist across providers.
```
otaviojava, line 477: Once we don't give annotations, we could talk about it.
I mean, the provider should only recover the annotations the provider has supported; otherwise, ignore it.

If I want to support [JMolecules](https://github.com/xmolecules/jmolecules), I should be accessible to recovery once I have support for it.

njr-11: > I mean, the provider should only recover the annotations the provider has supported; otherwise, ignore it.
> 
> If I want to support [JMolecules](https://github.com/xmolecules/jmolecules), I should be accessible to recovery once I have support for it.

I think we're saying the same thing here.  The previous wording was intended to allow custom entity annotations such as the JMolecules Entity that you linked to.  I'll try to clarify it more in the [commit](https://github.com/jakartaee/data/commit/55ef83d91ea130962f5a6e2ed4e07b1eb449ef61) to address review comments.

otaviojava (approved):

PR #135, opened 2023-04-04
Minimal identifier to enable vendor configuration

njr-11:
Have a simple identifier that can optionally be specified on a repository that Jakarta Data providers can use to distinguish which vendor-specific configuration is wanted for the repository.  In a future version of the spec, this could be reused to identify a Jakarta Config base configuration path, but should hopefully be something that Jakarta Data providers can use in a vendor-specific way as well.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Repository.java
+++ b/api/src/main/java/jakarta/data/repository/Repository.java
@@ -500,6 +500,37 @@ public @interface Repository {
     static final String ANY_PROVIDER = "";
 
     /**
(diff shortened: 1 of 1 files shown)


KyleAure (approved):
I think this is a good starting place for getting some sort of configuration available that will allow users to have multiple repositories sending data to different data stores. 
👍🏼

njr-11:
This has been approved for 2 weeks without any objections raised, so I'm merging it.

PR #137, opened 2023-04-12
Address the TODO to reserve keywords for future

njr-11:
There is a TODO comment to reserve any keywords that we might want to add meaning for in the future.  This is a good idea so that we avoid breaking changes to applications that are written with v1.0 so that they can move up and get identical behavior without experiencing breakages.  It informs them of particular keywords to avoid.  In this pull, I looked over the JPQL language and identified areas that query-by-method-name might possibly some day cover, but doesn't currently.  This includes abilities like extracting a particular component of a date, length of a string or size of a collection, rounding.  We can reserve the keywords now in case we ever want to leverage these capabilities.

Here are a few examples of things you cannot do now, but conceivably could be added via what Jakarta Persistence/JPQL supports:
```
packages.findByWidthRoundedDown(12.0)
people.findByBirthdateWithMonthAndBirthdateWithDay(Month.MAY.getValue(), 9);
people.findByNameCharCount(4);
cities.findByNameTrimmed("Rochester")
customers.findByEmailAddressesElementCountGreaterThan(1);
```

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Repository.java
+++ b/api/src/main/java/jakarta/data/repository/Repository.java
@@ -331,6 +331,9 @@ import java.lang.annotation.Target;
  * {@code WithQuarter}, {@code WithSecond}, {@code WithWeek}, {@code WithYear}.
  * </p>
  * <p>
+ * Reserved for find...By and count...By: {@code Distinct}.
+ * </p>
+ * <p>
  * Reserved for updates: {@code Add}, {@code Divide}, {@code Multiply}, {@code Set}, {@code Subtract}.
  * </p>
  *
--- a/api/src/main/java/jakarta/data/repository/Repository.java
+++ b/api/src/main/java/jakarta/data/repository/Repository.java
@@ -317,6 +317,23 @@ import java.lang.annotation.Target;
  *
  * </table>
  *
+ * <h3>Reserved for Future Use</h3>
+ * <p>
+ * The specification does not define behavior for the following keywords, but reserves
+ * them as keywords that must not be used as entity attribute names when using
+ * Query by Method Name. This gives the specification the flexibility to add them in
+ * future releases without introducing breaking changes to applications.
+ * </p>
+ * <p>
+ * Reserved for query conditions: {@code AbsoluteValue}, {@code CharCount}, {@code ElementCount},
+ * {@code Rounded}, {@code RoundedDown}, {@code RoundedUp}, {@code Trimmed},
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/data/repository/Repository.java
```diff
@@ -317,6 +317,23 @@
  *
  * </table>
  *
+ * <h3>Reserved for Future Use</h3>
```
otaviojava, line 320: Should we have something for the future?

I mean, how can we make sure about this?

I feel more comfortable with the current code support.

njr-11: > Should we have something for the future?

Given that it gives the potential benefit for forward compatibility with future releases without any real downside, yes.  Various specifications that have keywords do this sort of thing.  Java even has some currently unused keywords.  I think the better question is whether more keywords should be included in our list.

> I mean, how can we make sure about this?

What are you unsure about? It doesn't burden providers at all. They don't need to do anything.  If we ever want to add one of these in a future release, we'll be able to do it in a point release and won't be forced into a major release just to add it, and we won't need to worry about breaking applications.  If we don't ever want to any of them, we don't need to.

njr-11:
Added a commit to also reserve `Distinct` in case we ever decide to allow findDistinct... like Spring documents here:
https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.query-creation

KyleAure (approved):
I think this looks good.  I don't really have any suggestions of what else we could add to the list right now.  

My only question would be is there a way we could disallow the use of these keywords?  Or in other words, should an implementation throw an exception if it sees any of these keywords?

api/src/main/java/jakarta/data/repository/Repository.java
njr-11: Otavio indicated he is okay with merging this one on today's Jakarta Data call, so I'm resolving this conversation and merging it.

njr-11:
> I think this looks good. I don't really have any suggestions of what else we could add to the list right now.
> 
> My only question would be is there a way we could disallow the use of these keywords? Or in other words, should an implementation throw an exception if it sees any of these keywords?

This was more to ask applications (rather than providers) not to use the keywords as entity properties in order to allow for the future flexibility.  Implementations could certainly aid that by choosing to raise errors, but the specification isn't forcing them to do so because it would prevent experimentation if any implementations wanted to try any of these out in advance of future.

PR #139, opened 2023-04-21
Implement read-only repositories for TCK

### LINKED ISSUE #136: Need a TCK strategy to deal with eventual consistency in NoSQL databases
I don't have a good answer for this, so I'm going to write up what I can think of. Maybe others will be able to improve on it or be able to come up with better approaches.

1. As much as reasonably possible, isolate TCK tests that require updates from those that can be achieved with just reads.
2. Have a pre-population phase that inserts the data for the latter, which is then considered the read-only data by the TCK, before running tests.
3. Introduce an `EVENTUAL_CONSISTENCY_DELAY` TCK configuration option that can be configured when running against a database with eventual consistency.
4. For tests that perform updates (and for the pre-population step), and if a `EVENTUAL_CONSISTENCY_DELAY` is specified, then after performing a group of updates, the TCK must first poll the database for an updated result to appear, and after seeing it, must sleep for the full amount of the `EVENTUAL_CONSISTENCY_DELAY` before querying again and making assertions on the state of the data.

With this approach, it will always be possible, although hopefully unlikely, for tests to fail because the TCK did not wait long enough for eventual consistency.  If this happens, the tester can react to this by increasing the delay and/or re-running the TCK.

KyleAure:
Fixes #136 

Partially tested in standalone mode, but we don't have any implementations that support standalone deployment to do a full-scale test.  This will have to be sufficient for now until we know how other vendors intend to do DI.


tck/src/main/java/ee/jakarta/tck/data/framework/read/only/NaturalNumbers.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+@Repository
+public interface NaturalNumbers extends CrudRepository<NaturalNumber, Long> {
+    
+    // READ - add more read only queries for test cases
+
+    
+    // POPULATOR - Nested class that handles populating repository
+    public static final Class<?> populator = WholeNumberPopulator.class;
+    public static class WholeNumberPopulator implements Populator {
```
njr-11, line 41: The entity name NaturalNumber and populator name WholeNumberPopulator should be made consistent.  Either:
NaturalNumber / NaturalNumbers / NaturalNumberPopulator or WholeNumber / WholeNumbers / WholeNumberPopulator

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/ReadOnlyRepository.java
```diff
@@ -0,0 +1,44 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+package ee.jakarta.tck.data.framework.read.only;
+
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import jakarta.data.repository.DataRepository;
+
+//FIXME - Are user defined repository interfaces like this allowed via the Specification?
```
njr-11, line 23: Good question.  I'm guessing that you are asking about repository interfaces with type parameters for the entity and id, but which don't inherit from a built-in interface like DataRepository or CrudRepository. The spec does let you use your own interface, but doesn't clearly define for that case how the entity type and id type are determined, although it makes sense to infer them from the type parameters in the same way as built-in interfaces.  Probably we just need to add the extra clarity for this case, but until then, it should be avoided in the TCK and the well-defined approach used instead.

njr-11 (commented):
This is an excellent start. I added a few minor comments/questions to think about.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/AsciiCharacter.java
```diff
@@ -0,0 +1,81 @@
...
+@jakarta.persistence.Entity
+public class AsciiCharacter implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    @jakarta.nosql.Id
+    @jakarta.persistence.Id
+    private long id;
+
+    @jakarta.nosql.Column
+    private int decimal;
```
njr-11, line 30: `DECIMAL` is a reserved word in some databases.  Maybe rename the field to `numericValue` or use the annotations to give it a different name in the database?

njr-11 (approved):
Looks great.  I had one minor comment about a possible conflict with a database reserved word, and it might need updating to allow generic dependency injection for providers that don't support CDI, but those aspects can be covered later.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/AsciiCharacter.java
KyleAure: Good catch, fixed 👍🏼

PR #140, opened 2023-05-04
JPA specialization [Proposal]

otaviojava:
Hey everyone, this is one proposal.

I was concerned about missing the DDD repository's core idea during the last meeting.
Our API lost encapsulation and has several JPA details such as transaction, SQL, JPQL, JTA, etc.

I propose to create a subdomain JPA where the ideas around JPA can move there and keep the CrudDataRepository the most generic to work to any persistence layer engine, such as NoSQL types, web services, and so on.

A relational database is the most advanced and mature database. It does not make sense to wai others simultaneously; the CrudRepository should be more generic to work with any persistence layer. Thus, moving JPA forward without impacting the other types is an excellent benefit.

Please, let me know if it makes sense; it still left a lot of work on documentation, but first, let me know if this proposal makes sense.

:warning: It is unfinished work, the idea is to show the vision with some code, so it will be easier to understand.

```java
@JPARepository (dataStore="JNDI-REFERENCE")
public interface Airport extends JpaCrudRepository<AirPlane, Id>{
}
```

Some market references:

* [Spring](https://docs.spring.io/spring-data/jpa/docs/current/api/org/springframework/data/jpa/repository/JpaRepository.html)
* [Micronaut](https://micronaut-projects.github.io/micronaut-data/1.0.0/api/io/micronaut/data/jpa/repository/JpaRepository.html)
* [Quarkus](https://quarkus.io/guides/hibernate-orm-panache#defining-your-repository)


KyleAure (changes requested):
Overall, I do not agree with this change. 
I thought the point of Jakarta Data was to have a persistence layer that was agnostic (in your words) to the underlying database. If we separate off Persistence repositories into their own package then as a vendor or user what benefit do I get out of Jakarta Data? I could just switch and use Jakarta Persistence. 

The whole point of Jakarta Data is that I can write my application using a Data Repository.  Right now I want my underlying database to be Oracle, but in the future, I might want to switch to use MongoDB without having to do a massive refactor of my application. 

This change would break that goal and make this specification useless (IMHO)

api/src/main/java/jakarta/data/repository/jpa/SQL.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package jakarta.data.repository.jpa;
+
+import jakarta.data.repository.Page;
+import jakarta.data.repository.Sort;
+
+/**
+ * Defines the SQL string as SQL, JPA-QL that should be executed.
+ */
+public @interface SQL {
```
njr-11, line 26: JPQL is not the same as SQL. Our specification should not confuse the two.

api/src/main/java/jakarta/data/repository/Repository.java
```diff
@@ -537,38 +538,8 @@
...
-     * version of this specification. For providers backed by Jakarta Persistence,
-     * it might point to a {@code jakarta.annotation.sql.DataSourceDefinition} name
-     * or a {@code javax.sql.DataSource} JNDI name or resource reference,
-     * or other vendor-specific configuration.</p>
-     *
-     * <p>The default value of this attribute is {@link #DEFAULT_DATA_STORE}.</p>
-     *
-     * @return the name of a data store or {@link #DEFAULT_DATA_STORE}.
-     */
-    String dataStore() default DEFAULT_DATA_STORE;
```
njr-11, line 571: This is not specific to Jakarta Persistence.  It applies to every type of backend where there can be multiple, which is all types I'm aware of.

api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -43,24 +44,5 @@
...
-     *
-     * <p>The default value of empty string indicates that no counting query
-     * is provided. A counting query is unnecessary when pagination is
-     * performed with slices instead of pages and when pagination is
-     * not used at all.</p>
-     *
-     * @return a query for counting the number of elements across all pages.
-     *         Empty string indicates that no counting query is provided.
-     */
-    String count() default "";
```
njr-11, line 64: This is not specific to Jakarta Persistence.  Page counts and total element counts across all pages apply to every type of database that supports pagination.

api/src/main/java/jakarta/data/repository/jpa/KeysetAwareSlice.java
```diff
@@ -15,7 +15,11 @@
  *
  *  SPDX-License-Identifier: Apache-2.0
  */
-package jakarta.data.repository;
+package jakarta.data.repository.jpa;
```
njr-11, line 18: Keyset pagination is not specific to Jakarta Persistence.  It also applies to JDBC and to any database type that supports the `And` and `Or` keywords.  And in simple scenarios with only one sort criterion, it doesn't even need `Or`.

api/src/main/java/jakarta/data/repository/jpa/OrderBy.java
```diff
@@ -15,7 +15,11 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-package jakarta.data.repository;
+package jakarta.data.repository.jpa;
```
njr-11, line 18: Nothing about this is specific to Jakarta Persistence.

api/src/main/java/jakarta/data/repository/jpa/KeysetAwarePage.java
```diff
@@ -15,7 +15,10 @@
  *
  *  SPDX-License-Identifier: Apache-2.0
  */
-package jakarta.data.repository;
+package jakarta.data.repository.jpa;
```
njr-11, line 18: Keyset pagination is not specific to Jakarta Persistence.  It also applies to JDBC and to any database type that supports the `And` and `Or` keywords.  And in simple scenarios with only one sort criterion, it doesn't even need `Or`.

api/src/main/java/jakarta/data/repository/jpa/JPARepository.java
```diff
@@ -0,0 +1,598 @@
...
+ * environment where the Jakarta EE technology that provides the interceptor is available.</p>
+ */
+// TODO When can "By" be omitted and "All" added? Need to document that.
+// TODO Does Jakarta NoSQL have the same or different wildcard characters? Document this
+//       under: "Wildcard characters for patterns are determined by the data access provider"
+// TODO Ensure we have all required supported return types listed.
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface JPARepository {
```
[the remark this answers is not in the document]
otaviojava: We can use it: Jakarta Persistence API!

otaviojava:
@KyleAure 
> I thought the point of Jakarta Data was to have a persistence layer that was agnostic (in your words) to the underlying database.
We already broke it when we put a lot of transactions and JPA details on it.

Technically, I cannot affirm that all database engine has transaction and JPQL.

I'm proposing how to solve it.

njr-11 (changes requested):
The concept of having a Jakarta Persistence/JPA-specific extension to CrudRepository as Spring and Micronaut do could certainly be useful if done in the same manner as Spring and Micronaut, which is to add built-in repository methods that are unique to Jakarta Persistence.

The approach taken under this pull does not do that and instead appears to be a misguided attempt to group together and isolate aspects of the specification that are unfavored for NoSQL, calling these "JPA".

JPA itself is not a correct acronym in Jakarta. JPA stands for Java Persistence API, which Oracle owns as part of Java EE. In Jakarta, it is just "Jakarta Persistence". Refer to the [Jakarta Persistence specification document](https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1.html), in which the acronym JPA only appears when referencing Java EE's Java Persistence API.

Jakarta Data and the repository pattern within are much broader than Jakarta Persistence and Jakarta NoSQL and must not be characterized as one versus the other. Beyond these, a provider might instead implement repositories on JDBC, which shares some, but not all, aspects with Jakarta Persistence. A provider could alternatively be backed by REST (I think that was in Otavio's original proposal) or really anything that accesses and persists data. A database could even skip over Jakarta NoSQL, Jakarta Persistence, JDBC and REST altogether and directly supply a Jakarta Data provider for its database to be usable in a compliant way under the specification and TCK.

We should not arbitrarily forbid such implementations from providing features like keyset pagination, JTA transactions, or multiple database support under the specification just because they aren't Jakarta Persistence.  Jakarta Data should stay with its current approach of seeking to generalize and unify rather than fragment the programming model, especially where the fragmentation is inaccurate.

api/src/main/java/jakarta/data/repository/jpa/KeysetAwareSlice.java
otaviojava: If we renamed it to "relational" is that ok?

api/src/main/java/jakarta/data/repository/jpa/KeysetAwarePage.java
otaviojava: As above, if we renamed the package name to "relational"

api/src/main/java/jakarta/data/repository/Repository.java
otaviojava: Usually, we're using convention over configuration allied with The Twelve-Factor App:
https://12factor.net/
E.g.:

* [Spring](https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html#appendix.application-properties.data-migration)
* [Micronaut](https://guides.micronaut.io/latest/micronaut-jpa-hibernate-gradle-java.html#jpa-configuration)
* [Quarkus](https://quarkus.io/guides/config)

otaviojava:
> Jakarta Data and the repository pattern within are much broader than Jakarta Persistence and Jakarta NoSQL and must not be characterized as one versus the other. Beyond these, a provider might instead implement repositories on JDBC, which shares some, but not all, aspects with Jakarta Persistence. A provider could be backed by REST (I think that was in Otavio's original proposal) or anything that accesses and persists data. A database could even skip over Jakarta NoSQL, Jakarta Persistence, JDBC and REST altogether and directly supply a Jakarta Data provider for its database to be usable in a compliant way under the specification and TCK.

That is right! And that is our current bug :lady_beetle: 

I've created an issue:

https://github.com/jakartaee/data/issues/143

My question is: the relational work that does here makes, how to keep without impacting the other persistence engines?

api/src/main/java/jakarta/data/repository/Query.java
otaviojava: It is not valid.
You can paginate using a stream or any "pointer" or cursor of this data.

Samples:

* [Cassandra](https://docs.datastax.com/en/developer/java-driver/2.1/manual/paging/)
* [MongoDB](https://isotropic.co/how-to-implement-pagination-in-mongodb/)
* [Neo4J Cursos](https://neo4j.com/docs/graphql-manual/current/pagination/cursor-based/)



:warning: Neo4J also have [offset-based option](https://neo4j.com/docs/graphql-manual/current/pagination/offset-based/)

njr-11:
> We already broke it when we put a lot of transactions and JPA details on it.
> 
> Technically, I cannot affirm that all database engine has transaction and JPQL.

As we are all aware, not all databases do support transactions and JPQL.  However, you have it backwards here.  Defining how Jakarta Transactions and Jakarta Data specifications interact is obligatory of a well written specification in the Jakarta platform. Neglecting to define how Jakarta Transactions and Jakarta Data specifications interact is what would mean a broken specification.  It would be a broken specification if we were requiring databases without transaction support to participate in transactions.  But the spec does the opposite and applies requirements with respect to Jakarta Transactions only to data sources that are capable of transaction enlistment.  If the language isn't clear, it can be clarified further.  It's the wrong answer to say that only Jakarta Persistence-backed data sources are transactional and codify that into an API because it isn't true.

api/src/main/java/jakarta/data/repository/Query.java
njr-11: Right, I didn't word that very well. The point is that although not all databases can count total results, types other than Jakarta Persistence/JPA do have that ability and it is therefore wrong to restrict it to Jakarta Persistence/JPA alone.  We know for a fact that JDBC can do this via SQL independently of Jakarta Persistence.  Even MongoDB which you cited as a counter example exposes an [API for counting documents](https://www.mongodb.com/docs/manual/reference/method/db.collection.count/) so it is not unreasonable to allow for the possibility they might some day want to let the same thing to be done via whatever query language they support.  If a count query isn't useful for a particular type that doesn't support it, then just don't use it with those types.  But let it remain for all of those where it useful.  It seems like you are boxing it and limiting everything that isn't Jakarta Persistence where it isn't necessary and doesn't make sense.

api/src/main/java/jakarta/data/repository/Repository.java
njr-11: I don't follow this argument at all with respect to what it has to do with JPA being the deciding factor in whether or not you are able to use multiple databases.

njr-11:
> My question is: the relational work that does here makes sense and I don't want to remove it, but to keep it without impacting others' engine solutions. But how?

Can you point out specifically what is impacting other persistence types that you believe would prevent them from having a compliant Jakarta Data provider?  We have been very intentionally about writing language into the spec that doesn't ask for enlistment into JTA transactions or features like keyset pagination (relies on `Or` capability) when the database doesn't have the capability.  TCK tests don't exist yet, but our plan involves catching expected exceptions that indicate lack of support for `Or` and of course JTA-related tests will be limited to web profile and Jakarta Persistence-backed providers.  I'm not aware of any actual problems that these proposed changes fix.

otaviojava:
> Can you point out specifically what is impacting other persistence types that you believe would prevent them from having a compliant Jakarta Data provider? We have been very intentionally about writing language into the spec that doesn't ask for enlistment into JTA transactions or features like keyset pagination (relies on Or capability) when the database doesn't have the capability. TCK tests don't exist yet, but our plan involves catching expected exceptions that indicate lack of support for Or and of course JTA-related tests will be limited to web profile and Jakarta Persistence-backed providers. I'm not aware of any actual problems that these proposed changes fix.

Ok, I'll list those, hope it will be better than this PR.

PR #145, opened 2023-05-04
clarifications on inject and other entity models

### LINKED ISSUE #144: Simplify Repository annotation
Hey, some points would be nice to clarify at the Repository annotation:

> This class is a CDI bean-defining annotation when CDI is available....

Can we remove this point? Once it is DI agnostic, it needs to make sense.

>  Jakarta Persistence and Jakarta NoSQL define entity models you may use in Jakarta Data.

It is also not true; I can use JSON-B with Webservice, for example.

I prefer to keep the first two paragraphs because the other details are on the spec page.

njr-11:
fixes #144


njr-11:
@otaviojava this adds clarifications for #144

otaviojava (approved):

Issue #128, opened 2023-03-15
List of database limitations that TCK must allow for


njr-11:
To write a TCK that is valid for all types of providers, we need a list that can be referred to for identifying which capabilities might be limited by the underlying database, along with the spec-defined way to detect this and consider it a passing result.  Please add this information to the following table, or if you don't have edit authority, then add as a comment and one of the committers can put it in the table for you.

Please be as specific as possible when adding information. For example, if the limitation is only with NoSQL key-value, please indicate that information.

| Unavailable function | Type | How to Detect | Additional Detail |
| --------------------- | ---- | --------------- | ---------------- |
| `Or` operation | NoSQL | `MappingException` | |
| keyset pagination with multiple sorting (relies on `Or`) | NoSQL | `MappingException` | |
| `Page.totalElements` and `Page.totalPages` | NoSQL | `UnsupportedOperationException` from repository method that returns `Page` or `KeysetAwarePage` | |
| please add others that you know of ... | ... | ... | ... |

PR #146, opened 2023-05-04
Page.totalElements and Page.totalPages might not be supported

njr-11:
@otaviojava the issue you identified about some databases not being able to count results is a good one that does need to be addressed in a different place -- on the Page interface, where it will be impossible to implement these methods:

- https://github.com/jakartaee/data/blob/4f554badd219d63a18e6c04f31a3a92d65747d01/api/src/main/java/jakarta/data/repository/Page.java#L32
- https://github.com/jakartaee/data/blob/4f554badd219d63a18e6c04f31a3a92d65747d01/api/src/main/java/jakarta/data/repository/Page.java#L38

The spec needs to define the behavior here so that both users and the TCK can know what to expect.  This pulls updates JavaDoc to state that UnsupportedOperationException is raised from these methods when the database is incapable of counting the results.  The TCK will then be able to attempt the method and consider the UnsupportedOperationException and the correct number to both be passing results for the test, whereas returning an incorrect number will be a failure of the test.  We should update #128 for the TCK after merging this.


otaviojava:
@njr-11 
Maybe I'm not being clear; sorry for that.

I don't want to remove anything; I want to solve this miss encapsulation using the power of OOP.

Let's use Workers as samples where we have series such as Chef, Lawyer, Driver, and Doctor.

```java
public class Worker {
//it works if you are a driver
DriverLicense driverLicense();
//works if you are a doctor
void care(Person person);
//works if you are a Chef
void prepare(Food food);
//work with everybody
void work();
}
```

In this sample, the design is unsafe even with documentation because we need to read the documentation constantly. Another area for improvement is if I want to include another occupation, I need to change the Worker class, which breaks the open-close principle, which is not cool.


To solve it, we have two options:

## 1 polymorphism

1) The first option is to create an interface and let the implementation handle their details:

```java
interface Worker{
void work();
}

class Driver implements {

@Overwrite
void work():

DriverLicense driverLicense();
}

Docker implements Worker{

@Overwrite
void work():

void care(Person person);

}
```

In this scenario, I did not change the behavior; I moved the personal behavior to each implementation and got the Worker compatible with everyone; in this case, we're using **polymorphism**. That is fine, and Java can do it.

## 2 Specializations

Every implementation needs to define the documentation, and we want to centralize that behavior while isolating the worker. Thus, we can use specializations.

```java
interface Worker{
void work();
}

interface Driver extends Driver {

DriverLicense driverLicense();
}
```

We want the second option to document the JPA simultaneously not to impact the other persistence engine and keep the design safe.


```mermaid
graph TD;
    DataRepository-->CrudRepository;
    CrudRepository-->PageableRepository;
    PageableRepository-->RelationalPageableRepository-;
```

The CrudRepository must be generic enough to hold everyone with a persistence engine and use specializations or implementations to hold it.

Can a JPA use CrudRepository and JPA?
Yes, the user can because RelationalPageableRepository, or any name, is a specialization of both PageableRepository and CrudRepository.


:warning: I'm not against documentation; I love it, but we must also create a safe code design.

otaviojava:
The considerable debate concerns me a lot.

I have mixed feelings, such as a premature release if I'm too precious about it.

I've heard that we cannot break the API on Jakarta EE once we release a version, but I don't know if it is true.

But let's break problems into small pieces; we can focus on the TCK engine this week and then return to this.

otaviojava:
@njr-11 I'll do my best to create the meeting again :)

api/src/main/java/jakarta/data/exceptions/FunctionalityNotSupportedException.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public class FunctionalityNotSupportedException extends MappingException {
+    private static final long serialVersionUID = 4093503320173860847L;
+
+    /**
+     * Constructs a new FunctionalityNotSupportedException exception with the specified detail message.
+     *
+     * @param message the detail message.
+     */
+    public FunctionalityNotSupportedException(String message) {
```
otaviojava, line 34: why not use the [UnsupportedOperationException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/UnsupportedOperationException.html)?

njr-11: Good idea. No one on the call had thought of that. I agree, it's best to reuse a standard exception that fits.  It is updated now in [this commit](https://github.com/jakartaee/data/pull/146/commits/a700142e37878b94d0981ab3a383e3f60046f09d), please review again.

otaviojava (approved):

PR #189, opened 2023-07-12
Remove Reactive Repository and switching to single backticks

mswatosh:
The PR is split into to commits. The first covers minor updates to the spec:

- Removing reference to the ReactiveRepository since that is being pushed to a later spec version
    - Replacing the repositories image with a diagram using ditaa (an asciiart to diagram plugin for asciidoctor)
- A few grammer and clarification updates

The second commit switches from \`\`\` to \` for inline code. Generally \`\`\` is used for multiline code in asciidoctor.


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -32,17 +32,33 @@ A repository abstraction aims to significantly reduce the boilerplate code requi
 
 The parent interface in Jakarta Data repository abstraction is DataRepository.
 
-By default, Jakarta Data has support for three interfaces. However, the core is extensible. Therefore, a provider might extend one or more interfaces to a specific data target.
-
-image::02-repositories.png[alt=Repositories types]
```
KyleAure, line 37: Should this diagram be removed if it is no longer valid?

njr-11 (approved):

spec/src/main/asciidoc/repository.asciidoc
mswatosh: Good catch, I added a commit to remove it

PR #192, opened 2023-07-24
Release 1.0.0 b3

otaviojava:
## Changes

- Release version to 1.0.0-b3


njr-11 (approved):
Manual updates look good. For the next beta, we should try to use the automation and/or get it in place if it isn't already there.

PR #208, opened 2023-08-15
Enhanced Documentation: Explanation of Jakarta Data Extension Features

otaviojava:
## Pull Request Message

Enhanced Documentation: Explanation of Jakarta Data Extension Features

## Description

This Pull Request addresses the documentation enhancement regarding Jakarta Data's extension features for Java persistence. The changes provide a more comprehensive understanding of the features and their benefits to Java developers. Additionally, the modifications include examples and explanations of how these extension features can streamline code design and promote code reusability.

### Changes

- Added an introduction section providing context on the extension features.
- Expanded explanations for Default Methods, Interface Queries, and Custom Repositories.
- Highlighted the ability to combine Default Methods with Interface Queries for enhanced functionality.
- Included examples of code snippets illustrating each feature.
- Introduced bullet-point lists enumerating benefits for each extension feature.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -8,6 +8,10 @@ and this project adheres to https://semver.org/spec/v2.0.0.html[Semantic Version
 
 == [Unreleased]
 
+=== Added
+
+- Define Jakarta Data extensions
+
 == [1.0.0-b3] - 2022-07-24
 
 === Added
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -513,7 +513,7 @@ public interface BookRepository extends PageableRepository<Book, String> {
 }
 ----
 
-In this Interface, the `releaseThisYear` method is a default method that utilizes the `findByYear` method to retrieve books published in the current year. Additionally, the `register` method provides a default implementation for registering a book and an event.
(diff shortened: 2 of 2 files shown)


njr-11 (changes requested):
The sections on default methods and interfaces are great.  I suggested a few minor grammar fixes to example method names.

Let's split out the second on "Custom Repositories" to a separate issue/pull.  If we add this, it will need more work.

This proposal lacks the important detail of how a provider locates implementations and determines that ProductAuditRepositoryImpl is the right implementation to use versus any others that might be found, and how to disambiguate.  Similarly, it is missing the information for the application on how to indicate that ProductAuditRepositoryImpl is the correct implementation to pair with ProductAuditRepository.

Also, the term "Custom Repositories" is misleading because these aren't repositories.  They are a set of user-provided supplemental methods that can go along with a repository, but not an actual repository.  The term "Custom Repositories" would better be used to describe what customers do in the general case by composing their own repository interfaces.  In this example, that would be "ComputerRepository", which is actually a repository, not "ProductAuditRepository", which although named as though it is, is actually not a repository at all.

Also, given that the only advantage that I can see for what is termed as "Custom Repositories" here over default methods is the ability to use `@Inject` rather than CDI.instance.select to avoid the need for annotation literals, the example ought to include that, and even before that, we need to verify that it will actually work.

njr-11 (approved):

PR #211, opened 2023-08-17
obtaining the EntityManager and other helpful provider info

### LINKED ISSUE #202: Obtaining the EntityManager (or NoSQL equivalents) for a repository
The following pattern was discussed/proposed in the Jakarta Data Meeting on August 2nd:

```
@MyQualifier
@PersistenceContext(unitName = "MyPersistenceUnit")
@Produces
EntityManager entityMgr;

void dispose(@Disposes @MyQualifier EntityManager em) {
    em.close();
}

...

public interface Cars extends CrudRepository<Car, Integer> {
    default void someAdvancedOperation(int param1, int param2) {
        EntityManager em = CDI.current().select(EntityManager.class, MyQualifier.Literal.instance()).get();
        ... Use the entity manager for advanced JPA operations
    }
}
```

I later realized there is a problem with it in that it relies on the Jakarta Data user to supply the JPA configuration and corresponding EntityManager in the first place.  Some providers might externalize things this way, but Jakarta Data intentionally does not require it (by request of one of the providers and I also agree with them) because other JPA-based Jakarta Data providers might generate all of the JPA related stuff internally.  A spec solution to obtaining the entity manager (and other similar resources if NoSQL) needs to be something that will work for all providers.

It turns out there's a solution to obtaining underlying resources that is already in Jakarta EE and Java SE.

In Jakarta Persistence (JPA) you can invoke an unwrap method on an EntityManager to obtain the underlying resources that it uses, such as the java.sql.Connection,
```
    Connection connection = entityManager.unwrap(Connection.class);
```

Similarly, there is also an unwrap on several JDBC interfaces, such as,
```
    OracleConnection oc = connection.unwrap(OracleConnection.class);
```

We could continue this pattern to be able to obtain the underlying EntityManager for a repository with,
```
     EntityManager em = myRepository.unwrap(EntityManager.class);
```

which could be used from a default method:
```
public interface Cars extends CrudRepository<Car, Integer> {
    default void someAdvancedOperation(int param1, int param2) {
        try (EntityManager em = unwrap(EntityManager.class)) {
            ... Use the entity manager for advanced JPA operations
        }
    }
}
```

If we do this, we would need to decide whether to add the unwrap method to the base DataRepository interface, or to a separate mix-in interface (we would need a name for it).

@otaviojava 
@georgleber

njr-11:
Here is one possible solution to obtaining the EntityManager and other types of resources from the Jakarta Data provider for advanced scenarios.  It introduces a mix-in interface that applications can optionally include on a repository, that will be implemented by the Jakarta Data provider.  On this interface I included a generic method for obtaining types of underlying resources from the Jakarta Data provider as well as a few other helpful methods to query for the provider anme and whether or not a repository keyword is supported by the database.  This will be a much cleaner approach for the TCK to detect when keywords like "Or" are available to use or not, and will also enable the TCK to avoid the vulnerability which the TCK currently has with its approach of catching exceptions where it would wrongly consider a test to pass if a Jakarta Data provider raises one of these exceptions for the wrong reason.

fixes #202

### DIFF
--- a/api/src/main/java/jakarta/data/repository/RepositoryAssist.java
+++ b/api/src/main/java/jakarta/data/repository/RepositoryAssist.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.repository;
+
+import java.util.Optional;
+
+/**
+ * <p>A mix-in interface for repositories that provides access to additional information
+ * and other function from the Jakarta Data provider.</p>
+ *
+ * <p>This interface is implemented by the Jakarta Data provider when the application adds
+ * it to a repository, as follows:</p>
+ *
+ * <pre>
+ * &#64;Repository
+ * public interface Products extends CrudRepository&lt;Product, String&gt;, RepositoryAssist {
+ *     ...
+ *
+ *     default Collection&lt;Product&gt; lowPricedOrHighlyDiscounted(float maxPrice, float minDiscount) {
+ *         if (supportsKeyword("Or")) {
+ *             return findByPriceLessThanOrDiscountGreaterThan(maxPrice, minDiscount);
+ *         } else {
+ *             Set&lt;Product&gt; merged = new HashSet&lt;&gt;();
+ *             merged.addAll(findByPriceLessThan(maxPrice));
+ *             merged.addAll(findByDiscountGreaterThan(minDiscount));
+ *             return merged;
+ *         }
+ *     }
+ * }
+ * </pre>
+ */
+public interface RepositoryAssist {
+
+    // TODO are there any resource types for NoSQL that would be useful?
+    /**
+     * <p>Requests that the Jakarta Data provider supply a type of resource, such as
+     * {@code jakarta.persisence.EntityManager} (if backed by Jakarta Persistence) or
+     * {@code javax.sql.DataSource} or {@code java.sql.Connection} (if backed by JDBC),
+     * to a repository default method.</p>
+     *
+     * <p>If the resource type implements {@code AutoCloseable}, the repository
+     * default method is expected to close the resource instance before returning.
+     * For example,</p>
+     *
+     * <pre>
+     * &#64;Repository
+     * public interface Cars extends CrudRepository&lt;Car, Long&gt;, RepositoryAssist {
+     *     default Car[] advancedSearch(SearchOptions filter) {
+     *         try (EntityManager em = getResource(EntityManager.class).orElseThrow()) {
+     *             ... use entity manager
+     *             return results;
+     *         }
+     *     }
+     * }
+     * </pre>
+     *
+     * <p>If the repository default method does not close the resource instance,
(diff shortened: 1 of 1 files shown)


otaviojava:
Amazing this design discussion!

I believe that the [Custom Implementations for Spring Data Repositories](https://docs.spring.io/spring-data/data-commons/docs/current/reference/html/#repositories.custom-implementations) is one thing that we can take more advantage of.

The hard part is how to standardize, I enjoy this magic way, and with such a powerful engine as CDI, we can make it even more magic!

njr-11:
> I believe that the [Custom Implementations for Spring Data Repositories](https://docs.spring.io/spring-data/data-commons/docs/current/reference/html/#repositories.custom-implementations) is one thing that we can take more advantage of.
> 
> The hard part is how to standardize, I enjoy this magic way, and with such a powerful engine as CDI, we can make it even more magic!

There isn't anything magic in this pull, it's just an interface to let the application ask the Jakarta Data provider for what the application needs (the EntityManager in this case).
If you have ideas on how alternative approaches building on what you suggested can enable the application to obtain the EntityManager from the Jakarta Data Provider, please elaborate on the details.  Thus far, I've only seen how these other approaches can be used to let the application provide the EnityManager to itself if it already has the EntityManager.  That isn't the problem I'm trying solve.  The application doesn't already have the EntityManager and needs it.

otaviojava:
I am still thinking about this case @njr-11 
I am not sure.

If the goal is to obtain the EntityManager from the repository, why not inject it in the wrapper?

IMHO: Let's hold it for a while and let the vendors try it. For sure, we can go deep on it in the next meeting

njr-11:
> If the goal is to obtain the EntityManager from the repository, why not inject it in the wrapper?
> 
> IMHO: Let's hold it for a while and let the vendors try it. For sure, we can go deep on it in the next meeting

I won't be at the next meeting. Could you elaborate more on your idea here, I'm not sure what "inject it in the wrapper" means.  Are you thinking something like this?

```
    public default myDefaultMethod(int var1, String var2, EntityManager em) {
         ... use entity manager
    }
```

where the application would need to supply a `null` value and the Jakarta Data provider replaces it with the entity manager instance?

```
   myRepository.myDefaultMethod(1, "stringvalue", null);
```

That would be a valid option, although the null is awkward.  We could do that if that's what you are proposing.

otaviojava:
Indeed, let's dive into the explanations using the sample code you provided:

1. **Using a Service layer by the user itself:**

The proposed `UserService` class with injected `UserRepository` and `EntityManager` raises concerns about exposing repository core details:

```java
public class UserService {
    // Injection
    private UserRepository repository;
    // Injection
    private EntityManager em;
}
```


2. **EntityManager as the default method**

The approach of passing `EntityManager` as a parameter in a default method like this:

```java
public default void myDefaultMethod(EntityManager em) {
    Objects.requiredNonNull(em, "entity manager is required");
//the rules here
}
```


3. **The RepositoryAssist (this PR):**

Moreover, it's important to consider how these approaches align with the core principles of Domain-Driven Design (DDD). DDD emphasizes encapsulation and the separation of concerns to create a clear boundary between the domain logic and infrastructure details. This separation allows domain logic to remain independent and focused on business rules, promoting maintainability and adaptability over time.

The Optional getResource method, where the repository's internals are directly exposed, can be seen as violating these DDD principles. In DDD, repositories serve as gateways to the underlying data storage and should shield the domain layer from data access details. By allowing domain logic to access or manipulate the repository internals directly, you risk tightly coupling the domain logic to infrastructure details. It can lead to challenges when evolving the system, as changes to the repository's implementation may inadvertently impact the domain logic.

4. **Vendor Solution (My favorite, until we get more maturity on this, than we can move to the API):**

Allowing vendors to decide how to integrate with the available technologies is a reasonable suggestion. By giving vendors the flexibility to choose between different integration approaches, such as using the maturity of Spring or a specific solution like `RepositoryAssist`, you acknowledge the diversity of needs and preferences in the development community. This approach promotes adaptability and empowers vendors to make informed decisions based on their project requirements, existing knowledge, and technical constraints.

njr-11:
I'm confused by your answer. Were parts of it AI generated? It seems to be arguing against this particular approach on the grounds that it would expose the EntityManager, which is the whole point of what we're trying to achieve with this issue. All of the proposed solutions (except for 1 where the problem isn't applicable because it assumes you start by already having the EntityManager) are aiming to achieve that.

njr-11:
Close in favor of #248

Issue #216, opened 2023-08-22
Consider not requiring for repositories to trigger bean validation by default


dstepanov:
> When you invoke the save or saveAll method on repositories, the Jakarta Data provider ensures that the validation rules defined in the entity are automatically executed prior to updating the database if a Jakarta Validation provider is present. For instance, when persisting a Student object, the name field will be validated to ensure it is not blank.

This kind of validation might be unwanted for performance reasons, or validation might have happened before and is unnecessary.
The latest version of Micronaut Data requires the repository type arguments to be annotated for the entity or ID to be validated:

```
@Repository
public interface School extends PageableRepository<@Valid Student, String> {
}
```

otaviojava:
@dstepanov Hey, how are you?
Yeap, the goal is to make it optional as any integration with Jakarta EE specs. 

https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/jakarta-ee.adoc

dstepanov:
Sorry, my knowledge of the Jakarta EE is limited. I just saw that note and assumed the repository should always check for validation. We actually removed the implicit validation in Micronaut 4.

otaviojava:
There is no sorry; we are here to learn together!

What you said makes sense; it should not make this validation.

I will create a PR explicating the integration which is not mandatory. It is possible and period.

njr-11:
> This kind of validation might be unwanted for performance reasons, or validation might have happened before and is unnecessary. The latest version of Micronaut Data requires the repository type arguments to be annotated for the entity or ID to be validated:
> 
> ```
> @Repository
> public interface School extends PageableRepository<@Valid Student, String> {
> }
> ```

This is an excellent point from @dstepanov and I think we should look into doing the same in Jakarta Data.  `@Valid` is a spec-defined way under Jakarta [Bean] Validation for the user to explicitly declare when to opt in to validation.  It makes sense for that to be the way for users to opt in to validation.

Regarding `@Valid`, I could also see the possibility of someone using it on a more granular basis, like this, if using DataRepository where they define their own save/saveAll,

```
@Repository
public interface School extends DataRepository<Student, String> {
    void save(@Valid Student);
    ...
}
```
so we should also think about whether that pattern is allowed as well.

njr-11:
I sent a question to the Jakarta Bean Validation mailing list to check on our proposed usage of the `Valid` annotation, referencing this issue.

dstepanov:
To repeat my points from the email.

In this scenario:
[...]

PR #231, opened 2023-08-30
corrections and clarifications to Jakarta Validation section

njr-11:
This updates sections in the specification on Jakarta Validation based on conversations that we had with the Jakarta Validation team on the validation mailing list about method validation and cascading validation, which resulted from what is proposed under #216 .  This pull reflects what is currently possible with Jakarta Validation.  In order to fully achieve what was wanted, to have `@Valid` on the repository parameterized type be applicable to usage of that type on the repository methods, we need an enhancement from the Jakarta Validation spec, which we have asked for in Jakarta EE 11.  If we get that enhancement, this pull is written in a way so as to already be fully consistent with it, but would benefit from an additional code example showing the possibility of that approach.


gunnarmorling:
Note that JPA also performs validation of constraints on entities upon persist, so there's a risk of redundant validation here, when mandating the same at the repository level.

njr-11:
> Note that JPA also performs validation of constraints on entities upon persist, so there's a risk of redundant validation here, when mandating the same at the repository level.

I believe the JPA validation is configurable.  Do we need to include a statement that it should be turned off in order to avoid possible duplication or would that already be clear?

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -137,16 +136,18 @@ public class Student {
 
 In this example, the `name` field is annotated with `@NotBlank`, indicating that it must not be blank. The `age` field is annotated with both `@Positive` and `@Min(18)`, ensuring it is a positive integer greater than or equal to 18.
 
-The `School` repository interface, shown below, utilizes the validation rules defined in the `Student` entity:
+The `School` repository interface, shown below, uses the `jakarta.validation.Valid` annotation to cause the constraints from the `Student` entity to be applied to the `save` operation parameter, whereas the validation constraints are not applied to the `Student` entities that are the results of the `findByAgeLessThanEqual` operation because the example does not include any `jakarta.validation.Valid` annotation that applies to the return value.
```
mswatosh, line 139:
```suggestion
The `School` repository interface, shown below, uses the `jakarta.validation.Valid` annotation to cause the constraints from the `Student` entity to be validated during the `save` operation, whereas the validation constraints are not applied to the `Student` entities returned as a result of the `findByAgeLessThanEqual` operation because the `findByAgeLessThanEqual` method does not include a `jakarta.validation.Valid` annotation that applies to the return value.
```

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -137,16 +136,18 @@ public class Student {
...
 
-Regarding the delete and find operations, the integration with Jakarta Bean Validation can be included or made optional by the provider. It means that the validation process for delete and find operations may or may not be applied, depending on the configuration or implementation of a provider. It is up to the provider to determine whether or not to include validation for these database operations.
\ No newline at end of file
+Jakarta Data providers that are built using Jakarta Persistence might require the user to define persistence units for repositories or might handle the details of defining the persistence units internally. A user that defines the persistence unit for a Jakarta Data repository specifies the `validation-mode` as `NONE` per the "Enabling Automatic Validation" section of the Jakarta Persistence specification to avoid duplicate validation of entities. Similarly, the Jakarta Data provider uses the `validation-mode` of `NONE` or the `jakarta.persistence.validation.mode` map key with value of `none` that is defined in the "Enabling Automatic Validation" section of the Jakarta Persistence specification to avoid duplicate validation of entities.
```
mswatosh, line 153:
```suggestion
Jakarta Data providers that are built using Jakarta Persistence might require the user to define persistence units for repositories or might handle the details of defining the persistence units internally. A user that defines the persistence unit for a Jakarta Data repository must specify the `validation-mode` as `NONE` per the "Enabling Automatic Validation" section of the Jakarta Persistence specification to avoid duplicate validation of entities. Similarly, the Jakarta Data provider must specify either the `validation-mode` of `NONE` or the `jakarta.persistence.validation.mode` map key with value of `none` that is defined in the "Enabling Automatic Validation" section of the Jakarta Persistence specification to avoid duplicate validation of entities.
```

mswatosh (commented):
Looks good, just a few suggestions

njr-11:
fixes #237

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -137,16 +136,18 @@ public class Student {
...
 In this example, the `name` field is annotated with `@NotBlank`, indicating that it must not be blank. The `age` field is annotated with both `@Positive` and `@Min(18)`, ensuring it is a positive integer greater than or equal to 18.
 
-The `School` repository interface, shown below, utilizes the validation rules defined in the `Student` entity:
+The `School` repository interface, shown below, uses the `jakarta.validation.Valid` annotation to cause the constraints from the `Student` entity to be validated during the `save` operation, whereas the validation constraints are not applied to the `Student` entities returned as a result of the `findByAgeLessThanEqual` operation because the `findByAgeLessThanEqual` method does not include a `jakarta.validation.Valid` annotation that applies to the return value.
 
 [source,java]
 ----
 @Repository
-public interface School extends PageableRepository<Student, String> {
+public interface School extends DataRepository<Student, String> {
```
otaviojava, line 144:
```suggestion
public interface School extends DataRepository<@jakarta.validation.Valid Student, String> {
```
The way that Micronaut does sounds more semantic:

https://micronaut-projects.github.io/micronaut-data/latest/guide/

I mean, overwrite the method to include the valid annotation sounds noise in the method.

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -137,16 +136,18 @@ public class Student {
...
+    void save(@Valid Student s);
```
otaviojava, line 145:
```suggestion
````

Comment (otaviojava) [reply spec/src/main/asciidoc/jakarta-ee.adoc:145]: If this makes sense the semantic, I would remove this line.

File: spec/src/main/asciidoc/jakarta-ee.adoc (hunk 3 above)
Comment (njr-11) [reply spec/src/main/asciidoc/jakarta-ee.adoc:144]: > The way that Micronaut does sounds more semantic:

@otaviojava  I like the approach that Micronaut has as well.  But Jakarta Validation doesn't currently allow that pattern.  One of the Micronaut developers and myself have been in contact with the Jakarta Validation team about it, and I've opened issue https://github.com/jakartaee/validation/issues/186 officially requesting them to add this ability for Jakarta EE 11.  If they add it, we'll update our spec documentation to favor that approach.  But for now, our specification is limited to what is currently supported by Jakarta Validation.

File: spec/src/main/asciidoc/jakarta-ee.adoc (hunk 4 above)
Comment (njr-11) [reply spec/src/main/asciidoc/jakarta-ee.adoc:145]: See explanation in response to previous comment.

Comment (otaviojava) [reply spec/src/main/asciidoc/jakarta-ee.adoc:145]: Thanks.

**Comment by otaviojava** [comment]
> > Note that JPA also performs validation of constraints on entities upon persist, so there's a risk of redundant validation here, when mandating the same at the repository level.
> 
> I believe the JPA validation is configurable. I added an update to include a statement that it should be turned off in order to avoid possible duplication.


Hey Nathan, I got your point about disabling JPA and bringing it to the Jakarta Data level.
But why?
That point needs to be clarified for me because, in the end, the Provider will be a wrapper or decorator of either JPA or NoSQL implementation.

From the user perspective, the same result, a validation before a database interaction.

Sorry for drawing it; I hope to get myself cleaner this way.

In the first picture, we let the JPA or NoSQL provider follow the same flow:

```mermaid
flowchart TD
   JakartaDataAdapter --> JPAEngine
   JPAEngine --> Validation
   Validation--Is valid?--> Decision
   Decision--Yes--> Database
   Decision--No--> Exception
```


This one disables on JPA and moves it to the JakartaDataEngine, which will be an adapter or wrapper at the end.:

```mermaid
flowchart TD
   JakartaDataEngine --> Validation
   Validation--Is valid?--> Decision
   Decision--Yes--> JPAEngine
   Decision--No--> Exception
   JPAEngine --> Database
 JPAEngine --Disable the trigger--> Validation
```

Somehow expressing that the engine will handle that, avoiding the duplication, or bringing this to the "adapter" of the Jakarta Data Provider sounds better, mainly because it will not change anything at the JPA.

njr-11:
> Hey Nathan, I got your point about disabling JPA and bringing it to the Jakarta Data level.
> But why?
> That point needs to be clarified for me because, in the end, the Provider will be a wrapper or decorator of either JPA or NoSQL implementation.

Or it will have pure JDBC implementation, or some other implementation.

> From the user perspective, the same result, a validation before a database interaction.

It is not always the same result.  Your document is not correct.  There are 2 aspects to validation to consider.  One is the validation of the constraints on the entity, which is what JPA was doing, and defaulting a certain way.  The other is method validation, as defined by the Jakarta Validation specification. This includes constraints that are placed on repository methods themselves, to the parameters and/or return value.  JPA doesn't see these constraints.  Only Jakarta Data sees them.  They need to be validated according to the rules of Jakarta Validation, which defines its own way of controlling validation, covering both the method validation and the cascading validation to the entity constraints.  To prevent duplication, on the first aspect, we turn off validation that JPA does.  This allows us to fully honor the Jakarta Validation requirements that the Jakarta Validation spec has for an architecture like ours.

otaviojava:
In the query method, I got your point, but in the save method, I did not copy.

I mean, give an entity "Cat" with the validation annotations, and then at the CatRepository.

```java

@Entity
public class Contact {
@Email
 public String email;
@NotBlank
public String name;
}

@Repository
CatRepository extends CrudRepository<UUID, Cat> {
}


CatRepository repository;
repository.save(cat);
```

What is going to happen?

njr-11:
> What is going to happen?

It's going to follow the rules of the Jakarta Validation specification which means validating constraints on parameters and the return value (none are defined) and not cascade to validating the constraints on the entity because you didn't specify `@Valid`.  I understand in this case you cannot specify `@Valid` without declaring the save method yourself, and that's why we are requesting from Jakarta Validation the ability to put `@Valid` onto the type variable.  I realize you don't like what we have in the interim, but we cannot define something in our Jakarta specification that involves using another Jakarta specification contrary to how that specification defines itself to be used.  This is the best we can do until Jakarta Validation adds the function that we asked for, which hopefully will be Jakarta EE 11 lining up with our 1.0 version.

otaviojava (approved):

Issue #225, opened 2023-08-30
how can I define a concrete repository operation?


gavinking:
I might have missed something obvious (or non-obvious), but there does not appear to be an example in the spec of how to define a _concrete_ operation of a repository. It's extremely important that this be properly documented.

Naively, it looks a bit problematic: all the examples show repositories implemented as Java interfaces, and I don't see how one would inject an `EntityManager`, a `DataSource`, a `Connection`, or whatever, into an interface. Section 3.1 talks about injecting repositories into other things, but it doesn't seem to address injection into a repository.

Nor does there appear to be any operation of `DataRepository` or `CrudRepository` which would give me access to a connection.

(I've probably just missed something here.)

graemerocher:
since I have only been involved in this spec from an advisory perspective, I can say how this works in Micronaut Data and whether the spec wants to follow suite is up to the folks actively working on it. With Micronaut Data you can use an abstract class and inject the EntityManager (or whatever persistence technology specific interface) as constructor arguments, See:

https://micronaut-projects.github.io/micronaut-data/latest/guide/#repositories

gavinking:
@graemerocher Right, that's the most obvious solution.

njr-11:
This one is currently being discussed in other issues, but agreement hasn't been reached yet and so it hasn't made it into the spec.  A couple of ideas that I've thought of are:

1. Let users write default methods with an EntityManager (or DataSource or Connection) parameter where the invoking code sends in null and the repository provider substitutes in the actual entity manager for the method to use,

```
    default void myMethod(int arg1, String arg2, EntityManager em) {
        ... use entity manager
    }
```

An advantage here is that it gives the repository provider the ability to ensure the EntityManager gets closed upon completion of the method.  I'm not sure how important that is for EntityManager, but if it were a JDBC Connection, I know that's important.

2.  Add a method like `getResource(Class<T> type)` either one of the built-in repository interfaces or a mix-in interface for that purpose, that can be used to obtain the EntityManager,

```
    default void myMethod(int arg1, String arg2) {
        EntityManager em = getResource(EntityManager.class);
        ... use entity manager
    }
```

The repository provider could close resources here after the method completes, although it might be a little less straightforward to do it.

There are probably a variety of other options as well. It would be good to standardize one.

We don't have abstract classes as repositories defined in Jakarta Data yet, but that could get added.

gavinking:
The way this works in Hibernate 6 is as follows. I write:

```java
interface Queries {
[...]

Issue #224, opened 2023-08-30
explain how repositories work (or not) with entity lifecycle and persistence contexts


gavinking:
The spec does not mention the idea that an entity might have some sort of persistent lifecycle, nor does it mention the concept of a persistence context.

This hole is big enough to navigate an imperial star destroyer though, since there's a deep, fundamental difference in the programming model between:

- persistence context-free approaches, where updates require explicit API calls, and
- a solution like Jakarta Persistence, where updates are implicit, detected by the persistence context via dirty checking, and asynchronous.

It's impossible to imagine an API which abstracts over this difference!

Therefore, Jakarta Data needs to explain which of these programming models applies to a `CrudRepository`.

If the answer is that a `CrudRepository` is only useful with persistence context-free programming models, then the specification should face up to the fact that Jakarta Data is _not_ compatible with the Jakarta Persistence API, where every `EntityManager` is tied to a persistence context.

Now, you might be lucky, because I have proposed that Jakarta Persistence add a `StatelessEntityManager`, which would have no persistence context, but we have _not_ currently targeted this for JPA 4.0, because we want to take our time to get this functionality right.

On the other hand, if `CrudRepository` _is_ supposed to be used with a persistence context, the spec needs to explain how that works.

gavinking:
OK, so over in the discussion of #229, it's clear that there's overwhelming hostility to saying anything about persistence contexts and entity lifecycle, or toward engaging directly with concepts in the Jakarta Persistence spec.

If that's a correct understanding of the situation, then Jakarta Data should simply _remove_ references to Jakarta Persistence from the specification, and top claiming to support JPA.

Sanne:
This issue is critical: when dealing with managed entities in JPA, the scope of changes is tied to the transaction boundaries rather than the imperative operations a user would invoke on a repository.

This completely changes the semantics and has deep implications on how one designs the entire application, so if this isn't precisely defined by the spec it fundamentally violates the goal of the spec: introducing a "persistence agnostic API":
I would need to significantly rewrite and rethink my application when changing small details of the underlying persistance setup, so having this under-specified and still allow the application to compile just fine with the highly abstract API would mislead developers, most likely result in catastrophic problems at runtime.

So it's important to specify this aspect accurately, otherwise the promise of providing a general abstraction becomes a liability rather than a feature.

[...]

PR #235, opened 2023-09-01
Create a JPA domain

otaviojava:
JPA is a universe, and IMHO, having a unique domain in Jakarta Data makes sense.

I propose to have the JPA specialization and start to address these issues in this JPA specialization.

- https://github.com/jakartaee/data/issues/224
- https://github.com/jakartaee/data/issues/225

If it is okay, we can merge and request help from the Hibernate team to enhance those points.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/jpa/JPARepository.java
+++ b/api/src/main/java/jakarta/data/repository/jpa/JPARepository.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package jakarta.data.repository.jpa;
+
+import jakarta.data.repository.PageableRepository;
+
+public interface JPARepository<T, K> extends PageableRepository<T, K> {
+ //TODO https://github.com/jakartaee/data/issues/224
+ //TODO https://github.com/jakartaee/data/issues/225
+
+}
--- a/api/src/main/java/jakarta/data/repository/jpa/package-info.java
+++ b/api/src/main/java/jakarta/data/repository/jpa/package-info.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


graemerocher:
Doesn't this belong in the JPA spec if it adopts this spec?

otaviojava:
IMHO: Yes, that would be the perfect world where the spec team could address those issues better than us.

However, they may not want this approach.

I can put those behaviors on this JPA domain inside Jakarta Data.


But let's see what the other people think about it:

@gavinking @njr-11

gavinking:
So, look, "repositories" were first suggested as a feature for JPA on 2020, and so [back in 2021](https://github.com/jakartaee/persistence/issues/294#issuecomment-985054093) I responded that it didn't belong in the Persistence spec and that people who were interested should try it to write a separate spec, and that comment lead directly to the creation of the Jakarta Data project. And so here we are.

And so far my personal opinion hasn't changed at all: Jakarta Persistence is a self-contained API and doesn't require the use of "repositories", neither Jakarta Data repositories, nor any other sort of repository. @lukasj might very well have a different take, so naturally I'll let him speak for himself.

And at present it's completely unclear that Jakarta Data in its final form will be even _appropriate_ for use as a wrapper around Jakarta Persistence. In #229 there was extreme hostility to even _discussing_ the implementation difficulties and conceptual mismatches which arise here. Now, I do actually believe that it's _conceptually possible_ to evolve the Jakarta Data spec in a direction which would make a JPA-backed implementation _feasible_. But that will take work, and some time, and is certainly not yet a done deal.

And look, OK, please let me be a bit direct about this. If you folks hope for such a thing to ever happen, you'll have to work with JPA experts, and refrain from yelling at us when we talk about Jakarta Persistence in this issue tracker. If folks here continue with this _"get lost with your Jakarta Persistence concerns, we don't care about Jakarta Persistence here"_ attitude, in quite clear evidence on that other thread, then I'll let you guess what the inevitable outcome of that is going to be.

So no, in summary, there's a whole lot of work to be done here before we could even _contemplate_ adding anything Jakarta Data-related to the Persistence spec.

njr-11 (changes requested):
This interface doesn't have any methods on it or documentation. I realize your intent is probably to get things done in stages, which in general is a good approach, but let's not introduce a JpaRepository subinterface until we have decided to add at least one repository operation that is unique to Jakarta Persistence that could not also apply to JDBC or any other type of database.  We would also need to consider naming it something other than JPA which is owned by Oracle for Java EE.  Once we have an idea of what methods would belong on it, it might become clearer what to call it.

otaviojava:
@njr-11 What I had in my mind is with this specialization, answer those two issues:

- https://github.com/jakartaee/data/issues/224
- https://github.com/jakartaee/data/issues/225

What do you have in your mind? Only using the spec side?

njr-11:
> @njr-11 What I had in my mind is with this specialization, answer those two issues:
> 
>     *

225 is more general than JPA.  Note that Gavin also mentions `DataSource` and `Connection` which are JDBC.  There are ideas for solving this which have been discussed thus far which do not require a JPA-specific repository.

224 is an area that needs work for sure, and it is most likely to be what defines JPA-specific repository operations.  It remains to be determined whether that will be defined in a more general sense or very specific to JPA.  In either case, we should let the work that develops out of that determine how an additional repository sub interface gets defined and what best to name it.

otaviojava:
So, I will close this PR for while.

keilw:
Good idea, I would not rush into this, until it is clear, if spec wording is enough, or indeed some JPA integration was needed, and where.

I am not so actively involved in the Jakarta REST project other than the platform or Spec Committee, but it could be worth looking into the [changes since 2.1](https://jakarta.ee/specifications/restful-ws/3.1/jakarta-restful-ws-spec-3.1#changes-since-2.1-release) especially how optional support for XML/JAXB is handled there.

The use case is somewhat similar, not forcing users and implementers of the spec to support both:
- JSON
- XML

If they only ever need one.

Issue #236, opened 2023-09-01
Consider formalizing the Query by Method section


emmanuelbernard:
The spec mentions "first part" "second part". It also mentions "We also have the option to handle entities with another class on them" which does not seem defined. Examples use `And` but this is not defined by the spec what that means.

I think a BNF or some other form of formalization would be useful with a companion text description of the expected query output.

Likewise another class of ambiguity exists

> For queries by method name, the resolution algorithm starts by interpreting the whole part (AddressZipCode) as the property and checks the domain class for a property with that name (uncapitalized). If the algorithm succeeds, it uses that property.

This does not quite tell how the algorith would choose Address and then ZipCode as the path, and what happens when AddressZip also exists with a subproperty Code. Some unambiguity would make sense here. I understand there is an escape hatch via `_` but claifying the expected behavior for the user would be beneficial. 

> existsBy | Exists projection, returning typically a boolean result.

What is the non typical return type? Some clarification in the spec makes sense.

PR #239, opened 2023-09-01
Fix Consider formalizing the Query by Method section

otaviojava:
Changes: 

Fix: https://github.com/jakartaee/data/issues/236

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -261,7 +261,7 @@ class Address {
 - The method name includes an underscore (`_`) indicating manual traversal points.
 - The Property Resolution Algorithm will parse the method name and extract `Address_ZipCode`, recognizing the underscore as a traversal point.
 - It will then attempt to resolve the property named `Address` within the `Order` class, followed by the `zipcode` property within the `Address` class, following manual traversal points.
-- If properties `Address` and `ZipCode` are found in the appropriate classes or their nested objects, the query will filter `Order` records based on the provided `ZipCode` object within the `Address` object.
+- If properties `Address` and `ZipCode` are found in the appropriate classes or their nested objects, the query will filter `Order` records based on the provided `Zipcode` object within the `Address` object.
 
 WARNING: Define as a priority following standard Java naming conventions, camel case,  using underscore as the last resort.
 
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -265,6 +265,26 @@ class Address {
 - It will then attempt to resolve the property named `Address` within the `Order` class, followed by the `zipcode` property within the `Address` class, following manual traversal points.
 - If properties `Address` and `ZipCode` are found in the appropriate classes or their nested objects, the query will filter `Order` records based on the provided `Zipcode` object within the `Address` object.
 
+
+*Scenario 4: People Repository with findByAddressZipCode(String addressZipCode)*
+
+
+In this scenario, we have the following data model:
+
+[source,java]
+----
+class Person {
+  private Long id;
+  private String addressZipcode;
+}
+----
+
+- The query method `findByAddressZipCode` takes a `String` parameter named `addressZipCode`.
+- The Property Resolution Algorithm will parse the method name and extract `AddressZipCode`.
+- It will then attempt to resolve the property named `addressZipcode` in the `Person` class, following automatic class splitting by case-insensitive.
+- If a property named `addressZipCode` of type `String` exists in the `Person` class or its nested objects, the query will filter `Person` records based on the provided `addressZipCode` string.
+
+
 WARNING: Define as a priority following standard Java naming conventions, camel case,  using underscore as the last resort.
 
 In queries by method name, `Id` is an alias for the entity property that is designated as the id. Entity property names that are used in queries by method name must not contain reserved words.
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -182,30 +182,85 @@ public interface OrderRepository extends CrudRepository<Order, Long> {
 }
 ----
 
-For queries by method name, the resolution algorithm starts by interpreting the whole part (AddressZipCode) as the property and checks the domain class for a property with that name (uncapitalized). If the algorithm succeeds, it uses that property.
+The resolution algorithm for identifying properties in query methods by method name, with manual traversal points, is defined as follows:
+
+1. *Method Name Parsing*:: The query method's name is parsed to identify the property or properties being referenced. Method names in query methods typically follow a pattern of "findBy[Property]", where "[Property]" represents the name of the property you want to query by.
+
+2. *Property Extraction*:: The property name is extracted from the method name by removing the "findBy" prefix. For example, in the query method `findByAddressZipCode`, the property name extracted is `AddressZipCode`.
+
+3. *Property Name Capitalization*:: The extracted property name is treated as is, with its original capitalization. For example, if the property name is `AddressZipCode`, it remains in camel case.
+
+4. *Manual Traversal Points*:: To resolve ambiguity or to specify traversal through nested properties, underscores (`_`) can be used within the method name. Each underscore represents a traversal point to access nested properties. For example, `findByAddress_ZipCode` explicitly indicates traversal to the `Address` object's `ZipCode` property.
+
+5. *Domain Class Property Lookup*:: The framework checks the domain class associated with the repository for a property with the same name as the extracted property name (uncapitalized). If the domain class has a property named `addressZipCode`, this is considered a successful resolution.
+
+6. *Nested Property Handling*:: If the extracted property name includes underscores (`_`) indicating nested traversal, the framework follows the specified path to resolve the property.
+
+7. *Resolution Outcome*:: If the framework successfully identifies a property in the domain class or along the specified traversal path that matches the extracted property name, it uses that property in the query to filter data.
+
+
+*Scenario 1: Person Repository with findByAddressZipCode(ZipCode zipCode)*
+
+In this scenario, we have the following data models:
 
 [source,java]
 ----
-@Repository
-public interface OrderRepository extends CrudRepository<Order, Long> {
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -147,11 +147,23 @@ public interface ProductRepository extends CrudRepository<Product, Long> {
 }
 ----
 
-The parsing query method name has two parts: the subject and the property.
+The parsing of query method names follows a specific format:
+
+- The method name consists of the subject and the property.
+- The subject defines the query's condition or filtering criteria.
+- The property refers to a direct property of the managed entity.
```
njr-11, line 154:
```suggestion
- The subject, defines the action (such as `find` or `delete`) , optionally followed by an expression (for example, `First10`), followed by `By`
- The predicate defines the query's condition or filtering criteria, where multiple conditions are delimited by `And` or `Or`. For example, `PriceLessThanAndNameLike`.
- The order clause, which is optional, begins with `OrderBy` and consists of an ordered collection of entity attributes by which to sort results, delimited by `Asc` or `Desc` to specify the sort direction of the preceding attribute.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -147,11 +147,23 @@ public interface ProductRepository extends CrudRepository<Product, Long> {
...
+- The predicate is formed by combining the subject and property.
+
+Queries can also handle entities with related classes by specifying the relationship using dot notation.
+
+Example query methods:
+
+- `findByName(String name)`: Find entities by the 'name' property.
+- `findByAgeGreaterThan(int age)`: Find entities where 'age' is greater than the specified value.
+- `findByAuthorName(String authorName)`: Find entities by the 'authorName' property of a related entity.
+- `findByCategoryNameAndPriceLessThan(String categoryName, double price)`: Find entities by 'categoryName' and 'price' properties, applying an 'AND' condition.
```
njr-11, line 164:
```suggestion
- `findByCategoryNameAndPriceLessThan(String categoryName, double price)`: Find entities by 'categoryName' and 'price' properties, applying an 'And' condition.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -170,31 +182,86 @@ public interface OrderRepository extends CrudRepository<Order, Long> {
...
+
+3. *Property Name Capitalization*:: The extracted property name is treated as is, with its original capitalization. For example, if the property name is `AddressZipCode`, it remains in camel case.
+
+4. *Manual Traversal Points*:: To resolve ambiguity or to specify traversal through nested properties, underscores (`_`) can be used within the method name. Each underscore represents a traversal point to access nested properties. For example, `findByAddress_ZipCode` explicitly indicates traversal to the `Address` object's `ZipCode` property.
+
+5. *Domain Class Property Lookup*:: The framework checks the domain class associated with the repository for a property with the same name as the extracted property name (uncapitalized). If the domain class has a property named `addressZipCode`, this is considered a successful resolution.
+
+6. *Nested Property Handling*:: If the extracted property name includes underscores (`_`) indicating nested traversal, the framework follows the specified path to resolve the property.
```
njr-11, line 197: Should we state that you should not define entity attributes with underscore in them because it could lead to collisisons?

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -170,31 +182,86 @@ public interface OrderRepository extends CrudRepository<Order, Long> {
...
+class Address {
+  private Zipcode zipcode;
 }
 ----
 
+- The query method `findByAddressZipCode` takes a `ZipCode` object as a parameter.
+- The Property Resolution Algorithm will parse the method name and extract `AddressZipCode`.
+- It will then attempt to resolve the property named `addressZipCode` in the `Person` class, following automatic class splitting by camel case.
+- Since the `Person` class has an `Address` property, it will recursively follow the path to the `Address` class.
+- In the `Address` class, it will identify the `Zipcode` property and filter `Person` records based on the provided `ZipCode` object within the `Address` object.
```
njr-11, line 222: This might be further evidence that matching should be case insensitive.  As demonstrated here, it can be very difficult to remain consistent regarding case for compound words.

```suggestion
- In the `Address` class, it will identify the `zipcode` property and filter `Person` records based on the provided `Zipcode` object within the `Address` object.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -170,31 +182,86 @@ public interface OrderRepository extends CrudRepository<Order, Long> {
...
+  private String addressZipCode;
+  private Address address;
+}
+
+class Address {
+  private Zipcode zipcode;
 }
 ----
 
+- The query method `findByAddress_ZipCode` takes a `ZipCode` object as a parameter.
```
njr-11, line 259:
```suggestion
- The query method `findByAddress_ZipCode` takes a `Zipcode` object as a parameter.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -170,31 +182,86 @@ public interface OrderRepository extends CrudRepository<Order, Long> {
...
+- The method name includes an underscore (`_`) indicating manual traversal points.
+- The Property Resolution Algorithm will parse the method name and extract `Address_ZipCode`, recognizing the underscore as a traversal point.
+- It will then attempt to resolve the property named `Address` within the `Order` class, followed by the `ZipCode` property within the `Address` class, following manual traversal points.
```
njr-11, line 262:
```suggestion
- It will then attempt to resolve the property named `Address` within the `Order` class, followed by the `zipcode` property within the `Address` class, following manual traversal points.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -170,31 +182,86 @@ public interface OrderRepository extends CrudRepository<Order, Long> {
...
+- If properties `Address` and `ZipCode` are found in the appropriate classes or their nested objects, the query will filter `Order` records based on the provided `ZipCode` object within the `Address` object.
```
njr-11, line 263:
```suggestion
- If properties `Address` and `ZipCode` are found in the appropriate classes or their nested objects, the query will filter `Order` records based on the provided `Zipcode` object within the `Address` object.
```

njr-11 (changes requested):
Sorry about the large number of comments. This probably should have been separate pulls for the different areas covered.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -170,31 +182,86 @@ public interface OrderRepository extends CrudRepository<Order, Long> {
...
 ----
 
-For queries by method name, the resolution algorithm starts by interpreting the whole part (AddressZipCode) as the property and checks the domain class for a property with that name (uncapitalized). If the algorithm succeeds, it uses that property.
+The resolution algorithm for identifying properties in query methods by method name, with manual traversal points, is defined as follows:
+
+1. *Method Name Parsing*:: The query method's name is parsed to identify the property or properties being referenced. Method names in query methods typically follow a pattern of "findBy[Property]", where "[Property]" represents the name of the property you want to query by.
+
+2. *Property Extraction*:: The property name is extracted from the method name by removing the "findBy" prefix. For example, in the query method `findByAddressZipCode`, the property name extracted is `AddressZipCode`.
+
+3. *Property Name Capitalization*:: The extracted property name is treated as is, with its original capitalization. For example, if the property name is `AddressZipCode`, it remains in camel case.
```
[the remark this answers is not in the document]
otaviojava: IMHO: Nope

spec/src/main/asciidoc/repository.asciidoc
otaviojava: Yes, I could include that we are expecting the fields follow the Java convention name to the fields.

spec/src/main/asciidoc/repository.asciidoc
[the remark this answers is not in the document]
otaviojava: What do you think?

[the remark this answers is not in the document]
njr-11: Personally, I'd be in favor of case insensitive matching for property names unless the vendors have an issue with it.  Using the example from this pull, a lot of users who define an entity attribute called `zipcode` won't have a clear understanding of whether that means its name is "zipcode" or "Zipcode" and a lot of them will probably intuitively write method names like findByZipCode that doesn't match the case of either.  Being case insensitive would be more flexible and spare the user from having to experience errors in these cases.  It seems unfortunate to complain about differences like that.  Maybe the case insensitive parsing (or case conversion) makes it more expensive though.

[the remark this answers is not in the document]
otaviojava: This expensive part makes me afraid; that is why I suggest keeping the same following what we have with Spring and Micronout.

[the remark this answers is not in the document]
njr-11: I wouldn't just assume it's a problem.  Vendors like Micronaut would likely being doing their processing of methodNames at compile time.  Even a vendor that processes at run time probably only does so once and reuses.  It might not be much of a problem at all.

otaviojava:
@njr-11, there is no sorry, Thank you for your support and help. Please feel free to put as many suggestions as possible.

I created a paragraph about the name convention.

Please, let me know it it is ok.

njr-11:
> I created a paragraph about the name convention.
> 
> Please, let me know it it is ok.

The paragraph looks good.

The other overall comment to note is that some of your examples actually rely on case insensitive comparison, but you have the spec being updated to state that case sensitive is required.  This is making me think even more so that case insensitive is the better way to go, because it's so natural to subconsciously end up varying the case of compound words like this.  I shouldn't have made the comment about extra work earlier.  Just converting internally to either lower or upper case should be an easy way to make a case insensitive comparison.

otaviojava:
@njr-11, please let me know if got what you mean now.

njr-11 (approved):
There is a lot of good detail added in this pull. I'd be fine with merging it. In a different pull, it would be nice to add BNF notation to make things even more explicit at some point.

otaviojava:
@njr-11 I will put it in another PR.

Issue #244, opened 2023-09-05
Define basic terms: "entity", "repository", "identifier", etc.


gavinking:
A challenge this spec faces is the extremely high level of abstraction. At this level of abstraction we can't be sure that we all mean the same things by the words we use. There's a risk of accidentally smuggling the semantics of one sort of data storage technology into the specification without even realizing it.

To make a start on nailing down the meaning of words here, I propose the following starting point. Please let me know what you think.

Entities and repositories
====

Abstractly, an _entity_ (or _entity type_) is a schema for data. Here we assume that data is persistent, that is, that the data outlives any Java process which makes use of it.

- The schema may be as simple as a tuple of types, as is typical in the relational model, or it might be structured, as in document data stores.
- The schema might be explicit, as in the case of DDL which declares a relational table, or it might be implicit, as is commonplace in key/value stores.
- In either case, we assume that the entity is represented in Java as a class, which we call the _entity class_. (We will not consider generic programs which work with entity data via detyped representations.)
- Each persistent instantiation of the schema is distinguishable by a unique _identifier_. For example, a row of a relational database table is identifiable by the value of its primary key.
- Any persistent instantiation of the schema is representable by an instance of the entity class. In a given Java program, multiple entity class instances might represent the same persistent instance of the schema.

When there's no risk of confusion, we will often use the word “entity” to mean the entity class, or even an instance of the entity class.

A _repository_ is a Java interface which acts as a gateway for access to persistent data of one or more entity types.

- The repository usually exposes operations for querying and retrieving entity class instances representing persistent instances of the entities.
- The repository might also expose operations for modifying data: creating new persistent instances in the data store, removing persistent instances, or modifying the state of persistent instances. Conventionally, these operations are named _insert_, _delete_, and _update_, respectively.
- The repository and its entities might expose only a subset of the full data set available in the data store.
- Entities belonging to a given repository might have _associations_ between them. This is extremely common if the repository provides access to relational data. This specification does not define the semantics of associations between entities belonging to different repositories.
- Repositories are typically stateless. This specification does not address the definition of repositories backed by Jakarta Persistence-style stateful persistence contexts.

Jakarta Data allows the developer to define:

- a set of entity classes and mappings to a given data store, and
- one or more repository interfaces, written according certain rules, including the rules laid out by this specification.

Then an implementation of Jakarta Data which targets the same data store is responsible for providing an implementation of each repository interface.

otaviojava:
Thank you.

[...]

PR #245, opened 2023-09-06
Proposal: Define basic terms: "entity", "repository", "identifier", etc.

otaviojava:
Hey everyone, based on the Issue, I merged it into what we already have in our documentation.

Please let me know what you think about it.

Proposal work to: https://github.com/jakartaee/data/issues/244


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -87,17 +101,28 @@ public interface Garage extends CrudRepository<Car, String> {
...
 
-Applications are recommended not to mix Entity annotations from different models for the sake of clarity and to allow for the Entity annotation to identify which provider is desired in cases where multiple types of Jakarta Data providers are available.
+4. *Provider Differentiation*: To maintain clarity and specify the desired provider when using Jakarta Data, it's recommended that applications do not mix Entity annotations from different models. This practice allows the Entity annotation to indicate the desired provider, especially in cases where multiple types of Jakarta Data providers are available. Repository implementations are designed to search for and support specific Entity annotations while ignoring others.
```
njr-11, line 123: This softened the requirement providers rely on to avoid conflict into something that seems information rather than a requirement.

```suggestion
4. *Provider Differentiation*: To maintain clarity and specify the desired provider when using Jakarta Data, it's recommended that applications do not mix Entity annotations from different models. This practice allows the Entity annotation to indicate the desired provider, especially in cases where multiple types of Jakarta Data providers are available. A Jakarta Data provider must provide implementation of repositories for Entity types having the Entity annotations that it supports, ignoring Entity types only having identifiable Entity annotations that the Jakarta Data provider does not support. The latter are to be handled by other Jakarta Data providers that do support the other types of Entity annotations.
```

njr-11 (changes requested):
Mostly this looks really good.  I added comments for a few corrections to make.

Issue #227, opened 2023-08-30
Define the minimal set of portable functionality


gavinking:
Jakarta Data currently fails to specify a core minimum set of functionality which all implementations must support and which must work portably across implementations. 

- Much of the specified functionality appears to be optional, in that support for it depends on the capabilities of the data store.
- Most of the features have undefined or under-specified semantics.
- The specification almost never seems clearly to impose any hard requirement on implementations.

In its present state, I have no expectation of portable behavior _even for implementations which target the same data source_.

njr-11:
This area of the spec definitely needs work.  If everyone can help collect specific examples within this issue, it will improve the changes we identify all of the places and get them addressed.  I agree example code in specs does not convey requirements. The purpose of examples is to provide a helpful illustration to the reader to more quickly understand requirements and usage that are directly spelled out by the spec.  Let's improve this in all the places it is needed.

PR #247, opened 2023-09-06
Define the minimal set of portable functionality

otaviojava:
# Changes

- Include the portability file
- rename intro file to jakarta-data.adoc

I need help on this one. Are we getting close?

https://github.com/jakartaee/data/issues/227


njr-11 (changes requested):
There is a lot of good information here on limitations of types of different database types.  Ideally the limitations should be stated directly alongside where the requirements are declared in the specification.

We have a table of reserved keywords in the specification document, with columns,

Keyword | Description | Method signature Sample

Let’s add a fourth column, “Database Types Where Not Required”  (or the opposite if it’s more concise that way) to explicitly identify every keyword that is not required for a particular type of database. 

Based on what you wrote in this pull, for the `Or` keyword, you would list Wide-Column.

For the CrudRepository and PageableRepository methods, the information belongs in the JavaDoc of the specific method.  This is where users will first look (and it is the only place they should need to look) for identifying if they can rely on using the method.  Similarly, a Jakarta Data provider that implements the method will want to see the same information on the JavaDoc.  The method JavaDoc can also declare the exception that is raised (MappingException or UnsupportedOperationException or define a new one?) when the function is not available from the database.

Given that you can compose methods similar to those on the built-in repository interfaces, documentation also belongs in the spec document.  Add a similar “Database Types Where Not Required” column to the “Query Methods Keywords” table.

Based on what you wrote in this pull, for the `countBy` keyword, you would list Key-Value.

A footnote will be needed to address the part about the limitation of queries that do not use the identifier for Key-Value.

otaviojava:
@michael-simons could you help and give your opinions on these points, mainly the Graph part?

otaviojava:
@njr-11, what do you think?

spec/src/main/asciidoc/portability.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+
+==== Document Databases
+
+Document databases provide query flexibility akin to relational databases, offering robust query capabilities. They encourage denormalization for performance optimization. When interfacing with document databases, Jakarta Data goes a step further by supporting both built-in repositories: CrudRepository and PageableRepository. Additionally, method by query is implemented, though developers should be aware that some keywords may not be universally supported.
+
+These portability considerations reflect Jakarta Data's commitment to providing a consistent data access experience across diverse NoSQL database types. While specific capabilities and query support may vary, Jakarta Data aims to simplify data access, promoting flexibility and compatibility in NoSQL database interactions.
+
+==== Graph Database
+
+A Graph database, a specialized NoSQL variant, excels in managing intricate data relationships, rivaling even traditional relational databases. Its unique strength lies in its ability to handle complex relationships, enhanced by including direction and associated properties known as edges.
```
michael-simons, line 42:
```suggestion
A Graph database, a specialized NoSQL variant, excels in managing intricate data relationships, rivaling even traditional relational databases. Its unique strength lies in its ability to handle both directed and undirected edges  (or relationships) between vertices (or nodes), including the ability to store properties on both vertices and edges.```

File: spec/src/main/asciidoc/portability.asciidoc [hunk 2]
```diff
@@ -0,0 +1,66 @@
...
+
+In Jakarta Data, Graph databases support advanced querying capabilities, akin to relational databases, while providing extensive support for method by query keywords. Additionally, they fully implement CrudRepository and PageableRepository methods, ensuring seamless data access and manipulation within this specialized NoSQL database type.
```
michael-simons, line 44:
```suggestion
Graph databases usually can answer queries that returns rows with either flat objects, or collections or mixtures of them. Portability is guaranteed only for the mapping of rows to classes, for the fact that queries specified via annotations or other means are supported and that queries derived from keywords and combinations of mapped classes/properties will be translated into vendor specify query.
```

michael-simons (changes requested):

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -356,123 +356,159 @@ In queries by method name, `Id` is an alias for the entity property that is desi
...
 |Exists projection, returning as a `boolean` result.
+|Key-value, Wide-Column
 |===
 
-Jakarta Data implementations support the following list of predicate keywords to the extent that the database is capable of the behavior. A repository method will raise `jakarta.data.exceptions.DataException` or a more specific subclass of the exception if the database does not provide the requested functionality.
+[NOTE]
+.Note
+* The "Not Required For" column indicates the database types for which the respective keyword is not required or applicable.
+
+Jakarta Data implementations support the following list of predicate keywords to the extent that the database is capable of the behavior. A repository method will raise `java.lang.UnsupportedOperationException` or a more specific subclass of the exception if the database does not provide the requested functionality.
```
njr-11, line 385:
```suggestion
Jakarta Data implementations must support the following list of query-by-method keywords, except where indicated for a database type. A repository method must raise `java.lang.UnsupportedOperationException` or a more specific subclass of the exception if the database does not provide the requested functionality.
```

njr-11 (changes requested):
Thanks for adding the explicit information in table form. I looked over the rest of the pull and added some review comments.

otaviojava:
> Thanks for adding the explicit information in table form. I looked over the rest of the pull and added some review comments.

@njr-11 thank you. All comments fixed.

njr-11:
> > Thanks for adding the explicit information in table form. I looked over the rest of the pull and added some review comments.
> 
> @njr-11 thank you. All comments fixed.

Actually, there are still a number of review comments that are not addressed yet.  I think you may have missed them because due the quantity of comments, git defaults to minimizing some of them.  I had to click on a section to make git show them all.

otaviojava:
> > > Thanks for adding the explicit information in table form. I looked over the rest of the pull and added some review comments.
> > 
> > 
> > @njr-11 thank you. All comments fixed.
> 
> Actually, there are still a number of review comments that are not addressed yet. I think you may have missed them because due the quantity of comments, github defaults to minimizing some of them. I had to click on a section to make git show them all.

Thank you, I checked twice, I hope that I got all of those now.

njr-11 (approved):
The only remaining comment that wasn't addressed is to rewrite the following sentence,
"Portability is guaranteed only for the mapping of rows to classes, for the fact that queries specified via annotations or other means are supported and that queries derived from keywords and combinations of mapped classes/properties will be translated into vendor specify query."

There are getting to be a lot of commits on this pull already, so I'm fine if that happens under a separate pull.

PR #248, opened 2023-09-06
accessor method to obtain EntityManager

### LINKED ISSUE #202: Obtaining the EntityManager (or NoSQL equivalents) for a repository

njr-11:
Write up the approach for obtaining the EntityManager/DataSource/Connection that was discussed and decided upon during the Jakarta Data call.
fixes #202

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -511,6 +511,45 @@ import jakarta.data.repository.Sort;
  *                                 Sort.asc("name"));
  * </pre>
  *
+ * <h2>Repository Default Methods</h2>
+ *
+ * <p>You can compose default methods on your repository interface to supply
+ * user-defined implementation.</p>
+ *
+ * <p>For some advanced scenarios, you might need access to an
+ * underlying resource from the Jakarta Data provider, such as a
+ * {@code jakarta.persistence.EntityManager},
+ * {@code javax.sql.DataSource}, or
+ * {@code java.sql.Connection}</p>
+ *
+ * <p>To obtain the above, you can define accessor methods on your repository interface,
+ * where the method has no parameters and its result value is one of the
+ * aforementioned types. When you invoke the method, the Jakarta Data provider
+ * supplies an instance of the requested type of resource.</p>
+ *
+ * <p>For example,</p>
+ *
+ * <pre>
+ * &#64;Repository
+ * public interface Cars extends CrudRepository&lt;Car, Long&gt; {
+ *     ...
+ *
+ *     EntityManager getEntityManager();
+ *
+ *     default Car[] advancedSearch(SearchOptions filter) {
+ *         EntityManager em = getEntityManager();
+ *         ... use entity manager
+ *         return results;
+ *     }
+ * }
+ * </pre>
+ *
+ * <p>If the resource type inherits from {@link AutoCloseable} and you invoke the
+ * accessor method from a repository default method, the Jakarta Data provider
+ * automatically closes the resource after the default method ends.
+ * If you invoke the accessor method from outside the scope of a default method,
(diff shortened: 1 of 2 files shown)


otaviojava (approved):

otaviojava:
Way better, thank you.
I loved this approach.

keilw:
@njr-11  @otaviojava Why did that PR merge all sorts of HTML code into the `module-info`?
Was there a mix-up between files? I don't see how all of those aspects are relevant for the module-info/JPMS.

njr-11:
> @njr-11 @otaviojava Why did that PR merge all sorts of HTML code into the `module-info`? Was there a mix-up between files? I don't see how all of those aspects are relevant for the module-info/JPMS.

This information is included there as the main entry point for JavaDoc that developers will refer to when using Jakarta Data.  Note that if you go to the Jakarta EE page for the Jakarta Data spec, https://jakarta.ee/specifications/data/1.0/
and then scroll to the section where it links to the JavaDoc, "Jakarta Data 1.0 Specification Javadoc", this link, https://jakarta.ee/specifications/data/1.0/apidocs, goes directly to the page that is created by module-info.  At least this is how it works by default.  Let us know if there is some other convention/approach for Jakarta specs to have a different JavaDoc page be the entry point.  The information here is extremely important for developers as their primary reference and entry point for using Jakarta Data.  Typical users never read the spec document (although the spec document is very useful for Jakarta Data providers and very advanced users).  Most users will rely on JavaDoc alone.

keilw:
Ok, so that's a bit like the one for JShell: https://docs.oracle.com/javase/9/docs/api/jdk.jshell-summary.html

Btw, I think the "api" in the module name should not be included, All others I checked including REST or Persistence don't include it.

PR #258, opened 2023-09-12
type-safe entity attribute names

### LINKED ISSUE #230: migrate away from the use of stringly-typed field references
The specification has examples like `Sort.desc("yearBorn")`, `@OrderBy("age")`, etc.

In Hibernate and JPA we now strongly discourage this sort of untypesafe code. Instead, we leverage the JPA static metamodel generator to express _typesafe_ references to entity attributes, for example, `Order.asc(Book_.title)`.

You can see examples [here](https://docs.jboss.org/hibernate/orm/6.3/introduction/html_single/Hibernate_Introduction.html#paging-and-ordering).

Jakarta Data should carefully consider how it can integrate with the JPA static metamodel, or extend the idea to non-JPA entities.

njr-11:
Fixes #230

This implements the solution that is discussed in #230 to provide a StaticMetamodel annotation by which applications can access entity attribute names in a type-safe manner.


otaviojava:
Shall we discuss it after 1.0?

I checked the JPA, and it looks super fancy with several details. 
There is, for example the entire [metamodel](https://jakarta.ee/specifications/persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/metamodel/package-summary) package here

api/src/main/java/jakarta/data/repository/StaticMetamodel.java
```diff
@@ -0,0 +1,110 @@
...
+ * represents the unique identifier.</p>
+ *
+ * <p>In cases where multiple Jakarta Data providers provide repositories for the same
+ * entity type, no guarantees are made of the order in which the Jakarta Data providers
+ * assign values to the fields of the class that is annotated with {@code StaticMetamodel}.</p>
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface StaticMetamodel {
```
otaviojava, line 103: IMHO: it should be compatible not just with the repository domain but with all Data patterns that we might bring in the future; thus, it should be in the data package.

njr-11: That's an excellent comment.  I added a [commit](https://github.com/jakartaee/data/pull/258/commits/7d8b43c226f9f4e3212788a139a0bfc422d9ee83) to move it to the data package.

api/src/main/java/jakarta/data/StaticMetamodel.java
```diff
@@ -0,0 +1,110 @@
...
+ * }
+ * </pre>
+ *
+ * <p>You can define a static metamodel as follows,</p>
+ *
+ * <pre>
+ * &#64;StaticMetamodel(Person.class)
+ * public class Person_ {
+ *     public static volatile ssn; // or id
+ *     public static volatile name;
```
graemerocher, line 57: this does not look like syntactically correct java

njr-11: > this does not look like syntactically correct java

Oops. It looks like I left out the type.  I'll add it back in.

otaviojava:
> > Would you like to discuss it after 1.0?
> 
> I'm adding it to address one of the issues #230 to get Hibernate/Gavin on board with Jakarta Data.
> 
> You do raise a good point about wanting to be able to use it for other more complex patterns besides just getting entity attribute names in a type-safe manner. As I see it we have a couple of options here.
> 
> 1. We could have separate metamodel annotations where this current one (we might need to rename it accordingly) is only for getting the type-safe attribute names.  In the future, we could introduce another for more complex uses.
> 2. We could make StaticMetamodel extensible although currently only providing what corresponds to the function set that we have in v1.0.  The metamodel class field would be a general Attribute type, from which you can get the `name` and `asc`/`desc` Sort.  In the future we would be able to add more, and even attribute subclasses if we wanted.
> 
> I think the second option would be reasonable.


Shall we wait until the next call to discuss more about it?
I mean, we can work to solve in the first version, but it would amazing if we designed a little more those points.

P.s.: I enjoy the second option as well.

njr-11:
> Shall we wait until the next call to discuss more about it?

We can certainly discuss on the next call.  If anyone wants to offer ideas/thoughts prior to that they are certainly welcome to.

> P.s.: I enjoy the second option as well.

If option 2 is favored, I can look into switching this pull over to that, but I'll wait some time for more feedback to come in first just in case anyone is insisting on the current approach.

njr-11:
@otaviojava I have updated this per the discussion at the last Jakarta Data meeting, adding the extensible Attribute type, from which you can also get the type-safe name().  I wrote this in a way that allows the user to specify the fields of the metamodel as `public static final` (so no one can overwrite them), while allowing for one-time initialization by the Jakarta Data provider before the application gains control.

PR #261, opened 2023-09-14
refactor packages

### LINKED ISSUE #259: Ensure API classes are in a package that makes sense for them
Otavio raised this [comment](https://github.com/jakartaee/data/pull/258#discussion_r1324141108) during a code review

> IMHO: it should be compatible not just with the repository domain but with all Data patterns that we might bring in the future; thus, it should be in the data package.

Although the comment was directed at a specific class being introduced, it's good general point to consider for other API classes that we have already defined.  We have a number of classes in the `repository` package that shouldn't be limited to usage by repository pattern only. For example, the `Sort` class has general usefulness that could apply to other patterns that might be introduced in the future.

Here is a _one idea_ for arranging packages:

- jakarta.data
Streamable, Sort, Direction, Limit get moved here

- jakarta.data.exceptions
No changes

- jakarta.data.model
(TBD - if we add API classes for static metamodel)

- jakarta.data.page
Pageable, Page, Slice, KeysetAwarePage, KeysetAwareSlice, KeysetCursor, Pagination get moved here

- jakarta.data.repository
Other classes remain here

Any thoughts on this before we put a pull together for it?

njr-11:
Updates packages according to #259
fixes #259

The search/replace on package names included the signature test files, but those will need to be separately regenerated at somepoint to allow for the moved packages and other cases where classes are added/modified in other pulls.

### DIFF
--- a/api/src/main/java/jakarta/data/page/package-info.java
+++ b/api/src/main/java/jakarta/data/page/package-info.java
@@ -19,6 +19,27 @@
 /**
  * <p>Splits query results into slices or pages.</p>
  *
+ * <p>For query results that are expected to be large, it can be useful to
+ * read the results in separate parts instead of retrieving all of the results
+ * from the database at once.</p>
+ *
+ * <p>To accomplish this, ensure that the results are sorted in a consistent
+ * order. This is often done by having the final sort criterion be the
+ * unique identifier, but it can be achieved by other means as long as the
+ * order is guaranteed to be deterministic. Here is an example of sort criteria
+ * to allow data to be read in slices or pages,</p>
+ *
+ * <pre>
+ * Pageable.ofSize(25).sortBy(Sort.asc("lastName"),
+ *                            Sort.asc("firstName"),
+ *                            Sort.asc("id")); // unique identifier
+ * </pre>
+ *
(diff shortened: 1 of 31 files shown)


api/src/main/java/jakarta/data/page/package-info.java
```diff
@@ -0,0 +1,47 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * <p>Splits query results into slices or pages.</p>
```
otaviojava, line 20: shall we include what is pagination for Jakarta Data here and why it is important?

njr-11: > shall we include what is pagination for Jakarta Data here and why it is important?

That's a good idea.  I added it in https://github.com/jakartaee/data/pull/261/commits/ee9f257d02da1439ad043777b3b81d8ac2f5b64f

otaviojava (approved):

PR #260, opened 2023-09-14
Query by method name and Query by parameters

otaviojava:
It is the first part of the solution to: https://github.com/jakartaee/data/issues/255

Issue #223, opened 2023-08-30
consider alternatives to entity-specific repositories


gavinking:
The specification does not explain why entity-specific repositories were settled on, instead of other more recent designs. The casual reader is left with the impression that the answer to this question is simply _"because that's what Spring Data does"_. I hope that's not the true answer, because, if so, it would be an inadequate answer.

The case against "repositories" (or DAOs, to those of us who remember) is that:

- they're a solution which was proposed back when each CRUD operation for every entity required significant amounts of hand-coded logic, but
- that's no longer the case, not with "modern" persistence solutions that have been entrenched in the industry for the past two decades, and
- entity-specific repositories are an unnatural way to organize the _interesting_ part of the persistence logic, that is, queries which span multiple entities.

Therefore, we do not recommend the use of the DAO pattern to our users. In Hibernate 6, we propose a _far_ more flexible solution which lets the developer declare and organize "query methods" (with generated implementations) however the developer prefers. The developer can use this facility to implement entity-specific repositories if they like, or not, if they don't like.

gavinking:
So, for example, consider the following (speculative) solution, just for the sake of argument:


```java
@JpaRepository(unit="documents")
interface DocumentRepository {

    void insert(@Insert Document doc);
    void insert(@Insert Author book);
    void update(@Update Document doc);
    void update(@Update Author book);

    Document document(Long bookId, Type docType);
    Book author(Long authorId);

    @Query("where title like :titlePattern")
    List<Document> documentsByTitle(String titlePattern);

    @Query("where title like :titlePattern and type = :docType and author.id = authorId")
    List<Document> documentsByTitle(String titlePattern, Type docType, Long authorId);
}
```

This interface definition gives the implementation _everything it needs_ to fill in the implementations of those methods. And I didn't need to refer to any framework-defined super-interface.

Now, perhaps you dislike my `@Insert` and `@Update` annotations (and perhaps I do too, I just invented them now).

But then, if you _do_ hate them, you can write:

```java
interface EntityRepository<E> {
    void insert(@Insert E doc);
    void update(@Update E doc);
    void delete(@Delete E doc);
}
```

and you naturally recover an entity-specific repository. This `EntityRepository` could be defined by Jakarta Data, or it could just as easily be written by the user.

But this approach _does not impose entity-specific repositories on the user. It leaves that design decision up to them.

To me this is just much nicer.

gavinking:
> this already works in Micronaut Data in a similar way. 

[...]

PR #266, opened 2023-09-18
Specify parameter-based queries/methods in spec doc

njr-11:
Continue the work to fully specify parameter-based queries/methods.  This adds on to work done in #260 (precedence rules in spec) and #263 (include in JavaDoc) to ensure these are fully specified in the specification document and that the JavaDoc is consistent with it.


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -774,7 +775,37 @@ Wildcard characters for patterns are determined by the data access provider. For
 
 For relational databases, the logical operator `And` takes precedence over `Or`, meaning that `And` is evaluated on conditions before `Or` when both are specified on the same method. For other database types, the precedence is limited to the capabilities of the database. For example, some graph databases are limited to precedence in traversal order.
 
-=== Special Parameter Handling
+==== Query by Parameters
+
+The Query by Parameters pattern determines the query conditions from the names of the method's parameters that are not of type `Limit`, `Sort`, and `Pageable`. Each query condition is an equality comparison, comparing the parameter value against the value of the entity attribute whose name matches the method parameter name. For embedded attributes, the `_` character is used as the delimiter in the method parameter name. All query conditions are implicitly joined by the `And` operator, such that all must match for an entity to be considered matching.
```
otaviojava, line 780: I guess we got this discussion somewhere before; annotations such as "Find" could get more flexibility in this case:

https://docs.jboss.org/hibernate/orm/6.3/introduction/html_single/Hibernate_Introduction.html#generated-finder-methods

otaviojava: I found it: https://github.com/jakartaee/data/issues/223

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -808,6 +839,54 @@ first20 = products.findByNameLike(name, pageable);
...
+Repository methods with a name that begins with one of the prefixes, `insert`, `save`, `update`, `delete`, can have a single entity parameter that is one of the following types, where `E` is the entity type:
+
+- `E` - the entity type
+- `E[]` - an array of the entity type
+- `E...` - a variable arguments array of the entity type
+- `Iterable<E>` and subtypes of `Iterable<E>` - a collection of multiple entities
+
+Note: A form of `delete` and `update` can be defined in a different manner under the Query by Parameters and Query by Method Name patterns. In those cases, the method can have multiple parameters, none of which can be the entity type.
+
+==== Insert Methods
```
otaviojava, line 853: When discussing a repository, the save is the core method abstraction as a collection.

[Take Oliver's comments](https://github.com/jakartaee/data/issues/229#issuecomment-1701248061)

IMHO: The insert and update methods makes more sense in the DAO pattern.

spec/src/main/asciidoc/repository.asciidoc
njr-11: > I guess we got this discussion somewhere before; annotations such as "Find" could get more flexibility in this case:

It's certainly true that including annotations could add more flexibility here.  However, that's a separate proposal, or separate aspect of a proposal, to consider and discuss separately and is not intended to be addressed by this pull.  The scope of this pull is to specify what is described under your pull #260 where it states "determining the query from the supplied parameters and the `insert`, `update`, `delete`, `find`, `count`, or `exists` prefix. Query by parameters constructs queries based on method parameters and prefixes."

spec/src/main/asciidoc/repository.asciidoc
njr-11: > Take Oliver's comments

The link you posted to Oliver's comments appear to be an argument against allowing a single repository to have methods for a variety of types of entity.  This pull has no relation to that idea.

> IMHO: The insert and update methods makes more sense in the DAO pattern.

I'm documenting how insert and update work because you added insert and update under your pull #260, "determining the query from the supplied parameters and the `insert`, `update`, `delete`, `find`, `count`, or `exists` prefix."  Since we have them in the spec, we need to document how they work.  I just noticed that `save` is absent from that list, which seems like a mistake given that we have save throughout the rest of the spec.  I added https://github.com/jakartaee/data/pull/266/commits/ba3a536e5fe6b870ee05df0eabcdebacd36b5cb4 to this pull to include it.  Please review again.

spec/src/main/asciidoc/repository.asciidoc
otaviojava: Yes, I remember this PR, and that is my point: Does it make sense to have insert and update methods?
Should we update to use "DataAcessObject" instead of Repository?

IMHO: we should remove the insert and update and abstract with "save" for a while.

njr-11: > Yes, I remember this PR, and that is my point: Does it make sense to have insert and update methods? Should we update to use "DataAcessObject" instead of Repository?
> 
> IMHO: we should remove the insert and update and abstract with "save" for a while.

Well, I'm certainly not happy that I wasted all this effort documenting insert/update based on the fact that your pull added them so I thought you were in favor of them.  But if you insist on removing them now, I'll update this pull to do so.

njr-11: @otaviojava I removed insert and update for you: https://github.com/jakartaee/data/pull/266/commits/da6993ebbc64f90d379a79f677b1b6fa71df0db5

otaviojava (approved):

PR #273, opened 2023-09-25
repository programming model

gavinking:
This draft implements the ideas proposed in #223.

This is unfinished, since it updates only the specification, and not the Java APIs.

I'm sending it "early" for the purpose of obtaining feedback.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -136,6 +136,10 @@ A repository interface may inherit methods from a superinterface.
 A Jakarta Data implementation must treat inherited abstract methods as if they were directly declared by the repository interface.
 For example, a repository interface may inherit the `CrudRepository` interface defined by this specification.
 
+Repositories perform operations on entities. For repository methods that are annotated with `@Insert`, `@Update`, `@Save`, or `@Delete`, the entity type is determined from the method parameter type.  For `find` and `delete` methods where the return type is an entity, array of entity, or parameterized type such as `List<MyEntity>` or `Page<MyEntity>`, the entity type is determined from the method return type.  For `count`, `exists`, and other `delete` methods that do not return the entity or accept the entity as a parameter, the entity type cannot be determined from the method signature and a primary entity type must be defined for the repository.
+
+Users of Jakarta Data declare a primary entity type for a repository by inheriting from a built-in repository super interface, such as `BasicRepository`, and specifying the primary entity type as the first type variable.  The primary entity type is assumed for methods that do not otherwise specify an entity type, such as `countByPriceLessThan`. Methods that require a primary entity type raise `MappingException` if a primary entity type is not provided.
+
 [NOTE]
 A Jakarta Data provider might go beyond what is required by this specification and support abstract methods which do not fall into any of the above categories.
 Such functionality is not defined by this specification, and so programs with repositories which declare such methods are not portable between providers.
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -294,7 +294,7 @@ The purpose of this method is to provide the program with direct access to the d
 For example, if the Jakarta Data provider is based on JDBC, the return type might be `java.sql.Connection` or `javax.sql.DataSource`.
 Or, if the Jakarta Data provider is backed by Jakarta Persistence, the return type might be `jakarta.persistence.EntityManager`.
 
-The Jakarta Data implementation automatically recognizes the connection types it supports.
+The Jakarta Data provider recognizes the connection types it supports and implements the method such that it returns an instance of the type of resource. If the resource type implements `java.lang.AutoCloseable` and the resource is obtained within the scope of a default method of the repository, then the Jakarta Data provider automatically closes the resource upon completion of the default method. If the method for obtaining the resource is invoked outside the scope of a default method of the repository, then the user is responsible for closing the resource instance.
 
 [NOTE]
 A Jakarta Data implementation might allow a resource accessor method to be annotated with additional metadata providing information about the connection.
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -170,7 +170,7 @@ Lifecycle methods follow the general pattern:
 ReturnType lifecycle(Entity e);
 ----
 
-where `lifecycle` is the arbitrary name of the method, `Entity` is a concrete entity class, `Lifecycle` is a lifecycle annotation, and `ReturnType` is a return type that is permitted by the lifecycle annotation JavaDoc.
+where `lifecycle` is the arbitrary name of the method, `Entity` is a concrete entity class or an `Iterable` or array of this entity, `Lifecycle` is a lifecycle annotation, and `ReturnType` is a return type that is permitted by the lifecycle annotation JavaDoc.
 
 This specification defines four built-in lifecycle annotations: `@Insert`, `@Update`, `@Delete`, and `@Save`.
 
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -160,28 +160,19 @@ Additional portability guarantees may be provided by specifications which extend
 A _lifecycle method_ is an abstract method annotated with a _lifecycle annotation_.
 Lifecycle methods allow the program to make changes to persistent data in the data store.
 
-A lifecycle method must:
+A lifecycle method must be annotated with a lifecycle annotation. The method signature of the lifecycle method, including its return type, must follow the requirements that are specified by the JavaDoc of the lifecycle annotation.
 
-- have a single parameter annotated with a lifecycle annotation, whose type is a concrete entity class `E`, and
-- be declared `void`, or return exactly the same entity type `E`.
-
-That is, its signature must be:
+Lifecycle methods follow the general pattern:
 
 [source,java]
 ----
-void lifecycle(@Lifecycle Entity e);
+@Lifecycle
+ReturnType lifecycle(Entity e);
 ----
 
-or:
+where `lifecycle` is the arbitrary name of the method, `Entity` is a concrete entity class, `Lifecycle` is a lifecycle annotation, and `ReturnType` is a return type that is permitted by the lifecycle annotation JavaDoc.
 
-[source,java]
-----
-Entity lifecycle(@Lifecycle Entity e);
-----
-
-where `lifecycle` is the arbitrary name of the method, `Entity` is a concrete entity class, and `Lifecycle` is a lifecycle annotation.
-
-This specification defines three built-in lifecycle annotations: `@Insert`, `@Update`, and `@Delete`.
+This specification defines four built-in lifecycle annotations: `@Insert`, `@Update`, `@Delete`, and `@Save`.
 
 For example:
 
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -157,7 +157,7 @@ Additional portability guarantees may be provided by specifications which extend
 
 ==== Lifecycle methods
 
-A _lifecycle method_ is an abstract method annotated with a _lifecycle annotation_.
+A *lifecycle method* is an abstract method annotated with a _lifecycle annotation_.
 Lifecycle methods allow the program to make changes to persistent data in the data store.
 
 A lifecycle method must be annotated with a lifecycle annotation. The method signature of the lifecycle method, including its return type, must follow the requirements that are specified by the JavaDoc of the lifecycle annotation.
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -112,6 +112,246 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
 
 Jakarta Data empowers developers to shape their data access strategies by defining entity classes and repositories, with implementations seamlessly adapting to the chosen data store. This flexibility and Jakarta Data's persistence-agnostic approach promote robust data management within Java applications.
 
+=== Repository interfaces
+
+A Jakarta Data repository is a Java interface annotated `@Repository`.
+A repository interface may declare:
+
+- _abstract_ (non-`default`) methods, and
+- _concrete_ (`default`) methods.
+
+A concrete method may call other methods of the repository, including abstract methods.
+
+Every abstract method of the interface is usually either:
+
+- an entity instance _lifecycle method_,
+- an _annotated query method_,
+- an _automatic query method_, or
+- a _resource accessor method_.
+
+A repository may declare lifecycle methods for a single entity type, or for multiple related entity types.
+Similarly, a repository might have query methods which return different entity types.
+
+A repository interface may inherit methods from a superinterface.
+A Jakarta Data implementation must treat inherited abstract methods as if they were directly declared by the repository interface.
+For example, a repository interface may inherit the `CrudRepository` interface defined by this specification.
+
+[NOTE]
+A Jakarta Data provider might go beyond what is required by this specification and support abstract methods which do not fall into any of the above categories.
+Such functionality is not defined by this specification, and so programs with repositories which declare such methods are not portable between providers.
+
+The subsections below specify the rules that an abstract method declaration must observe so that the Jakarta Data implementation is able to provide an implementation of the abstract method.
+
+- If every abstract method of a repository complies with the rules specified below, then the Jakarta Data implementation must provide an implementation of the repository.
+- Otherwise, if a repository declares an abstract method which does not comply with the rules specified below, or makes use of functionality which is not supported by the Jakarta Data implementation, then an error might be produced by the Jakarta Data implementation at build time or at runtime.
+
+The portability of a given repository interface between Jakarta Data implementations depends on the portability of the entity types it uses.
+If an entity class is not portable between given implementations, then any repository which uses the entity class is also unportable between those implementations.
+
+[NOTE]
+Additional portability guarantees may be provided by specifications which extend this specification, specializing to a given class of datastore.
+
+==== Lifecycle methods
+
+A _lifecycle method_ is an abstract method annotated with a _lifecycle annotation_.
+Lifecycle methods allow the program to make changes to persistent data in the data store.
+
+A lifecycle method must:
+
+- have a single parameter annotated with a lifecycle annotation, whose type is a concrete entity class `E`, and
+- be declared `void`, or return exactly the same entity type `E`.
+
+That is, its signature must be:
+
+[source,java]
+----
+void lifecycle(@Lifecycle Entity e);
+----
+
+or:
+
+[source,java]
+----
+Entity lifecycle(@Lifecycle Entity e);
+----
+
+where `lifecycle` is the arbitrary name of the method, `Entity` is a concrete entity class, and `Lifecycle` is a lifecycle annotation.
+
+This specification defines three built-in lifecycle annotations: `@Insert`, `@Update`, and `@Delete`.
+
+For example:
+
+[source,java]
+----
+void insertBook(@Insert Book book);
+----
+
+Lifecycle methods are not guaranteed to be portable between all providers.
+
+[NOTE]
+Any Jakarta Data implementation is strongly encouraged to support lifecycle methods declared using the built-in lifecycle annotations, to enable portability of user code between Jakarta Data providers.
+However, since these lifecycle operations are not universal across all possible data stores, support for `@Insert`, `@Update`, and `@Delete` is not required by this specification.
+
+There is no special programming model for lifecycle annotations.
+The Jakarta Data implementation automatically recognizes the lifecycle annotations it supports.
+
+[NOTE]
+====
+A Jakarta Data provider might extend this specification to define additional lifecycle annotations, or to support lifecycle methods with signatures other than the usual signatures defined above. For example, a provider might support "merge" methods declared as follows:
+
+[source,java]
+----
+Book mergeBook(@Merge Book book);
+----
+
+Such lifecycle methods are not portable between Jakarta Data providers.
+====
+
+==== Annotated query methods
+
+An _annotated query method_ is an abstract method annotated by a _query annotation_ type.
+The query annotation specifies a query in some datastore-native query language.
+
+Each parameter of an annotated query method must either:
+
+- have exactly the same name and type as a named parameter of the query,
+- have exactly the same type and position within the parameter list of the method as a positional parameter of the query, or
+- be of type `Limit`, `Pageable`, or `Sort`.
+
+A repository with annotated query methods with named parameters should be compiled so that parameter names are preserved in the class file (for example, using `javac -parameters`), or the parameter names must be specified explicitly using the `@Param` annotation.
+
+A parameter of an annotated query method may not be annotated with lifecycle annotation.
+
+The return type of the annotated query method must be consistent with the result type of the query specified by the query annotation.
+
+[NOTE]
+====
+The result type of a query depends on datastore-native semantics, and so the return type of an annotated query method cannot be specified here.
+However, Jakarta Data implementations are strongly encouraged to support the following return types:
+
+- for a query which returns a single result of type `T`, the type `T` itself, or `Optional<T>`,
+- for a query which returns many results of type `T`, the types `List<T>`, `Page<T>`, and `T[]`.
+
+Furthermore, implementations are encouraged to support `void` as the return type for a query which never returns a result.
+====
+
+This specification defines the built-in `@Query` annotation, which may be used to specify a query in an arbitrary query language understood by the Jakarta Data provider.
+
+For example, using a named parameter:
+
+[source,java]
+----
+@Query(“where title like :title order by title”)
+Page<Book> booksByTitle(String title, Pageable page);
+----
+
+Or, using a positional parameter:
+
+[source,java]
+----
+@Query(“delete from Book where isbn = ?1”)
+void deleteBook(String isbn);
+----
+
+Programs which make use of annotated query methods are not portable between providers.
+
+[NOTE]
+====
+A Jakarta Data provider might extend this specification to define its own query annotation types.
+For example, a provider might define a `@SQL` annotation for declaring queries written in SQL.
+====
+
+There is no special programming model for query annotations.
+The Jakarta Data implementation automatically recognizes the query annotations it supports.
+
+==== Automatic query methods
+
+An _automatic query method_ is an abstract method with return type `E`, `Optional<E>`, `Page<E>`, or `List<E>`, where `E` is an entity class, and where each parameter either:
+
+- has exactly the same type and name as a persistent field or property of the entity class, or
+- is of type `Limit`, `Pageable`, or `Sort`.
+
+A repository with automatic query methods should be compiled so that parameter names are preserved in the class file (for example, using `javac -parameters`), or the parameter names must be specified explicitly using the `@Param` annotation.
+
+A parameter of an automatic query method may not be annotated with lifecycle annotation.
+
+For example:
+
+[source,java]
+----
+Book bookByIsbn(String isbn);
+
+List<Book> booksByYear(Year year, Sort order, Limit limit)
+----
+
+Automatic query methods _are_ portable between providers.
+
+==== Resource accessor method
+
+A _resource accessor method_ is a method with no parameters which returns a type supported by the Jakarta Data provider.
+The purpose of this method is to provide the program with direct access to the data store.
+
+For example, if the Jakarta Data provider is based on JDBC, the return type might be `java.sql.Connection` or `javax.sql.Datasource`.
+Or, if the Jakarta Data provider is backed by Jakarta Persistence, the return type might be `jakarta.persistence.EntityManager`.
+
+The Jakarta Data implementation automatically recognizes the connection types it supports.
+
+[NOTE]
+A Jakarta Data implementation might allow a resource accessor method to be annotated with additional metadata providing information about the connection.
+
+For example:
+
+[source,java]
+----
+Connection connection();
+
+default void cleanup() {
+    try (Statement s = connection().createStatement()) {
+        s.executeUpdate("truncate table books");
+    }
+}
+----
+
+A repository may have at most one resource accessor method.
+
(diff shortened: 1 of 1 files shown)


njr-11:
At first glance, there is a lot of good information here and lots of it looks good, but I highly recommend splitting this up and delivering it in smaller pieces so that we can ensure it is sufficiently well-reviewed.

gavinking:
@njr-11 I don't want to split it up into bits because it's a single internally self-consistent model. The "bits" don't make sense in isolation from each other.

And the core if this change is only 200 lines of contiguous text. So I would much prefer to take feedback as comments on the  relevant lines. (Or as comments on issue #223, of course.)

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,246 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+[NOTE]
+Any Jakarta Data implementation is strongly encouraged to support lifecycle methods declared using the built-in lifecycle annotations, to enable portability of user code between Jakarta Data providers.
+However, since these lifecycle operations are not universal across all possible data stores, support for `@Insert`, `@Update`, and `@Delete` is not required by this specification.
+
+There is no special programming model for lifecycle annotations.
+The Jakarta Data implementation automatically recognizes the lifecycle annotations it supports.
+
+[NOTE]
+====
+A Jakarta Data provider might extend this specification to define additional lifecycle annotations, or to support lifecycle methods with signatures other than the usual signatures defined above. For example, a provider might support "merge" methods declared as follows:
```
njr-11, line 200: We had agreement to add lifecycle methods, but not to include the annotations on them. I would like to ensure these annotations are not used up in this manner where they do not seem to add much additional benefit and will not later be available for the spec to use for other purposes.  Some possible revisions to this section follow:

```suggestion
A _lifecycle method_ is an abstract method that accepts an entity as the method's only parameter.
Lifecycle methods allow the program to make changes to persistent data in the data store.

A lifecycle method must:

- have a single parameter whose type is a concrete entity class `E`, and
- be declared `void`, or return exactly the same entity type `E`.

That is, its signature must be:

[source,java]
----
void lifecycle(Entity e);
----

or:

[source,java]
----
Entity lifecycle(Entity e);
----

where `lifecycle` is a method name that starts with `insert`, `update`, `save`, or `delete`, and `Entity` is a concrete entity class.

For example:

[source,java]
----
void insertBook(Book book);
----

[NOTE]
Jakarta Data providers must support lifecycle methods to the extent that the data store is capable of the corresponding operation. If the data store is not capable of the operation, the Jakarta Data provider must raise `UnsupportedOperationException` if the operation is attemped.

There is no special programming model for lifecycle annotations.
The Jakarta Data implementation automatically recognizes the lifecycle annotations it supports.

[NOTE]
====
A Jakarta Data provider might extend this specification to define lifecycle annotations, or to support lifecycle methods with signatures other than the usual signatures defined above. For example, a provider might support "merge" methods declared as follows:
```

We also need to specify the behavior that users can expect from each lifecycle method that is defined from the specification.  There's a pull where I have included this information, but it's been held up over other arguing and hasn't gone in.  After this goes in, that pull will likely be a massive merge conflict and we'll need to cancel it, but maybe we can reuse some of the specifics from there in defining what insert and update are and what the user can expect from them.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,246 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+An _annotated query method_ is an abstract method annotated by a _query annotation_ type.
+The query annotation specifies a query in some datastore-native query language.
+
+Each parameter of an annotated query method must either:
+
+- have exactly the same name and type as a named parameter of the query,
+- have exactly the same type and position within the parameter list of the method as a positional parameter of the query, or
+- be of type `Limit`, `Pageable`, or `Sort`.
+
+A repository with annotated query methods with named parameters should be compiled so that parameter names are preserved in the class file (for example, using `javac -parameters`), or the parameter names must be specified explicitly using the `@Param` annotation.
```
njr-11, line 221:
```suggestion
A repository with annotated query methods with named parameters must be compiled so that parameter names are preserved in the class file (for example, using `javac -parameters`), or the parameter names must be specified explicitly using the `@Param` annotation.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,246 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+
+A parameter of an annotated query method may not be annotated with lifecycle annotation.
```
njr-11, line 223: No longer needed after previous suggested changes,
```suggestion
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,246 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+For example, a repository interface may inherit the `CrudRepository` interface defined by this specification.
+
+[NOTE]
+A Jakarta Data provider might go beyond what is required by this specification and support abstract methods which do not fall into any of the above categories.
+Such functionality is not defined by this specification, and so programs with repositories which declare such methods are not portable between providers.
+
+The subsections below specify the rules that an abstract method declaration must observe so that the Jakarta Data implementation is able to provide an implementation of the abstract method.
+
+- If every abstract method of a repository complies with the rules specified below, then the Jakarta Data implementation must provide an implementation of the repository.
+- Otherwise, if a repository declares an abstract method which does not comply with the rules specified below, or makes use of functionality which is not supported by the Jakarta Data implementation, then an error might be produced by the Jakarta Data implementation at build time or at runtime.
```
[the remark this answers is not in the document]
starksm64: Component specs can have optional features. It is the platform and profiles that cannot have optional features, and so have to either explicitly state that such optional features from a component spec is either not required and excluded from the platform/profile, or to state that it is required behavior.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,246 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+
+A parameter of an automatic query method may not be annotated with lifecycle annotation.
+
+For example:
+
+[source,java]
+----
+Book bookByIsbn(String isbn);
+
+List<Book> booksByYear(Year year, Sort order, Limit limit)
```
njr-11, line 284: When I previously saw "Automatic Query methods" in the list of specification-defined types of repository methods,

> - an entity instance _lifecycle method_,
> - an _annotated query method_,
> - an _automatic query method_, or
> - a _resource accessor method_.

I assumed that the term "automatic query method" encompassed the Query by Method Name pattern as well as the pattern described in this section because both automatically generate a query.  This section either needs a more specific name and the previously mentioned list needs to be updated, or we need to update this section to include Query by Method Name.  I'll explore the latter with a suggestion,

```suggestion
==== Automatic query methods

An _automatic query method_ is an abstract method that either follows the Query by Method Name pattern where the Jakarta Data provider generates a query based on the name of the method, or follows a pattern for whereby the Jakarta Data provider automatically generates a find query based on the names of parameters that are supplied to the method, where the method has return type that identifies the entity, such as `E`, `Optional<E>`, `Page<E>`, or `List<E>`, where `E` is an entity class. Each parameter must either:

- have exactly the same type and name as a persistent field or property of the entity class, or
- be of type `Limit`, `Pageable`, or `Sort`.

A repository with automatic query methods that are based on parameters must either be compiled so that parameter names are preserved in the class file (for example, using `javac -parameters`), or the parameter names must be specified explicitly using the `@Param` annotation.

For example:

[source,java]
----
Book bookByIsbn(String isbn);

List<Book> booksByYear(Year year, Sort order, Limit limit)
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,246 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+
+==== Resource accessor method
+
+A _resource accessor method_ is a method with no parameters which returns a type supported by the Jakarta Data provider.
+The purpose of this method is to provide the program with direct access to the data store.
+
+For example, if the Jakarta Data provider is based on JDBC, the return type might be `java.sql.Connection` or `javax.sql.Datasource`.
+Or, if the Jakarta Data provider is backed by Jakarta Persistence, the return type might be `jakarta.persistence.EntityManager`.
+
+The Jakarta Data implementation automatically recognizes the connection types it supports.
```
njr-11, line 297: It looks like the paragraph that describes how the specification arranges for these connections and entity managers to be closed was lost when the section was moved here.  The following update adds it back in,

```suggestion
The Jakarta Data provider recognizes the connection types it supports and implements the method such that it returns an instance of the type of resource. If the resource type implements `java.lang.AutoCloseable` and the resource is obtained within the scope of a default method of the repository, then the Jakarta Data provider automatically closes the resource upon completion of the default method. If the method for obtaining the resource is invoked outside the scope of a default method of the repository, then the user is responsible for closing the resource instance.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,246 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+public interface ProductRepository extends CrudRepository<Product, Long> {
+
+  // Assumes that the Product entity has attributes: yearProduced
+  List<Product> findMadeIn(int yearProduced, Sort... sorts);
+
+  // Assumes that the Product entity has attributes: name, status.
+  boolean existsWithStatus(String name, Status status);
+
+  // Assumes that the Product entity has attributes: yearProduced
+  void deleteOutdated(int yearProduced);
```
njr-11, line 351: These examples don't obey the rules that you defined for automatic queries in that they don't return the entity or multiple entities.  Also, I don't see any way that your proposal distinguishes between exists vs delete vs find.  As I understand from the way you wrote it, it will only do find.  If you want to cover these other things, we would need to include more details from the Query by Parameter Names pattern.

njr-11 (changes requested):

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -839,44 +1015,6 @@ first20 = products.findByNameLike(name, pageable);
...
-System.out.println("Saved version " + product.getVersion() + " of " + product);
-----
-
-==== Delete Methods
-
-Delete methods remove entities from the database based on the unique identifier of the entity parameter value. If the entity is versioned (for example, with `@code jakarta.persistence.Version` or by another convention from the entity model such as having an attribute named `version`), then the version must also match. Other entity attributes do not need to match. The the unique identifier of an entity is not found in the database or its version does not match, a `delete` method with a return type of `void` or `Void` must raise `OptimisticLockingFailureException`.
-
-==== Return Types
-
 Refer to the Jakarta Data module JavaDoc section on "Return Types for Repository Methods" for a listing of valid return types for methods with entity parameters.
```
njr-11, line 1018: Most of the detail in this section with the specification requirements for delete(entity) and save(entity) operations was lost in your rewrite.  The easiest fix would be to put the whole section back.

(FYI - This is where I have a separate pull that has been held up that will add update/insert requirements as well).

njr-11 (changes requested):

spec/src/main/asciidoc/repository.asciidoc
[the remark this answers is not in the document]
njr-11: > Component specs can have optional features. It is the platform and profiles that cannot have optional features, and so have to either explicitly state that such optional features from a component spec is either not required and excluded from the platform/profile, or to state that it is required behavior.

That means we could technically include this as an optional feature, but it will burden the Jakarta EE Platform and Web Profile with having to call it out as not required and excluded.  For something this minor, I'd say it's definitely not worth it to include this as an optional feature.  It's only restating what vendors with the optional capability would already be assumed to be doing.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,246 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+- an _annotated query method_,
+- an _automatic query method_, or
+- a _resource accessor method_.
+
+A repository may declare lifecycle methods for a single entity type, or for multiple related entity types.
+Similarly, a repository might have query methods which return different entity types.
+
+A repository interface may inherit methods from a superinterface.
+A Jakarta Data implementation must treat inherited abstract methods as if they were directly declared by the repository interface.
+For example, a repository interface may inherit the `CrudRepository` interface defined by this specification.
```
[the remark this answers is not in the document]
gavinking: > we need to specify how the user identifies the entity types to the Jakarta Data provider

My preference would be that: (at least for ):

- there simply is no primary entity for repositories which don't inherit `BasicRepository`, and
- instead, there's a way to infer the entity type for each abstract method of the repository. 

I _believe_ that's doable. The return type of a query method should be enough to infer the queried entity  type.

But we need to talk that through because I might be missing something.

Otherwise your suggested change is fine. 👍

spec/src/main/asciidoc/repository.asciidoc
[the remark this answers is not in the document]
gavinking: @njr-11 I'm not sure I understand. What precisely is the optional feature you mean?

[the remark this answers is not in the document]
gavinking: > Beyond the optional features, we have extensibility. I can create a Repository specialization that might work with a provider or a specific database.

Yes, it's more this extensibility that I'm trying to describe here. I'm not trying to actually _define_ any optional features, just kind of lay out a framework for how Jakarta Data can be extended by datastore-specific "extensions".

spec/src/main/asciidoc/repository.asciidoc
gavinking: @njr-11 the last comment I saw from @otaviojava was [this one](https://github.com/jakartaee/data/issues/223#issuecomment-1707757922), where he said:

> I would include the annotations for semantics and flexibility reasons.
>
> Thus, Insert, Update, Delete, and Save annotations.

And I agree on this: the annotations mean that there different kinds of lifecycle methods e.g. `@Save` vs `@Persist` can be technology-specific and not hardcoded in as primitives.

Note that anyone extending `BasicRepository` would not need to use the annotations, since they would just be overriding methods already annotated. (Though we should clarify that, it's not well-defined in what I've written!)

Anyway, let's discuss this more.

spec/src/main/asciidoc/repository.asciidoc
gavinking: > I assumed that the term "automatic query method" encompassed the Query by Method Name pattern as well as the pattern described in this section because both automatically generate a query. 

It could, I still have not thought this through completely.

spec/src/main/asciidoc/repository.asciidoc
gavinking: Yes, good point, nice catch. This was a pre-existing example that I just moved from where it was before, without proper review.

So the thing is that it looks like you guys had added this language:

> A method name prefix, either `find` or `exists` or `count` or `delete`, specifies the type of query to be performed.

Now, FWIW, I would simply remove that feature.  Neither `exists`, `count`, or `delete` queries are common enough in my experience to be privileged in this way, and, IMO, we should get away from assigning special magical semantics via method naming conventions. It's easy enough to handle these things using `@Query`.

But, again, let's discuss it.

spec/src/main/asciidoc/repository.asciidoc
gavinking: Oops.

spec/src/main/asciidoc/repository.asciidoc
gavinking: Anyway, I fixed the code example to use `@Query`.

spec/src/main/asciidoc/repository.asciidoc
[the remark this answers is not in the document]
njr-11: > The return type of a query method should be enough to infer the queried entity type.

That should work for `find` queries, and for queries with `@Query` that return some other type, the entity type wouldn't be needed because the query language is hard-coded.  The only places I can think of where the entity type is needed but isn't determinable from the method signature are the `countBy` and `existsBy` methods, and  I agree it doesn't make sense to be able to define such methods when there isn't a primary entity type.  I'll see if I can update my prior code suggestion with that.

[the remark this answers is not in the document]
njr-11: That worked. My above suggestion is updated now.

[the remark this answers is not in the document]
gavinking: OK cool, I will take a look later.

spec/src/main/asciidoc/repository.asciidoc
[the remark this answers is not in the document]
njr-11: This is my mistake. I had misread the section and thought it was referring to something else.  When I re-read it now, it looks great.  There is no reason to take this out.  It's only calling out that Jakarta Data providers might choose to extend the spec by allowing additional abstract methods, which is fine.  Please disregard my prior comment on this one. I'll mark the conversion resolved so it doesn't hold up the pull.

spec/src/main/asciidoc/repository.asciidoc
njr-11: I don't have any issue with what the annotations would provide, I just don't want them used up for this, which I don't see providing all that much value, because I think we could do different, more valuable things with some of the same annotations later.  In my suggestion, I tried to illustrate how you can achieve basically the same pattern (relying on method name prefix) without the annotations.  This would give us the flexibility to decide on what the annotations will be used for in the future, while remaining compatible with using them for this purpose if we decide to do so.  Vendors could also supply their own annotations in the mean time if they believe it's important.

spec/src/main/asciidoc/repository.asciidoc
njr-11: > Now, FWIW, I would simply remove that feature. Neither `exists`, `count`, or `delete` queries are common enough in my experience to be privileged in this way, and, IMO, we should get away from assigning special magical semantics via method naming conventions. It's easy enough to handle these things using `@Query`.

I don't see those as important either.  I think they were only added based on a misunderstanding that you wanted them.  The user already has alternatives in `@Query` as you mentioned, as well as `existsBy`, `countBy`, and `deleteBy`, all of which provide more function than magical `exists`, `count`, or `delete` based on parameter names.  Your update to switch the examples to `@Query` looks good, so I'll mark this resolved.

spec/src/main/asciidoc/repository.asciidoc
njr-11: Resolving this per discussion that occurred on this week's Jakarta Data meeting.  I'll re-review it in terms of the decision that was made to use the lifecycle annotations.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,246 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+[source,java]
+----
+void insertBook(@Insert Book book);
+----
+
+Lifecycle methods are not guaranteed to be portable between all providers.
+
+[NOTE]
+Any Jakarta Data implementation is strongly encouraged to support lifecycle methods declared using the built-in lifecycle annotations, to enable portability of user code between Jakarta Data providers.
+However, since these lifecycle operations are not universal across all possible data stores, support for `@Insert`, `@Update`, and `@Delete` is not required by this specification.
```
njr-11, line 193:
```suggestion
Jakarta Data providers must support lifecycle methods to the extent that the data store is capable of the corresponding operation. If the data store is not capable of the operation, the Jakarta Data provider must raise `UnsupportedOperationException` when the operation is attempted, per the requirements of the JavaDoc for the lifecycle annotation.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,246 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+
+Each parameter of an annotated query method must either:
+
+- have exactly the same name and type as a named parameter of the query,
+- have exactly the same type and position within the parameter list of the method as a positional parameter of the query, or
+- be of type `Limit`, `Pageable`, or `Sort`.
+
+A repository with annotated query methods with named parameters must be compiled so that parameter names are preserved in the class file (for example, using `javac -parameters`), or the parameter names must be specified explicitly using the `@Param` annotation.
+
+A parameter of an annotated query method may not be annotated with lifecycle annotation.
```
njr-11, line 223:
```suggestion
An annotated query method must not also be annotated with a lifecycle annotation.
```

njr-11 (changes requested):
@gavinking I revised my review comments based on the decision that was made at this past week's Jakarta Data meeting.  It seems like we reached agreement on most of what is covered in this pull, so hopefully once the updates from the review comments are included, we'll be able to merge in your pull.

spec/src/main/asciidoc/repository.asciidoc
gavinking: @njr-11 I'm fine with this change, but I would like it to also mention that a provider might alternatively choose to detect and report the error at compile time.

njr-11: > @njr-11 I'm fine with this change, but I would like it to also mention that a provider might alternatively choose to detect and report the error at compile time.

Good point. I added that to the code suggestion.

spec/src/main/asciidoc/repository.asciidoc
njr-11: It no longer matters that the sections on Delete and Save were lost because the requirements are now documented on the annotations themselves (in the JavaDoc), so the issue I originally raised here no longer applies and I'll mark this conversation as resolved.

njr-11:
@gavinking with Otavio's pull to add the lifecycle annotations now being merged, I was able to resolve one of the 4 remaining review comments.  Could you incorporate the remaining 3 suggestions (I don't think these should be controversial) and rebase your pull to resolve merge conflicts?  After that I think we should be ready to merge this.  Thanks!

njr-11:
We need to make progress on getting this issue in because it is accumulating merge conflicts with other pulls that update the main spec document.  I'll look into merging in the suggestions, which I don't see any opposition to, resolve the current set of merge conflicts, switch the pull to Ready for Review, and approve my own review of the pull.  If I'm able to do these things successfully, then I'll post a second comment to let others know to review it, and then hopefully we can get it merged.

njr-11 (approved):

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,239 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+
+- an entity instance _lifecycle method_,
+- an _annotated query method_,
+- an _automatic query method_, or
+- a _resource accessor method_.
+
+A repository may declare lifecycle methods for a single entity type, or for multiple related entity types.
+Similarly, a repository might have query methods which return different entity types.
+
+A repository interface may inherit methods from a superinterface.
```
otaviojava, line 135: Could you explain more about this point?

What are you talking about here?

Are you talking about buil-in repositories? Or non-built-in ones?

njr-11:
This is ready for review now if anyone else want to review it.  I was able to successfully resolve the merge conflicts and add in the suggestions.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,239 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+
+The portability of a given repository interface between Jakarta Data implementations depends on the portability of the entity types it uses.
+If an entity class is not portable between given implementations, then any repository which uses the entity class is also unportable between those implementations.
+
+[NOTE]
+Additional portability guarantees may be provided by specifications which extend this specification, specializing to a given class of datastore.
+
+==== Lifecycle methods
+
+A _lifecycle method_ is an abstract method annotated with a _lifecycle annotation_.
```
otaviojava, line 160: Do you think we should use bold instead?
```suggestion
A *lifecycle method* is an abstract method annotated with a _lifecycle annotation_.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,239 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+
+==== Annotated query methods
+
+An _annotated query method_ is an abstract method annotated by a _query annotation_ type.
+The query annotation specifies a query in some datastore-native query language.
+
+Each parameter of an annotated query method must either:
+
+- have exactly the same name and type as a named parameter of the query,
+- have exactly the same type and position within the parameter list of the method as a positional parameter of the query, or
```
otaviojava, line 213: Should we include it otherwise?

spec/src/main/asciidoc/repository.asciidoc
njr-11: That seems trivial. I'll commit it to make progress.

spec/src/main/asciidoc/repository.asciidoc
njr-11: > Are you talking about buil-in repositories? Or non-built-in ones?

Given that all the built-in repositories only allow for a single entity type, per the parameterized type variable for the entity, we can assume that Gavin is targeting these statements at user-defined (non-built-in) repository interfaces.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,239 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+- a _resource accessor method_.
+
+A repository may declare lifecycle methods for a single entity type, or for multiple related entity types.
+Similarly, a repository might have query methods which return different entity types.
+
+A repository interface may inherit methods from a superinterface.
+A Jakarta Data implementation must treat inherited abstract methods as if they were directly declared by the repository interface.
+For example, a repository interface may inherit the `CrudRepository` interface defined by this specification.
+
+Repositories perform operations on entities. For repository methods that are annotated with `@Insert`, `@Update`, `@Save`, or `@Delete`, the entity type is determined from the method parameter type.  For `find` and `delete` methods where the return type is an entity, array of entity, or parameterized type such as `List<MyEntity>` or `Page<MyEntity>`, the entity type is determined from the method return type.  For `count`, `exists`, and other `delete` methods that do not return the entity or accept the entity as a parameter, the entity type cannot be determined from the method signature and a primary entity type must be defined for the repository.
```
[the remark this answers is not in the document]
njr-11: > > the entity type is determined from the method parameter type
> 
> Should we explain that it also works with an array or iterable parameter?

The language "determined from the method parameter type" already allows for for that.  If you have a specific suggestion to improve the language, please propose it.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,240 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+
+- an entity instance _lifecycle method_,
+- an _annotated query method_,
+- an _automatic query method_, or
+- a _resource accessor method_.
+
+A repository may declare lifecycle methods for a single entity type, or for multiple related entity types.
+Similarly, a repository might have query methods which return different entity types.
+
+A repository interface may inherit methods from a superinterface.
```
otaviojava, line 135:
```suggestion
A non-built-in repository interface may inherit methods from a superinterface.
```

Can we procedure this way? 
WDYT?

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,240 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+
+The subsections below specify the rules that an abstract method declaration must observe so that the Jakarta Data implementation is able to provide an implementation of the abstract method.
+
+- If every abstract method of a repository complies with the rules specified below, then the Jakarta Data implementation must provide an implementation of the repository.
+- Otherwise, if a repository declares an abstract method which does not comply with the rules specified below, or makes use of functionality which is not supported by the Jakarta Data implementation, then an error might be produced by the Jakarta Data implementation at build time or at runtime.
+
+The portability of a given repository interface between Jakarta Data implementations depends on the portability of the entity types it uses.
+If an entity class is not portable between given implementations, then any repository which uses the entity class is also unportable between those implementations.
+
+[NOTE]
```
otaviojava, line 156:
```suggestion
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -112,6 +112,240 @@ Subsequently, an implementation of Jakarta Data, specifically tailored to the ch
...
+
+Lifecycle methods follow the general pattern:
+
+[source,java]
+----
+@Lifecycle
+ReturnType lifecycle(Entity e);
+----
+
+where `lifecycle` is the arbitrary name of the method, `Entity` is a concrete entity class, `Lifecycle` is a lifecycle annotation, and `ReturnType` is a return type that is permitted by the lifecycle annotation JavaDoc.
```
otaviojava, line 174:
```suggestion
where `lifecycle` is the arbitrary name of the method, `Entity` is a concrete entity class or an `Iterable` or array of this entity, `Lifecycle` is a lifecycle annotation, and `ReturnType` is a return type that is permitted by the lifecycle annotation JavaDoc.
```

spec/src/main/asciidoc/repository.asciidoc
njr-11: My vote would be against adding that language because it's true that built-in repository interfaces can inherit methods from a superinterface as well.  This currently exists in the spec where PageableRepository inherits from BasicRepository and inherits its methods as well.  And where CrudRepository inherits from BasicRepository and inherits its methods as well.

otaviojava:
We have a long time in this PR; I will merge it and open PRs to improve it.

spec/src/main/asciidoc/repository.asciidoc
gavinking: Yes, that's right.

njr-11:
> If you let me know about any concerns that were left unresolved, I will send a PR with those edits later today.

Thanks for all of your contributions on this @gavinking !  I believe we resolved everything and am not aware of anything else to follow up on from this pull.

otaviojava:
@gavinking thank you for all the support.
If you don´t mind review it again.

Please, let us know if we are missing anything.

gavinking:
@otaviojava yes I will do that, I promise!

PR #274, opened 2023-09-26
Rename BasicRepository

otaviojava:
# Changes

- Rename CrudRepository to BasicRepository; thus, we can make sure this is for representing the repository pattern.


njr-11 (approved):
All of the changes here look correct.  After it is merged we can check again if any were missed, such as in JavaDoc/spec doc.

PR #281, opened 2023-10-02
Corrections to requirements around CDI extensions

### LINKED ISSUE #278: Erroneous statements in sections on CDI Full/Lite Integration
The sections on CDI Full Integration/CDI Lite Integration in the Jakarta Data spec are written under the erroneous assumption that BuildCompatibleExtension only works on CDI Lite and not CDI Full and makes overly restrictive requirements on providers aligning with that.  As brought up on a couple of the Jakarta Data calls, I had our CDI team review the section and they also objected to some other aspects of it, such as implying that CDI does certain things that are specific to Jakarta Data when it's actually Jakarta Data defining how to build upon CDI to do so.  I'll look into updating this section of the spec to address these things and maybe even include snippets of an example extension.

njr-11:
Git diff was doing a poor job of matching up what changed, making it look like more changed than really did, so I tried splitting the changes into separate commits.  It might help to look at those individually when reviewing.

Fixes #278

### DIFF
--- a/spec/src/main/asciidoc/jakarta-ee.adoc
+++ b/spec/src/main/asciidoc/jakarta-ee.adoc
@@ -40,25 +40,9 @@ repository.save(ferrari);
 In the above snippet, we inject the `CarRepository` instance into our code using the `@Inject` annotation. Once injected, we can use the repository object to invoke various data access methods provided by the `CarRepository` interface, such as `save()`, `findByType()`, and `findByName()`.
 This integration between CDI and Jakarta EE Data allows for seamless management of repository instances and enables easy data access and manipulation within your Jakarta EE applications.
 
-==== CDI Extension
+=== Jakarta Data Providers
 
-Jakarta Data and CDI integrate through CDI extensions, which provide a mechanism to customize and enhance the behavior of CDI in your Jakarta EE application. Jakarta Data integration with Jakarta CDI applies to both CDI full and CDI Lite.
-
-*CDI Full Integration*:
-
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -40,43 +40,86 @@ repository.save(ferrari);
...
 
-The entity annotation class is the primary strategy to avoid conflicts between Jakarta Data providers. In most cases, it is sufficient to differentiate between providers. However, there may be situations where more than the entity annotation class is needed. In such cases, the application can specify the name of the desired Jakarta Data provider using the optional `provider` attribute of the `Repository` annotation.
+The entity annotation class is the primary strategy to avoid conflicts between Jakarta Data providers. In most cases, it is sufficient to differentiate between providers. In situations where multiple Jakarta Data providers support the same entity annotation class, the application can specify the name of the desired Jakarta Data provider using the optional `provider` attribute of the `Repository` annotation.
+
+To ensure compatibility and prevent conflicts, Jakarta Data providers must disregard all `Repository` annotations that specify a different provider's name through the `Repository.provider()` attribute. By ignoring these annotations, Jakarta Data providers allow other Jakarta Data providers to handle them.
+
+==== Example CDI Extension
```
otaviojava, line 73: Does it make sense to have this?
IMHO: we should remove this example extension.

njr-11: > Does it make sense to have this?
> IMHO: we should remove this example extension.

I included it because I thought it was helpful, but have removed it now under https://github.com/jakartaee/data/pull/281/commits/296d6ab831e44fc8b1582753a8975253842c6718 because you don't think it makes sense to include.  Please re-review the pull.

PR #282, opened 2023-10-04
Adding CRUD Annotations for Dynamic Repositories

otaviojava:
This Pull Request (PR) proposes the addition of new annotations for dynamic Repositories within the Jakarta Data framework: `Insert`, `Delete`, `Update`, and `Save`. These annotations simplify and enhance the interaction with data entities in dynamic Repository interfaces. They provide a more intuitive way to declare methods for various CRUD (Create, Read, Update, Delete) operations.

**Benefits:**

1. **Improved Code Readability:** The new annotations make it easier to identify the intended operation of each method within dynamic Repositories, improving code readability and maintainability.

2. **Consistency Across Operations:** Each annotation enforces a specific operation type, ensuring consistency in method declarations and reducing potential errors.

3. **Enhanced Documentation:** The Javadoc documentation for these annotations offers comprehensive explanations, helping developers understand their usage and behavior.

4. **Type-Safe Returns:** The annotations provide type safety by ensuring that the return type matches the parameter type, enhancing code quality.

**Sample Dynamic Repository:**

As an example, consider a dynamic Repository interface for a `Garage` that manages `Car` entities:

```java
@Repository
interface Garage {

    @Insert
    Car park(Car car);  // Insert a car entity and return it

    @Delete
    void unpark(Car car);  // Delete a car entity

}
```

In this example, the `@Save` annotation indicates that the `park` method performs an insert operation and returns the inserted `Car` entity. Similarly, the `@Update`, `@Delete`, and `@Save` annotations indicate their respective operations.

**Usage:**

Developers can now use these annotations to create dynamic Repository interfaces that are more expressive and self-documenting, leading to cleaner and more maintainable code.

### DIFF
--- a/api/src/main/java/jakarta/data/Delete.java
+++ b/api/src/main/java/jakarta/data/Delete.java
@@ -33,7 +33,10 @@ import java.lang.annotation.Target;
  *     <li>An {@code Iterable} of entities to be deleted.</li>
  *     <li>An array of entities to be deleted.</li>
  * </ul>
- * <p>The return type of the annotated method must be {@code void}, {@code boolean}, a numeric primitive type (such as {@code int}), or a corresponding primitive wrapper type (such as {@link Integer}). A boolean return type indicates whether or not an entity was deleted from the database. A numeric return type indicates how many entities were deleted from the database.
+ * <p>The return type of the annotated method must be {@code void}, {@code boolean}, a numeric primitive type
+ * (such as {@code int}), or a corresponding primitive wrapper type (such as {@link Integer}).
+ * A boolean return type indicates whether or not an entity was deleted from the database.
+ * A numeric return type indicates how many entities were deleted from the database.
  * </p>
  * <p>Deletion of a given entity is performed by matching the entity's Id. If the entity is versioned (e.g.,
  * with {@code jakarta.persistence.Version}), the version is also checked for consistency during deletion.
@@ -42,7 +45,9 @@ import java.lang.annotation.Target;
  *
  * * <p>If this annotation is combined with other operation annotations (e.g., {@code @Insert}, {@code @Update},
  *  * {@code @Save}), it will throw an {@link IllegalStateException} as only one operation type can be specified.</p>
- * <p>If the unique identifier of an entity is not found in the database or its version does not match, and the return type of the annotated method is {@code void} or {@code Void}, the method must raise {@link jakarta.data.exceptions.OptimisticLockingFailureException}.
+ * <p>If the unique identifier of an entity is not found in the database or its version does not match, and the return
+ * type of the annotated method is {@code void} or {@code Void}, the method must
+ * raise {@link jakarta.data.exceptions.OptimisticLockingFailureException}.
  */
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
--- a/api/src/main/java/jakarta/data/Delete.java
+++ b/api/src/main/java/jakarta/data/Delete.java
@@ -42,7 +42,14 @@ import java.lang.annotation.Target;
  * with {@code jakarta.persistence.Version}), the version is also checked for consistency during deletion.
  * Properties other than the Id and version do not need to match for deletion.
  * </p>
- *
+ * <p>For example, consider an interface representing a garage:</p>
+ * <pre>
+ * {@literal @}Repository
+ * interface Garage {
+ *     {@literal @}Save
+ *     Car unpark(Car car);
+ * }
+ * </pre>
  * * <p>If this annotation is combined with other operation annotations (e.g., {@code @Insert}, {@code @Update},
  *  * {@code @Save}), it will throw an {@link IllegalStateException} as only one operation type can be specified.</p>
  * <p>If the unique identifier of an entity is not found in the database or its version does not match, and the return
--- a/api/src/main/java/jakarta/data/Delete.java
+++ b/api/src/main/java/jakarta/data/Delete.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data;
+
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface Delete {
+}
--- a/api/src/main/java/jakarta/data/Delete.java
+++ b/api/src/main/java/jakarta/data/Delete.java
@@ -41,7 +41,7 @@ import java.lang.annotation.Target;
  * </p>
  *
  *
- * @see jakarta.data.exceptions.OptimisticLockingFailureException
+ * <p>If the unique identifier of an entity is not found in the database or its version does not match, and the return type of the annotated method is {@code void} or {@code Void}, the method must raise {@link jakarta.data.exceptions.OptimisticLockingFailureException}.
  */
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
--- a/api/src/main/java/jakarta/data/Delete.java
+++ b/api/src/main/java/jakarta/data/Delete.java
@@ -24,6 +24,28 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+/**
+ * <p>The {@code Delete} annotation indicates to dynamic templates or repositories that the annotated method
+ * will perform a delete operation. This method should have a unique parameter whose type can be one of the following:
+ * </p>
+ * <ul>
+ *     <li>The entity to be deleted.</li>
+ *     <li>An {@code Iterable} of entities to be deleted.</li>
+ *     <li>An array of entities to be deleted.</li>
+ * </ul>
+ * <p>The return type of the annotated method should be {@code void}, as the delete operation does not return a value.
+ * </p>
+ * <p>Deletion of a given entity is performed by matching the entity's Id. If the entity is versioned (e.g.,
+ * with {@code jakarta.persistence.Version}), the version is also checked for consistency during deletion.
+ * Properties other than the Id and version do not need to match for deletion.
+ * </p>
+ *
+ * @throws OptimisticLockingFailureException if the entity is not found in the database for deletion
+ *         or has a version for optimistic locking that is inconsistent with the version in the database.
+ * @throws NullPointerException when the entity is null.
+ *
+ * @see OptimisticLockingFailureException
+ */
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.METHOD)
--- a/api/src/main/java/jakarta/data/Delete.java
+++ b/api/src/main/java/jakarta/data/Delete.java
@@ -51,7 +51,7 @@ import java.lang.annotation.Target;
  * }
  * </pre>
  * * <p>If this annotation is combined with other operation annotations (e.g., {@code @Insert}, {@code @Update},
- *  * {@code @Save}), it will throw an {@link IllegalStateException} as only one operation type can be specified.</p>
+ *  * {@code @Save}), it will throw an {@link UnsupportedOperationException} as only one operation type can be specified.</p>
  * <p>If the unique identifier of an entity is not found in the database or its version does not match, and the return
  * type of the annotated method is {@code void} or {@code Void}, the method must
  * raise {@link jakarta.data.exceptions.OptimisticLockingFailureException}.
--- a/api/src/main/java/jakarta/data/Delete.java
+++ b/api/src/main/java/jakarta/data/Delete.java
@@ -40,11 +40,11 @@ import java.lang.annotation.Target;
  * Properties other than the Id and version do not need to match for deletion.
  * </p>
  *
- * @throws OptimisticLockingFailureException if the entity is not found in the database for deletion
+ * @throws jakarta.data.exceptions.OptimisticLockingFailureException if the entity is not found in the database for deletion
  *         or has a version for optimistic locking that is inconsistent with the version in the database.
  * @throws NullPointerException when the entity is null.
  *
- * @see OptimisticLockingFailureException
+ * @see jakarta.data.exceptions.OptimisticLockingFailureException
  */
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
--- a/api/src/main/java/jakarta/data/Delete.java
+++ b/api/src/main/java/jakarta/data/Delete.java
@@ -40,9 +40,6 @@ import java.lang.annotation.Target;
  * Properties other than the Id and version do not need to match for deletion.
  * </p>
  *
- * @throws jakarta.data.exceptions.OptimisticLockingFailureException if the entity is not found in the database for deletion
- *         or has a version for optimistic locking that is inconsistent with the version in the database.
- * @throws NullPointerException when the entity is null.
  *
  * @see jakarta.data.exceptions.OptimisticLockingFailureException
  */
--- a/api/src/main/java/jakarta/data/Delete.java
+++ b/api/src/main/java/jakarta/data/Delete.java
@@ -25,8 +25,8 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * <p>The {@code Delete} annotation indicates to dynamic templates or repositories that the annotated method
- * will perform a delete operation. This method should have a unique parameter whose type can be one of the following:
+ * <p>The {@code Delete} annotation indicates that the annotated repository method requests one or more
+ * entities to be removed from the database. This method must have a single parameter whose type must be one of the following:
  * </p>
  * <ul>
  *     <li>The entity to be deleted.</li>
--- a/api/src/main/java/jakarta/data/Delete.java
+++ b/api/src/main/java/jakarta/data/Delete.java
@@ -46,7 +46,7 @@ import java.lang.annotation.Target;
  * <pre>
  * {@literal @}Repository
  * interface Garage {
- *     {@literal @}Save
+ *     {@literal @}Delete
  *     Car unpark(Car car);
  * }
  * </pre>
--- a/api/src/main/java/jakarta/data/Delete.java
+++ b/api/src/main/java/jakarta/data/Delete.java
@@ -33,7 +33,7 @@ import java.lang.annotation.Target;
  *     <li>An {@code Iterable} of entities to be deleted.</li>
  *     <li>An array of entities to be deleted.</li>
  * </ul>
- * <p>The return type of the annotated method should be {@code void}, as the delete operation does not return a value.
+ * <p>The return type of the annotated method must be {@code void}, {@code boolean}, a numeric primitive type (such as {@code int}), or a corresponding primitive wrapper type (such as {@link Integer}). A boolean return type indicates whether or not an entity was deleted from the database. A numeric return type indicates how many entities were deleted from the database.
  * </p>
  * <p>Deletion of a given entity is performed by matching the entity's Id. If the entity is versioned (e.g.,
  * with {@code jakarta.persistence.Version}), the version is also checked for consistency during deletion.
--- a/api/src/main/java/jakarta/data/Delete.java
+++ b/api/src/main/java/jakarta/data/Delete.java
@@ -40,7 +40,8 @@ import java.lang.annotation.Target;
  * Properties other than the Id and version do not need to match for deletion.
  * </p>
  *
- *
+ * * <p>If this annotation is combined with other operation annotations (e.g., {@code @Insert}, {@code @Update},
+ *  * {@code @Save}), it will throw an {@link IllegalStateException} as only one operation type can be specified.</p>
  * <p>If the unique identifier of an entity is not found in the database or its version does not match, and the return type of the annotated method is {@code void} or {@code Void}, the method must raise {@link jakarta.data.exceptions.OptimisticLockingFailureException}.
  */
 @Documented
--- a/api/src/main/java/jakarta/data/Insert.java
+++ b/api/src/main/java/jakarta/data/Insert.java
@@ -25,8 +25,8 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * <p>The {@code Insert} annotation indicates to dynamic templates or repositories that the annotated method
- * will perform an insert operation. This method should have a unique parameter whose type can be one of the following:
+ * <p>The {@code Insert} annotation indicates that the annotated repository method requests that one or more entities
+ * be inserted into the database. This method must have a single parameter whose type must be one of the following:
  * </p>
  * <ul>
  *     <li>The entity to be inserted.</li>
--- a/api/src/main/java/jakarta/data/Insert.java
+++ b/api/src/main/java/jakarta/data/Insert.java
@@ -63,8 +63,8 @@ import java.lang.annotation.Target;
  * </p>
  *
  * <p>If this annotation is combined with other operation annotations (e.g., {@code @Update}, {@code @Delete},
- *  {@code @Save}), it will throw an {@link IllegalStateException} as only one operation type can be specified, where
- *  a Jakarta Data provider implementation choose to detect (and report) this error at compile time or at runtime..</p>
+ *  {@code @Save}), it will throw an {@link UnsupportedOperationException} because only one operation type can be specified.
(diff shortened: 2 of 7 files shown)


otaviojava:
@gavinking @njr-11 it still WIP, I hope finishi until our meeting today.

njr-11 (commented):
> I think this is reasonable, though would still expect a method called `save` or `delete` to just do the right thing

I agree with Graeme. This pattern is reasonable, but unnecessary.

It would also prevent us from being able to better define these annotations for other purposes in the future, so I am not in favor of it.

If the consensus of the group is to put these changes in now, there are a couple of minor errors in the JavaDoc that will need correcting.  I do appreciate that you've put in the effort to fully specify the requirements that the user will need to know within the JavaDoc.

api/src/main/java/jakarta/data/Delete.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * <p>The {@code Delete} annotation indicates to dynamic templates or repositories that the annotated method
+ * will perform a delete operation. This method should have a unique parameter whose type can be one of the following:
```
njr-11, line 29:
```suggestion
 * <p>The {@code Delete} annotation indicates that the annotated repository method requests one or more
 * entities to be removed from the database. This method must have a single parameter whose type must be one of the following:
```

api/src/main/java/jakarta/data/Insert.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * <p>The {@code Insert} annotation indicates to dynamic templates or repositories that the annotated method
+ * will perform an insert operation. This method should have a unique parameter whose type can be one of the following:
```
njr-11, line 29:
```suggestion
 * <p>The {@code Insert} annotation indicates that the annotated repository method requests that one or more entities
 * be inserted into the database. This method must have a single parameter whose type must be one of the following:
```

api/src/main/java/jakarta/data/repository/Save.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * <p>The {@code Save} annotation indicates to dynamic templates or repositories that the annotated method
+ * will perform a save operation. This method should have a unique parameter whose type can be one of the following:
```
njr-11, line 29:
```suggestion
 * <p>The {@code Save} annotation indicates that the annotated repository method
 * updates one or more entities if found in the database
 * and inserts entities into the database that are not found.
 * This method must have a single parameter whose type must be one of the following:
```

api/src/main/java/jakarta/data/repository/Save.java
```diff
@@ -0,0 +1,67 @@
...
+ * </p>
+ * <ul>
+ *     <li>The entity to be saved.</li>
+ *     <li>An {@code Iterable} of entities to be saved.</li>
+ *     <li>An array of entities to be saved.</li>
+ * </ul>
+ * <p>The return type of the annotated method should be the same as the parameter type, ensuring consistency
```
njr-11, line 36:
```suggestion
 * <p>The return type of the annotated method must be the same as the parameter type, ensuring consistency
```

api/src/main/java/jakarta/data/repository/Save.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * <p>For example, consider an interface representing a garage:</p>
+ * <pre>
+ * @Repository
+ * interface Garage {
+ *     {@literal @}Save
+ *     Car parking(Car car);
+ * }
+ * </pre>
+ * <p>The {@code @Insert} annotation can be used to indicate that the {@code parkCar} method is responsible for inserting
```
njr-11, line 57:
```suggestion
 *     Car park(Car car);
 * }
 * </pre>
 * <p>The {@code @Insert} annotation can be used to indicate that the {@code park(Car)} method is responsible for updating the entity in database if it already exists there and otherwise inserting
```

njr-11 (changes requested):
It turns out there were more issues than I realized in the JavaDoc. Hopefully I found all of them.

api/src/main/java/jakarta/data/Insert.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data;
```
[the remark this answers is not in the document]
otaviojava: IMHO: Nope, because those are not exclusive to the repository as Save is.

otaviojava:
@njr-11

njr-11 (changes requested):
It looks like you missed some review comments from before (Github probably collapsed/hid them) and I added a couple of new ones, at least one of which is my fault from introducing a grammar mistake (left out the word "the") in one of my previous suggestions.

api/src/main/java/jakarta/data/Insert.java
```diff
@@ -0,0 +1,65 @@
...
+ * </p>
+ * <ul>
+ *     <li>The entity to be inserted.</li>
+ *     <li>An {@code Iterable} of entities to be inserted.</li>
+ *     <li>An array of entities to be inserted.</li>
+ * </ul>
+ * <p>The return type of the annotated method should match the type of the parameter. For example, if the method is
+ * annotated with {@code @Insert} and takes a parameter of type {@code Car car}, the return type should be {@code Car}.
+ * Similarly, if the parameter is an {@code Iterable<Car>} or an array of {@code Car}, the return type should be the
+ * corresponding type.
```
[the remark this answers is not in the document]
otaviojava: in case of EntityManager it return the entity self, but not all databases have EntityManager.

api/src/main/java/jakarta/data/Insert.java
```diff
@@ -0,0 +1,65 @@
...
+ * </p>
+ * <p>After invoking this method, it is recommended not to use the entity value supplied as a parameter, as this method
+ * makes no guarantees about the state of the entity value after insertion.
+ * </p>
+ * <p>If the entity uses optimistic locking, and the version differs from the version in the database, an
+ * {@link jakarta.data.exceptions.OptimisticLockingFailureException} may be thrown.
```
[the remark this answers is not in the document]
otaviojava: That is wrong because it does not work with the NoSQL database. Once not all databases have consistency, they work with eventual consistency, it should  be might return the EntityExistsException.

api/src/main/java/jakarta/data/Insert.java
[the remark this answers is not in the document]
njr-11: > in case of EntityManager it return the entity self, but not all databases have EntityManager.

You have it backwards.  Please refer to the [EntityManager.persist](https://jakarta.ee/specifications/persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/entitymanager#persist(java.lang.Object)) JavaDoc which clearly shows a return type of `void`.  It does not return the entity.  If you want the entity returned, you need to use `save` (which corresponds to EntityManager.merge and is the update-or-insert-if-not-present pattern).  The whole point of offering the insert vs save is to let the user optimize for different patterns.  Forcing an insert to also fetch an updated entity (if it's even possible - it won't be in JPA if the id is generated) would be contrary to that.

api/src/main/java/jakarta/data/Insert.java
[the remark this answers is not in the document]
njr-11: Thanks for pointing that out. I'll add another suggestion to cover eventual consistency.

api/src/main/java/jakarta/data/Delete.java
```diff
@@ -0,0 +1,63 @@
...
+ * <p>For example, consider an interface representing a garage:</p>
+ * <pre>
+ * {@literal @}Repository
+ * interface Garage {
+ *     {@literal @}Delete
+ *     Car unpark(Car car);
+ * }
+ * </pre>
+ * * <p>If this annotation is combined with other operation annotations (e.g., {@code @Insert}, {@code @Update},
+ *  * {@code @Save}), it will throw an {@link IllegalStateException} as only one operation type can be specified.</p>
```
njr-11, line 54:
```suggestion
 *  * {@code @Save}), it will throw an {@link UnsupportedOperationException} as only one operation type can be specified.</p>
```

api/src/main/java/jakarta/data/Insert.java
```diff
@@ -0,0 +1,66 @@
...
+ *     {@literal @}Insert
+ *     Car parking(Car car);
+ * }
+ * </pre>
+ * <p>The {@code @Insert} annotation can be used to indicate that the {@code parkCar} method is responsible for inserting
+ * a car entity into a database.
+ * </p>
+ *
+ * <p>If this annotation is combined with other operation annotations (e.g., {@code @Update}, {@code @Delete},
+ *  {@code @Save}), it will throw an {@link IllegalStateException} as only one operation type can be specified.</p>
```
njr-11, line 60:
```suggestion
 *  {@code @Save}), it will throw an {@link UnsupportedOperationException} as only one operation type can be specified.</p>
```

api/src/main/java/jakarta/data/Update.java
```diff
@@ -0,0 +1,64 @@
...
+ * <p>For example, consider an interface representing a garage:</p>
+ * <pre>
+ * &#64;Repository
+ * interface Garage {
+ *     {@literal @}Update
+ *     Car update(Car car);
+ * }
+ * </pre>
+ * <p>If this annotation is combined with other operation annotations (e.g., {@code @Insert}, {@code @Delete},
+ * {@code @Save}), it will throw an {@link IllegalStateException} as only one operation type can be specified.</p>
```
njr-11, line 57:
```suggestion
 * {@code @Save}), it will throw an {@link UnsupportedOperationException} as only one operation type can be specified.</p>
```

njr-11 (changes requested):
Recommend switching the exception type from `IllegalStateException` to `UnsupportedOperationException`

Looking at the JavaDoc for IllegalStateException would imply that the operation is valid, but being invoked at the wrong time: 
> Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation.

That isn't what we want to convey.  The point we are trying to get across with the exception is that combining multiple operation annotations isn't supported.

otaviojava:
> Recommend switching the exception type from `IllegalStateException` to `UnsupportedOperationException`
> 
> Looking at the JavaDoc for IllegalStateException would imply that the operation is valid, but being invoked at the wrong time:
> 
> > Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation.
> 
> That isn't what we want to convey. The point we are trying to get across with the exception is that combining multiple operation annotations isn't supported.

@njr-11 I mean when this code happened:

```java
@Insert
@Update 
void operation(Car car);
```

Given a relational database, both have support, don't they? But is it one I will execute?
In my mind, it should be Illegal.

njr-11:
> ```java
> @Insert
> @Update 
> void operation(Car car);
> ```
>   
> Given a relational database, both have support, don't they? But is it one I will execute?
> In my mind, it should be Illegal.
> ```

I thin we agree here.  A combination `@Insert @Update` operation or combination `@Save @Delete` operation or whatever other combination, is not supported and so must raise an exception.  I was only pointing out that the definition of IllegalStateException didn't seem to make sense here and suggested UnsupportedOperationException instead, which you already added the commits for, so I think we're fine on this.

otaviojava:
> > ```java
> > @Insert
> > @Update 
> > void operation(Car car);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Given a relational database, both have support, don't they? But is it one I will execute?
> > In my mind, it should be Illegal.
> 
> I think we agree here. A combination `@Insert @Update` operation or combination `@Save @Delete` operation or whatever other combination, is not supported and so must raise an exception. I was only pointing out that the description of IllegalStateException indicates it isn't the right exception for this and I suggested UnsupportedOperationException instead, which you already added the commits for, so I think we're fine on this.

I will improve the documentation, but I hold this one, mainly, because the insert and update behavior will move to here as well.

api/src/main/java/jakarta/data/Insert.java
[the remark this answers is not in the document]
gavinking: FWIW I would have thought they belonged in `jakarta.data.repository`.

gavinking:
>  I was only pointing out that the description of `IllegalStateException` indicates it isn't the right exception for this and I suggested `UnsupportedOperationException` instead, which you already added the commits for, so I think we're fine on this.

It might be worth pointing out that an implementation of Jakarta Data might also choose to detect this error at compile time.

api/src/main/java/jakarta/data/Update.java
```diff
@@ -0,0 +1,68 @@
...
+ * &#64;Repository
+ * interface Garage {
+ *     {@literal @}Update
+ *     Car update(Car car);
+ * }
+ * </pre>
+ * <p>If this annotation is combined with other operation annotations (e.g., {@code @Insert}, {@code @Delete},
+ * {@code @Save}), it will throw an {@link IllegalStateException} as only one operation type can be specified, where
+ * a Jakarta Data provider implementation choose to detect (and report) this error at compile time or at runtime.</p>
+ * @see jakarta.data.exceptions.OptimisticLockingFailureException
```
njr-11, line 62:
```suggestion
 * {@code @Save}), it will throw an {@link UnsupportedOperationException} because only one operation type can be specified.
 * A Jakarta Data provider implementation must detect (and report) this error at compile time or at runtime.</p>
```

api/src/main/java/jakarta/data/repository/Save.java
```diff
@@ -0,0 +1,73 @@
...
+ * and inserts entities into the database that are not found.
+ * This method must have a single parameter whose type must be one of the following:
+ * </p>
+ * <ul>
+ *     <li>The entity to be saved.</li>
+ *     <li>An {@code Iterable} of entities to be saved.</li>
+ *     <li>An array of entities to be saved.</li>
+ * </ul>
+ * <p>The return type of the annotated method must be the same as the parameter type, ensuring consistency
+ * with the saved entity or entities.
```
njr-11, line 39:
```suggestion
 * <p>The return type of an annotated method that requires a single entity as the parameter
 * must have a return type that is {@code void}, {@code Void}, or the same type as the parameter.
 * The return type of an annotated method that accepts an {@code Iterable} or array of entities
  * as the parameter must have a return type that is {@code void}, {@code Void},
  * or an {@code Iterable} or array of the entity.
```

api/src/main/java/jakarta/data/repository/Save.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * <p>For example, consider an interface representing a garage:</p>
+ * <pre>
+ * {@literal @}Repository
+ * interface Garage {
+ *     {@literal @}Save
+ *     Car park(Car car);
+ * }
+ * </pre>
+ * <p>The {@code @Update} annotation can be used to indicate that the {@code park(Car)} method is responsible
```
njr-11, line 59:
```suggestion
 * <p>The {@code @Save} annotation can be used to indicate that the {@code park(Car)} method is responsible
```

api/src/main/java/jakarta/data/repository/Save.java
```diff
@@ -0,0 +1,73 @@
...
+ * </ul>
+ * <p>The return type of the annotated method must be the same as the parameter type, ensuring consistency
+ * with the saved entity or entities.
+ * </p>
+ * <p>Saving an entity involves persisting it in the database. If the entity has an ID or key that already exists
+ * in the database, the method will update the existing record. If the entity does not exist in the database or has a
+ * null ID, this method will insert a new record. The entity instance returned by this method will be updated with
+ * any automatically generated or incremented values that changed due to the save operation.
+ * </p>
+ * <p>After invoking this method, avoid using the entity value that was supplied as a parameter, as it may not accurately
```
njr-11, line 46:
```suggestion
 * <p>Entities that are returned by the annotated method must include all values that were
 * written to the database, including all automatically generated values and incremented values
 * that changed due to the save. The position of entities within an {@code Iterable} or array return value
 * must correspond to the position of entities in the parameter based on the unique identifier of the entity.</p>
 * <p>After invoking this method, avoid using the entity value that was supplied as a parameter, because it might not accurately
```

api/src/main/java/jakarta/data/repository/Save.java
```diff
@@ -0,0 +1,73 @@
...
+ * }
+ * </pre>
+ * <p>The {@code @Update} annotation can be used to indicate that the {@code park(Car)} method is responsible
+ * for updating the entity in the database if it already exists there and otherwise inserting
+ * a car entity into a database.
+ * </p>
+ *
+ * <p>If this annotation is combined with other operation annotations (e.g., {@code @Update}, {@code @Delete},
+ *  {@code @Insert}), it will throw an {@link IllegalStateException} as only one operation type can be specified, where
+ *  a Jakarta Data provider implementation choose to detect (and report) this error at compile time or at runtime.</p>
```
njr-11, line 66:
```suggestion
 *  {@code @Insert}), it will throw an {@link UnsupportedOperationException} because only one operation type can be specified.
 *  A Jakarta Data provider implementation must detect (and report) this error at compile time or at runtime.</p>
```

njr-11 (changes requested):
I'll agree to the same compromise here as on 287 - the entity return value is okay for Insert, but not for Update, where it is inefficient and forces extra retrieval from the database.

api/src/main/java/jakarta/data/Insert.java
```diff
@@ -0,0 +1,67 @@
...
+ *     Car parking(Car car);
+ * }
+ * </pre>
+ * <p>The {@code @Insert} annotation can be used to indicate that the {@code parkCar} method is responsible for inserting
+ * a car entity into a database.
+ * </p>
+ *
+ * <p>If this annotation is combined with other operation annotations (e.g., {@code @Update}, {@code @Delete},
+ *  {@code @Save}), it will throw an {@link IllegalStateException} as only one operation type can be specified, where
+ *  a Jakarta Data provider implementation choose to detect (and report) this error at compile time or at runtime..</p>
```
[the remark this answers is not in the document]
otaviojava: Thank you for pointing it out, Nathan.
But at the it shows: **[UnsupportedOperationException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/UnsupportedOperationException.html)**

> Thrown to indicate that the requested operation is not supported.

What I understand: Both annotations are supported, but in this case the user is doing something wrong! Should we have an exception for this?
What do you think?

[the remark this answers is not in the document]
njr-11: > Thank you for pointing it out, Nathan. But at the it shows: **[UnsupportedOperationException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/UnsupportedOperationException.html)**
> 
> > Thrown to indicate that the requested operation is not supported.
> 
> What I understand: Both annotations are supported, but in this case the user is doing something wrong! Should we have an exception for this? What do you think?

The Jakarta Data provider would need to include a meaningful message for it to make sense, for example:

`UnsupportedOperationException: The YourRepository.doSomething repository method has conflicting annotations: @Save and @Delete.`

It would be fine with me if we instead introduce a new exception type for this. For example,

`AnnotationConflictException: @Save and @Delete cannot be combined on the YourRepository.doSomething repository method.`

My point here was just that `IllegalStateException` doesn't fit.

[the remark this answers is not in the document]
otaviojava: I checked in the exception present in Java; the IllegalArgumentException was the closest that makes sense for us.

Maybe create one might sound better, such as IllegalConfigurationException...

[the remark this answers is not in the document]
otaviojava: Ahhh, that makes sense.
thank you.

api/src/main/java/jakarta/data/Delete.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data;
```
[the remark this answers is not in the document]
otaviojava: I moved, done.

otaviojava:
> I'll agree to the same compromise here as on 287 - the entity return value is okay for Insert, but not for Update, where it is inefficient and forces extra retrieval from the database.

As we did with insert, how about allowing both?
So, in a case that does not need such as a mutable entity or the entity does not have any kind of version is fine.

However, if the user works with a Java record of any version control they can work with that as well.

api/src/main/java/jakarta/data/repository/Insert.java
```diff
@@ -0,0 +1,75 @@
...
+ * Entities that are returned by the annotated method must include all values that were
+ * written to the database, including all automatically generated values and incremented values
+ * that changed due to the insert. The position of entities within an {@code Iterable} or array return value
+ * must correspond to the position of entities in the parameter based on the unique identifier of the entity.
+ * </p>
+ * <p>After invoking this method, it is recommended not to use the entity value supplied as a parameter, as this method
+ * makes no guarantees about the state of the entity value after insertion.
+ * </p>
+ * <p>If the entity uses optimistic locking, and the version differs from the version in the database, an
+ * {@link jakarta.data.exceptions.OptimisticLockingFailureException} may be thrown.
```
njr-11, line 53: Re-posting this review comment because git marked it outdated and was hiding it,
```suggestion
 * <p>If an entity of this type with the same unique identifier already exists in the database
 * and the databases performs ACID (atomic, consistent, isolated, durable) transactions,
 * then annotated method raises {@link jakarta.data.exceptions.EntityExistsException}.
 * In databases that follow the BASE model or use an append model to write data,
 * this exception is not thrown.
```

api/src/main/java/jakarta/data/repository/Insert.java
```diff
@@ -0,0 +1,75 @@
...
+ * <p>For example, consider an interface representing a garage:</p>
+ * <pre>
+ * {@literal @}Repository
+ * interface Garage {
+ *     {@literal @}Insert
+ *     Car parking(Car car);
+ * }
+ * </pre>
+ * <p>The {@code @Insert} annotation can be used to indicate that the {@code parkCar} method is responsible for inserting
+ * a car entity into a database.
```
njr-11, line 64: Re-posting this review comment because git was hiding it as outdated:
```suggestion
 *     Car park(Car car);
 * }
 * </pre>
 * <p>The {@code @Insert} annotation can be used to indicate that the {@code park(Car)} method is responsible for inserting
 * a {@code Car} entity into a database.
```

api/src/main/java/jakarta/data/repository/Update.java
```diff
@@ -0,0 +1,67 @@
...
+ * A boolean return type indicates whether a matching entity was found in the database to update.
+ * A numeric return type indicates how many matching entities were found in the database to update.
+ * </p>
+ * <p>Updating an entity involves modifying its existing data in the database. The method will search for the entity
+ * in the database using its ID (and version, if versioned) and then update the corresponding record with the new data. After invoking
+ * this method, do not continue to use the entity value that is supplied as a parameter, as it may not accurately
+ * reflect the changes made during the update process.
+ * </p>
+ * <p>If the entity uses optimistic locking and its version differs from the version in the database,
+ * an {@link jakarta.data.exceptions.OptimisticLockingFailureException} will be thrown.
```
njr-11, line 48: Re-posting another review comment that git is hiding as outdated:
```suggestion
 * <p>If the entity does not exist in the database or it is versioned and its version differs from the version in the database,
 * no update is made and no error is raised.
```

njr-11:
> > I'll agree to the same compromise here as on 287 - the entity return value is okay for Insert, but not for Update, where it is inefficient and forces extra retrieval from the database.
> 
> As we did with insert, how about allowing both? So, in a case that does not need such as a mutable entity or the entity does not have any kind of version is fine.
> 
> However, if the user works with a Java record of any version control they can work with that as well.

Yes, that's a good point, in the case of the annotation, we don't need to choose a single return type, and can be more flexible like this.  I would be fine with that, but please let's use a different pull after this one goes in.  It is already big enough and I think we're otherwise almost ready to merge this pull.  I highlighted 3 code review comments from before (hopefully non-controversial) that hid had marked outdated but still apply.

otaviojava:
> > > I'll agree to the same compromise here as on 287 - the entity return value is okay for Insert, but not for Update, where it is inefficient and forces extra retrieval from the database.
> > 
> > 
> > As we did with insert, how about allowing both? So, in a case that does not need such as a mutable entity or the entity does not have any kind of version is fine.
> > However, if the user works with a Java record of any version control they can work with that as well.
> 
> Yes, that's a good point, in the case of the annotation, we don't need to choose a single return type, and can be more flexible like this. I would be fine with that, but please let's use a different pull after this one goes in. It is already big enough and I think we're otherwise almost ready to merge this pull. I highlighted 3 code review comments from before (hopefully non-controversial) that hid had marked outdated but still apply.

You are right it became huge!
Please review it and I will open a new PR to move this update discussion.

njr-11 (approved):
> You are right it became huge!
> Please review it and I will open a new PR to move this update discussion.

It looks good now. Thanks!

PR #283, opened 2023-10-04
insert and update methods on built-in repository

njr-11:
It was just decided on today's Jakarta Data call that we will include the insert and update methods on the built-in repository interface.  This pull copies over the updates for it from the #270 commits.

### DIFF
--- a/api/src/main/java/jakarta/data/exceptions/EntityExistsException.java
+++ b/api/src/main/java/jakarta/data/exceptions/EntityExistsException.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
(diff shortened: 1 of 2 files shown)


otaviojava (approved):

otaviojava:
It does not work in some NoSQL databases, it should work with both, I will think a solution that return to work to both.

njr-11:
> Also, It might be more valuable if it returns the entity; there are cases in which I work with an immutable class that returns the entity with a new value, such as an ID is critical.
> 
> Relational might not be hard once it returns the interface itself, but if I don't have the option to return the entity result there is no way for the user to find it by the id, for example.

Relational databases would have the same problem.  In those scenarios where the id is being generated, you should be using `save`, not `insert`.

PR #284, opened 2023-10-05
Return the CrudRepository method with the insert and update methods

otaviojava:
# Changes

- Return the CrudRepository interface
- Move the insert and update methods to the CrudRepository (I understood that is the goal to return this interface)
- I added to return with the entities, mainly to be compatible with some NoSQL database or even SQL data that might work with Java records eventually


api/src/main/java/jakarta/data/repository/CrudRepository.java
```diff
@@ -0,0 +1,103 @@
...
+     * unique identifier already exists in the database, then this method raises
+     * {@link EntityExistsException} for.</p>
+     *
+     * @param entity the entity to insert. Must not be {@code null}.
+     * @return the inserted entity.
+     * @throws EntityExistsException if the entity is already present in the database.
+     * @throws NullPointerException if the entity is null.
+     * @throws UnsupportedOperationException for databases that use an append model to write data.
+     */
+    T insert(T entity);
```
otaviojava, line 50: @njr-11 is it fine to SQL, isn't it?

In NoSQL we can work with record, so, return the void, we will lost the link. That is why I would like this return.

```java
record Book(UUID databaseGenerated, String name){}

var newBook= repository.insert(book);
```

api/src/main/java/jakarta/data/repository/CrudRepository.java
```diff
@@ -0,0 +1,103 @@
...
+     * then this method raises {@link EntityExistsException}.</p>
+     *
+     * @param entities entities to insert.
+     * @return an iterable containing the inserted entities.
+     * @throws EntityExistsException if any of the entities are already present in the database.
+     * @throws NullPointerException if either the iterable is null or any element is null.
+     * @throws UnsupportedOperationException for Key-Value and Wide-Column databases
+     *         that use an append model to write data.
+     */
+    Iterable<T> insertAll(Iterable<T> entities);
```
njr-11, line 64: The return value should not be switched to the inserted entities, which would require additional retrieval from the database. When users want the pattern of performing the additional retrieval, they will use saveAll, not insertAll.

```suggestion
     * @throws EntityExistsException if any of the entities are already present in the database.
     * @throws NullPointerException if either the iterable is null or any element is null.
     * @throws UnsupportedOperationException for Key-Value and Wide-Column databases
     *         that use an append model to write data.
     */
    void insertAll(Iterable<T> entities);
```

njr-11 (changes requested):
I'm fine with the approach of moving the insert and update methods to a CrudRepository rather than having them on BasicRepository.  However, I don't agree with changing their return types.  Not needing to fetch the entities is one of the distinctions that these patterns have that make them different from save/saveAll which already offers that ability.

api/src/main/java/jakarta/data/repository/CrudRepository.java
njr-11: > @njr-11 is it fine to SQL, isn't it?
> 
> In NoSQL we can work with record, so, return the void, we will lost the link. That is why I would like this return.
> 
> ```java
> record Book(UUID databaseGenerated, String name){}
> 
> var newBook= repository.insert(book);
> ```

There isn't any difference between record and other types of entities here because Jakarta Data doesn't require or in any way even suggest that an entity that is supplied as a parameter will be updated.  There also isn't any difference between relational and NoSQL here.  Whenever you want the pattern of being able know the state of thing you are inserting or update, you use `save`, which provides that ability, not `insert`.  Inserts are not supposed to be queries that fetch information.

otaviojava:
> I'm fine with the approach of moving the insert and update methods to a CrudRepository rather than having them on BasicRepository. However, I don't agree with changing their return types. Not needing to fetch the entities is one of the distinctions that these patterns have that make them different from save/saveAll which already offers that ability.

let's take a step back:

- insert: create an information into any database.
- update: change information that already does exist. With NoSQL it might be different. Please, let me know if what I put in the documentation makes sense.
- save: it is an abstraction: As our documentation says: If the entity has an ID or key that exists in the database, the method will update the existing record. Otherwise, it will insert a new record.

If I want to explicitly insert and update I should use those methods instead of "save".

Does it make sense?

njr-11:
@otaviojava It's fine to make things unsupported where the database isn't capable of it, but the user needs to be able to rely on guarantees. Using ambiguous language as is added under https://github.com/jakartaee/data/pull/284/commits/62cc8d34e58bd51bf2e149831b6fa37713cf10b5 and https://github.com/jakartaee/data/pull/284/commits/56b7c42ad818f16e2608de1e37c747f62be6edf1 will deprive them of the ability to rely on what the method does and make it useless.  Could you revert those commits and not write it in terms of "might work for relational databases".  Relational databases shouldn't be mentioned at all since some other databases can be capable of the same.  Instead, specifically call out what types of databases are not capable as raising UnsupportedOperationException.

api/src/main/java/jakarta/data/repository/CrudRepository.java
otaviojava: Why? I mean, with SQL I can return the same values from the parameter, it might be fine, isn't it?

The point is with Record:

So, how do we handle records in any persistence capability?

```java
record Book(@AutoIncrement @Id Long id, String name){
}
var book = Book.of(null, "Effective Java");
library.insert(book);

````

njr-11:
> > I'm fine with the approach of moving the insert and update methods to a CrudRepository rather than having them on BasicRepository. However, I don't agree with changing their return types. Not needing to fetch the entities is one of the distinctions that these patterns have that make them different from save/saveAll which already offers that ability.
> 
> let's take a step back:
> 
>     * insert: create an information into any database.
> 
>     * update: change information that already does exist. With NoSQL it might be different. Please, let me know if what I put in the documentation makes sense.
> 
>     * save: it is an abstraction: As our documentation says: If the entity has an ID or key that exists in the database, the method will update the existing record. Otherwise, it will insert a new record.
> 
> 
> If I want to explicitly insert and update I should use those methods instead of "save".
> 
> Does it make sense?

Your summary here sounds great.  It's your changes in the pull regarding return types that aren't consistent with it.  There is nothing about create/insert and update that should involve fetching entities.

otaviojava:
> @otaviojava It's fine to make things unsupported where the database isn't capable of it, but the user needs to be able to rely on guarantees. Using ambiguous language as is added under [e52fbce](https://github.com/jakartaee/data/commit/e52fbce00e631ac1b05fb7099812f24061bcf8b5) and [56b7c42](https://github.com/jakartaee/data/commit/56b7c42ad818f16e2608de1e37c747f62be6edf1) will deprive them of the ability to rely on what the method does and make it useless. Could you revert those commits and not write it in terms of "might work for relational databases". Relational databases shouldn't be mentioned at all since some other databases can be capable of the same. Instead, specifically call out what types of databases are not capable as raising UnsupportedOperationException.

hum, I got it. 

Could you help me to write this point?

Those databases can insert and update but using BASE instead of ACID.

In theory, both can use those methods, however, with different data consistency models.

We also have append model, it is inserting and updating, but different from SQL...

api/src/main/java/jakarta/data/repository/CrudRepository.java
njr-11: > Why? I mean, with SQL I can return the same values from the parameter, it might be fine, isn't it?
> 
> The point is with Record:
> 
> So, how do we handle records in any persistence capability?
> 
> ```java
> record Book(@AutoIncrement @Id Long id, String name){
> }
> var book = Book.of(null, "Effective Java");
> library.insert(book);
> ```

I'm pretty sure you asked this same question elsewhere and I answered it, but I'll do so here as well.  Record entities vs Class entities and SQL vs NoSQL makes absolutely no difference here and will have the same problems either way.  What matters for this scenario is the fact that you requested an automatically generated id.  You don't know what id was generated unless the method returns it to you.  In Jakarta Data, we have a method that does exactly that: `save`, and that's what you should be using for this pattern.  Querying the database is not what `insert` is for.

otaviojava:
> > I am glad that you enjoyed the divide-and-conquer strategy :)
> > The insert and update methods are a Pandora's box. I need help making it work for SQL and NoSQL and, if possible, allow us to use Java Record on those methods.
> 
> How about using divide-and-conquer strategy in an even more fine grained way? In this pull, only move the methods to CrudRepository without changing the return types. We can all approve that and merge it. Then, in a separate pull, work on whatever precise wording we need to state when UnsupportedOperationException is raised or when EntityExistsException should not be raised. If you still believe you need insert and update retrieving entities from the database, open a separate issue for that. I'm not going to approve a pull that includes making insert and update perform retrieval of entities because that isn't CRUD.

Okay, you are right, I returned the PR to only extract those methods, once we merged we can discuss more about insert and update.

api/src/main/java/jakarta/data/repository/CrudRepository.java
otaviojava: I will close this one here once we agree to go deep on this in the next PR.

njr-11 (approved):
This looks good now. We can discuss the related topics that came up in separate issues.

PR #287, opened 2023-10-07
Make insert and update method work also with NoSQL databases

otaviojava:
# Changes

- Make the return an instance instead of a void or boolean
- Enhance documentation explaining that returning the same instance to both methods is ok.
- Enable Data-driven domain-supporting records in all methods.
- Explaining more about the exceptions when the database does not support ACID.

## Challenges

- It would be amazing to have support to record even if JPA does not support it.
- Non-ACID databases can have insert and update methods but with different consistency strategies, such as clock-vector, read-repair, or any async handshake.
- I cannot assume anymore the NoSQL type as we did with the data structure, 

⚠️ it is a Pandora's. I need as much help as possible to support both database engines.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -47,7 +47,7 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      * @throws UnsupportedOperationException for Key-Value and Wide-Column databases
      *         that use an append model to write data.
      */
-    void insert(T entity);
+    <S extends T> S insert(S entity);
 
     /**
      * <p>Inserts multiple entities into the database. If an entity of this type with the same
@@ -60,7 +60,7 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      * @throws UnsupportedOperationException for Key-Value and Wide-Column databases
      *         that use an append model to write data.
      */
-    void insertAll(Iterable<T> entities);
+    <S extends T> Iterable<S> insertAll(Iterable<S> entities);
 
     /**
      * <p>Modifies an entity that already exists in the database.</p>
@@ -79,7 +79,7 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      * @return true if a matching entity was found in the database to update, otherwise false.
      * @throws NullPointerException if the entity is null.
      */
-    boolean update(T entity);
+    <S extends T> S update(S entity);
 
     /**
      * <p>Modifies entities that already exists in the database.</p>
@@ -98,5 +98,5 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      * @return the number of matching entities that were found in the database to update.
      * @throws NullPointerException if either the iterable is null or any element is null.
      */
-    int updateAll(Iterable<T> entities);
+    <S extends T> Iterable<S> updateAll(Iterable<S> entities);
 }
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -48,7 +48,7 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      *
      * @param entity the entity to insert. Must not be {@code null}.
      * @param <S> Type of the entity to insert.
-     * @return the inserted entity, which may or may not be a different instance depending on the Jakarta Data provider.
+     * @return the inserted entity, which may or may not be a different instance depending on whether the insert caused values to be generated or automatically incremented.
      * @throws EntityExistsException if the entity is already present in the database (in ACID-supported databases).
      * @throws NullPointerException if the entity is null.
      */
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -104,8 +104,7 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      *
      * <p>For an update to be made to an entity, a matching entity with the same unique identifier
      * must be present in the database. In databases that use an append model to write data or
-     * follow the BASE model, this method  behavior as the {@link #insertAll} method,
-     * especially if the database does not support ACID transactions.</p>
+     * follow the BASE model, this method behaves the same as the {@link #insertAll} method.</p>
      *
      * <p>If the entity is versioned (for example, with {@code jakarta.persistence.Version} or by
      * another convention from the entity model such as having an attribute named {@code version}),
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -95,7 +95,7 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      *
      * <p>Non-matching entities are ignored and do not cause an error to be raised.</p>
      *
-     * @param entity the entity to update. Must not be {@code null}.
+     * @return true if a matching entity was found in the database to update, otherwise false.
 @return true if a matching entity was found in the database to update, otherwise false.
      * @throws NullPointerException if the entity is null.
      */
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -38,27 +38,38 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
 
     /**
      * <p>Inserts an entity into the database. If an entity of this type with the same
-     * unique identifier already exists in the database, then this method raises
-     * {@link EntityExistsException}.</p>
+     * unique identifier already exists in the database and the database supports ACID transactions,
+     * then this method raises {@link EntityExistsException}. In databases that follow the BASE model
+     * or use an append model to write data, this exception may not be thrown.</p>
+     *
+     * <p>The entity instance returned as a result of this method may be the same instance as the one
+     * supplied as a parameter, especially in non-Java record classes. However, for Jakarta Data providers
+     * that support Java records, a different instance may be returned.</p>
      *
      * @param entity the entity to insert. Must not be {@code null}.
-     * @throws EntityExistsException if the entity is already present in the database.
+     * @param <S> Type of the entity to insert.
+     * @return the inserted entity, which may or may not be a different instance depending on the Jakarta Data provider.
+     * @throws EntityExistsException if the entity is already present in the database (in ACID-supported databases).
      * @throws NullPointerException if the entity is null.
-     * @throws UnsupportedOperationException for Key-Value and Wide-Column databases
-     *         that use an append model to write data.
      */
     <S extends T> S insert(S entity);
 
     /**
-     * <p>Inserts multiple entities into the database. If an entity of this type with the same
-     * unique identifier as any of the given entities already exists in the database,
-     * then this method raises {@link EntityExistsException}.</p>
+     * <p>Inserts multiple entities into the database. If any entity of this type with the same
+     * unique identifier as any of the given entities already exists in the database and the database
+     * supports ACID transactions, then this method raises {@link EntityExistsException}.
+     * In databases that follow the BASE model or use an append model to write data, this exception
+     * may not be thrown.</p>
+     *
+     * <p>The entities within the returned iterable may be the same instances as those supplied
+     * as parameters, especially in non-Java record classes. However, for Jakarta Data providers
+     * that support Java records, different instances may be returned.</p>
      *
      * @param entities entities to insert.
-     * @throws EntityExistsException if any of the entities are already present in the database.
-     * @throws NullPointerException if either the iterable is null or any element is null.
-     * @throws UnsupportedOperationException for Key-Value and Wide-Column databases
-     *         that use an append model to write data.
+     * @param <S> Type of the entities to insert.
+     * @return an iterable containing the inserted entities, which may or may not be different instances depending on the Jakarta Data provider.
+     * @throws EntityExistsException if any of the entities are already present in the database (in ACID-supported databases).
+     * @throws NullPointerException if the iterable is null or any element is null.
      */
     <S extends T> Iterable<S> insertAll(Iterable<S> entities);
 
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -96,10 +96,7 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      * <p>Non-matching entities are ignored and do not cause an error to be raised.</p>
      *
      * @param entity the entity to update. Must not be {@code null}.
-     * @param <S> Type of the entity to update.
-     * @return the updated entity. The entity instance returned as a result of this method may be the same
-     * instance as the one supplied as a parameter, especially in non-Java record classes. However,
-     * for Jakarta Data providers that support Java records, a different instance may be returned.
+@return true if a matching entity was found in the database to update, otherwise false.
      * @throws NullPointerException if the entity is null.
      */
     @Update
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -67,7 +67,7 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      *
      * @param entities entities to insert.
      * @param <S> Type of the entities to insert.
-     * @return an iterable containing the inserted entities, which may or may not be different instances depending on the Jakarta Data provider.
+     * @return an iterable containing the inserted entities, which may or may not be different instances depending on whether the insert caused values to be generated or automatically incremented.
      * @throws EntityExistsException if any of the entities are already present in the database (in ACID-supported databases).
      * @throws NullPointerException if the iterable is null or any element is null.
      */
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -120,10 +120,7 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      * <p>Non-matching entities are ignored and do not cause an error to be raised.</p>
      *
      * @param entities entities to update.
-     * @param <S> Type of the entities to update.
-     * @return an iterable containing the updated entities. In the case of Jakarta Data providers
-     *         that support Java records, this may return a different instance from the input,
-     *         preserving immutability.
+     * @return the number of matching entities that were found in the database to update.
      * @throws NullPointerException if either the iterable is null or any element is null.
      */
     @Update
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -121,6 +121,6 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      * @throws NullPointerException if either the iterable is null or any element is null.
      */
     @Update
-    <S extends T> Iterable<S> updateAll(Iterable<S> entities);
+    int updateAll(Iterable<T> entities);
 
 }
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -42,9 +42,11 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      * then this method raises {@link EntityExistsException}. In databases that follow the BASE model
      * or use an append model to write data, this exception is not thrown.</p>
      *
-     * <p>The entity instance returned as a result of this method may be the same instance as the one
-     * supplied as a parameter, especially in non-Java record classes. However, for Jakarta Data providers
-     * that support Java records, a different instance may be returned.</p>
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/data/repository/CrudRepository.java
```diff
@@ -38,35 +38,48 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
...
 
     /**
-     * <p>Inserts multiple entities into the database. If an entity of this type with the same
-     * unique identifier as any of the given entities already exists in the database,
-     * then this method raises {@link EntityExistsException}.</p>
+     * <p>Inserts multiple entities into the database. If any entity of this type with the same
+     * unique identifier as any of the given entities already exists in the database and the database
+     * supports ACID transactions, then this method raises {@link EntityExistsException}.
+     * In databases that follow the BASE model or use an append model to write data, this exception
+     * may not be thrown.</p>
```
njr-11, line 62:
```suggestion
     * is not thrown.</p>
```

api/src/main/java/jakarta/data/repository/CrudRepository.java
```diff
@@ -38,35 +38,48 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
...
+    <S extends T> Iterable<S> insertAll(Iterable<S> entities);
 
     /**
      * <p>Modifies an entity that already exists in the database.</p>
      *
      * <p>For an update to be made, a matching entity with the same unique identifier
-     * must be present in the database.</p>
+     * must be present in the database. In databases that use an append model to write data or
+     * follow the BASE model, this method will work similarly to the {@link #insert} method,
+     * especially if the database does not support ACID transactions.</p>
```
njr-11, line 82: A specification cannot have ambiguous language like "will work similarly to the {@link #insert} method, especially".  This doesn't have any enforceable meaning whatsoever because "similarly" is a matter of interpretation.  A specification needs to be precise about stating requirements. Otherwise each vendor will implement its own interpretation, and customers will have no guarantees against which to code and so be unable to use the API.  Also, language like "especially" should never be used in a spec requirement because it suggests that it is actually a non-requirement.

njr-11 (changes requested):

api/src/main/java/jakarta/data/repository/CrudRepository.java
```diff
@@ -38,35 +38,48 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
...
-     * @throws EntityExistsException if the entity is already present in the database.
+     * @param <S> Type of the entity to insert.
+     * @return the inserted entity, which may or may not be a different instance depending on the Jakarta Data provider.
+     * @throws EntityExistsException if the entity is already present in the database (in ACID-supported databases).
      * @throws NullPointerException if the entity is null.
-     * @throws UnsupportedOperationException for Key-Value and Wide-Column databases
-     *         that use an append model to write data.
      */
-    void insert(T entity);
+    <S extends T> S insert(S entity);
```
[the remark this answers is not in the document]
otaviojava: Please, allow me to give more context for both:

# Insert

We can create an entity, and this database can generate an ID using a strong consistency autoincrement or an ID based on a hash such as UUID.

In NoSQL, this is trivial. 
E.g.: 

```java
record Book(UUID id, String title, String author){} 

var book = Book.of("Effective Java", "Josh");
repository.insert(book);
```

In this case, it will map and convert to the database, and the database will return to the current stage:

```mermaid
    sequenceDiagram
    Entity->>Mapper: The entity from the user
    Mapper->>Database: the conversion
    Database->>Mapper: the database will generate the ID
    Mapper->>Entity: the return will either keep the same instance to non-record or a new instance to record
```

I am not reading but returning to the current stage from the database.

We have a similar case with the version or timestamp at the update method; this may vary the consistency strategy and the algorithm, for example, the clock vector.


I believe that SQL has a similar process.
I mean, they have GeneratedValue and Version annotations.

Do you think it makes sense?

api/src/main/java/jakarta/data/repository/CrudRepository.java
otaviojava: Thank you, I will update this point.

otaviojava: Could you check again?

njr-11: Switching from "especially" to "particularly" and from "will work" to "behaves" doesn't change anything about clearly stating requirements in the spec.  This still lacks explicit requirements and leaves the reader guessing what it means that works/behaves similarly to insert.   My prior comment still applies.

api/src/main/java/jakarta/data/repository/CrudRepository.java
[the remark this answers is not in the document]
njr-11: > I believe that SQL has a similar process.
> I mean, they have GeneratedValue and Version annotations.

`GeneratedValue` and `Version` annotations are from JPA/Jakarta Persistence, not from SQL, and they are more limited than what SQL can do.  We cannot assume that every Jakarta Data provider will want to limit itself to a JPA-based implementation (and even if they did, the JPA insert operation entitymanager.persist actually returns `void`, not an updated entity!).  SQL allows columns to be generated by the database beyond the Id and Version, and these would need to be retrieved.

In general, I'm fine with having some API in Jakarta Data that is intended for a particular subset of databases and not others.  However, you already have the `save`/`saveAll` operations to cover the pattern that you are trying to achieve for the databases that can optimize this way.  It is not right to force that same pattern upon `insert` and `update` where doing so will prevent this other other subset of databases from being usable or cause them to perform badly.  The advantage of having separate methods for `save` vs `insert` (and `update`) is that we can make the behavior different, and that way cover both subsets of databases with the combination of different behaviors that we offer.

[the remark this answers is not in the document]
otaviojava: > > I believe that SQL has a similar process.
> > I mean, they have GeneratedValue and Version annotations.
> 
> `GeneratedValue` and `Version` annotations are from JPA/Jakarta Persistence, not from SQL, and they are more limited than what SQL can do. We cannot assume that every Jakarta Data provider will want to limit itself to a JPA-based implementation (and even if they did, the JPA insert operation entitymanager.persist actually returns `void`, not an updated entity!). SQL allows columns to be generated by the database beyond the Id and Version, and these would need to be retrieved.
> 
> In general, I'm fine with having some API in Jakarta Data that is intended for a particular subset of databases and not others. However, you already have the `save`/`saveAll` operations to cover the pattern that you are trying to achieve for the databases that can optimize this way. It is not right to force that same pattern upon `insert` and `update` where doing so will prevent this other other subset of databases from being usable or cause them to perform badly. The advantage of having separate methods for `save` vs `insert` (and `update`) is that we can make the behavior different, and that way cover both subsets of databases with the combination of different behaviors that we offer.


When I mentioned the GeneratedValue and Version and mentioned NoSQL databases, for example:
- MongoDB: that provides [ObjectID](https://www.mongodb.com/docs/manual/reference/method/ObjectId/), so when I **insert** the database will provide this for me.
- ArangoDB: when you **update** it will provide a new [Document revision](https://docs.arangodb.com/3.11/concepts/data-structure/documents/#document-revisions)


SQL also provides this behavior with, for example, when I insert with [PostgreSQL with serial](https://www.geeksforgeeks.org/postgresql-serial/), also with [MySQL provider auto increment](https://dev.mysql.com/doc/refman/8.0/en/example-auto-increment.html).


What I don't understand is, why it is specific to NoSQL if SQL has the same approach, could you explain?
The goal is not to support Version and autoincrement. If yes, we need to put this on the insert and update methods.

api/src/main/java/jakarta/data/repository/CrudRepository.java
otaviojava: Do you have any suggestions?
Otherwise, it is time to discuss a SQLRepository.

otaviojava: I did a new update, the challenge here:

Appendedable database and BASE databases, both have update methods, but while Appendable works as an alias to both update and insert with BASE the update will be eventual consistency,

api/src/main/java/jakarta/data/repository/CrudRepository.java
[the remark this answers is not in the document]
njr-11: @otaviojava I am not arguing against the usefulness of the one pattern you keep bringing up.  The point is that we already covered that pattern with `save`/`saveAll`, which is inherited from the `BasicRepository`.  We should not be getting rid of the other useful patterns that `insert` and `update` offer and replacing them with a duplication of `save`.  Maybe it will help to list different patterns/goals that a user might have:

| Goal | Method(s) |
| ---- | --------- |
| Create a new entity that has one or more generated values, where the Id is one of the generated values, and continue using the entity afterward | `E save(E)` |
| Create a new entity that has one or more generated values, where the Id is one of the generated values, without using the entity afterward | `void insert(E)` |
| Create a new entity that has no generated values | `void insert(E)` |
| Create a new entity that has one or more generated values, where the Id is not one of the generated values, without using the entity afterward | `void insert(E)` |
| Create a new entity that has one or more generated values, where the Id is not one of the generated values, and continue using the entity afterward | `void insert(E)`, then `E findById(id)` |
| Create a new entity that has a specified Id or otherwise update the existing entity | `E save(E)` |
| If an entity (with generated values) with the specified id exists, modify the entity and continue using the entity afterward | boolean `update(E)`, then `E findById(id)` |
| If an entity (without generated values) with the specified id exists, modify the entity and continue using the entity afterward | boolean `update(E)` |
| If an entity with the specified id exists, modify the entity and do not use the entity afterward | boolean `update(E)` |

The different signatures for `E save(E)`, `void insert(E)`, and `boolean update(E)` allow the user to achieve every one of the scenarios above and lead the user to naturally select which is most optimal to their pattern.  The proposal to make insert return the entity will cause the 4th, 8th, and 9th above to become inefficient wherever the database or programming model doesn't have a special optimization to return **all** of the generated values.  It will require additional retrievals from the database, and wherever this happens, users will despise it.  If you don't believe it will be very common, I've already pointed out JPA's only pure insert operation, [EntityManager.persist](https://jakarta.ee/specifications/persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/entitymanager#persist(java.lang.Object)), returns a `void` type, not an updated entity.  Furthermore, JPA doesn't not have this optimization for pure [updates](https://jakarta.ee/specifications/persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/query#executeUpdate()) either.  A JDBC-based implementation might or might not have an optimization available, depending on the capabilities of the database and JDBC driver.  It is wrong to make core API force inefficient behavior here which users will be very displeased with.  I don't agree with this change.  If there needs to be any further discussion of it, let's schedule a call to go over it because we keep restating the same things over and over in the comments.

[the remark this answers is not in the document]
otaviojava: The EntityManager.persist will return it because it changes the class that must be mutable as the spec says:

> An update to the state of an entity includes both the assignment of a new value to a persistent property or field of the entity as well as the modification of a mutable value of a persistent property or field

https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0#a1955

In that case, it will be easier to implement JPA IMHO:

```java
@Overwrite
public Movie insert(Movie movie){
//no id, because it will be auto-generated by the database
entityManager.insert(movie); that changes will happen in this class because it is mutable
return movie;
}
```
Thus, the generated ID and version will happen in this instance. Did I get it right?

Imagine a scenario in which I want to **insert** it, but with a mutable class using record.

```java
@Overwrite
public Movie insert(Movie movie){
//no id, because it will be auto-generated by the database
otherDatabaseprovider.insert(movie);// I must create a new instance because this is a record.
return movieUpdated;
}
```

If I don't have this return, I will lose track, mainly if this is a hashed ID.

If we work with JPA, I agree that it can directly change the entity itself once it does not support records, but if any provider does, this insert will become impossible to use. 

If you take a look at the Micronaut, you can see that it will return the entity:

https://micronaut-projects.github.io/micronaut-data/2.3.0/api/index.html?io/micronaut/data/repository/CrudRepository.html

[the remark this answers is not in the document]
njr-11: > The EntityManager.persist will return void, because it changes the class that **must** be mutable as the spec says:
> 
> > An update to the state of an entity includes both the assignment of a new value to a persistent property or field of the entity as well as the modification of a mutable value of a persistent property or field
> 
> https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0#a1955

I understood this to be saying that you can update entities by assigning a new value to a field/property,

```
movie.title = "Updated Title";
```

or modify the value of a mutable field or property,

```
movie.actors.add("Someone Famous");
```

But if your interpretation is correct, (it appears that it might be based on trying it out on one JPA provider which does modify the supplied entity instance) then we would be able to allow `E insert(E)`, but I still don't see a way to allow the entity return type on `update`.

api/src/main/java/jakarta/data/repository/CrudRepository.java
```diff
@@ -95,8 +113,11 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
...
      * @param entities entities to update.
-     * @return the number of matching entities that were found in the database to update.
+     * @param <S> Type of the entities to update.
+     * @return an iterable containing the updated entities. In the case of Jakarta Data providers
+     *         that support Java records, this may return a different instance from the input,
+     *         preserving immutability.
      * @throws NullPointerException if either the iterable is null or any element is null.
      */
-    int updateAll(Iterable<T> entities);
+    <S extends T> Iterable<S> updateAll(Iterable<S> entities);
```
njr-11, line 122: It is not acceptable to force retrieval from the database on update.  Please restore this to how it was, as follows:
```suggestion
    int updateAll(Iterable<T> entities);
```

api/src/main/java/jakarta/data/repository/CrudRepository.java
```diff
@@ -95,8 +113,11 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      * <p>Non-matching entities are ignored and do not cause an error to be raised.</p>
      *
      * @param entities entities to update.
-     * @return the number of matching entities that were found in the database to update.
+     * @param <S> Type of the entities to update.
+     * @return an iterable containing the updated entities. In the case of Jakarta Data providers
+     *         that support Java records, this may return a different instance from the input,
+     *         preserving immutability.
```
njr-11, line 119: It is not acceptable to force retrieval from the database on update.  Please restore this to how it was, as follows:
```suggestion
     * @return the number of matching entities that were found in the database to update.
```

njr-11 (changes requested):
@otaviojava 
So that we can make progress and still have a chance to get our current set of proposals into an M1 with the other Jakarta specs, I'm agreeing to compromise and have the entity return type for `insert`/`insertAll`, but not `update`/`updateAll`.  Please accept this and let's more forward from here.

otaviojava:
> @otaviojava So that we can make progress and still have a chance to get our current set of proposals into an M1 with the other Jakarta specs, I'm agreeing to compromise and have the entity return type for `insert`/`insertAll`, but not `update`/`updateAll`. Please accept this and let's more forward from here.

Sorry for the delay.
Nathan, we need this return to update to a similar case, but instead of generating the id we use for version for SQL and timestamp, clock-version, and document review on NoSQL.

> The version attribute is updated by the persistence provider runtime when the object is written to the database. All non-relationship fields and properties and all relationships owned by the entity are included in version checks

https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#a2059

otaviojava:
> > @otaviojava So that we can make progress and still have a chance to get our current set of proposals into an M1 with the other Jakarta specs, I'm agreeing to compromise and have the entity return type for `insert`/`insertAll`, but not `update`/`updateAll`. Please accept this and let's more forward from here.
> 
> Sorry for the delay. Nathan, we need this return to update to a similar case, but instead of generating the id we use for version for SQL and timestamp, clock-version, and document review on NoSQL.
> 
> > The version attribute is updated by the persistence provider runtime when the object is written to the database. All non-relationship fields and properties and all relationships owned by the entity are included in version checks
> 
> https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#a2059

@njr-11 WDYT?

njr-11:
> The version attribute is updated by the persistence provider runtime when the object is written to the database.

The problem is that to get the entity into the managed state where JPA updates the version requires a retrieval from the database.  That's what we're trying to avoid.  There are other ways to do update with JPA (and JDBC) which don't require fetching the entity at all and perform a single update operation to the database, qualified by the id and version, which is much more efficient for detached entities as we are dealing with.  In that case, JPA doesn't even need to see the entity object at all.

If you change the return type to entity on `CrudRepository`, users will get the inefficient behavior with extra retrieval by default.  Instead, we should let the API guide the users to do the right thing.  In the case where someone actually wants the updated entity returned, they will see that they can use the `save` operation instead which already does this, or use the `@Update` method to define their own update method returning the entity, or just issue a find after performing the update to get the entity.

njr-11:
I think you keep confusing `update` with `save`.

> ```java
> public T update(T entity){
> entitymanager.merge(entity);
> ```

It would be wrong to implement it as above because `merge` will create the entity in the database if it didn't exist, violating the contract for `update` and instead making it behave like a `save`.  To guarantee it didn't exist in the database in order to make `merge` safe, you would first need to query the database for it, and do so in a way that keeps a lock on the entry so that it doesn't disappear before you issue the `merge`.  This can be done, but it's inefficient, and it isn't a pure update given that it includes extra retrieval from the database.

You are right that there is a trade-off here.  The best choice is what the API already does - not forcing the entity to be retrieved - because the user has at least 3 other good options to choose from in that case, which I listed in my previous comment:

- Use `save` instead. It returns the entity.
- Define an `@Update` method explicitly asking for the entity to be retrieved where you actually need it.
- Invoke a find operation to retrieve the entity where you actually need it.

All of those are natural approaches that the spec provides.

I should also point out that I see update typically being performed as the final database operation when you are typically done with the entity, so in most cases the user doesn't even want the entity after having made the update.  They mainly want to confirm that their update worked and be done.

The CrudRepository.update return type is already optimized toward the predominant usage pattern and is fully aligned with CRUD.  We should not change it.

otaviojava:
Nope, we agree on what is update, insert, and update.

Let me create a scenario, where we have the Nathan entity and JPA:

```java
@Entity
public class Person {    

    @Id
    @GeneratedValue
    private Long id;
    @Column
    private String name;

    @Version
    private Long version;
}
```

In the first scenario we agree that it will generate an ID in the insert:

```java
Person person = Person.of("Nathan");
repository.insert(person);\\it will generate an ID
```

My point is in the update with the version.

```java
person.name("Nathan Rauh");
repository.update(person);\\it should be thrown an exception because we should not support version here?
```

In this case is easy, because the class allows mutability, but how about record?

My pain is the consistency with the record, once, I cannot change the object, just return a new instance.

We need to solve this in some way or document that this update method does not support either record or database modification such as version.

[Micronaut solves it return the instance](https://micronaut-projects.github.io/micronaut-data/2.3.0/api/index.html?io/micronaut/data/repository/CrudRepository.html)

njr-11:
The example you gave is not valid regardless of whether a record is involved.  The specification states that after the method ends, "not to use the entity value supplied as a parameter, as this method makes no guarantees about the state of the entity value after insertion."  Given this, the user should choose a variant of `insert` -- either a repository method they defined themselves with the desired return type, or `CrudRepository.insert`, which we have already changed in this pull to have the  entity return type:

```
Person person = Person.of(name1);
person = repository.insert(person); // it will generate an ID and version

person.name(name2);
repository.update(person);
```

This will work just fine, once you commit the remaining review comments to avoid altering the `CrudRepository.update` method and merge this pull.  If do you want this pattern with `CrudRepository.insert` in the Jakarta Data M1, I would recommend merging what we have agreed on thus far, and then use a separate pull/issue if you want to continue to debate about the return value of update.

otaviojava:
> The example you gave is not valid regardless of whether a record is involved. The specification states that after the method ends, "not to use the entity value supplied as a parameter, as this method makes no guarantees about the state of the entity value after insertion." Given this, the user should choose a variant of `insert` -- either a repository method they defined themselves with the desired return type, or `CrudRepository.insert`, which we have already changed in this pull to have the entity return type:
> 
> ```
> Person person = Person.of(name1);
> person = repository.insert(person); // it will generate an ID and version
> 
> person.name(name2);
> repository.update(person);
> ```
> 
> This will work just fine, once you commit the remaining review comments to avoid altering the `CrudRepository.update` method and merge this pull. If do you want this pattern with `CrudRepository.insert` in the Jakarta Data M1, I would recommend merging what we have agreed on thus far, and then use a separate pull/issue if you want to continue to debate about the return value of update.

You are right, let's merge and open a PR only to discuss the update point.

api/src/main/java/jakarta/data/repository/CrudRepository.java
```diff
@@ -77,18 +95,19 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      *
      * <p>Non-matching entities are ignored and do not cause an error to be raised.</p>
      *
-     * @param entity the entity to update.
      * @return true if a matching entity was found in the database to update, otherwise false.
+@return true if a matching entity was found in the database to update, otherwise false.
```
njr-11, line 99: I'm sorry, I had an error in my prior suggestion, which accidentally caused the `@return` to be duplicated. This will fix it:
```suggestion
     * @param entity the entity to update. Must not be {@code null}.
     * @return true if a matching entity was found in the database to update, otherwise false.
```

njr-11 (approved):

Issue #133, opened 2023-03-31
TCK scenarios to cover


njr-11:
I'm starting this list to record potential ideas for TCK scenarios. Please feel free to edit, or if you aren't a committer just add a comment so that a committer can add/update for you,

- [x] #179 Request the first Page of 10 results, expecting to find all 10. From the Page, verify the totalElements and totalPages expected.

- [x] #179 Request the third Page of 10 results, expecting to find all 10. Request the next Page via nextPageable, expecting page number 4 and another 10 results.

- [x] #179 Request the last Page of up to 10 results, expecting to find the final 3.

- [x] #179 Request a Page higher than the last Page, expecting an empty Page with 0 results.

- [x] #179 Request a Page of results where none match the query, expecting an empty Page with 0 results.

- [x] #147 Request the first Slice of 5 results, expecting to find all 5.

- [x] #147 Request the third Slice of 5 results, expecting to find all 5. Request the next Slice via nextPageable, expecting page number 4 and another 5 results.

- [x] #147 Request the last Slice of up to 5 results, expecting to find the final 2.

- [x] #147 Request a Slice higher than the last Slice, expecting an empty Slice with 0 results.

- [x] #147 Request a Slice of results where none match the query, expecting an empty Slice with 0 results.

- [x] #214 Request the first KeysetAwarePage of 8 results, expecting to find all 8, then request the next KeysetAwarePage after the keyset of the last entry of the page, expecting to find the next 8. Or validate that UnsupportedOperationException is raised to indicate that the underlying database is incapable of keyset pagination.

- [x] #214 Request a KeysetAwarePage of 7 results after the keyset of the 20th result, expecting to find the next 7 results. Then request the KeysetAwarePage before the keyset of the first entry of the page, expecting to find the previous 7 results. Then request the KeysetAwarePage after the last entry of the page, expecting to find the next 7. Or validate that UnsupportedOperationException is raised to indicate that the underlying database is incapable of keyset pagination.

- [x] #214 Request a KeysetAwarePage of results where none match the query, expecting an empty KeysetAwarePage with 0 results. Or validate that UnsupportedOperationException is raised to indicate that the underlying database is incapable of keyset pagination.

- [x] #150 Request the first KeysetAwareSlice of 6 results, expecting to find all 6, then request the next KeysetAwareSlice after the keyset of the last entry of the slice, expecting to find the next 6. Or validate that MappingException is raised to indicate that the underlying database is incapable of keyset pagination.

[...]

PR #289, opened 2023-10-11
Create CDI extension tests

KyleAure:
Related to #133 

```
If CDI is available (Jakarta EE profile tests), include a CDI extension to handle a custom entity annotation. Inject at least one repository of this entity type and verify that a method can be invoked on the repository verifying that this extension is not interfered with by other Jakarta Data providers. Include this under Jakarta EE profile tests and skip if CDI is not available.
```

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/core/cdi/ExtensionTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/core/cdi/ExtensionTests.java
@@ -13,7 +13,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  */
-package ee.jakarta.tck.data.full.cdi;
+package ee.jakarta.tck.data.core.cdi;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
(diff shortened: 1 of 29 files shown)


tck/src/main/java/ee/jakarta/tck/data/core/cdi/ExtensionTests.java
```diff
@@ -0,0 +1,61 @@
...
+import ee.jakarta.tck.data.framework.junit.anno.AnyEntity;
+import ee.jakarta.tck.data.framework.junit.anno.Assertion;
+import ee.jakarta.tck.data.framework.junit.anno.CDI;
+import ee.jakarta.tck.data.framework.junit.anno.Core;
+import jakarta.enterprise.inject.spi.Extension;
+import jakarta.inject.Inject;
+
+@Core
+@AnyEntity
+@CDI
```
njr-11, line 37: This test is only valid in web profile (or platform), where CDI Full is available, providing `jakarta.enterprise.inject.spi.Extension` which is used for the extension.
In core profile, only CDI Lite is required, which is not required to include `jakarta.enterprise.inject.spi.Extension`.
Note that `jakarta.enterprise.inject.build.compatible.spi.BuildCompatibleExtension` must be available in both CDI Lite and CDI Full, meaning it can be relied upon for all of Core Profile, Web Profile, and Jakarta EE Platform.

njr-11 (approved):

PR #291, opened 2023-10-11
Enhance the Update annotation

otaviojava:
# Changes

- Include more documentation to Update annotation.

It is a start to discuss.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Update.java
+++ b/api/src/main/java/jakarta/data/repository/Update.java
@@ -61,7 +61,10 @@ import java.lang.annotation.Target;
  * <p>If the entity does not exist in the database or it is versioned and its version differs from the version in the database,
  * no update is made and no error is raised.
  * </p>
- *
+ *<p>
+ * In databases that use an append model to write data or follow the BASE model, this method
+ * behaves the same as the {@code @Insert} method.
+ *  </p>
  * <p>For example, consider an interface representing a garage:</p>
  * <pre>
  * &#64;Repository
--- a/api/src/main/java/jakarta/data/repository/Update.java
+++ b/api/src/main/java/jakarta/data/repository/Update.java
@@ -35,9 +35,23 @@ import java.lang.annotation.Target;
  *     <li>An array of entities to be updated.</li>
  * </ul>
  * <p>The return type of the annotated method must be {@code void}, {@code boolean}, a numeric primitive type
- * (such as {@code int}), or a corresponding primitive wrapper type (such as {@link Integer}).
+ * (such as {@code int}), a corresponding primitive wrapper type (such as {@link Integer}), or the same type as the parameter.
+ * </p>
+ * <p>
+ * The return type of an annotated method that accepts an {@code Iterable} or array of entities as the parameter must
+ * have a return type that is {@code void}, {@code Void}, or an {@code Iterable} or array of the entity.
  * A boolean return type indicates whether a matching entity was found in the database to update.
  * A numeric return type indicates how many matching entities were found in the database to update.
+ * The return type of an annotated method that accepts an {@code Iterable} or array of entities as the parameter must
+ * have a return type that is {@code void}, {@code Void}, or an {@code Iterable} or array of the entity.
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/data/repository/Update.java
```diff
@@ -35,9 +35,23 @@
...
  * A boolean return type indicates whether a matching entity was found in the database to update.
  * A numeric return type indicates how many matching entities were found in the database to update.
+ * The return type of an annotated method that accepts an {@code Iterable} or array of entities as the parameter must
+ * have a return type that is {@code void}, {@code Void}, or an {@code Iterable} or array of the entity.
+ * For example, if the method is annotated with {@code @Update} and takes a parameter of type {@code Car car},
+ * the return type can be {@code Car}.
+ * Similarly, if the parameter is an {@code Iterable<Car>} or an array of {@code Car}, the return type can be
+ * {@code Iterable<Car>}.
+ * Entities that are returned by the annotated method must include all values that were
+ * written to the database, including all automatically generated values, update versions and incremented values
```
njr-11, line 52:
```suggestion
 * written to the database, including all automatically generated values, updated versions and incremented values
```

api/src/main/java/jakarta/data/repository/Update.java
```diff
@@ -35,9 +35,23 @@
...
- * (such as {@code int}), or a corresponding primitive wrapper type (such as {@link Integer}).
+ * (such as {@code int}), a corresponding primitive wrapper type (such as {@link Integer}), or the same type as the parameter.
+ * </p>
+ * <p>
+ * The return type of an annotated method that accepts an {@code Iterable} or array of entities as the parameter must
+ * have a return type that is {@code void}, {@code Void}, or an {@code Iterable} or array of the entity.
  * A boolean return type indicates whether a matching entity was found in the database to update.
  * A numeric return type indicates how many matching entities were found in the database to update.
+ * The return type of an annotated method that accepts an {@code Iterable} or array of entities as the parameter must
+ * have a return type that is {@code void}, {@code Void}, or an {@code Iterable} or array of the entity.
```
njr-11, line 46: The same text is duplicated in two places here, probably a copy/paste error from the `@Insert` method given that it leaves out the numeric and boolean types. The correct return type requirements were just stated prior to this so we can just remove this.
We do need to say something about what happens with the return value when some of the entities are not found. It will make sense to include that here,
```suggestion
 * A boolean return type indicates whether a matching entity was found in the database to update.
 * A numeric return type indicates how many matching entities were found in the database to update.
 * An entity return type indicates the updated entity if found in the database. If the entity is not found
 * in the database or has a non-matching version, then {@code null} is returned.
 * An {@code Iterable} or array return type includes all matching entities that are found in the database,
 * skipping over entities that are not present in the database or have a non-matching version.
```

api/src/main/java/jakarta/data/repository/Update.java
```diff
@@ -35,9 +35,23 @@
...
+ * For example, if the method is annotated with {@code @Update} and takes a parameter of type {@code Car car},
+ * the return type can be {@code Car}.
+ * Similarly, if the parameter is an {@code Iterable<Car>} or an array of {@code Car}, the return type can be
+ * {@code Iterable<Car>}.
+ * Entities that are returned by the annotated method must include all values that were
+ * written to the database, including all automatically generated values, update versions and incremented values
+ * that changed due to the update. The position of entities within an {@code Iterable} or array return value
+ * must correspond to the position of entities in the parameter based on the unique identifier of the entity.
```
njr-11, line 54:
```suggestion
 * that changed due to the update. The order of entities within an {@code Iterable} or array return value
 * must correspond to the position of entities in the parameter based on the unique identifier of the entity,
 * leaving out those that did not match the unique identifier and version that is in the database.
```

njr-11 (changes requested):
This should work.  I added some suggestions to describe what happens with the return value when entities to update are not found in the database or have a version that doesn't match.

njr-11 (approved):
This looks good now. Approving.

PR #293, opened 2023-10-12
Align JavaDoc with spec updates from 273

njr-11:
Update the JavaDoc to match up with the specification updates that are being made under 273 and include the life cycle methods

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -156,7 +156,7 @@ import java.util.List;
  *     List&lt;Order&gt; forZipCode(int zipCode);
  *
  *     &#64;Save
- *     Order save(Order order);
+ *     Order checkout(Order order);
  * }
  * </pre>
  *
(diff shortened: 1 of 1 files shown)


api/src/main/java/module-info.java
```diff
@@ -143,7 +155,8 @@
  *     &#64;Query("SELECT o FROM Order o WHERE o.address.zipCode=?1")
  *     List&lt;Order&gt; forZipCode(int zipCode);
  *
- *     void save(Order order);
+ *     &#64;Save
+ *     Order save(Order order);
```
otaviojava, line 159:
```suggestion
 *     Order checkout(Order order);
```

How about this? I used Amazon as sample, we checkout an order.

njr-11: Good suggestion. I added it.

otaviojava (approved):

Issue #296, opened 2023-10-16
Question about the `Attribute` class in the `StaticMetamodel`


otaviojava:
**Goal:**
I am seeking clarification on the usage and necessity of the `Attribute` class in the `StaticMetamodel` and whether it is essential for the Jakarta Data specification.

**Context:**
I have been working on creating a Java Annotations Processor for the `StaticMetamodel`, specifically focusing on the `Attribute` class. However, after carefully examining the class and its usage within the specification, I am questioning its relevance and potential drawbacks.

**Concerns:**

1. **Safety:** The `Attribute` class appears to have a potential safety issue. The `init` method allows direct modification of attributes. For instance, one can do something like:
   
   ```java
   Product_.price.init(null);
   ```
   
   This seems to bypass encapsulation and allows users to modify attributes inadvertently or intentionally, which raises concerns about data integrity.

2. **Delay Layer:** The `Attribute` class seems to act as a delay layer, wrapping the `AttributeInfo` without introducing new information. This abstraction layer could increase memory usage without significant benefit, affecting efficiency.

3. **Extensibility:** The current approach using the `Attribute` class may not allow easy extensibility. If there is a need to introduce new capabilities, such as NoSQL-specific functionalities (e.g., greater than or equal conditions), the current approach might not be conducive to such extensions.

**Questions:**

Considering these concerns, I would like to ask if there is a valid and essential reason for the existence of the `Attribute` class within the Jakarta Data specification. Can we consider removing this class and allowing Java annotation processors to handle these operations directly?

**Suggested Alternative:**

An alternative approach could be to provide interfaces or abstract classes for specific attribute types, allowing for extensibility. For example:

```java
public interface NumberAttributeInfo extends AttributeInfo {
    Criteria eq(Number number);
    Criteria gt(Number number);
    Criteria gte(Number number);
}

public class Person_ {
    public static final AttributeInfo name = new Person_IDField(); // or any generated terminology
    public static final NumberAttributeInfo age = new Person_AgeField();
}
```

This approach might offer more flexibility without needing the `Attribute` class.

**Additional Notes:**

I'm not sure if the limitations I mentioned are due to the use of reflection, but the Jakarta Data specification should not unduly limit the capabilities of Java Annotation Processors.

njr-11:
> 1. **Safety:** The `Attribute` class appears to have a potential safety issue. The `init` method allows direct modification of attributes

[...]

PR #299, opened 2023-10-18
Update the StaticModel

otaviojava:
# Changes

- move to the package metamodel
- give the option to use it on both run time and build time
- Remove the condition to look for the `StaticMetamodel` as mandatory. Because at the Java Annotation Processor, we will generate those classes from the `Entity`annotation, e.g.: [Hibernate](https://github.com/hibernate/hibernate-metamodelgen/blob/master/src/main/java/org/hibernate/jpamodelgen/JPAMetaModelEntityProcessor.java#L56-L58)
- Allow dynamic `StaticMetamodel ' search using the Metamodel interface.

I read a lot of the [Jakarta Persistence Spec](https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#staticmetamodel-annotation). What I don't understand is the way to allow a mutable entity instead of create once it reads from the annotations at build time. 

Is there anything that must happen in real-time?

Update: I find the reason:
> When the entity manager factory for a persistence unit is created, it is the responsibility of the persistence provider to initialize the state of the metamodel classes of the persistence unit. Any generated metamodel classes must be accessible on the classpath.

https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#bootstrapping


api/src/main/java/jakarta/data/metamodel/Metamodel.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.metamodel;
+
+
+/**
+ * Provides access to the metamodel of the entity.
+ */
+public interface Metamodel {
```
njr-11, line 24: I don't see anything specifically wrong with this interface, but the specification and JavaDoc do not appear to provide any indication of how the application obtains implementations of it in order to use it.  It also doesn't appear to be very useful, allowing access only to the entity name, which is not something that the application currently ever needs to provide to the Jakarta Data API.

api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -15,16 +15,14 @@
...
 import jakarta.data.Sort;
 
 /**
- * Implemented by the Jakarta Data provider to
- * {@link Attribute#init(AttributeInfo) initialize} an attribute field
- * in the {@link StaticMetamodel}.
+ * Represents an attribute of a Java type.
  */
-public interface AttributeInfo {
+public interface Attribute {
```
njr-11, line 25: Elsewhere in the pull you refer to 
`{@link Attribute#init(Attribute) initialize}`
but you have not included an `init` method on `Attribute`.

api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
```diff
@@ -105,9 +105,7 @@
 @Target(ElementType.TYPE)
 public @interface StaticMetamodel {
     /**
-     * An entity class.
-     *
-     * @return the entity class.
+     * Class being modelled by the annotated class.
```
njr-11, line 108: This is a nice improvement to the JavaDoc, but you lost the `@return`

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/AsciiCharacter_.java
```diff
@@ -15,14 +15,14 @@
...
-    public static final Attribute id = Attribute.get();
-    public static final Attribute hexadecimal = Attribute.get();
-    public static final Attribute isControl = Attribute.get();
-    public static final Attribute numericValue = Attribute.get();
-    public static final Attribute thisCharacter = Attribute.get();
+    Attribute id = Attribute.get();
+    Attribute hexadecimal = Attribute.get();
+    Attribute isControl = Attribute.get();
+    Attribute numericValue = Attribute.get();
+    Attribute thisCharacter = Attribute.get();
```
njr-11, line 27: This is fine to remove the `public static final` which is implicit for interface fields.  However, with your changes elsewhere in this pull, the `Attribute.get()` part of the code won't compile anymore because you removed that method.  This makes it difficult for me to understand what your intent is with this pull.  If it was just a mistake that `get` and `init` were lost, they will need to be put back and you'll need to figure out how `init` works without the `AttributeInfo` which was removed.  If it was intentional that `get` and `init` are removed, that leaves me wondering what the user actually puts here.  They cannot leave it blank because the field is implicitly final, being on an interface.  If they put `null` there, it will never have a value because the field is final.  Hopefully you do not want the user to directly reference generated code here.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -644,41 +644,51 @@ In instances where a Jakarta Data provider for NoSQL databases encounters a recu
...
+Jakarta Data's Static Metamodel empowers developers with multiple options to access entity attributes safely, whether through automatic generation, manual creation, or dynamic exploration via the 'jakarta.data.metamodel.Metamodel' interface.
+
+===== Application Requirements for a Metamodel Class
 
-- The application defines a class (the metamodel class) and annotates it with the `@StaticMetamodel` annotation.
-- The application specifies the `value` of the `@StaticMetamodel` annotation to be an entity class that the application uses in a repository as the result type of a find method or the parameter type of an insert, update, save, or delete method.
+This specification introduces the concept of a canonical metamodel, which entails the structure and conventions for creating metamodel classes.
 
-For each field of the metamodel class,
+- For every entity class "X" located within package "p," a corresponding metamodel class "X_" within package "p" is automatically generated.
```
njr-11, line 665: There are multiple problems here.
Forcing every entity class in the package to have a metamodel is wasteful and inefficient, contradicting your requirement under #296 not to "increase memory usage without significant benefit, affecting efficiency."
The existing design is far superior, because it limits to entities that are actually referenced by repositories, and furthermore limits to only what the application asks for via `@StaticMetamodel`.

Secondly, this says that the metamodel class is automatically generated -- yet application code references it. This is unacceptable because application code should not need to compile against generated code, only spec API.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -644,41 +644,51 @@ In instances where a Jakarta Data provider for NoSQL databases encounters a recu
...
+- The name of the metamodel class is derived from the entity class by appending an underscore " _ " to the entity class's name.
+- The metamodel class "X_" is required to be annotated with the `jakarta.persistence.StaticMetamodel` annotation.
```
njr-11, line 667: Is this copied from the Jakarta Persistence specification?  It refers to their annotation, `jakarta.persistence.StaticMetamodel`, rather than ours.

njr-11 (changes requested):
There might be other comments on this, but I'll stop here for now, given that it isn't clear to me from the pull how this is expected to work.

otaviojava:
@njr-11 I will close mine, once you created this one here: https://github.com/jakartaee/data/pull/302

PR #302, opened 2023-10-23
TCK test for metamodel class generated at build time

njr-11:
This pull adds a TCK test to include a generated static metamodel class.  It also adds spec language to require the `@Generated` annotation on such classes so that Jakarta Data providers can optimize not to attempt initialization of generated metamodel classes that already have implementation.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -67,6 +67,12 @@
             <version>${assertj.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>jakarta.annotation</groupId>
+            <artifactId>jakarta.annotation-api</artifactId>
+            <version>${jakarta.annotation.version}</version>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
 
 </project>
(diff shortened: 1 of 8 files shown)


otaviojava:
One item left is compatibility; for example, not necessarily a Jakarta Data provider will be compatible with another one.

E.g.: Jakarta Data might focus only on build time, or they are different; for example, one is NoSQL, and another is SQL.

njr-11:
> or they are different; for example, one is NoSQL, and another is SQL.

I think we already have that possibility covered in the spec where we say that the metamodel is supplied by the "Jakarta Data provider that provides a repository for the entity class".  The JavaDoc also states this requirement.  Given that the SQL-based Jakarta Data provider and NoSQL-based Jakarta Data provider are used in different repositories with different types of entity classes, there is no overlap in providing the metamodels.

> Jakarta Data might focus only on build time

Build time steps really ought to be an optimization here, not a requirement, so that a user can compile once and run against any compatible Jakarta Data implementation that supports the same database.  If that isn't so, then we would need to remove the TCK tests that cover run time and only have coverage for the mock pre-generated path.

otaviojava (approved):

otaviojava:
@njr-11 can we use the same package name that we have in Jakarta Persistence?

I mean, it makes more sense since this model comes from metadata, from annotations.

PR #307, opened 2023-10-24
Update package name to metamodel to follow the Jakarta Persistence

otaviojava:
# Change

- Rename Package to follow the same name as Jakarta Persistence


KyleAure (approved):
This change looks good to me.

njr-11 (approved):
The changes all look fine to me as well, although it seems unnecessary to rename it. If others like the longer name, this is fine.

otaviojava:
> The changes all look fine to me as well, although it seems unnecessary to rename it. If others like the longer name, this is fine.

Yeap, it is longer, but it more cleaner and makes explicit what this package does.

PR #308, opened 2023-10-24
Modify method signature to handle with changes in the database

otaviojava:
# Change

- modify the update methods to return an instance instead of int.

⚠️ The challenge here is that Persistence can impact performance with Jakarta. However, the API has support for `Version` that implies changes during the update method.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -96,11 +96,12 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      * <p>Non-matching entities are ignored and do not cause an error to be raised.</p>
      *
      * @param entity the entity to update. Must not be {@code null}.
+     * @param <S> Type of the entities to update.
      * @return true if a matching entity was found in the database to update, otherwise false.
      * @throws NullPointerException if the entity is null.
      */
     @Update
-    boolean update(T entity);
+    <S extends T> S update(S entity);
 
     /**
      * <p>Modifies entities that already exist in the database.</p>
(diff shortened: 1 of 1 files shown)


njr-11 (changes requested):
I don't agree with switching the return type of CrudRepository.update/updateAll to return entities.  I've experimented with implementing this, and returning the entity can require retrieval from the database, which is not what the Update in CRUD means, so I am not in favor of it.

otaviojava:
> I don't agree with switching the return type of CrudRepository.update/updateAll to return entities. I've experimented with implementing this, and returning the entity can require retrieval from the database, which is not what the Update in CRUD means, so I am not in favor of it.

I know that you don't agree, but indeed, we need to solve this issue.

There is a versioning in both: Jakarta Persistence and Jakarta NoSQL.

How to procedure? throw an exception where there is database interaction? 
Should we forbid record on this method?

njr-11:
> How to procedure? throw an exception where there is database interaction in the update?
> Should we forbid Java records on this method?

Those seem like overreactions to what isn't even a problem in the first place.  The CrudRepository.update and updateAll methods are correct as is and already describe the ideal behavior, matching a CRUD update, and useful for the purpose of knowing whether your update was successful.

When a user has a scenario where it wants to know the updated version number it can choose any of:

- Use the `CrudRepository.save` method that exists to serve that purpose. For example, to update if version 5: `myEntity = repo.save(new MyEntity(id, value1, value2, 5))`
- Use `@Update` instead of `CrudRepository` to define your own `update` method returning an entity.
- Use `CrudRepository.update` and follow it up with a query for the version.
- Just know that the version number was incremented after a successful `CrudRepository.update` because the result of the method tells you that the update operation was successful.

There are a lot of choices that a user has when the above pattern is actually desired.  As I've pointed out elsewhere, typical usage involves the update being the last thing you do and you don't need an updated entity afterward, so don't penalize the typical usage case by forcing it to incur bad performance from extra retrievals for the sake of a less common usage pattern that can be solved other ways.

PR #310, opened 2023-10-25
Rename Metadata classes

otaviojava:
# Changes

- update the class's nomenclature to get closer to the Jakarta Persistence.
- In this case, `Attribute` instead of `AttributeInfo`
- `AttributeMetadata` instead of `Attribute`

References:

https://github.com/jakartaee/persistence/tree/master/api/src/main/java/jakarta/persistence/metamodel


njr-11 (changes requested):
I don't understand how these renames add value or move us closer into alignment with Jakarta Persistence.  It would seem to do the opposite given that Jakarta Persistence's static metamodel involves the use of *Attribute fields, not AttributeMetadata, so this pull takes us further away from Jakarta Persistence's static metamodel and adds churn to spec API, introducing some errors along the way where documentation is only partially updated.

otaviojava:
> I don't understand how these renames add value or move us closer into alignment with Jakarta Persistence. It would seem to do the opposite given that Jakarta Persistence's static metamodel involves the use of *Attribute fields, not AttributeMetadata, so this pull takes us further away from Jakarta Persistence's static metamodel and adds churn to spec API, introducing some errors along the way where documentation is only partially updated.

Ok, could we swap to the interface be `Attribute we have in Jakarta Persistence?

https://jakarta.ee/specifications/persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/metamodel/attribute

njr-11:
> Ok, could we swap to the interface be `Attribute we have in Jakarta Persistence?
> 
> https://jakarta.ee/specifications/persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/metamodel/attribute

There might be some typos in your question. I don't understand it.

The Jakarta Persistence 3.1 spec gives [this example](https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#a6961),
```
@StaticMetamodel(Order.class)
public class Order {
    public static volatile SingularAttribute<Order, Integer> orderId;
    public static volatile SingularAttribute<Order, Customer> customer;
    public static volatile SetAttribute<Order, Item> lineItems;
    public static volatile SingularAttribute<Order, Address> shippingAddress;
    public static volatile SingularAttribute<Order, BigDecimal> totalCost;
}
```

In Jakarta Data, you would currently either write (or have it generated for you),
```
@StaticMetamodel(Order.class)
public class Order {
    public static final Attribute orderId = Attribute.get();
    public static final Attribute customer = Attribute.get();
    public static final Attribute lineItems = Attribute.get();
    public static final Attribute shippingAddress = Attribute.get();
    public static final Attribute totalCost = Attribute.get();
}
```

These seem to already directly correspond, except that we have made it safer by replacing `volatile` with `final`, and simpler because there is currently only `Attribute` to use here rather than subtypes of it.

otaviojava:
> > Ok, could we swap to the interface be `Attribute we have in Jakarta Persistence?
> > https://jakarta.ee/specifications/persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/metamodel/attribute
> 
> There might be some typos in your question. I don't understand it.
> 
> The Jakarta Persistence 3.1 spec gives [this example](https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#a6961),
> 
> ```
> @StaticMetamodel(Order.class)
> public class Order {
>     public static volatile SingularAttribute<Order, Integer> orderId;
>     public static volatile SingularAttribute<Order, Customer> customer;
>     public static volatile SetAttribute<Order, Item> lineItems;
>     public static volatile SingularAttribute<Order, Address> shippingAddress;
>     public static volatile SingularAttribute<Order, BigDecimal> totalCost;
> }
> ```
> 
> In Jakarta Data, you would currently either write (or have it generated for you),
> 
> ```
> @StaticMetamodel(Order.class)
> public class Order {
>     public static final Attribute orderId = Attribute.get();
>     public static final Attribute customer = Attribute.get();
>     public static final Attribute lineItems = Attribute.get();
>     public static final Attribute shippingAddress = Attribute.get();
>     public static final Attribute totalCost = Attribute.get();
> }
> ```
> 
> These seem to already directly correspond, except that we have made it safer by replacing `volatile` with `final`, and simpler because there is currently only `Attribute` to use here rather than subtypes of it.
> 
> It seems like you are suggesting that just renaming `Attribute` to `AttributeMetadata` or `AttributeInfo` is going to make it closer to Jakarta Persistence, but I don't see how that is so.


We agree that in build time, we can generate the implementation, thus not using the wrapper class:

```java 
@StaticMetamodel(Order.class)
@Generated
public class Order {
public static final Attribute orderId = // an implementation where the Attribute is an interface not a wrapper class
}
```

Another point even in runtime, it might be better if we keep using the interface instead of breaking the interface segregation principle (ISP).

```java 
 @StaticMetamodel(Order.class)
 public class Order {
     public static final Attribute orderId = AttributeInfo.get();
     public static final Attribute customer = AttributeInfo.get();
     public static final Attribute lineItems = AttributeInfo.get();
     public static final Attribute shippingAddress = AttributeInfo.get();
     public static final Attribute totalCost = AttributeInfo.get();
 }
```

Where the Attribute is the interface and AttributeInfo is the wrapper class.

PR #316, opened 2023-10-26
Static metamodel should disallow sort for unsortable types

njr-11:
After reviewing what we have for the StaticMetamodel attributes, I realized that including sort methods on the one-size-fits-all `Attribute` class doesn't make sense because some types of entity attributes aren't capable of being sorted, for example a collection or an embeddable type.  Also, the special methods to sort ignoring case don't make sense on numeric types and several other other sortable types.

A solution to this could be a minimal hierarchy of Attribute subtypes, where the StaticMetamodel class can use whichever is appropriate to the entity attribute type:

Attribute  - general attribute, only offers the type-safe `name()`
SortableAttribute (extends Attribute)  - adds `asc()` and `desc()` sort
TextAttribute (extends SortableAttribute)  - adds `ascIgnoreCase()` and `descIgnoreCase()`

Also while working on this pull, I found a valid way to switch `Attribute` from a class to an interface as I believe Otavio has been wanting to do, so I switched that as well.


api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -25,111 +25,93 @@
...
      *
-     * @return a request for a descending, case insensitive sort on the entity attribute.
-     * @throws MappingException if an entity attribute with this name does not exist or
-     *                          if no Jakarta Data provider provides a repository for the entity type.
+     * @param attr attribute implementation provided by the Jakarta Data provider.
+     * @return true if initialization was successful; false if already initialized.
      */
-    public final Sort descIgnoreCase() {
-        return attrInfo().descIgnoreCase();
+    default boolean init(Attribute attr) {
```
otaviojava, line 51: IMHO: This method is an encapsulation, that lets the implementation define they way to initialize such as runtime or build time.

api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -25,111 +25,93 @@
...
      * <pre>
-     * &#64;StaticMetamodel(Vehicle.class)
-     * public class Vehicle_ {
-     *     public static final Attribute vin = Attribute.get();
-     *     public static final Attribute make = Attribute.get();
-     *     public static final Attribute model = Attribute.get();
-     *     public static final Attribute year = Attribute.get();
+     * &#64;StaticMetamodel(Person.class)
+     * public interface Person {
+     *     SortableAttribute id = SortableAttribute.get();
```
otaviojava, line 68: We need to be open to both runtime and build time.
Once you are not defining it as final, you can keep it open as Jakarta Persistence does.

api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -25,111 +25,93 @@
...
+     *     TextAttribute name = TextAttribute.get();
      * }
      * </pre>
      *
      * @return a new instance for an entity attribute.
      */
-    public static final Attribute get() {
-        return new Attribute();
-    }
+    public static Attribute get() {
```
otaviojava, line 77: This method does not make sense either.
Let the vendor do it their way.

otaviojava:
I like the approach; however, the interface could be more immutable and open to use both runtime and build time.

I also would use CollectionAttribute because sometimes we need the type inside the collection.

api/src/main/java/jakarta/data/metamodel/Attribute.java
njr-11: > We need to be open to both runtime and build time. Once you are not defining it as final, you can keep it open as Jakarta Persistence does.
> 
> Ref: https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#a6961

No, it **is** defined as `final` because `final` is implicit for attributes of interfaces.  That hasn't changed.  You need to explain what you mean that this is not open to build time.  We still have a TCK test that covers build time, and that test still compiles and passes with this approach, so your comment there makes no sense to me.

api/src/main/java/jakarta/data/metamodel/Attribute.java
njr-11: > IMHO: This method is an encapsulation issue; letting the implementation define their way to initialize, such as runtime or build time sounds better.

I don't see how I introduced any encapsulation issue here.

Before change this code was:
```    public final boolean init(AttributeInfo attrInfo)```

and now it is:
```   default boolean init(Attribute attr)```

`public` is implicit as an interface method, so there is no change there.  The changes that I made to the signature are switching the type of the parameter and making the method `default` so that the Jakarta Data provider doesn't need to bother implementing it when the attribute doesn't require initialization at run time.

api/src/main/java/jakarta/data/metamodel/Attribute.java
njr-11: > This method does not make sense either. Let the vendor do it their way.

This method is essential for allowing the user to initialize the metamodel field for the case where the metamodel is initialized at run time,

```public static final Attribute price = Attribute.get();```

By telling me to get rid of it and let the vendor do it means that build-time will be the only approach possible for the static metamodel, and that is completely unacceptable.

Also, it should be pointed out that we already had a `public static final Attribute get()` method prior to this pull, and I've only removed the `final` from it because `final` is implicit after that class was switched to an interface.

njr-11:
> I like the approach; however, the interface could be more immutable and open to use both runtime and build time.

You need to explain how have made things less open to build-time.  You gave no reasons for your general statement.  As I see it, the opposite is true, and this pull makes things more open to build-time because prior to this it was not possible for the build-time approach to supply custom implementations and extensions of attribute types, but now this has become possible (see the example in the TCK).  Before this pull, build-time was forced to use `Attribute.get` and `Attribute.init`, and after it, the build-time approach doesn't need to use those as all, and can just provide implementation of the Attribute interface.

> I also would use CollectionAttribute because sometimes we need the type inside the collection.

If we are going to pursue this approach, I can add a CollectionAttribute to it now if you like.  I was expecting we would add more subtypes like that in a future version what they provide is useful with the Jakarta Data API.

njr-11:
@otaviojava I added CollectionAttribute in https://github.com/jakartaee/data/pull/316/commits/a2f333d23942900f30917711f13216c62c5d9f64
I can definitely think of things to put there after version 1.0, but it isn't clear to me what, if any, additional methods, should be put there now.  Would you be willing to add onto it what you are thinking of in a follow-on pull?

api/src/main/java/jakarta/data/metamodel/Attribute.java
otaviojava: I will create a prototype for both scenarios: build time and runtime, thus, we can discuss this in check it in action.

api/src/main/java/jakarta/data/metamodel/Attribute.java
otaviojava: But why not set a field instance?

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/reflect/Field.html#set(java.lang.Object,java.lang.Object)

otaviojava:
> @otaviojava I added CollectionAttribute in [a2f333d](https://github.com/jakartaee/data/commit/a2f333d23942900f30917711f13216c62c5d9f64) I can definitely think of things to put there after version 1.0, but it isn't clear to me what, if any, additional methods should be put there now. Would you be willing to add onto it what you are thinking of in a follow-on pull?

I will create a side project to test both: runtime and compile time, this way, we can check together those side coners and imagine the scenarios.

I will do it on this weekend.

api/src/main/java/jakarta/data/metamodel/Attribute.java
njr-11: > I will create a prototype for both scenarios: build time and runtime, thus, we can discuss this in check it in action.

Yes, go ahead. But please understand that the TCK already has a prototype of both build time and run time styles.  And I've already updated an implementation accordingly and have seen both scenarios working in the TCK with the changes under this pull, so it has been verified in action.  Hopefully it helps when you try it out as well.

api/src/main/java/jakarta/data/metamodel/Attribute.java
njr-11: > But why not set a field instance?
> 
> https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/reflect/Field.html#set(java.lang.Object,java.lang.Object)

Java does not allow setting a field instance when the field is `final`.  You also cannot leave a `final` value unset and set it for the first time later.  Java `final` doesn't allow it.   If you want to set it later, you need non-final, and you become vulnerable to being overwritten and contaminated by any code that decides to change it, as was the concern in #296 .  Again, this pull is the wrong place for that discussion.  Go back and reopen #296 if you want to have that discussion.  The requirement for`final` is already present and we should not abuse this pull to change that.

otaviojava:
Can we close this PR?

njr-11:
> Can we close this PR?

I had moved it to Jakarta Data Future rather than 1.0, but yes it makes more sense to close it because the pull itself cannot be merged, and in the next release we will end up using a new issue for static metamodel, which likely will need a lot of discussion before creating any pulls.

PR #317, opened 2023-10-27
Overloaded use of Param is confusing

njr-11:
I came across the [following requirement](https://github.com/jakartaee/data/blob/26c8b42c7443dd7c059c637a4272e474fb48e350/spec/src/main/asciidoc/repository.asciidoc?plain=1#L295) while reviewing the specification,

> A repository with automatic query methods that are based on parameters must either be compiled so that parameter names are preserved in the class file (for example, using `javac -parameters`), or the parameter names must be specified explicitly using the `@Param` annotation.

This is certainly useful to be able to specify the entity attribute name via annotation when you are unable to ensure compilation with `-parameters` and we want there to be a way to do that.  However, `@Param` to specify the entity attribute name is confusing because `@Param` already has a different purpose in the spec: to specify the names of query language parameters.  For example,

```
@Query("UPDATE Product o SET o.price=o.price * (1-:rate) WHERE o.prodnum=:id")
boolean reducePrice(@Param("id") long productId,
                    @Param("rate") float discountRate);
```

versus the overloaded meaning of an entity attribute name,

```
List<Person> findNamed(@Param("firstName") String first,
                       @Param("lastName") String last);
```

With the right amount of documentation we could certainly define in the spec that when `@Param` annotates a parameter of a method, if the `@Query` annotation is on the method, then `@Param` refers to a Query Language named parameter, but when no `@Query` annotation is on the method, then `@Param` refers to the name of an entity attribute that is being compared against the value of the method parameter.  However, it would be clearer just to use separate annotations for the two different meanings.  This pull adds a separate annotation for the latter case (designating the name of the entity attribute for comparison against).

This will keep the two usage patterns separate,

Query Language named parameter:

```
@Query("UPDATE Product o SET o.price=o.price * (1-:rate) WHERE o.prodnum=:id")
boolean reducePrice(@Param("id") long productId,
                    @Param("rate") float discountRate);
```

Entity attribute name:

```
List<Person> findNamed(@By("firstName") String first,
                       @By("lastName") String last);
```


otaviojava:
In both contexts, we are talking about parameters for a query. It would be easier for keep an unique annotation and explore the documentation.
Otherwise, we need to document when we mix documentation, such as:

```java
List<Person> findNamed(@Param("firstName") String first,
                       @Param("lastName") String last);
```

```java
@Query("UPDATE Product o SET o.price=o.price * (1-:rate) WHERE o.id=:id")
boolean reducePrice(@By("id") long productId,
                    @By("rate") float discountRate);
```

njr-11:
I don't agree with your argument.  There are already lots of ways for badly-behaving users to do nonsensical things with annotations and other patterns. For example,
```
@Save
List<Car> deleteByName(@Param("IDontFollowRules") Product prod);
```

A user who intentionally violates the rules of the spec like this is going to end up with useless code, and it's their own fault for doing so, and these users behavior badly should not be our priority in writing the spec.  Instead, we should prioritize making the specification meaningful and clear for users who want to do the right thing.  What I am pointed out here is that making `@Param(name)` sometimes apply to a Query Language named parameter name and other times apply to an entity attribute name depending on some external annotation or lack thereof is just confusing and only hurts the users who are wanting to do the right thing, but have trouble understanding documentation because we have overloaded the meaning of things and made it unnecessarily complex.

Rather than the example you gave which looks like someone doing the opposite of what the clearly separated documentation for the two different scenarios told them to, the real risk is in overloaded `Param` annotation trying to be two different things at once confuses someone into unintentionally misunderstanding the overloaded documentation as follows and setting the `Param` value to the entity attribute name where in a case where it ought to have been the Query Language named parameter name,

```
// incorrect because the user misunderstands overloaded Param documentation:
@Query("UPDATE Product o SET o.price=o.price * (1-:rate) WHERE o.prodnum=:id")
boolean reducePrice(@Param("prodnum") long productId,
                    @Param("price") float discountRate);
```

otaviojava:
IMHO: We need to document this case as well and define, somehow, that the Param annotations work only for query, otherwise it will be ignored.

```java
@Save
List<Car> deleteByName(@Param("IDontFollowRules") Product prod);
```

otaviojava:
I am reading the spec now; I will answer it later tomorrow.
So, please, give me more time.

otaviojava (approved):

PR #318, opened 2023-10-30
Document both scenarios that the Param annotation is used for

njr-11:
This pull is an inferior solution versus #317 and it is preferable that #317 be merged and this pull closed without merging.  However, if #317 is not accepted, then this pull is needed to at least documente the confusing overloaded behavior that currently exists in the spec for using `@Param` to represent both Query Language named parameter names AND entity attribute names depending the presence or absence of an external annotation on the method of which one or more parameters are annotated with `@Param`.


njr-11:
closing in favor of the solution under #317

PR #321, opened 2023-10-31
TCK tests for life cycle methods

njr-11:
Add TCK for Insert, Update, and Delete life cycle methods.
Cover both single entity as well as multiple entities in the parameter.
Use an entity with a generated version and confirm it gets updated and is used to make updates.
Confirm that exceptions are raised per the spec requirements.

I also removed two extra `*` characters that I noticed in JavaDoc.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/PersistenceEntityTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/PersistenceEntityTests.java
@@ -16,12 +16,15 @@
 package ee.jakarta.tck.data.standalone.persistence;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.fail;
 
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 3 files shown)


tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/PersistenceEntityTests.java
```diff
@@ -131,6 +136,22 @@ public void testIdAttributeWithDifferentName() {
         assertEquals(5L, catalog.deleteByProductNumLike("TEST-PROD-%"));
     }
 
+    @Assertion(id = "133", strategy = "Attempt to insert and entity that already exists in the database and expect EntityExistsException.")
```
mswatosh, line 139:
```suggestion
    @Assertion(id = "133", strategy = "Attempt to insert an entity that already exists in the database and expect EntityExistsException.")
```

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/PersistenceEntityTests.java
```diff
@@ -169,6 +190,66 @@ public void testNotRunOnNOSQL() {
...
     }
 
+    @Assertion(id = "133", strategy = "Life cycle methods to insert, update, and delete multiple entities.")
+    public void testMultipleInsertUpdateDelete() {
+        catalog.deleteByProductNumLike("TEST-PROD-%");
+
+        Product[] added = catalog.addMultiple(Product.of("blueberries", 2.49, "TEST-PROD-95", Department.GROCERY),
+                                              Product.of("strawberries", 2.29, "TEST-PROD-96", Department.GROCERY),
+                                              Product.of("raspberries", 2.39, "TEST-PROD-97", Department.GROCERY));
+
```
mswatosh, line 200:
```suggestion
assertEquals(3, added.length);

```
Should we verify that added has a length of exactly 3, like is being done below to confirm modified is exactly 2?

mswatosh (commented):
Overall the tests look good, just added a few suggestions.
(This still needs approval from a committer)

jmstephensgit (approved):

PR #335, opened 2023-11-06
More complete documentation of primary entity type

njr-11:
Documentation of how the primary entity type is determined is lacking for when a built-in repository interface is not used.  This pull adds it, along with a few example code snippets to help illustrate.


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -436,7 +436,7 @@ For example, a repository interface may inherit the `CrudRepository` interface d
...
+Users of Jakarta Data declare a primary entity type for a repository by inheriting from a built-in repository super interface, such as `BasicRepository`, and specifying the primary entity type as the first type variable. For repositories that do not inherit from a super interface with a type parameter to indicate the primary entity type, life cycle methods on the repository determine the primary entity type if all life cycle methods where the method parameter is a type, an array of type, or is parameterized with a type that is annotated as an entity, correspond to the same entity type. The primary entity type is assumed for methods that do not otherwise specify an entity type, such as `countByPriceLessThan`. Methods that require a primary entity type raise `MappingException` if a primary entity type is not provided.
```
mswatosh, line 439:
```suggestion
Users of Jakarta Data declare a primary entity type for a repository by inheriting from a built-in repository super interface, such as `BasicRepository`, and specifying the primary entity type as the first type variable. For repositories that do not inherit from a super interface with a type parameter to indicate the primary entity type, life cycle methods on the repository determine the primary entity type. To do so, all life cycle methods where the method parameter is a type, an array of type, or is parameterized with a type that is annotated as an entity, must correspond to the same entity type. The primary entity type is assumed for methods that do not otherwise specify an entity type, such as `countByPriceLessThan`. Methods that require a primary entity type raise `MappingException` if a primary entity type is not provided.
```
Took me a few reads of this sentence to understand it, so splitting it in to might help. Also, what happens if a User doesn't inherit from a built-in repository interface, and also doesn't use the same entity type? Presumably an exception, but should we specify what kind?

mswatosh (commented):
Good additions, just added one suggestion in the spec doc

spec/src/main/asciidoc/repository.asciidoc
njr-11: That's a good suggestion to improve the clarify. I'll add it.
Regarding your question,
> Also, what happens if a User doesn't inherit from a built-in repository interface, and also doesn't use the same entity type? Presumably an exception, but should we specify what kind?

That is covered further down in documentation that was already there before this pull. It states:
"Methods that require a primary entity type raise `MappingException` if a primary entity type is not provided."

otaviojava (approved):

PR #336, opened 2023-11-07
PoC create Pagination specialization

otaviojava:
Currently, we have a unique interface to care for the two types of pagination.

This approach may lead to mistakes because there is a method that makes sense to OFF-SET, the next(), and methods that make sense to Key, cursor(). Also, breaking the Liskov Principle.

The proposal is, using the semantics that we have in our documentation, to create two interfaces:

```mermaid
classDiagram
    Pagination <|-- Pageable
    Pagination <|-- KeySetPageable
  class Pageable{
       Pageable next();
    }

  class KeySetPageable{
        Cursor cursor();
    }
```

I started to show the proposal, but I did not finish. I want to hear if it makes sense.


njr-11 (changes requested):
The proposed specialization is overly complex and confusing.  I don't see any violation of the Liskov principle in the first place. The Liskov principle requires that subtypes can be used in place of the supertype without breaking the program.  Prior to this change, there was just the Pageable interface with its built-in implementation that follows the rules outlined by the Pageable JavaDoc precisely.
Subtypes would only come from optional extension by vendors, who are responsible for abiding by the principle. They should already be able to do so with no difficulty.

Your proposal here renames Pageable to Pagination (that part is fine) and adds subinterfaces of it, and then implements all of the subinterfaces with a single implementation, seemingly defeating the purpose of the specialization.  The user ends up with 3 different pagination-related interfaces which they need to be concerned with when composing repository methods, and the Jakarta Data provider similarly needs to be able to handle 3 different pagination-related interfaces that could appear on a repository method signature.  It is unnecessary complexity and creates churn in the spec that runs the risk of introducing errors.  I don't think the pull as it currently is will compile cleanly, but there is also the risk of JavaDoc and spec errors.

Think of the current Pageable interface as a simple configuration-like object for a page request.  Some of the configuration is optional, such as whether to provide a keyset cursor, whether to provide a starting page, and whether to provide additional Sorts.  We don't need a specialized SortsPageable when someone chooses to do the latter.  Neither should we need a specialized KeysetPageable when someone chooses to do the former.

PR #339, opened 2023-11-08
Make Optional return the cursor

otaviojava:


njr-11 (approved):
This mostly works, but it seems like it would be better to do:
```
record Pagination(long page, int size, List<Sort> sorts, Mode mode, Optional<Cursor> cursor) implements Pageable {
```
so that cursor is actually a component of the record and will show up that way (instead of type) if someone invokes Pageable.ofSize(10).getClass().getRecordComponents();

PR #347, opened 2023-11-12
Include offset pagination

otaviojava:
# Changes

- Include at the spec off set pagination


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -1109,6 +1111,106 @@ The critical differences between Offset and Keyset pagination lie in their retri
...
+
+Offset pagination offers several key features that make it a valuable approach for managing and retrieving large datasets in a controlled and efficient manner:
+
+- *Page Size*: The maximum number of elements to be included in each page is known as the page size. This parameter determines the subset of data retrieved with each pagination request.
+
+- *Page Number*: The page number indicates which subset of the dataset to retrieve. It typically starts from 1, representing the first page, and increments with each subsequent page.
+
+- *Efficient Navigation*: Offset pagination allows efficient dataset navigation. By specifying the desired page and page size, developers can control the data retrieved, optimizing memory usage and processing time.
+
+- *Sequential Order*: Elements are retrieved sequentially based on a predefined criterion, such as ascending or descending order of a specific attribute, like an ID.
```
njr-11, line 1128: The plural form should be used here because the pattern allows ordering by multiple.
```suggestion
- *Sequential Order*: Elements are retrieved sequentially based on predefined criteria, such as ascending or descending order of a specific attribute, like an ID.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -1109,6 +1111,106 @@ The critical differences between Offset and Keyset pagination lie in their retri
...
+
+Code Execution:
+
+[source,java]
+----
+
+@Inject
+People people;
+
+Page<Person> page = people.findAll(Pageable.ofPage(1).size(2));
```
njr-11, line 1176: Sorting by id is required to guarantee the order that is shown in the example,
```suggestion
Page<Person> page = people.findAll(Pageable.ofPage(1).size(2).sortBy(Sort.asc("id")));
```

njr-11 (changes requested):
Sort by Id needs to be added to the example to make it valid.

otaviojava:
> Sort by Id needs to be added to the example to make it valid.

Thank you to put it.

njr-11 (approved):

otaviojava:
@njr-11 what do you think if we do something pretty similar to key-set?

njr-11:
> @njr-11 what do you think if we do something pretty similar to key-set?

The section for keyset cursor pagination already exists and I think does a better job of being explicit about requirements and expectations.

PR #348, opened 2023-11-13
Create validation tests

KyleAure:


tck/src/main/java/ee/jakarta/tck/data/web/validation/PersistenceTests.java
```diff
@@ -0,0 +1,238 @@
...
+        // Update
+        Rectangle updatedRect = new Rectangle("RECT-020", 10L, 10L, 10, 10);
+        rectangles.save(updatedRect);
+        
+        assertEquals(1, rectangles.count(), "Number of results was incorrect");
+        
+        // Verify
+        resultRects = getResults();
+        Rectangle resultRect = resultRects.get(0);
+        assertRectangleFields(updatedRect, resultRect);
```
njr-11, line 129: Although in practice this probably works in almost all cases, Jakarta makes no guarantees about the state of an entity parameter that is supplied to a save operation after the method, so we shouldn't be asserting based on it. The result of save needs to be used instead of the parameter supplied to it,
```suggestion
        updatedRect = rectangles.save(updatedRect);
        
        assertEquals(1, rectangles.count(), "Number of results was incorrect");
        
        // Verify
        resultRects = getResults();
        Rectangle resultRect = resultRects.get(0);
        assertRectangleFields(updatedRect, resultRect);      
```

tck/src/main/java/ee/jakarta/tck/data/web/validation/PersistenceTests.java
```diff
@@ -0,0 +1,238 @@
...
+                new Rectangle("RECT-032", 1800L, 999L, 120, 80),
+                new Rectangle("RECT-033", 1L, 1L, 1, 2)
+                );
+        rectangles.saveAll(updatedRects);
+        assertEquals(4, rectangles.count(), "Number of results was incorrect");
+        
+        // Verify All
+        resultRects = getResults();
+        for(int i = 0; i < resultRects.size(); i++) {
+            assertRectangleFields(updatedRects.get(i), resultRects.get(i));
```
njr-11, line 159: Similar to above comment:
```suggestion
        updatedRects = rectangles.saveAll(updatedRects);
        assertEquals(4, rectangles.count(), "Number of results was incorrect");
        
        // Verify All
        resultRects = getResults();
        for(int i = 0; i < resultRects.size(); i++) {
            assertRectangleFields(updatedRects.get(i), resultRects.get(i));
```

tck/src/main/java/ee/jakarta/tck/data/web/validation/PersistenceTests.java
```diff
@@ -0,0 +1,238 @@
...
+        }     
+    }
+    
+    @Assertion(id = "133", strategy="Ensure that entities that violate the constraints cannot be updated in the database")
+    public void testUpdateWithInvalidConstraints() {
+        ConstraintViolationException resultingException;
+        
+        // Save
+        Rectangle validRect = new Rectangle("RECT-040", 5L, 5L, 5, 5);
+        rectangles.save(validRect);
```
njr-11, line 169:
```suggestion
        validRect = rectangles.save(validRect);
```

tck/src/main/java/ee/jakarta/tck/data/web/validation/PersistenceTests.java
```diff
@@ -0,0 +1,238 @@
...
+        ConstraintViolationException resultingException;
+        
+        // Save All
+        List<Rectangle> validRects = List.of(
+                new Rectangle("RECT-050", 5L, 5L, 5, 5),
+                new Rectangle("RECT-051", 5L, 5L, 5, 5),
+                new Rectangle("RECT-052", 5L, 5L, 5, 5),
+                new Rectangle("RECT-053", 5L, 5L, 5, 5)
+                );
+        rectangles.saveAll(validRects);
```
njr-11, line 197:
```suggestion
        validRects = rectangles.saveAll(validRects);
```

tck/src/main/java/ee/jakarta/tck/data/web/validation/Rectangles.java
```diff
@@ -0,0 +1,32 @@
...
+
+import jakarta.data.repository.DataRepository;
+import jakarta.data.repository.Repository;
+import jakarta.validation.Valid;
+
+@Repository
+public interface Rectangles extends DataRepository<Rectangle, String> {
+    
+    void save(@Valid Rectangle rect);
+    void saveAll(@Valid Iterable<Rectangle> rects);
```
njr-11, line 28: You'll need the following updates to make the prior suggestions valid,
```suggestion
    Rectangle save(@Valid Rectangle rect);
    Iterable<Rectangle> saveAll(@Valid Iterable<Rectangle> rects);
```

njr-11 (changes requested):
This looks great!  I added a few suggestions around a minor technicality where Jakarta Data makes no guarantees about the ability to continue using entity instances that are supplied as parameters after the repository method and instead requires you to obtain them from the method result if you want to continue using them as the tests are doing in places

tck/src/main/java/ee/jakarta/tck/data/web/validation/ValidationTests.java
```diff
@@ -0,0 +1,342 @@
...
+        return ShrinkWrap.create(WebArchive.class)
+                .addClasses(Rectangle.class, Rectangles.class);
+    }
+    
+    @Inject
+    private Rectangles rectangles;
+    
+    @BeforeEach
+    public void cleanup() {
+        rectangles.deleteAll();
```
njr-11, line 54: Does this also need to wait for eventual consistency?  Some tests count the total number of rectangles so if some from a previous test can sometimes still be seen, it seems like this might cause an intermittent failure.

otaviojava: It does for NoSQL, however, we can keep the focus on SQL while we don't have the NoSQL spec yet.

KyleAure: Good catch, fixed :)

Issue #228, opened 2023-08-30
Proofread and copyedit the spec


gavinking:
A native English speaker with specification-writing experience needs to spend some time rewording the specification.

gavinking:
I'm informed by @maxandersen that this issue report could be taken the wrong way. To clarify: as a native speaker with lots of experience at writing specs, I am of course volunteering to do this work.

otaviojava:
Hey, I got your point, and you are right; there are a lot of issues for improvement. I have almost zero experience writing a spec from scratch and am not a native speaker.


Thank you. I will need as much help as needed, mainly because I do it in my free time as a community person.

njr-11:
Additional work that is needed under this issue:
- [ ] The **Copyright** section has a number of variables that are not being substituted when I build locally, as well as in the officially-built M1 spec doc, https://jakarta.ee/specifications/data/1.0/data-1.0.0-m1.
- [ ] Section **1.3. Conventions** is empty. We should consider removing it unless there are plans to author something for that section.
- [ ] Section **1.4. Jakarta Data Project Team** should be updated to include the names of people who have made significant contributions to the spec. We should consider voting in some of these people as committers and add them under that section as well.

PR #349, opened 2023-11-13
Review section 4 on Repository Interfaces, except for pagination

njr-11:
Proofread section 4 on Repository Interfaces and all subsection except for 4.8 Pagination.

Section 4.5.2 on Entity Property Names has examples where the discussion is ambiguous because the entity property name and type are the same: `Address address` and `ZipCode zipCode`.  In order to make it clear which between the property name vs the property type is being referred to, I am changing the types to `MailingAddress address` and `int zipCode`.  Beyond this pull, this section needs some consideration as to whether its content is appropriate to a specification.  It seems to describe how a particular implementation's algorithm deals with certain examples rather than specifying requirements that implementations must meet.  I'll note this separately under #228 so that we remember to revisit it.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -430,7 +430,7 @@ In instances where a Jakarta Data provider for NoSQL databases encounters a recu
 
 == Repository Interfaces
 
-A Jakarta Data repository is a Java interface annotated `@Repository`.
+A Jakarta Data repository is a Java interface annotated with `@Repository`.
 A repository interface may declare:
 
 - _abstract_ (non-`default`) methods, and
@@ -442,7 +442,7 @@ Every abstract method of the interface is usually either:
 
 - an entity instance _lifecycle method_,
 - an _annotated query method_,
-- an _automatic query method_, or
+- an _automatic query method_ (with Parameter-based conditions or Query by Method Name), or
 - a _resource accessor method_.
 
 A repository may declare lifecycle methods for a single entity type, or for multiple related entity types.
@@ -457,9 +457,7 @@ Repositories perform operations on entities. For repository methods that are ann
 Users of Jakarta Data declare a primary entity type for a repository by inheriting from a built-in repository super interface, such as `BasicRepository`, and specifying the primary entity type as the first type variable. For repositories that do not inherit from a super interface with a type parameter to indicate the primary entity type, life cycle methods on the repository determine the primary entity type. To do so, all life cycle methods where the method parameter is a type, an array of type, or is parameterized with a type that is annotated as an entity, must correspond to the same entity type. The primary entity type is assumed for methods that do not otherwise specify an entity type, such as `countByPriceLessThan`. Methods that require a primary entity type raise `MappingException` if a primary entity type is not provided.
 
 
-NOTE: A Jakarta Data provider might go beyond what is required by this specification and support abstract methods which do not fall into any of the above categories.
-
-Such functionality is not defined by this specification, and so programs with repositories which declare such methods are not portable between providers.
+NOTE: A Jakarta Data provider might go beyond what is required by this specification and support abstract methods which do not fall into any of the above categories. Such functionality is not defined by this specification, and so programs with repositories which declare such methods are not portable between providers.
 
 The subsections below specify the rules that an abstract method declaration must observe so that the Jakarta Data implementation is able to provide an implementation of the abstract method.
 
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -457,9 +457,7 @@ Repositories perform operations on entities. For repository methods that are ann
...
 
 
-NOTE: A Jakarta Data provider might go beyond what is required by this specification and support abstract methods which do not fall into any of the above categories.
-
-Such functionality is not defined by this specification, and so programs with repositories which declare such methods are not portable between providers.
+NOTE: A Jakarta Data provider might go beyond what is required by this specification and support abstract methods which do not fall into any of the above categories. Such functionality is not defined by this specification, and so programs with repositories which declare such methods are not portable between providers.
```
mswatosh, line 460:
```suggestion
NOTE: A Jakarta Data provider might go beyond what is required by this specification and support abstract methods which do not fall into any of the above categories. Such functionality is not defined by this specification, and so applications with repositories which declare such methods are not portable between providers.
```
Do we want this as generic as programs, or would applications read better here?

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -583,20 +581,24 @@ The Jakarta Data implementation automatically recognizes the query annotations i
 
 === Parameter-based automatic query methods
```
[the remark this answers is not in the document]
njr-11: This section does not include Query by Method Name.  It says, "the Query by Method Name pattern is discussed separately", and  there is a separate section, at the same level, for that:

```
=== Query by Method Name
```

It would be confusing to call this section Automatic query methods when it is not the place to find information on both types of Automatic query method, and is only intended for Parameter-based conditions on automatic query methods.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -786,19 +791,19 @@ In this scenario, we have the following data models:
...
+- The query method `findByAddress_ZipCode` takes an `int` object as a parameter.
 - The method name includes an underscore (`_`) indicating manual traversal points.
 - The Property Resolution Algorithm will parse the method name and extract `Address_ZipCode`, recognizing the underscore as a traversal point.
-- It will then attempt to resolve the property named `Address` within the `Order` class, followed by the `zipcode` property within the `Address` class, following manual traversal points.
-- If properties `Address` and `ZipCode` are found in the appropriate classes or their nested objects, the query will filter `Order` records based on the provided `Zipcode` object within the `Address` object.
+- It will then attempt to resolve the property named `address` within the `Order` class, followed by the `zipcode` property within the `MailingAddress` class, following manual traversal points.
+- If properties `address` and `zipCode` are found in the appropriate classes or their nested objects, the query will filter `Order` records based on the provided `zipcode` integer within the `MailingAddress` object.
```
mswatosh, line 806:
```suggestion
- If properties `address` and `zipcode` are found in the appropriate classes or their nested objects, the query will filter `Order` records based on the provided `zipcode` integer within the `MailingAddress` object.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -987,11 +992,15 @@ Wildcard characters for patterns are determined by the data access provider. For
...
+==== Return Types
+
+Refer to the Jakarta Data module JavaDoc section on "Return Types for Repository Methods" for a listing of valid return types for methods using Query by Method Name.
+
 === Special Parameter Handling
 
 Jakarta Data also supports particular parameters to define pagination and sorting.
 
-Jakarta Data recognizes, when specified on a repository method after the query parameters, specific types, like `Limit`, `Pageable`, and `Sort`, to dynamically apply limits, pagination, and sorting to queries. The following example demonstrates these features:
+Jakarta Data recognizes, when specified on a repository method after the query parameters, the specific types, `Limit`, `Pageable`, and `Sort`, to dynamically apply limits, pagination, and sorting to queries. The following example demonstrates these features:
```
mswatosh, line 1003:
```suggestion
Jakarta Data recognizes, when specified on a repository method after the query parameters, the specific types, `Limit`, `Pageable`, and `Sort`, to dynamically apply limits, pagination, and sorting to queries, respectively. The following example demonstrates these features:
```

mswatosh (commented):
Lots of good updates, I added a few suggestions

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -652,12 +654,14 @@ public interface ProductRepository extends BasicRepository<Product, Long> {
...
 - The order clause, which is optional, begins with `OrderBy` and consists of an ordered collection of entity attributes by which to sort results, delimited by `Asc` or `Desc` to specify the sort direction of the preceding attribute.
 - The method name is formed by combining the subject, predicate, and order clause, in that order.
 
-Queries can also handle entities with related classes by specifying the relationship using dot notation.
+NOTE: This specification uses the terms subject and predicate in a way that aligns with industry terminology rather than how they are defined in English grammar. Aligning with the rules of English grammar, the subject would be the repository instance upon which the method is invoked, and the predicate would start with the action (such as `find`) and include the prepositional phrase `By...`.
+
+Queries can also handle entities with relation attributes by specifying the relationship using dot notation, with the dot converted to underscore so that it is a valid character within the method name.
```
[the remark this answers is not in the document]
njr-11: Good idea. I'll add your suggestion.

spec/src/main/asciidoc/repository.asciidoc
njr-11: Yes, application reads better. I'll add your suggestion.

spec/src/main/asciidoc/repository.asciidoc
njr-11: Good catch with the capitalization. I'll add your suggestion.

spec/src/main/asciidoc/repository.asciidoc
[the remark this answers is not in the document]
mswatosh: Ah good point, I think I glossed over the part in parenthesis, but it's pretty clearly stated.

otaviojava (approved):

PR #351, opened 2023-11-13
Removing Paging Justification

mswatosh:
This section gives a really good explanation of why pagination is important, but I don't think it's necessary for the specification. This could be re-used if someone writes a sample or blog post covering paging though.


njr-11 (approved):
> @njr-11 WDYT?

The paragraph right before this section already gives a good introduction, even mentioning its purpose and benefits, so these bullet points do seem unnecessary. I don't see anything within them that specifies any requirements on the user or Jakarta Data provider that we would be losing, so I agree with removing.

otaviojava:
@mswatosh could you fix the conflict?

mswatosh:
The merge conflict is fixed 👍

PR #355, opened 2023-11-14
review of section 7 on Portability

njr-11:
The section on Portability in Jakarta Data needed some more specifics, including calling out a minimal list of JPA annotations that must be supported.  We can't just say everything you can define on a JPA entity is supported, because our specification doesn't have the detail to define how the integration would work. For example, JPA will let you define composite IDs, but we don't currently say anything about how that would integrate with Jakarta Data and what it would mean to repository queries and so forth, so we should leave that one out for now.

### DIFF
--- a/spec/src/main/asciidoc/portability.asciidoc
+++ b/spec/src/main/asciidoc/portability.asciidoc
@@ -1,22 +1,46 @@
+
 == Portability in Jakarta Data
 
-Portability is a critical aspect of Jakarta Data, ensuring flexibility and adaptability across different data stores while maintaining consistent functionality. Jakarta Data offers a comprehensive approach to portability, which is discussed separately for relational databases and NoSQL databases.
+Jakarta Data offers varying degrees of portability depending on the database and capabilities used. A subset of function is standardized across all database types, while other subsets of function are standardized only for the specific types of databases to which the capability pertains. These requirements are explicitly called out in documentation throughout the specification, such as the "Unavailable In" column of the tables of repository keywords in the Jakarta Data module JavaDoc. The Jakarta Data specification levies requirements against applications and Jakarta Data providers, but not against databases. The Jakarta Data specification requires the Jakarta Data provider to offer function to the extent that the database is capable and to raise an exception where the database is not capable. It is not the aim of Jakarta Data to offer the ability to switch between different databases, but to standardize a common starting point for data access from which capability that is specific to the various types of databases is able to build upon.
 
-=== Portability in Relational Databases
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/portability.asciidoc
```diff
@@ -1,22 +1,46 @@
...
+Jakarta Data's support for query methods, including pagination, ordering, and limiting, is designed to work seamlessly with relational databases.
 
-3. **Query Methods:** Jakarta Data's support for query methods, including pagination, ordering, and limiting, is designed to work seamlessly with relational databases.
+All limitations of Jakarta Persistence, the respective Jakarta Persistence provider, JPQL, SQL, and of the database apply when using entities that are annotated with `jakarta.persistence.Entity`. Repository methods must correspond to operations that are valid via `JPQL` and `SQL` queries. For example, although one can write a repository method that asks for sorting by a collection attribute or attempts to perform a Like operation on a numeric type rather than a String, there is no expectation for a Jakarta Data provider to support doing so because Jakarta Persistence and JPQL and SQL do not offer the ability.
```
mswatosh, line 37:
```suggestion
All limitations of Jakarta Persistence, the respective Jakarta Persistence provider, JPQL, SQL, and of the database apply when using entities that are annotated with `jakarta.persistence.Entity`. Repository methods must correspond to operations that are valid via `JPQL` and `SQL` queries. For example, although one can write a repository method that asks for sorting by a collection attribute or attempts to perform a Like operation on a numeric type rather than a String, there is no expectation for a Jakarta Data provider to support doing so because Jakarta Persistence, JPQL, and SQL do not offer the ability.
```

spec/src/main/asciidoc/portability.asciidoc
```diff
@@ -1,22 +1,46 @@
...
+=== Portability for Relational Databases
 
 All functionality defined by Jakarta Data must be supported when using relational databases.
 
-1. **Support for Jakarta Persistence Annotations:** Jakarta Data, when used in conjunction with a Jakarta Persistence provider, ensures comprehensive support for all Jakarta Persistence entity annotations. This support covers a wide range of functionality, including entity definitions, primary keys, and relationships.
+==== Jakarta Persistence Annotations
+
+Jakarta Data, when used in conjunction with a Jakarta Data provider that is backed by Jakarta Persistence, requires support for the `jakarta.persistence.Entity` annotation on entity classes and requires support for the following minimal set of Jakarta Persistence annotations for entity properties:
+
+* `jakarta.persistence.Basic`
```
otaviojava, line 16: How about saying that all Jakarta Persistence annotations should be supported when used in conjunction with Jakarta Data?

njr-11: > How about saying that all Jakarta Persistence annotations should be supported when used in conjunction with Jakarta Data?

I don't think we can do that because the spec would first need to define how to interoperate with the particular capabilities.  For example, with a composite id, the meaning of methods like findByIdBetween is unclear as is their usage.  We cannot claim that vendors should support things that are undefined.

otaviojava: hum, I got it.

otaviojava (approved):

PR #357, opened 2023-11-15
Include example with data for keyset pagination

njr-11:
On the Jakarta Data call, Otavio asked that I add a keyset cursor-based pagination example scenario similar to what he wrote for offset-based pagination.  This pull adds it, using the same example data from the offset scenario. I updated one of the entities so that it has the same name as another, and the keyset scenario includes removing the duplicate as it is stepping through the pages of results such that an entity would be skipped (due to shifted offset) after deletion if using offset-based pagination, but with keyset cusror-based pagination, it finds the entity based on the cursor position and does not skip.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -1147,7 +1147,7 @@ The dataset contains the following elements:
    {"id":3, "name":"Alyse Dadson"},
    {"id":4, "name":"Orelle Roughey"},
    {"id":5, "name":"Jaquith Wealthall"},
-   {"id":6, "name":"Debor Bracegirdle"},
+   {"id":6, "name":"Boothe Martinson"},
    {"id":7, "name":"Patten Bedell"},
    {"id":8, "name":"Danita Pilipyak"},
    {"id":9, "name":"Harlene Branigan"},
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -1308,3 +1308,87 @@ for (Pageable p = Pageable.ofSize(25).sortBy(Sort.desc("yearBorn"),
...
+----
+
+Code Execution:
+
+[source,java]
+----
+@Inject
+People people;
+
+Pageable firstPageRequest = Pageable.ofSize(4).sortBy(Sort.asc("name"), Sort.asc("id"));
```
otaviojava, line 1347: This will generate a Pageable of offset, is not it?

njr-11: Yes, it requests the first keyset aware page with an offset from the beginning of the data set.  Then, once it has that page, it requests the next page, which is via a cursor based on the keyset of the last element of the first page.  This is typical usage if you wanted to start at the beginning and read through all of the results making sure you don't miss any due to deletions of other entities.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -1308,3 +1308,87 @@ for (Pageable p = Pageable.ofSize(25).sortBy(Sort.desc("yearBorn"),
...
+// The user decides to remove one of the entities that has the same name,
+people.deleteById(10);
+----
+
+
+Next Page Execution:
+
+[source,java]
+----
+Pageable nextPageRequest = page.nextPageable();
```
otaviojava, line 1376: @njr-11 we are using this method; should we use cursor instead?

njr-11: > @njr-11 we are using this method; should we use cursor instead?

No, `nextPagageable` on a KeysetAwarePage/KeysetAwareSlice implicitly uses a cursor, per the JavaDoc:

```
     * <p>Returns pagination information for requesting the next page
     * in a forward direction from the current page. This method computes a
     * keyset from the last entity of the current page and includes the
     * keyset in the pagination information so that it can be used to
     * obtain the next slice in a forward direction according to the
     * sort criteria and relative to that entity.</p>
```

otaviojava (approved):

PR #359, opened 2023-11-16
review of pagination JavaDoc

njr-11:
Proofread all of the JavaDoc in the jakarta.data.page package.
I also cleaned up a few compile warnings that I noticed in unit test (missing cast of null value to varargs method parameter type) and another annotation class (missing JavaDoc).

### DIFF
--- a/api/src/main/java/jakarta/data/page/Slice.java
+++ b/api/src/main/java/jakarta/data/page/Slice.java
@@ -50,7 +50,7 @@ public interface Slice<T> extends Streamable<T> {
      * sorting first by the sort criteria of the repository method,
      * and then by the sort criteria of the page request.
      *
-     * @return the page content as a {@link List}; will never be {@literal null}.
+     * @return the page content as a {@link List}; will never be {@code null}.
      */
     List<T> content();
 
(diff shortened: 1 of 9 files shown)


api/src/main/java/jakarta/data/page/Slice.java
```diff
@@ -21,16 +21,36 @@
...
     /**
-     * Returns the page content as {@link List}.
+     * Returns the page content as a {@link List}.
+     * The list is sorted according to the combined sort criteria of the repository method
+     * and the sort criteria of the page request that is supplied to the repository method,
+     * sorting first by the sort criteria of the repository method,
+     * and then by the sort criteria of the page request.
      *
-     * @return the page content as {@link List}; will never be {@literal null}.
+     * @return the page content as a {@link List}; will never be {@literal null}.
```
mswatosh, line 53:
```suggestion
     * @return the page content as a {@link List}; will never be {@code null}.
```
Should this be `@code` instead of `@literal`? null doesn't seem to need `@literal`, at least not how VSCode renders the javadoc, but `@code` could make it clear what we mean by not returning null.

mswatosh (commented):
Looks good to me, added one suggestion that I caught while reviewing

otaviojava (approved):

PR #365, opened 2023-11-18
Replace @literal with @code in javadoc

mswatosh:
replaces the use of `@literal` with `@code`

There are also quite a few uses of `&#64;`, `&lt;`, `&gt;`, and just `<code>` that could be updated to use `@code` as well. I think this could improve readability in the javadocs, particularly cases like `module-info.java`, but it would be a large effort, so I would like to know what other people think before attempting it.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/BasicRepository.java
+++ b/api/src/main/java/jakarta/data/repository/BasicRepository.java
@@ -94,12 +94,12 @@ public interface BasicRepository<T, K> extends DataRepository<T, K> {
      * <p>If the entity uses optimistic locking and the version differs from the version in the database,
      * an {@link OptimisticLockingFailureException} will be thrown.</p>
      *
-     * @param entity The entity to be saved. Must not be {@literal null}.
+     * @param entity The entity to be saved. Must not be {@code null}.
      * @param <S> Type of the entity to save.
-     * @return The saved entity; never {@literal null}.
+     * @return The saved entity; never {@code null}.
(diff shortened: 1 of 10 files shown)


api/src/main/java/jakarta/data/repository/BasicRepository.java
```diff
@@ -131,25 +131,25 @@ public interface BasicRepository<T, K> extends DataRepository<T, K> {
     /**
      * Retrieves an entity by its Id.
      *
-     * @param id must not be {@literal null}.
-     * @return the entity with the given Id or {@literal Optional#empty()} if none found.
+     * @param id must not be {@code null}.
+     * @return the entity with the given Id or {@code Optional#empty()} if none found.
```
njr-11, line 135: Thus one was probably intended to be `@link`,
```suggestion
     * @return the entity with the given Id or {@link Optional#empty()} if none is found.
```

njr-11 (approved):
Changes look good.  I added one suggestion where I suspect that `@link` was intended rather than `@literal`.

Regarding your question,

> There are also quite a few uses of &#64;, &lt;, &gt;, and just `<code>` that could be updated to use `@code` as well. I think this could improve readability in the javadocs, particularly cases like module-info.java, but it would be a large effort, so I would like to know what other people think before attempting it.

I'd be fine with it, but it might not be worth the effort given that the readability improvement would be limited to looking at the source code.  The JavaDoc for the end user should render the same either way.

PR #368, opened 2023-11-20
Entity Property Names section belongs under Entity Classes, not Query by Method Name

njr-11:
The section on Entity Property Names applies across the spec and is not specific to Query by Method Name, so it should be moved from that section to the Entity Classes section.  Also, this section has implementation detail mixed in, which I'll replace with stating what the requirements are.


otaviojava (approved):

mswatosh (commented):
LGTM

Issue #353, opened 2023-11-14
Review the Spec for conciseness


mswatosh:
With my last build, the Spec is currently at 47 pages! It's amazing how far the spec has come over the last year, and given that there are more ideas being planned for the next version of Jakarta Data, the spec will only continue to grow. Given that, I think it's important that we keep the spec concise where we can, so it remains manageable for future versions.

- [ ] Section 1
- [ ] Section 2
- [ ] Section 3
- [ ] Section 4
- [ ] Section 5

otaviojava:
@mswatosh can we close this one?

mswatosh:
Not yet, sorry I haven't had a chance to review this. I'll try to get to it soon

gavinking:
@mswatosh I happen to agree that there's currently a fair bit of repetition in the spec, including quite a lot of throat-clearing at the beginning.

Overall the spec is now quite well-written, so this doesn't bother me as much as it did before, and I don't think it's a critical issue, but, that said:

- I wonder if chapters 1 and 2 could be merged into a single chapter, and possibly condensed a little.
- I wonder if chapters 5, 6, and 7, being very short, could be reorganized as a single chapter (or possibly as two chapters), and reviewed for repetition.

WDYT?

KyleAure:
Closing issue as completed for the Jakarta Data 1.0 milestone

PR #372, opened 2023-11-22
preliminary review for conciseness

njr-11:
Read over the spec at a high level and identify some of the more obvious places where it could be more concise.  This is being done for #353


mswatosh (commented):
Good improvements, lgtm

otaviojava (approved):

PR #383, opened 2023-11-29
Update committer list

njr-11:
Update the committer list in the spec doc according to the successful committer votes that just went through.  Note that the links won't resolve yet, but are very predictable and it is better to have the accurate list of names shown.  I also corrected the link to one of the exists names which had a copy/paste error that was pointing to a different person.


KyleAure (approved):
Looks good

PR #394, opened 2023-12-11
reintroduce static metamodel?

njr-11:
This is where we left off at on the static metamodel.  At one point, it was removed and deferred to next release, which was supposed to have been with the approval of @gavinking who had opened the original issue for it: #230  but based on recent comments in that issue I suspect that was missed and didn't happen.

If we do decide to bring this back, this pull includes all of the latest that we had for static metamodel, including the working TCK tests.  If I recall correctly, one of the items that remained under discussion was whether we should keep the pattern where the metamodel fields are static (as in this pull) versus switch to one where the fields are volatile.  The former is safer, but required additional code in the spec to facilitate it.  This was discussed in a meeting with Gavin on January 3, 2024 and it was decided to go with the non-final/volatile approach for metamodel attributes that is consistent with Jakarta Persistence.  Also discussed during this meeting - to avoid conflicts on generated class names with Jakarta Persistence, the convention was switched to place the underscore first, and also bring back the ability to have String fields on the metamodel for name values that can be used within annotations.

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/Attribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/Attribute.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.metamodel;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+import jakarta.data.Sort;
+import jakarta.data.exceptions.MappingException;
+
+/**
+ * Represents an entity attribute in the {@link StaticMetamodel}.
+ */
+public interface Attribute {
+    /**
+     * Obtain the entity attribute name, suitable for use wherever the specification requires
+     * an entity attribute name. For example, as the parameter to {@link Sort#asc(String)}.
+     *
+     * @return the entity attribute name.
+     * @throws MappingException if an entity attribute with this name does not exist or
+     *                          if no Jakarta Data provider provides a repository for the entity type.
+     */
+    String name();
+
+    /**
+     * <p>Used by the Jakarta Data provider to initialize this {@code Attribute} with implementation.
+     * When initializing a subtype of {@code Attribute} with this method, the Jakarta Data provider
+     * must supply an instance of the subtype.</p>
+     *
+     * <p>The default value of <code>false</code> is appropriate for instances that are
+     * pre-initialized, such as those that are provided by a Jakarta Data provider
+     * or are generated in advance with an annotation processor.</p>
+     *
+     * @param attr attribute implementation provided by the Jakarta Data provider.
+     * @return true if initialization was successful; false if already initialized.
+     */
+    default boolean init(Attribute attr) {
+        return false;
+    }
+
+    /**
+     * <p>Obtains a new instance for the Jakarta Data provider to initialize.</p>
+     *
+     * <p>The Jakarta Data provider automatically initializes the instance when
+     * used as the value of a {@code public}, {@code static}, {@code final} field
+     * of a class or interface that is annotated with {@link StaticMetamodel},
+     * where the field name matches the name of an entity attribute.
+     * For example, a {@code Person} entity with attributes {@code id}, {@code emails},
+     * {@code phoneNumbers}, and {@code name} could have the following static metamodel,</p>
+     *
+     * <pre>
+     * &#64;StaticMetamodel(Person.class)
+     * public interface Person {
+     *     SortableAttribute id = SortableAttribute.get();
+     *     Attribute emails = Attribute.get();
+     *     Attribute phoneNumbers = Attribute.get();
+     *     TextAttribute name = TextAttribute.get();
+     * }
+     * </pre>
+     *
+     * @return a new instance for an entity attribute.
+     */
+    public static Attribute get() {
+        return new Attribute() {
+            // This allows the Jakarta Data provider to initialize the static final
+            // Attribute instance with the provider's own implementation of it.
+            private final AtomicReference<Attribute> impl = new AtomicReference<Attribute>();
+
+            private final Attribute attr() {
+                Attribute attr = impl.get();
+                if (attr == null) {
+                    throw new MappingException("The static metamodel for this attribute has not been initialized by a Jakarta Data provider.");
+                }
+                return attr;
+
+            }
+
+            @Override
+            public boolean init(Attribute attr) {
+                if (attr == null) {
+                    throw new NullPointerException();
+                } else {
+                    return impl.compareAndSet(null, attr);
+                }
+            }
+
+            @Override
+            public String name() {
+                return attr().name();
+            }
+
+            @Override
+            public String toString() {
+                Attribute attr = impl.get();
+                String attrName = attr == null ? "[uninitialized]" : attr.name();
+                String className = getClass().getSimpleName();
+                return new StringBuilder(className.length() + 1 + attrName.length())
+                        .append(className).append(':').append(attrName)
+                        .toString();
+            }
+        };
+    }
+}
--- a/api/src/main/java/jakarta/data/metamodel/Attribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/Attribute.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,10 +17,7 @@
  */
 package jakarta.data.metamodel;
 
-import java.util.concurrent.atomic.AtomicReference;
-
 import jakarta.data.Sort;
-import jakarta.data.exceptions.MappingException;
 
 /**
  * Represents an entity attribute in the {@link StaticMetamodel}.
@@ -31,87 +28,6 @@ public interface Attribute {
      * an entity attribute name. For example, as the parameter to {@link Sort#asc(String)}.
      *
      * @return the entity attribute name.
-     * @throws MappingException if an entity attribute with this name does not exist or
-     *                          if no Jakarta Data provider provides a repository for the entity type.
      */
     String name();
-
-    /**
-     * <p>Used by the Jakarta Data provider to initialize this {@code Attribute} with implementation.
-     * When initializing a subtype of {@code Attribute} with this method, the Jakarta Data provider
-     * must supply an instance of the subtype.</p>
-     *
-     * <p>The default value of <code>false</code> is appropriate for instances that are
-     * pre-initialized, such as those that are provided by a Jakarta Data provider
-     * or are generated in advance with an annotation processor.</p>
-     *
-     * @param attr attribute implementation provided by the Jakarta Data provider.
-     * @return true if initialization was successful; false if already initialized.
-     */
-    default boolean init(Attribute attr) {
-        return false;
-    }
-
-    /**
-     * <p>Obtains a new instance for the Jakarta Data provider to initialize.</p>
-     *
-     * <p>The Jakarta Data provider automatically initializes the instance when
-     * used as the value of a {@code public}, {@code static}, {@code final} field
-     * of a class or interface that is annotated with {@link StaticMetamodel},
-     * where the field name matches the name of an entity attribute.
-     * For example, a {@code Person} entity with attributes {@code id}, {@code emails},
-     * {@code phoneNumbers}, and {@code name} could have the following static metamodel,</p>
-     *
-     * <pre>
-     * &#64;StaticMetamodel(Person.class)
-     * public interface Person {
-     *     SortableAttribute id = SortableAttribute.get();
-     *     Attribute emails = Attribute.get();
-     *     Attribute phoneNumbers = Attribute.get();
-     *     TextAttribute name = TextAttribute.get();
-     * }
-     * </pre>
-     *
-     * @return a new instance for an entity attribute.
-     */
-    public static Attribute get() {
-        return new Attribute() {
-            // This allows the Jakarta Data provider to initialize the static final
-            // Attribute instance with the provider's own implementation of it.
-            private final AtomicReference<Attribute> impl = new AtomicReference<Attribute>();
-
-            private final Attribute attr() {
-                Attribute attr = impl.get();
-                if (attr == null) {
-                    throw new MappingException("The static metamodel for this attribute has not been initialized by a Jakarta Data provider.");
-                }
-                return attr;
-
-            }
-
-            @Override
-            public boolean init(Attribute attr) {
-                if (attr == null) {
-                    throw new NullPointerException();
-                } else {
-                    return impl.compareAndSet(null, attr);
-                }
-            }
-
-            @Override
-            public String name() {
-                return attr().name();
-            }
-
-            @Override
-            public String toString() {
-                Attribute attr = impl.get();
-                String attrName = attr == null ? "[uninitialized]" : attr.name();
-                String className = getClass().getSimpleName();
-                return new StringBuilder(className.length() + 1 + attrName.length())
-                        .append(className).append(':').append(attrName)
-                        .toString();
-            }
-        };
-    }
 }
--- a/api/src/main/java/jakarta/data/metamodel/SortableAttribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/SortableAttribute.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,10 +17,7 @@
  */
 package jakarta.data.metamodel;
 
-import java.util.concurrent.atomic.AtomicReference;
-
 import jakarta.data.Sort;
-import jakarta.data.exceptions.MappingException;
 
 /**
  * Represents a sortable entity attribute in the {@link StaticMetamodel}.
@@ -40,8 +37,6 @@ public interface SortableAttribute extends Attribute {
      * Obtain a request for an ascending {@link Sort} based on the entity attribute.
      *
      * @return a request for an ascending sort on the entity attribute.
-     * @throws MappingException if an entity attribute with this name does not exist or
-     *                          if no Jakarta Data provider provides a repository for the entity type.
      */
     Sort asc();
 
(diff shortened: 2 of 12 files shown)


gavinking:
> If I recall correctly, one of the items that remained under discussion was whether we should keep the pattern where the metamodel fields are static (as in this pull) versus switch to one where the fields are volatile.

@njr-11 I would love it if you guys could find time to talk me through this. In JPA it's totally the responsibility of the provider to initialize fields of the metamodel, and so the metamodel interfaces are just interfaces. But it looks like you're proposing something a bit different here. That's fine, but I would like to understand how it impacts existing metamodel generators and interop with JPA.

njr-11:
> I would love it if you guys could find time to talk me through this. In JPA it's totally the responsibility of the provider to initialize fields of the metamodel, and so the metamodel interfaces are just interfaces. But it looks like you're proposing something a bit different here. That's fine, but I would like to understand how it impacts existing metamodel generators and interop with JPA.

Relying on the provider was the original plan, and is still an option.  In this case the provider needs to initialize it if it wasn't already initialized (indicated by the Generated annotation).  I believe Otavio was thinking that some IDEs would offer a way to define and generate the fully implemented metamodel during development time.

@otaviojava and @gavinking Would it work to move the Jakarta Data meeting from Wednesday, Dec 20 to be at the same time this week on Wednesday, Dec 13?  We could use it to sort out what should happen with the static metamodel and whether some of it should be included in version 1.0.

gavinking:
> In this case the provider needs to initialize it if it wasn't already initialized

Yeah right that's the way it works in JPA.

> I believe Otavio was thinking that some IDEs would offer a way to define and generate the fully implemented metamodel during development time.

Yeah so I would like to understand the thinking there.

> Would it work to move the Jakarta Data meeting from Wednesday, Dec 20 to be at the same time this week on Wednesday, Dec 13?

I can make it, the only thing is I have a hard cutoff at the 55 minute mark.

njr-11:
> > Would it work to move the Jakarta Data meeting from Wednesday, Dec 20 to be at the same time this week on Wednesday, Dec 13?
> 
> I can make it, the only thing is I have a hard cutoff at the 55 minute mark.

@otaviojava If this works for you as well (it does for Gavin and myself), go ahead and move the Jakarta Data meeting to Dec 13 (I'll make sure to log on early enough to check if I need to start it up) for discussion.  If not, we can aim to discuss at the scheduled meeting on the following week.

otaviojava:
> > > Would it work to move the Jakarta Data meeting from Wednesday, Dec 20 to be at the same time this week on Wednesday, Dec 13?
> > 
> > 
> > I can make it, the only thing is I have a hard cutoff at the 55 minute mark.
> 
> @otaviojava If this works for you as well (it does for Gavin and myself), go ahead and move the Jakarta Data meeting to Dec 13 (I'll make sure to log on early enough to check if I need to start it up) for discussion. If not, we can aim to discuss at the scheduled meeting on the following week.

Sorry, I saw the message now.

otaviojava:
Another point, IMHO: a good point to start should be the Jakarta Persistence API.
Basically, copy/paste what they have and start from there.

But, not to this version, to the following version.

njr-11:
> Sorry, I saw the message now.
> 
> Did you talk?

Otavio - we did not have any meeting to discuss this last week because you did not respond beforehand and you need to be there to work out with Gavin what is acceptable and which release it will be in. We can discuss it this week instead.  The update that I just pushed is only a rebase onto latest, not any additional changes.

njr-11:
> Folks, do you think we could set up a time to talk through this after the xmas period?

There is a Jakarta Data meeting scheduled for Wednesday, January 6, where we would discuss this if both of you are available for it.

njr-11:
The latest updates to the proposed metamodel reflect changes that were discussed at the January 3 meeting.

api/src/main/java/jakarta/data/metamodel/package-info.java
```diff
@@ -0,0 +1,60 @@
...
+ * public class _Product {
+ *     SortableAttribute id = SortableAttribute.get();
+ *     TextAttribute name = TextAttribute.get();
+ *     SortableAttribute price = SortableAttribute.get();
+ * }
+ *
+ * ...
+ *
+ * &#64;Repository
+ * Product products;
```
otaviojava, line 46:
```suggestion
 * Products products;
```

njr-11: Thanks for catching this! Both you and Kyle spotted it.

njr-11:
@otaviojava and @KyleAure This pull is updated with the review comment that you spotted as well as the JavaDoc correction that we noted on the Jakarta Data meeting earlier today.

KyleAure (approved):

njr-11:
@otaviojava could you complete your review of this so that we can merge it to enable Kyle to move forward with creating Milestone 3?

otaviojava:
> @otaviojava could you complete your review of this so that we can merge it to enable Kyle to move forward with creating Milestone 3?

As I said, I will need more time to test it deeply.
I am ok to move it forward, remembering our agreement:

- Move this static metamodel forward.
- The spec also allows all the info to be generated at build time.
- We agree to be open to breaking compatibility in the next version if needed.

api/src/main/java/jakarta/data/metamodel/SortableAttribute.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * <ul>
+ * <li>numeric attributes</li>
+ * <li>enum attributes</li>
+ * <li>time attributes</li>
+ * <li>boolean attributes</li>
+ * <li>{@link TextAttribute textual attributes}</li>
+ * </ul>
+ */
+public interface SortableAttribute extends Attribute {
```
otaviojava, line 34: IMHO: by default, all attributes are sortable because the database defines this aspect, not the Java programming.

I mean, I don't have any scenario where the Java side will define if the attribute is sortable or not.

api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
```diff
@@ -0,0 +1,128 @@
...
+ *     public String first;
+ *     public String last;
+ * }
+ * </pre>
+ *
+ * <p>You can define a static metamodel as follows,</p>
+ *
+ * <pre>
+ * &#64;StaticMetamodel(Person.class)
+ * public class _Person {
```
otaviojava, line 58: I know that `Person_`is already taken from Jakarta Persistence.

Can we use another suffix instead of having a prefix?

I mean, for us, is more natural, using IDE, explore suffix at the code, for example, given a Person in a code we might have:

- PersonEntity
- PersonService
- PersonRepository
- PersonController


In general, by conversion, we work mostly with suffixes instead of prefixes. Some suggestions:

- Person_Metadata
- Person__
- PersonMetadata

api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -0,0 +1,33 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.metamodel;
+
+import jakarta.data.Sort;
+
+/**
+ * Represents an entity attribute in the {@link StaticMetamodel}.
+ */
+public interface Attribute {
```
otaviojava, line 25: I would merge this interface with `SortableAttribute`, I am assuming that the minimum of an attribute is being sortable.

api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
```diff
@@ -0,0 +1,128 @@
...
+ *     // These can be uninitialized and non-final if you don't need to access them from annotations.
+ *     public static final String SSN = "ssn";
+ *     public static final String NAME = "name";
+ *     public static final String NAME_FIRST = "name.first";
```
otaviojava, line 62: What happens with Entity with relationship?

For example,

```java
class Person {

private Address address;
}

class Address {
private String street;
}
```
Thus, the metadata:

```java

class PersonMetadata {
public static final String Address_STREET = "address.street";
}

class AddressMetadata {
public static final String STREET = "name";
}
```

Please ignore the terminology. My point is. This constant might have a substantial recursive constant.

I would skip this constant.

For example, Jakarta Persistence does not have this:

https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0#staticmetamodel-annotation

api/src/main/java/jakarta/data/metamodel/TextAttribute.java
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.metamodel;
+
+import jakarta.data.Sort;
+
+/**
+ * Represents an textual entity attribute in the {@link StaticMetamodel}.
+ */
+public interface TextAttribute extends SortableAttribute {
```
otaviojava, line 25: I would remove this interface and keep the static metamodel as simple as possible.

api/src/main/java/jakarta/data/metamodel/SortableAttribute.java
gavinking: Hrrrm. Good point. That's probably right. I mean, I bet we could find some sort of data store where binary data is not sortable, but I guess we need to model that in the API.

api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
gavinking: My motivation for suggesting the prefix `_` was that I figured it was actually more ergonomic in the IDE: i.e. it's nicer to type `_Bo<ctrl-space>` than `Bo<ctrl-space>_`.

If you guys really hate the prefix underscore, I would suggest something like `Entity_Person`.

njr-11: > My motivation for suggesting the prefix `_` was that I figured it was actually more ergonomic in the IDE: i.e. it's nicer to type `_Bo<ctrl-space>` than `Bo<ctrl-space>_`.
> 
> If you guys really hate the prefix underscore, I would suggest something like `Entity_Person`.

My own preference is that I liked the single underscore because it also makes the name as concise as possible, making the code easier to read and understand,

```
page = people.findByAgeBetween(20, 65, Pageable.ofSize(10).sortBy(_Person.lastName.asc(),
                                                                  _Person.firstName.asc(),
                                                                  _Person.id.asc());
```

rather than

```
page = people.findByAgeBetween(20, 65, Pageable.ofSize(10).sortBy(Person_Metadata.lastName.asc(),
                                                                  Person_Metadata.firstName.asc(),
                                                                  Person_Metadata.id.asc());
```

api/src/main/java/jakarta/data/metamodel/Attribute.java
gavinking: > Similarly, if an attribute is an embeddable or collection type, it might not make sense to be able to sort it either.

Ah yeah, of course. Forgot about that.

api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
gavinking: > What happens with Entity with relationship?

An entity only has these fields for its _direct_ members. So no `HEADQUARTER_STREET`, etc.

[Note that I guess we didn't say anything about embeddables yet, but in JPA embeddables have their own static metamodel types.]

> Jakarta Persistence does not have this

It's coming in JPA 3.2

njr-11: Your example mostly looks good but has an error/typo here:

> ```java
> class AddressMetadata {
> public static final String STREET = "name";
> }
> ```

The value should be "street".

> Please ignore the terminology. My point is. This constant might have a substantial recursive constant.
> As much as we have a relationship, it might grow. E.g.: `person.address.country.city.street`
>
> I would skip this constant.

Your particular hierarchy doesn't make any sense. You wouldn't find a city embeddable inside of the country or a street field on a city.  Let's use `person.address.city.state.country` as a better example.
If the user defines their entities and embeddable objects that way, it certainly could have a lengthy relationship. But that isn't a problem of the static metamodel. The Jakarta Data specification already defines this behavior with embeddables.  A user can already do this if they want to:

```
people.findByYearOfBirth(int birthYear, Sort.asc("person.address.city.state.country"), ...);
```

The only part the static metamodel is adding is allowing them to do it in a type safe way.

The example with that many levels of embeddables isn't as realistic. However, one or two levels of embeddables is something users will want to do.  This is something we support in the spec, and I don't see why we should skip over having a type safe solution for it like we do for other usage.

> For example, Jakarta Persistence does not have this

That's fine. We don't need to match Persistence for the sake of copying. We should include it because referencing embeddables in this way is part of the Jakarta Data programming model and we should make our whole programming model type safe.

api/src/main/java/jakarta/data/metamodel/TextAttribute.java
njr-11: > I would remove this interface and keep the static metamodel as simple as possible.

If we consolidate it all onto SortableAttribute, then anything that is a SortableAttribute will be able to request  .ascIgnoreCase() and .descIgnoreCase() which means letting users write nonsense code for numeric fields.  By having these specific types, it makes a hierarchy of interfaces that control what you can actually do with the attributes based on what types they are.  That is a superior model because it stops errors from being introduced while code is being developed.

api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
otaviojava: 🤔 I did not know that. Thank you for explaining.
If it is more ergonomic to the IDE, let's keep this way.

api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
gavinking: For the record I'm not sure this is necessary:

    public static final String NAME_FIRST = "name.first";

But I need to think it through some more.

otaviojava: > The example with that many levels of embeddables isn't as realistic.

Unfortunately, it is more natural than we wish. But let me create a real scenario here, and I will simplify it.

Unfortunately, it happens more than I wish, for example, in e-commerce:

```java
class Customer {
    private long id;
    private Address billingAddress;
    private List<Order> orders;
}

class Order {
    private long id;
    private Customer customer;
    private List<Product> products;
    private Address shipping;
    private Payment payment;

}

class Address {

}

class Payment {
Address address;
}
```


So, there are two different paths for Addresses, where to client this is for billing and the order the address is to shipping the product:
* `billingAddress.street`
* `shipping.street`
My point is only the static and how to solve this point.


⚠️ We have an even worse scenario in finance, especially with credit-card providers integration.

njr-11: > So, there are two different paths for Addresses, where to client this is for billing and the order the address is to shipping the product:
> 
>     * `billingAddress.street`
> 
>     * `shipping.street`
>       My point is only the static and how to solve this point.

Otavio, yes, that's an excellent scenario. But why are you trying to use that scenario as an argument against the spec updates in this pull?  This pull provides the type-safe solution to your scenario.  If you follow what is in this pull and apply it to your example scenario, everything is very straightforward,

```
@StaticMetamodel(Customer.class)
class _Customer {
    public static final String ID = "id";
    public static final String BILLINGADDRESS = "billingAddress";
    public static final String BILLINGADDRESS_STREET = "billingAddress.street";
    public static final String ORDERS = "orders";
}

@StaticMetamodel(Order.class)
class _Order {
    public static final String ID = "id";
    public static final String CUSTOMER = "customer";
    public static final String CUSTOMER_BILLINGADDRESS = "customer.billingAddress";
    public static final String CUSTOMER_BILLINGADDRESS_STREET = "customer.billingAddress.street";
    public static final String PRODUCTS = "products";
    public static final String SHIPPING = "shipping";
    public static final String SHIPPING_STREET = "shipping.street";
    public static final String PAYMENT = "payment";
    public static final String PAYMENT_ADDRESS = "payment.address";
    public static final String PAYMENT_ADDRESS_STREET = "payment.address.street";
}

@StaticMetamodel(Address.class)
class _Address {
    public static final String STREET = "street";
}

@StaticMetamodel(Payment.class)
class _Payment {
    public static final String ADDRESS = "address";
    public static final String ADDRESS_STREET = "address.street";
}
```

Example usage:

```
orders.findAll(Sort.asc(_Order.CUSTOMER_BILLINGADDRESS_STREET));
orders.findAll(Sort.asc(_Order.SHIPPING_STREET));
orders.findAll(Sort.asc(_Order.PAYMENT_ADDRESS_STREET));
```

The above usage equates to what you could already write as strings without the static metamodel,

```
orders.findAll(Sort.asc("customer.billingAddress.street"));
orders.findAll(Sort.asc("shipping.street"));
orders.findAll(Sort.asc("payment.address.street"));
```

If the relationship was embeddables, this will all work cleanly. If it is some other sort of relationship that isn't required by Jakarta Data, then it's up to the provider what will happen.  This pull doesn't change any of that.  The only part added in this pull is the ability to use type safe values in place of the strings.

That said, if for some reason everyone else is against specifying and standardizing how to define the type constants and attribute fields for embeddable attributes, we could certainly leave that part of the static metamodel for now.

gavinking: OK, so now I see the thing I missed.

In Hibernate, the equivalent type to `Sort` is parameterized by the entity type. So it's `Sort<Customer>` which allows a query method to accept only the right kind of `Sort` objects.

We don't have that here yet. since `Sort` is not declared as a parameterized type.

So what I would like us to work towards is:

```java
repository.allCustomers(_Customer.name.asc())
```

which we already have, but I would like it to be more typesafe, and:

```java
repository.allCustomers(_Address.name.of(_Customer.billingAddress).asc())
```

or something like that ... just trying things out here.

@njr-11 @otaviojava Do you follow what I'm trying to get at here?

gavinking: i.e you would declare things like:

```java
Employee[] findByYearHired(int yearYired, Limit maxResults, Sort<Employee>... sortBy);
```

njr-11: > i.e you would declare things like:
> 
> ```java
> Employee[] findByYearHired(int yearYired, Limit maxResults, Sort<Employee>... sortBy);
> ```
This is nice because during development time it helps prevent the error of accidentally using a Sort that was intended for a different entity type.

> ```java
> repository.allCustomers(_Address.name.of(_Customer.billingAddress).asc())
> ```

That's a good example of where a parameterized sort would be necessary to disambiguate which entity the query pertains to, which is another scenario where the parameterized sort would be useful.  Let's open a separate issue for this idea because it applies regardless of whether we have the static metamodel.

I should have written up my earlier example with both the String constants and the Attribute fields because the latter should already cover the type-safe sort objects:

```
repository.allCustomers(_Customer.billingAddress_street.asc())
```

Relying on the following metamodel classes...

```
@StaticMetamodel(Address.class)
class _Address {
    public static final String STREET = "street";

    public static volatile SortableAttribute street;
}

@StaticMetamodel(Customer.class)
class _Customer {
    public static final String ID = "id";
    public static final String BILLINGADDRESS = "billingAddress";
    public static final String BILLINGADDRESS_STREET = "billingAddress.street";

    public static volatile SortableAttribute id;
    public static volatile Attribute billingAddress;
    public static volatile SortableAttribute billingAddress_street;
}
```

It should be noted that if we add parameterized sort, then SortableAttribute would likely be parameterized as well,
```
    public static volatile SortableAttribute<Customer> id;
```
so that _Customer.id.asc() will return a Sort<Customer> that forces type-safe usage.

otaviojava:
it is became huge.
So I will merge this one and create threads with scenarios.

api/src/main/java/jakarta/data/metamodel/Attribute.java
otaviojava: > > I would merge this interface with `SortableAttribute`, I am assuming that the minimum of an attribute is being sortable.
> 
> We should not assume that every attribute is sortable. For example, the Jakarta Data specification lists `byte[]` as a data type that represents binary data. I would not expect that to be sortable. Similarly, if an attribute is an embeddable or collection type, it might not make sense to be able to sort it either. Having a interface to designate non-sortable types helps prevent the user during development time from introducing errors of requesting sort on things that cannot be sorted.

@njr-11 My point is that the database will define what is sortable or not, non the spec or the Java App.

api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
otaviojava: I am fine to merge the PR, but we need to cover all scenarios once we include this metadata, furthermore, it should cover both SQL and NoSQL databases. Including the feature is the first step; make sure it works is the second step.

otaviojava: About the path, I would go to the same path that we do in code as metadata, thus:

`_Customer. _Address.street.asc()`

gavinking: Yeah, I guess that's about right @otaviojava .... except probably this: `_Customer.address.street.asc()`

api/src/main/java/jakarta/data/metamodel/Attribute.java
njr-11: > @njr-11 My point is that the database will define what is sortable or not, non the spec or the Java App.

I agree. The database makes the decision. The Java App must be aware of that -- what is and is not sortable -- in order to write valid code asking to sort on some attributes and not others.  If the Java App tries to declare something sortable that cannot be sorted, that should result in an error.  Maybe we should write that somewhere in the spec.

api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
njr-11: Excellent - I think that's exactly how we defined it, so we just need to confirm that what we wrote is sufficiently clear.

njr-11: Disregard my comment above about that exactly matching how we defined it.  I misread it as `_Customer.address_street.asc()` (how we defined it), but the difference is in the `.` vs `_` between `address` and `street`.  It does look nicer the way you wrote it.  However, for that to work, the `_Customer.address` Attribute would also need to also be a static metamodel class, so it would need to have a type that is both `Attribute` and `_Address`.

```java
@StaticMetamodel(Customer.class)
public class _Customer {
    public static final String ID = "id";
    public static final String BILLINGADDRESS = "billingAddress";
    public static final String BILLINGADDRESS_STREET = "billingAddress.street";

    public static volatile SortableAttribute id;
    public static volatile _Address billingAddress; // <-- Both Attribute and _Address
}

@StaticMetamodel(Address.class)
public class _Address implements Attribute { // <-- added implements
    public static final String STREET = "street";

    public static volatile SortableAttribute street;

    @Override
    public String name() { // <-- who provides/implements the Attribute methods?
}
```

This approach seems to make things more complex and has some issues that would need to be worked through. It might be possible for the metamodel class for the embeddable to become abstract and then avoid declaring the Attribute methods, but ultimately the method does need to be implemented, and wherever that happens will run into the problem that the name() is different depending on which entity embeds it, because they can each name the Address field whatever they like.

PR #411, opened 2024-01-04
Fix TCK test errors

KyleAure:
While running the TCK I found two errors that need to be fixed:

- The EntityTest application did not include the correct classes to run.
- The validation test repository used invalid query methods.


njr-11 (approved):
Thanks for catching these errors!

Issue #412, opened 2024-01-04
[clarification]: Should non-standard query by methods be supported by user defined repositories


KyleAure:
### Specification

https://github.com/jakartaee/data/blob/e7dbeab00950ed6bd14d4d0cd90df595ef9c908e/api/src/main/java/module-info.java#L245-L248

### Problem

While writing tests for Validation in the TCK I copied methods from the BasicRepsoitory interface and added validation annotations.  Later while running the TCK I ran into issues because of the non-standard query by methods that exist in the BasicRepository.

The BasicRepsoitory has the following methods that are presumably query by method:
https://github.com/jakartaee/data/blob/e7dbeab00950ed6bd14d4d0cd90df595ef9c908e/api/src/main/java/jakarta/data/repository/BasicRepository.java#L178
https://github.com/jakartaee/data/blob/e7dbeab00950ed6bd14d4d0cd90df595ef9c908e/api/src/main/java/jakarta/data/repository/BasicRepository.java#L231

I might write my own Repository interface which has:
```java
public interface MyRepository<MyEntity, Long> extends DataRepository<MyEntity, Long> {
  long count();
}
```

### Ambiguity

Currently the javadoc / spec fails to define behavior around this: 

If a custom repository is written with the method `count()` should the implementation respect this method because it also exists within the built in repositories, or should a DataException be thrown to alert the user that their query is invalid?

### Additional information

I can think of three solutions here: 

### Documentation

- Update documentation to define behavior when these methods are used in custom repositories
- Update the javadoc for these methods to point out that they are special uses of invalid syntax that need to be supported by implementations.

### Use proper syntax

- Update these methods to use proper syntax:

```java
long countBy();
void deleteAllBy();
```

### Swap to query by parameter

- Create `@Count` annotation
- Modify `@Delete` annotation to accept no parameters
- Update these methods to use these annotations:

```java
@Count
long count();

@Delete
void deleteAll();
```

njr-11:
It's a good point that `count()` and `deleteAll()` from `BasicRepository` don't follow the specification's own rules per the BNF for Query by Method Name, and thus when users copy them to their own repositories that don't inherit from BasicRepository, they won't behave the same as they would if the repository had inherited from it.

njr-11:
A simple way to fix half of the problem (the `deleteAll()` part) is covered by #413
The `count()` method will need a different solution.

otaviojava:
IMHO: those are methods from the interface.
Checking the spec: 

[...]

PR #414, opened 2024-01-05
[Fix]Should non-standard query by methods be supported by user defined repositories

otaviojava:
Changes:

- https://github.com/jakartaee/data/issues/412


njr-11 (approved):

otaviojava:
> This looks good. We'll also need to update a few TCK tests to match, but that can be a separate pull.

Let me try to fix it at the same PR.

njr-11 (approved):
TCK tests are matching now. Thanks!

PR #423, opened 2024-01-16
Explicit mention of persistence unit is missing from javadoc

njr-11:
Explicitly state in the JavaDoc that the dataStore value can be a persistence unit reference when using JPA rather than expecting users to assume this is the case.


otaviojava (approved):

gavinking:
Well, I guess you could say I only need to go down the JNDI route if the name begins with `java:`. And so I would generate different code depending on the format of the string value.

PR #439, opened 2024-02-07
Experiment with Sort with type parameter [Solution 2]

otaviojava:
# Changes

The goal is to create a Parameterized Sort that keeps what we currently have. Thus, update sort as an interface and share it for both: 

- Sort
- ParamitizedSort

```mermaid
classDiagram
    Sort <|-- DefaultSort
    Sort <|-- ParameterizedSort
  
````

⚠️ I did not update the TCK; my point here is: We can create new capabilities without breaking the core API all the time.

Reference: https://github.com/jakartaee/data/pull/432

### DIFF
--- a/api/src/main/java/jakarta/data/DefaultSort.java
+++ b/api/src/main/java/jakarta/data/DefaultSort.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2022,2023 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by njr-11** [comment]
> I did not update the TCK; my point here is: We can create new capabilities without breaking the core API all the time.

Not updating the TCK or showing other usage examples means that you didn't demonstrate the approach works.  It turns out it doesn't work because it doesn't achieve the goal of enabling a user to write a repository in a way that enforces type safety of Sort.

On example is that with this proposal, there is no way for the repository to enforce a type for the Sorts of a Pageable,

```java
Page<Employee> findByYearHired(int yearHired, Pageable pageRequest); 
```

allowing the user to mix up types without warning or error,

```java
page1 = employees.findByYearHired(2024, Pageable.ofSize(10).page(1).sortBy(
                  _Employee.lastName.asc(),
                  _Employer.name.asc(), // attribute from wrong entity used here!
                  _Employee.id.asc()));
```

otaviojava:
> Not updating the TCK or showing other usage examples means that you didn't demonstrate the approach works. It turns out it doesn't work because it doesn't achieve the goal of enabling a user to write a repository in a way that enforces type safety of Sort.

I had 20 mins this morning. Sorry for that. But I can update you in a few days.

> One example is that with this proposal, there is no way for the repository to enforce a type for the Sorts of a Pageable,

ah, I got your point; the goal is not to give another option but to give this as the unique option.

In this case, I would vote for -1.

PR #453, opened 2024-02-15
Fix small typo at repository spec at the UUID explanation

otaviojava:


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -244,7 +244,7 @@ Every Jakarta Data provider must support the following basic types within its pr
 | Represent date and time-related data.
 
 | `java.util.UUID`
-| Universally Unique IDentifier for identifying entities.
```
mswatosh, line 247: I think this is intentional, since it's for **UUID**: **U**niversally **U**nique **ID**entifier

otaviojava: Good point!

otaviojava: https://www.ietf.org/rfc/rfc4122.txt

PR #467, opened 2024-02-21
Remove BasicRepository.deleteAll()

njr-11:
It was discussed on today's Jakarta Data call to remove the built-in deleteAll method from BasicRepository.  Users could still write their own deleteAll if they wish, but this operation is seen as something that is not likely to be typically wanted.


gavinking:
😍

otaviojava (approved):

PR #468, opened 2024-02-21
rework section on provider/repository

gavinking:
Folks I would like to propose a bit of a rework of chapter 5.

WDYT? 

@njr-11 I believe you wrote the current version. Do you think this is an improvement? Does it miss anything important?

### DIFF
--- a/spec/src/main/asciidoc/jakarta-ee.adoc
+++ b/spec/src/main/asciidoc/jakarta-ee.adoc
@@ -4,6 +4,8 @@ A Jakarta Data provider might come as an integrated component of a Jakarta EE co
 For example, a Jakarta Data provider might use a CDI portable extension to integrate with dependency injection.
 The role of the Jakarta Data provider is to supply implementations of application-defined repositories.
 
+=== Providers and Repositories
+
 A Jakarta Data provider must supply an implementation of each repository interface which either:
 
 - explicitly specifies the name of the provider via the `provider` member of the `@Repository` annotation, or
--- a/spec/src/main/asciidoc/jakarta-ee.adoc
+++ b/spec/src/main/asciidoc/jakarta-ee.adoc
@@ -1,24 +1,56 @@
 == Jakarta Data Providers
 
-Jakarta Data providers implement repository interfaces, performing queries and other operations related to entities per the rules of the Jakarta Data specification. A Jakarta Data provider makes the repository implementation available to the application via dependency injection. The Jakarta Data specification defines the rules by which Jakarta Data providers must abide to ensure that multiple Jakarta Data providers are able to coexist in a system without interfering or overlapping on the same injection points.
+A Jakarta Data provider might come as an integrated component of a Jakarta EE container, or it might be a separate component that integrates with the Jakarta EE container via standard or proprietary SPIs.
+For example, a Jakarta Data provider might use a CDI portable extension to integrate with dependency injection.
+The role of the Jakarta Data provider is to supply implementations of application-defined repositories.
+
+A Jakarta Data provider must supply an implementation of each repository interface which either:
+
+- explicitly specifies the name of the provider via the `provider` member of the `@Repository` annotation, or
+- does not explicitly specify a `provider`, but whose entity classes are annotated with an entity-defining annotation type supported by the provider.
+
+The repository implementation and the Jakarta Data provider are responsible for mediating interaction between the application program and the datastore, often taking advantage of other Jakarta EE services such as datasource provisioning and transaction management.
+
+An application usually obtains a reference to an implementation of a repository via dependency injection.
+Therefore, the provider must make its repository implementations available to the bean container responsible for dependency injection.
+
+Multiple Jakarta Data providers may coexist within a single program and provide repository implementations.
+It is the responsibility of the application developer to ensure that each repository is eligible for implementation by exactly one provider, according to the rules specified below.
+
+[NOTE]
+====
+If the application developer fails to disambiguate the provider for every repository in the application, the behavior is undefined by this specification.
+====
 
 === Provider support for Entities
 
-A Jakarta Data provider supplies an implementation of repository interfaces in Jakarta Data for one or more types of entities. An entity refers to a class that represents objects in a storage engine, such as SQL or NoSQL databases.
+A Jakarta Data provider typically supports one entity-defining annotation type, but it may support multiple entity-defining annotation types.
+A provider may even have no entity-defining annotation and feature a programming model for entity classes where the entity classes are unannotated.
 
-The `jakarta.persistence.Entity` annotation from the Jakarta Persistence specification can be used by repository entity classes for Jakarta Data providers that are backed by a Jakarta Persistence provider. Other Jakarta Data providers must not support the `jakarta.persistence.Entity` annotation.
+In particular:
 
-The `jakarta.nosql.Entity` annotation from the Jakarta NoSQL specification can be used by repository entity classes for Jakarta Data providers that are backed by NoSQL databases. Other Jakarta Data providers must not support the `jakarta.nosql.Entity` annotation.
+- The `jakarta.persistence.Entity` annotation from the Jakarta Persistence specification is an entity-defining annotation for Jakarta Data providers backed by a Jakarta Persistence provider. Other Jakarta Data providers must not support the use of `jakarta.persistence.Entity` as an entity-defining annotation.
 
-Jakarta Data providers that define custom entity annotations must follow the convention that the class name of all supported entity annotation types ends with `Entity`. This enables Jakarta Data providers to identify if a repository entity class contains entity annotations from different Jakarta Data providers so that the corresponding `Repository` can be ignored by Jakarta Data providers that should not provide it.
+- The `jakarta.nosql.Entity` annotation from the Jakarta NoSQL specification is an entity-defining annotation for Jakarta Data providers backed by NoSQL databases. Other Jakarta Data providers must not support the use of `jakarta.nosql.Entity` as an entity-defining annotation.
 
-Jakarta Data providers must ignore all `Repository` annotations where annotations for the corresponding entity are available at run time and none of the entity annotations are supported by the Jakarta Data provider. Ignoring these `Repository` annotations allows other Jakarta Data providers to handle them.
+A Jakarta Data provider might define a custom entity-defining annotation.
+The name of such a custom entity-defining annotation type must end with `Entity`. This convention allows any Jakarta Data provider to identify when a repository is associated with entity classes declared using an unrecognized entity-defining annotation.
+
+A Jakarta Data provider must ignore every repository associated with entity classes declared using an entity-defining annotation which is available at runtime, but not recognized by the provider, allowing another Jakarta Data provider to supply the implementation.
+
+On the other hand, if an entity class is annotated with an entity-defining annotation which is _not_ available at runtime, the Jakarta Data provider must completely ignore the annotation, and treat the entity as if that entity-defining annotation were missing.
+
+[NOTE]
+====
+This allows an entity class to be compiled with multiple entity-defining annotations, each targeting a different Jakarta Data provider, and have the provider automatically disambiguated at runtime based on the availability of the entity-defining annotations.
+====
 
 === Provider Name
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -1,24 +1,42 @@
...
+A Jakarta Data provider might define a custom entity-defining annotation.
+The name of such a custom entity-defining annotation type must end with `Entity`. This convention allows any Jakarta Data provider to identify when a repository is associated with entity classes declared using an unrecognized entity-defining annotation.
 
-Jakarta Data providers must ignore all `Repository` annotations where annotations for the corresponding entity are available at run time and none of the entity annotations are supported by the Jakarta Data provider. Ignoring these `Repository` annotations allows other Jakarta Data providers to handle them.
+A Jakarta Data provider must ignore every repository associated with entity classes declared using an unrecognized entity-defining annotation, allowing another Jakarta Data provider to supply the implementation.
```
njr-11, line 32: There are some subtle, and probably unintentional, changes in meaning here.  The previous language allows for the scenario where an entity class is compiled with multiple entity-defining annotations but only one is available at run time (The TCK actually does this so it can run against either type of provider).  The previous language also allows for a provider that supports a simpler model with unannotated entities.  I believe we can adjust your rewrite as follows to preserve the meaning:
```suggestion
A Jakarta Data provider must ignore every repository associated with entity classes where entity-defining annotations are available at run time but none are recognized by the provider, allowing another Jakarta Data provider to supply the implementation.
```

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -1,24 +1,42 @@
...
+Therefore, the provider must make its repository implementations available to the bean container responsible for dependency injection.
+
+Multiple Jakarta Data provider may coexist within a single program, and provide repository implementations.
+It is the responsibility of the application developer to ensure that each repository is eligible for implementation by exactly one provider, according to the rules specified below.
 
 === Provider support for Entities
 
-A Jakarta Data provider supplies an implementation of repository interfaces in Jakarta Data for one or more types of entities. An entity refers to a class that represents objects in a storage engine, such as SQL or NoSQL databases.
+A Jakarta Data provider supports one or more entity-defining annotation types.
+For example:
```
njr-11, line 23: The language "For example" implies the following paragraphs are examples and not requirements.
```suggestion
In the case of Jakarta Data providers that are built on Jakarta Persistence and Jakarta NoSQL:
```

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -1,24 +1,42 @@
...
+An application usually obtains a reference to an implementation of a repository via dependency injection.
+Therefore, the provider must make its repository implementations available to the bean container responsible for dependency injection.
+
+Multiple Jakarta Data provider may coexist within a single program, and provide repository implementations.
+It is the responsibility of the application developer to ensure that each repository is eligible for implementation by exactly one provider, according to the rules specified below.
 
 === Provider support for Entities
 
-A Jakarta Data provider supplies an implementation of repository interfaces in Jakarta Data for one or more types of entities. An entity refers to a class that represents objects in a storage engine, such as SQL or NoSQL databases.
+A Jakarta Data provider supports one or more entity-defining annotation types.
```
njr-11, line 22: This is another subtle change that inadvertently rules out providers supporting unannotated entities.  We can adjust it as follows to preserve the meaning:
```suggestion
A Jakarta Data provider typically supports one entity-defining annotation type, but can support any number of entity-defining annotation types.
```

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -1,24 +1,42 @@
...
+For example:
 
-The `jakarta.persistence.Entity` annotation from the Jakarta Persistence specification can be used by repository entity classes for Jakarta Data providers that are backed by a Jakarta Persistence provider. Other Jakarta Data providers must not support the `jakarta.persistence.Entity` annotation.
+- The `jakarta.persistence.Entity` annotation from the Jakarta Persistence specification is the entity-defining annotation for Jakarta Data providers backed by a Jakarta Persistence provider. Other Jakarta Data providers must not support the use of `jakarta.persistence.Entity` as an entity-defining annotation.
```
njr-11, line 25: To allow for providers that support unannotated entities:
```suggestion
- The `jakarta.persistence.Entity` annotation from the Jakarta Persistence specification is an entity-defining annotation for Jakarta Data providers backed by a Jakarta Persistence provider. Other Jakarta Data providers must not support the use of `jakarta.persistence.Entity` as an entity-defining annotation.
```

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -1,24 +1,42 @@
...
 
-The `jakarta.nosql.Entity` annotation from the Jakarta NoSQL specification can be used by repository entity classes for Jakarta Data providers that are backed by NoSQL databases. Other Jakarta Data providers must not support the `jakarta.nosql.Entity` annotation.
+- The `jakarta.nosql.Entity` annotation from the Jakarta NoSQL specification is the entity-defining annotation for Jakarta Data providers backed by NoSQL databases. Other Jakarta Data providers must not support the use of `jakarta.nosql.Entity` as an entity-defining annotation.
```
njr-11, line 27: To allow for providers that support unannotated entities:
```suggestion
- The `jakarta.nosql.Entity` annotation from the Jakarta NoSQL specification is an entity-defining annotation for Jakarta Data providers backed by NoSQL databases. Other Jakarta Data providers must not support the use of `jakarta.nosql.Entity` as an entity-defining annotation.
```

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -1,24 +1,42 @@
 == Jakarta Data Providers
 
-Jakarta Data providers implement repository interfaces, performing queries and other operations related to entities per the rules of the Jakarta Data specification. A Jakarta Data provider makes the repository implementation available to the application via dependency injection. The Jakarta Data specification defines the rules by which Jakarta Data providers must abide to ensure that multiple Jakarta Data providers are able to coexist in a system without interfering or overlapping on the same injection points.
+A Jakarta Data provider might come as an integrated component of a Jakarta EE container, or it might be a separate component that integrates with the Jakarta EE container via standard or proprietary SPIs.
+For example, a Jakarta Data provider might use a CDI portable extension to integrate with dependency injection.
```
njr-11, line 4:
```suggestion
A Jakarta Data provider can use a CDI portable extension to integrate with dependency injection.
```

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -1,24 +1,42 @@
...
+The role of the Jakarta Data provider is to supply implementations of application-defined repositories.
+
+A Jakarta Data provider must supply an implementation of each repository interface which either:
+
+- explicitly specifies the name of the provider via the `provider` member of the `@Repository` annotation, or
+- does not explicitly specify a `provider`, but whose entity classes are annotated with an entity-defining annotation type supported by the provider.
+
+The repository implementation, and the provider itself, is responsible for mediating interaction between the application program and the datastore, often taking advantage of other Jakarta EE services such as datasource provisioning and transaction management.
```
njr-11, line 12:
```suggestion
The repository implementation and the Jakarta Data provider are responsible for mediating interaction between the application program and the datastore, often taking advantage of other Jakarta EE services such as datasource provisioning and transaction management.
```

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -1,24 +1,42 @@
...
+
+An application usually obtains a reference to an implementation of a repository via dependency injection.
+Therefore, the provider must make its repository implementations available to the bean container responsible for dependency injection.
+
+Multiple Jakarta Data provider may coexist within a single program, and provide repository implementations.
```
njr-11, line 17:
```suggestion
Multiple Jakarta Data providers may coexist within a single program and provide repository implementations.
```

njr-11 (commented):
There are some good improvements here and also a few cases where changes resulted in a subtly different meaning. I added suggestions to clarify those back to the original meaning wherever that occurred, plus a few minor grammar updates.

spec/src/main/asciidoc/jakarta-ee.adoc
gavinking: I see. So that scenario wasn't obvious. I would add language explicitly talking about this usage pattern (unless it's already somewhere else in the spec that I missed).

spec/src/main/asciidoc/jakarta-ee.adoc
gavinking: Not sure about this change. I was trying to be illustrative here, not prescriptive.

spec/src/main/asciidoc/jakarta-ee.adoc
gavinking: I changed it to "In particular:"

spec/src/main/asciidoc/jakarta-ee.adoc
gavinking: You really want to allow providers which support _no_ entity-defining annotation?

OK, I guess.

spec/src/main/asciidoc/jakarta-ee.adoc
gavinking: Applied the change.

spec/src/main/asciidoc/jakarta-ee.adoc
gavinking: Applied the change.

spec/src/main/asciidoc/jakarta-ee.adoc
njr-11: I remember it being discussed a while back, and wanted to be sure we weren't accidentally removing it.  I did do some experimenting with the idea myself and it did seem workable, albeit limiting.

spec/src/main/asciidoc/jakarta-ee.adoc
gavinking: Applied the change.

spec/src/main/asciidoc/jakarta-ee.adoc
njr-11: Sure, that's fine. I'm pretty sure we state this elsewhere, so we can leave this as you wrote it and it will be fine.  I'll mark this comment resovled.

spec/src/main/asciidoc/jakarta-ee.adoc
gavinking: I have added language making this more explicit, WDYT, Nathan?

gavinking:
One thing I don't think is very clear (unless I've missed it) is to what extent a single repository can have entities annotated with a mix of entity-defining annotations, and what happens if the provider recognizes only _some_ of them.

njr-11 (approved):
This looks good - Thanks for improving this section!

gavinking:
> One thing I don't think is very clear (unless I've missed it) is to what extent a single repository can have entities annotated with a mix of entity-defining annotations, and what happens if the provider recognizes only some of them.

Perhaps we should just say that all entities belonging to a given repo should have the same annotation, and it's undefined behavior if they don't.

njr-11:
> > One thing I don't think is very clear (unless I've missed it) is to what extent a single repository can have entities annotated with a mix of entity-defining annotations, and what happens if the provider recognizes only some of them.
> 
> Perhaps we should just say that all entities belonging to a given repo should have the same annotation, and it's undefined behavior if they don't.

Good idea. That way, a provider need only check the primary entity type (or the first entity found) when making the determination of whether that provider handles the repository.

gavinking:
> That way, a provider need only check the primary entity type (or the first entity found) when making the determination of whether that provider handles the repository.

Oh yeah, that would be nice. OK.

gavinking:
> Perhaps we should just say that all entities belonging to a given repo should have the same annotation, and it's undefined behavior if they don't.

I've added a new paragraph to that effect.

PR #469, opened 2024-02-21
rewrite section on Portability for Relational Databases

gavinking:
I was not happy with section 7.1, so here's what I think it should say.

At least, something like this.

Thoughts?

### DIFF
--- a/spec/src/main/asciidoc/jakarta-ee.adoc
+++ b/spec/src/main/asciidoc/jakarta-ee.adoc
@@ -82,15 +82,14 @@ When a repository method is annotated with an interceptor binding annotation, th
 
 === Jakarta Persistence
 
-When integrating Jakarta Data with Jakarta Persistence, developers can leverage the JPA annotations to define the mapping of entities in repositories. Entities in Jakarta Persistence are typically annotated with `jakarta.persistence.Entity` to indicate their persistence capability.
+When integrating Jakarta Data with Jakarta Persistence, a developer uses the annotations defined in `jakarta.persistence` to specify the object/relational mapping metadata for entity types stored in a relational database.
 
-A Jakarta Data provider that supports Jakarta Persistence allows you to define repositories for classes marked with the `jakarta.persistence.Entity` annotation.
+According to the Jakarta Persistence specification, every entity class should be annotated `jakarta.persistence.Entity`. Jakarta Data places an additional interpretation on this annotation, treating it as an entity-defining annotation.
 
-By supporting Jakarta Persistence annotations, Jakarta Data providers enable Java developers to utilize familiar and standardized mapping techniques when defining entities in repositories, ensuring compatibility and interoperability with the respective technologies.
+Thus, a Jakarta Data provider which supports Jakarta Persistence is able to supply an implementation of any repository whose associated entity classes are marked with the `jakarta.persistence.Entity` annotation. Typically, such an implementation simply delegates operations declared by the repository interface to the Jakarta Persistence `EntityManager`. Management of persistence contexts and integration with Jakarta Transactions remains the responsibility of the Jakarta Persistence provider.
 
-==== Persistence Context
+By supporting Jakarta Persistence, a Jakarta Data provider enables Java developers to utilize familiar and standardized mapping techniques when defining entities associated with Jakarta Data repositories, ensuring compatibility and interoperability with the respective technologies.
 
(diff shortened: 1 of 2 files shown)


gavinking:
By the way, at this point, with so many of the foundations already in place, it would take me approximately an hour or two to add `@Persist`, `@Merge`, `@Refresh`, `@Lock`, `@Remove`, `@Detach`, `@JPQL` and `@SQL` annotations to Jakarta Data, thus providing fully portable support for JPA.

Just a thought. I know we said we weren't going to do it in this release, but we've come a long way since then.

njr-11 (approved):
The additions look good, and it also does a nice job of preserving the important parts of what was previously there.

gavinking:
I have pushed a second commit, which reworks section 6.4 on interop with JPA.

This *does* change semantics, since the current version says that JPA-backed repositories must by stateless, but JPA doesn't support that.

TANGENTIAL: the TCK has a test repository (`ee.jakarta.tck.data.standalone.persistence.Catalog`) which I believe is intended for use with JPA-backed providers. Now, this test is perfectly fine for my Hibernate-based implementation since I have stateless sessions. It's not, in my opinion, appropriate for a _generic_ JPA-backed provider. This problem is easily fixed if we do #470.

otaviojava (approved):

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -82,15 +82,14 @@ When a repository method is annotated with an interceptor binding annotation, th
...
 
-==== Persistence Context
+By supporting Jakarta Persistence, a Jakarta Data provider enables Java developers to utilize familiar and standardized mapping techniques when defining entities associated with Jakarta Data repositories, ensuring compatibility and interoperability with the respective technologies.
 
-Repository operations must behave as though backed by a stateless Entity Manager in that persistence context is not preserved across the end of repository methods. All entities that are returned by repository methods must be in a detached state such that modifications to these entities are not persisted to the database unless the application explicitly invokes a `Save` or `Update` life cycle method for the entity.
```
njr-11, line 93: I realize the stateless EntityManager didn't make it into Jakarta Persistence 3.2, but it would be nice to be able to standardize the behavior in anticipation of it.  I think it should be possible to simulate the behavior with what is currently in the Jakarta Persistence API unless I'm missing something.  Although I suppose it will be somewhat inefficient to do so.  I don't really like leaving it so unclear in the Jakarta Data spec, which leaves the user wondering what behavior they will end up with, or maybe assuming a behavior that is not the right one.

gavinking: OK so as I've argued elsewhere the implementations don't in principle need to wait for permission from the JPA spec to offer stateless sessions. So if you want me put back in something about this approach, that's totally fine, but I want it to be clear that you can also have a repository backed by a regular stateful session, and the previous language sounded like it was disallowing that.

njr-11: I think @otaviojava already merged this one on the basis of my review of the first commit. Apparently git doesn't reset the review status when more commits are added. We will need to ensure this discussion doesn't get lost because the change has been applied already.

gavinking: All this is easier to specify if I can talk about it in terms of `@Insert` vs `@Persist` if you get my drift.

gavinking: > I think @otaviojava already merged this one on the basis of my review of the first commit. Apparently git doesn't reset the review status when more commits are added. We will need to ensure this discussion doesn't get lost because the change has been applied already.

Don't worry Nathan I can write this up tonight if we agree on the basic idea.

gavinking: @njr-11 So how about something like this:

A Jakarta Data provider which supports Jakarta Persistence entities bearing the annotation `jakarta.persistence.Entity` may support either or both of the following repository styles:

- A repository might be backed by a Jakarta Persistence `EntityManager` with a stateful persistence context. The lifecycle operations of the repository are annotated with the Jakarta Persistence-specific annotations `@Persist`, `@Merge`, and friends, and follow the semantics of the corresponding standard operations of `jakarta.persistence.EntityManager`. In particular, the repository observes the semantics of the Jakarta Persistence `CascadeType`.
- Alternatively, a repository might by "stateless", in the sense that it does not maintain a persistence context which outlives the invocation of a single repository method. The lifecycle operations of this kind of repository are annotated `@Insert`, `@Delete`, or `@Update`, and adhere to the usual semantics of these annotations. In particular, these operations never cascade to related entities.

That's just a first cut, of course.

otaviojava: My bad, I did not see this discussion.

Issue #264, opened 2023-09-17
Repository to (optionally) return managed JPA entities (JPA integration)


lprimak:
**Description**

As far as I understand it, the repository finder methods return POJOs and not managed entities. In order to integrate with existing JPA codebases, it would be very useful if managed entities were returned.

As a:

- [ ] Application user/user of the configuration itself
- [x] API user (application developer)
- [ ] SPI user (container or runtime developer)
- [ ] Specification implementer

All that would be needed is a reference to EntityManager. I propose to infer EntityManager from CDI,
similar to what's implemented here: https://docs.flowlogix.com/#section-daohelper
```
@Inject
// Inject default EntityManager
@EntityManagerSelector
MyRepository<Entity, Key> repository;
```
```
@Inject
@EntityManagerSelector(NonDefaultEM.class)
MyRepository<Entity, Key> repository;
```
where `NonDefaultEM` is a custom `@Qualifier` annotation that points to a producer method that injects the appropriate EM.
Default would be no-qualifier injection, which would work in most cases.

otaviojava:
Hello @lprimak, how are you?

I updated the spec to return the EntityManager using convention methods.

Does it help?

https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/repository.asciidoc#resource-accessor-methods

lprimak:
Hi, Otavio!
First of all, thanks for your great work on this spec. It looks very promising!

* Is my assumption that the finder methods don't return managed entities correct?

If my assumption is correct, I would want the ability to pick which `EntityManager` manages the returned entities. Since an application may have multiple databases and entity managers, it's not enough for the implementation to pick one. I want to be able to pick one in application code, thus `@EntityManagerSelector` annotation.

This would also permit the return of managed entities from finder methods, using the *correct* entity manager.

Hopefully this makes things more clear.

Thank you!

lprimak:
>  I think we should defer discussion over managed entities to the subsequent version of Jakarta Data.

Understood.

I think in that case, spec needs to be updated to say so explicitly, with a caveat that the feature may be added in the future to integrate more tightly with JPA

Does spec mention how a persistence unit is picked currently, it there are multiple ones? I might have missed it or it's not there.
Does it use persistence.xml file or not?

njr-11:
> Does spec mention how a persistence unit is picked currently, it there are multiple ones? I might have missed it or it's not there.
> Does it use persistence.xml file or not?

[...]

Issue #22, opened 2022-09-22
Should we have support to configuration


otaviojava:
Once we're talking about data sources, it is natural to have settings such as user, password, and so on.

My question is: should we have support or a convention about configuration on this version?

#### JPA

```properties
# datasource configuration
quarkus.datasource.db-kind = postgresql
quarkus.datasource.username = hibernate
quarkus.datasource.password = hibernate
quarkus.datasource.jdbc.url = jdbc:postgresql://localhost:5432/hibernate_db

# drop and create the database at startup (use `update` to only update the schema)
quarkus.hibernate-orm.database.generation=drop-and-create
```

#### Mongodb

```properties
# configure the mongoDB client for a replica set of two nodes
quarkus.mongodb.connection-string = mongodb://mongo1:27017,mongo2:27017

# configure the mongoDB client for a single instance on localhost
quarkus.mongodb.connection-string = mongodb://localhost:27017
```

#### JPA

```properties
spring.datasource.driver-class-name=com.mysql.cj.jdbc.Driver
spring.datasource.username=mysqluser
spring.datasource.password=mysqlpass
spring.datasource.url= jdbc:mysql://localhost:3306/myDb?createDatabaseIfNotExist=true
```

#### MongoDB
```properties
spring.data.mongodb.uri=mongodb+srv://<username>:<pwd>@<cluster>.mongodb.net/mygrocerylist
spring.data.mongodb.database=mygrocerylist
```

#### JPA

```properties
jpa.default.properties.hibernate.hbm2ddl.auto= update
jpa.default.properties.hibernate.hbm2ddl.show_sql=true
```

njr-11:
+1 to supporting configuration.
We also need to consider that in a Jakarta EE environment, many users of relational databases/Jakarta Persistence are using data sources either configured by vendor-specific means or [DataSourceDefinition](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/annotation/sql/datasourcedefinition), and are possibly accessed through resource references, and possibly with support for forms of authentication beyond just basic user/password.  The granularity of some of these existing definitions concerns me (for example java:comp names that mean different things based on which component does the lookup) with respect to a repository that is shared across the whole application or even multiple applications.

There could also be multiple data sources and/or multiple Jakarta NoSQL providers pointing at different databases, so configuration might be needed on a per-repository basis rather than system wide (although the latter would certainly be more convenient).

[...]

PR #474, opened 2024-02-22
allow disambiguating annotation on resource accessor methods

gavinking:
Here's a proposal that in my opinion solves #264, as well as providing an alternative to "configuration" as envisaged in issue #22.

What do ya'll think of this approach?

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -853,6 +853,30 @@ default void cleanup() {
 
 A repository may have at most one resource accessor method.
 
+A resource accessor method may be annotated with:
+
+- one or more CDI qualifier types,
+- `jakarta.persistence.PersistenceContext` or `jakarta.persistence.PersistenceUnit`, or
+- `jakarta.annotation.Resource`.
+
+In this case, the repository implementation should use such annotations to disambiguate the resource it uses to access the data store, according to the usual rules defined by CDI, Jakarta EE, or Jakarta Persistence.
+
+For example, `@Resource` might be used to disambiguate a `DataSource` used to obtain JDBC ``Connection``s:
+
+[source,java]
+----
+@Resource(name="documentDatabase")
+DataSource connection();
+----
+
+Similarly, `@PersistenceContext` might be used to disambiguate a Jakarta Persistence persistence unit:
+
+[source,java]
+----
+@PersistenceContext(unitName="documents")
+EntityManager entityManager();
+----
+
 === Query by Method Name
 
 The Query by method mechanism allows for creating query commands by naming convention.


gavinking:
The issue is that currently, if I have, say, two persistence units in my EE application, we have no clear way to tell a repository which one to use.

otaviojava:
Once we are changing this part.
IMHO: 
The entire part of [Resource accessor methods](https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/repository.asciidoc#resource-accessor-methods) regarding methodology and dynamics should be included in the [Jakarta Persistence Context](https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/jakarta-ee.adoc#jakarta-persistence).

As a reader, it sounds weird to maintain the rhythm and cadence for the reader.

gavinking:
> The entire part of [Resource accessor methods](https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/repository.asciidoc#resource-accessor-methods) regarding methodology and dynamics should be included in the [Jakarta Persistence Context](https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/jakarta-ee.adoc#jakarta-persistence).

@otaviojava I'm not sure what you mean. None of that section is specific to Jakarta Persistence or to relational databases. Mongo DB or whatever has some sort of native "connection" or "session" type.

JDBC is just an example.

njr-11:
> The issue is that currently, if I have, say, two persistence units in my EE application, we have no clear way to tell a repository which one to use.

Oh I see what you are trying to do here. 
We do already have this, via the `dataStore` attribute of `@Repository`.

```
@Repository(dataStore = "persistence/MyPersistenceUnit")
public interface Employees {
```

gavinking:
OK, yeah, I forgot about that when I was looking at https://github.com/jakartaee/data/issues/264, but the thing is that the `@Repository` approach doesn't let me specify CDI qualifier types. (Which was the main thing I was really targeting here.)

gavinking:
(And I don't think it would be correct to just stick the CDI qualifiers on the repository itself.)

njr-11:
How important is it to be able to obtain the resource via qualifiers?  The `Respository(dataStore=` approach is more general and can work for any resource type, is self-documenting, and more concise.  But you are right, it doesn't cover if the user wants to have the resource be located via qualifiers.  Unfortunately, none of the qualifier patterns seem very intuitive.

gavinking:
Well it's something we (JPA) have been getting pressure from Jakarta EE folks and some individuals about so apparently it matters to some people.

I think there's a desire to unify everything around a CDI-based approach and deemphasize the more "legacy" JNDI stuff. And I'm not going to argue against that.

gavinking:
Is it a "have to have" 🤷‍♂️ well, for me personally, no, I guess??

njr-11:
> Well it's something we (JPA) have been getting pressure from Jakarta EE folks and some individuals about so apparently it matters to some people.
> 
> I think there's a desire to unify everything around a CDI-based approach and deemphasize the more "legacy" JNDI stuff. And I'm not going to argue against that.

It sounds like it's part of the CDI-centric push from Jakarta EE platform, which I agree in general is a good direction.  But I don't think some of the specs we might utilize are far enough along with it for us to figure out what the best approach will be for Jakarta Data.  For example, I don't see a spec-defined solution yet for a CDI alternative to resource references to JDBC DataSource.  I would rather wait for these other specifications and Jakarta EE to get further along with a clearer picture of where it all ends up and then have Jakarta Data tie into that, especially when the approaches we currently would have with qualifiers are a bit awkward and not intuitive.  I would propose that we go with what is currently in the spec here for 1.0 and assign this issue to Future, where we can revisit when there is better direction.

gavinking:
>  I don't see a spec-defined solution yet for a CDI alternative to resource references to JDBC `DataSource`.

Just to be clear, this has been available and well-defined since CDI 1.0. The way to do it is to write:

```java
@Produces @Resource(lookup="java:global/env/jdbc/CustomerDatasource")
@CustomerDatabase DataSource customerDatabase;
```

or:

```java
@Produces @PersistenceContext(unitName="CustomerDatabase")
@CustomerDatabase EntityManager customerDatabasePersistenceContext;
```
anywhere that CDI will look.

This is currently in section 9.7.1 of the CDI 4 spec, but it's been there forever.

I'm pointing this out because some people don't seem to be aware that it exists, or for some reason view it as some sort of "workaround".

gavinking:
> I would propose that we go with what is currently in the spec here for 1.0 and assign this issue to Future, where we can revisit when there is better direction.

OK, so look, how about we close issue https://github.com/jakartaee/data/issues/264, since I think that the issue description is a bit outdated/confusing/not-quite-right, and I will open a new issue specifically describing the problem with CDI qualifiers, and laying out what I see as the range of possible solutions. (There aren't really many options, since interfaces are a bit limited in terms of places you can but an annotation.)

njr-11:
> I'm pointing this out because some people don't seem to be aware that it exists, or for some reason view it as some sort of "workaround".

I'm one of the people that thought of that as a workaround.  I didn't realize that was the intended solution.

gavinking:
> I'm one of the people that thought of that as a workaround. I didn't realize that was the intended solution.

I mean, to me it's a great solution. I find it more natural to write the mapping from a string-based name to a typesafe annotation in Java, than to write annotations in, say `persistence.xml`. But people apparently want to write them in `persistence.xml`, so that's what we're going to let them do 🤷‍♂️🤷‍♂️

njr-11:
I've done more investigation of this proposal and even got some of it sort of working, although this has also raised more concerns because I don't see it working in all cases.

First, some background. We currently let the user configure either a persistence unit reference or a data source resource reference, as you can do with `@Repository(dataStore=...)`.  We believe this is important because many users will not be JPA users and they shouldn't need to configure anything JPA specific to use Jakarta Data if they don't want to or know how to.  What we have been doing is to take the data source resource reference and essentially write the JPA config for them.  That works because a data source resource reference fits into that model.  If you consider what happens with the resource accessor method (maybe better termed a resource supplier method for this usage) we now end up with a set of qualifiers and a resource type of DataSource.

One problem is how to supply that to the Jakarta Persistence provider. Even if the Data CDI extension is running at a point where it has access to do CDI.current().select(DataSource.class, qualifierAnnos), this doesn't help us any because Jakarta Persistence config doesn't accept DataSource instances.  It takes JNDI names to resource references.  At this point, I can locate the producer definition, grab the `@Resource` definition on it and look at the `name()` and use that, but only if `name` was specified. This actually worked when I tried it, sort of, because I did everything in a particular way. First, I actually specified a JNDI name value on `@Resource`. To me, it seems to defeat the purpose of being CDI-centric if you still need JNDI names.  Second, the name I specified was intentionally in java:app so that the Jakarta Persistence provider would be able to look it up even when not running in the same module/component.  I expect most users would want to omit the resource reference JNDI `name` value.  When that happens, there are some defaults that apply, based on field name and so forth.  Technically I could follow those same rules to figure out the name.  But then there's another problem -- I expect these defaulted names are in java:comp and they won't always be accessible.

Another problem I worry about is CDI scopes. In this case of the hacky approach I described above, I swizzled it back to JNDI and the scope doesn't even apply.  That seems wrong.  If taking another approach, maybe with obtaining the EntityManager/EntityManagerFactory rather than DataSource, do we need to be concerned with it going out of scope and the Jakarta Data provider/Jakarta Persistence provider still trying to use it across the application?  Maybe the users needs to always need to specify `@ApplicationScoped`.  If they don't, this could be another place for things to go wrong.

Another issue is with obtaining EntityManager.  I like how the existing approach can give you an EntityManagerFactory, not EntityManager. In the future I see this being quite useful if a Jakarta Data provider wants to in turn request a specific type of EntityManager from the Jakarta Persistence provider.  I'm imagining a emf.getStatelessEntityManager method or something to that effect.  If, on the other hand, we have a model where the user, via their repository, supplies the EntityManager and not the EntityManagerFactory, then the Jakarta Data provider is stuck with whatever sort of EntityManager the user's gave it.

To summarize, the proposed approach has some aspects that might be workable, but introduces complexity and unanswered questions. We shouldn't add it in version 1.0. Maybe there are ways we can improve on it and/or be able to resolve some of the issues related to it, and if so, maybe it will be something we can add later on.

gavinking:
@njr-11 I don't quite understand what you're trying to do here.

Are you trying to obtain a `DataSource` and somehow pass that directly to JPA? That's not the way JPA is supposed t work. A JPA `EntityManagerFactory` has its own `DataSource` configured in `persistence.xml`, or programmatically via the new `PersistenceConfiguration` object. You don't pass a `DataSource` to an `EMF`.

With respect to CDI qualifier:

- For a repository backed by JPA the idea is that these would be CDI qualifier annotations that identify a _persistence unit_, not a _datasource_.
- The case of qualifier annotations which identify a _datasoure_ is for when you have a repository backed by direct JDBC access or whatever.

gavinking:
That is to say:

```java
interface DocumentRepo {
    @DocumentDatabase 
    EntityManager entityManager()
}
```

Would by implemented by a class like this, say:
```
class DocumentRepo_ {
    private final EntityManager entityManager;

    @Inject
    DocumentRepo_(@DocumentDatabase EntityManager entityManager) {
        this.entityManager = entityManager;
    }
    
    @Override
    EntityManager entityManager() { return entityManager; }
}
```

Or whatever. There's lot of possible variations on this.

Anyway the point is I don't see what a repository based on JPA would be messing around with the `DataSource` directly.

njr-11:
One of the scenarios I was looking it is when a user writes a repository resource accessor/supplier method to produce a DataSource, associating that repository to a Jakarta Data provider that is backed by Jakarta Persistence.  I noted that we can already handle the equivalent with `@Repository(dataStore=...)` but I do not see any good way of doing so for the resource accessor/supplier pattern.  Maybe we need to specify that is an error path.  Even if we do that, my other concerns still apply about the EntityManager.

gavinking:
> One of the scenarios I was looking it is when a user writes a repository resource accessor/supplier method to produce a `DataSource`, associating that repository to a Jakarta Data provider that is backed by Jakarta Persistence. 
>  I noted that we can already handle the equivalent with `@Repository(dataStore=...)`

I suppose you're doing it by passing the datasource name via config property `"jakarta.persistence.jtaDataSource"` to `Persistence.createEntityManagerFactory()`.

I believe that it's possible to directly pass an _instance_ of `DataSource` using the property `"jakarta.persistence.dataSource"`. That's how the Jakarta EE container passes the datasource to the JPA provider I believe. (Not really an expert on this particular aspect.)

Quote from the JPA spec:

<blockquote>
The following additional standard properties
are defined by this specification for the configuration of the entity
manager factory:

* `jakarta.persistence.jdbc.driver` — value is
the fully qualified name of the driver class.
* `jakarta.persistence.jdbc.url` — string
corresponding to the driver-specific URL.
* `jakarta.persistence.jdbc.user` — value is the
username used by database connection.
* `jakarta.persistence.jdbc.password` — value is
the password for database connection validation.
* `jakarta.persistence.dataSource` — value is
instance of `javax.sql.DataSource` to be used for the specified
persistence unit.
* `jakarta.persistence.validation.factory` —
value is instance of `jakarta.validation.ValidatorFactory`.
</blockquote>

njr-11:
Nice - I should try out "jakarta.persistence.dataSource" and see how that works. That would reduce the issues with DataSource to being more similar to those with EntityManager, where I'm wondering what if it goes out of scope?

gavinking:
I mean I guess we could close this pull request but issue #476 is still identifying a very real problem that we currently have no solution for.

We need a solution.

PR #482, opened 2024-02-23
add Order.getSorts()

gavinking:
Since `Iterable` isn't at all convenient to work with in generated code.

Weird but true.

### DIFF
--- a/api/src/main/java/jakarta/data/Order.java
+++ b/api/src/main/java/jakarta/data/Order.java
@@ -96,6 +96,15 @@ public class Order<T> implements Iterable<Sort<T>> {
     }
 
     /**
+     * The instances of {@link Sort} belonging to this {@code Order}.
+     *
+     * @return the instances of {@link Sort}
+     */
+    public List<Sort<T>> getSorts() {
+        return sorts;
+    }
+
+    /**
      * Determines whether this instance specifies matching {@link Sort} criteria
      * in the same order of precedence as another instance.
      *
--- a/api/src/main/java/jakarta/data/Order.java
+++ b/api/src/main/java/jakarta/data/Order.java
@@ -101,7 +101,7 @@ public class Order<T> implements Iterable<Sort<T>> {
      * @return the instances of {@link Sort}, from highest precedence to lowest precedence.
      */
     public List<Sort<T>> getSorts() {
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/data/Order.java
```diff
@@ -95,6 +95,15 @@ public static final <T> Order<T> by(Sort<T>... sorts) {
         return new Order<T>(List.of(sorts));
     }
 
+    /**
+     * The instances of {@link Sort} belonging to this {@code Order}.
+     *
+     * @return the instances of {@link Sort}
```
njr-11, line 101: Updating JavaDoc to let the user know the list of Sorts follows the order of precedence that they provided just in case they aren't comfortable assuming so.
```suggestion
     * @return the instances of {@link Sort}, from highest precedence to lowest precedence.
```

njr-11 (commented):
This looks good. I added one minor update to JavaDoc

njr-11 (approved):

api/src/main/java/jakarta/data/Order.java
```diff
@@ -95,6 +95,15 @@ public static final <T> Order<T> by(Sort<T>... sorts) {
         return new Order<T>(List.of(sorts));
     }
 
+    /**
+     * The instances of {@link Sort} belonging to this {@code Order}.
+     *
+     * @return the instances of {@link Sort}, from highest precedence to lowest precedence.
+     */
+    public List<Sort<T>> getSorts() {
```
otaviojava, line 103:
```suggestion
    public List<Sort<T>> sorts() {
```

Once we are using the record style, could we keep this method as well?

https://github.com/jakartaee/data/pull/75

api/src/main/java/jakarta/data/Order.java
```diff
@@ -95,6 +95,15 @@ public static final <T> Order<T> by(Sort<T>... sorts) {
...
+        return sorts;
```
otaviojava, line 104: Could you make this return immutable?

```suggestion
        return java.util.Collections.unmodifiableList(sorts);
```

gavinking: Ah, yes, of course, nice catch.

PR #486, opened 2024-02-24
Jakarta persistence module proposal

otaviojava:
# Changes

- Create a Jakarta Persistence Module


api/persistence/src/main/java/jakarta/data/persistence/JakartaPersistenceRepository.java
```diff
@@ -0,0 +1,72 @@
...
+import jakarta.persistence.RefreshOption;
+
+/**
+ * <p>A built-in repository supertype for repositories backed by a Jakarta Persistence
+ * {@code EntityManager}.</p>
+ *
+ * @param <T> the type of the primary entity class of the repository.
+ * @param <K> the type of the Id attribute of the primary entity.
+ */
+public interface JakartaPersistenceRepository <T, K> extends DataRepository<T, K> {
```
otaviojava, line 35: Once we agree with `persistence`I belive the  `JakartaPersistence` might be fine.

Open for suggestion, I am not good at name proposals

njr-11:
Otavio - before you get too far along this path, this is not something that we can get into version 1.0.  We can look at additional repositories in the next version.

Issue #473, opened 2024-02-22
[clarification]: Resource enlistment with JTA


gavinking:
### Specification

Jakarta Data / Jakarta Transactions

### I need clarification on ...

section 6.2. Jakarta Transactions Usage.

### Additional information

I've been talking with @Sanne about what the spec says here.

First:

> When running in an environment where Jakarta Transactions is available and a global transaction is active on the thread of execution for a repository operation and the data source backing the repository is capable of transaction enlistment, the repository operation enlists the data source resource as a participant in the transaction. 

It was never _traditionally_ the responsibility of technologies in this layer to manage resource enlistment. That was the job of some interaction between&mdash;I dunno the details&mdash;JCA and JTA, probably, I suppose.

In particular, if we're talking about a repository backed by a container JDBC `Datasource`, it would not _literally_ be the repository itself that enlists the connection with JTA. Now, sure, it might happen by a side-effect of the "repository operation", and I could read the spec this way, and maybe that's fine.

But for the sake of not creating confusion, I wonder if there's something clearer/more explicit we could say here?

> The repository operation ... might cause the transaction to be marked as rollback only (`jakarta.transaction.Status.STATUS_MARKED_ROLLBACK`) if the repository operation fails.

There's quite a lot of ambiguity hiding in that "might". Should we be more prescriptive here? Should the repository mark the transaction for roll back _every time it throws_? That seems too heavy-handed, perhaps. I note that JPA has a set of exception types, and each one specifies whether it triggers mark-for-rollback.

I wonder what we could say that would not be too much of a blunt instrument?

njr-11:
#483 clarifies that language to avoid implying that the Jakarta Data provider is responsible for transaction enlistment.

Regarding the second part, I agree it would be too heavy handed for the repository to force every exception to mark the transaction for rollback.  I'm not sure what else to say though.  The user will want to know about the possibility but I don't know how to give them any more guarantees here, only to identify that being marked for roll back is a possibility.  Application code that traps an exception and wants to know if it can keep performing operations in the transaction does have the ability through the JTA API to invoke `tran.getStatus()` to see if the transaction is `Status.STATUS_MARKED_ROLLBACK`, so at least they have a way to find out

PR #489, opened 2024-02-25
expand spec sections on JTA and Jakarta Interceptors

gavinking:
I'm trying to make very clear who is responsible for what, and which things are prescriptive requirements on the implementation of Jakarta Data, vs just being information about how Jakarta Data can take advantage of CDI/JTA/Interceptors.

### DIFF
--- a/spec/src/main/asciidoc/jakarta-ee.adoc
+++ b/spec/src/main/asciidoc/jakarta-ee.adoc
@@ -107,16 +107,42 @@ NOTE: CDI Lite (corresponding to Jakarta Core profile) does not include a requir
 
 NOTE: Jakarta Data providers that wish to provide both extensions can use CDI's `@SkipIfPortableExtensionPresent` to prevent the `BuildCompatibleExtension` from colliding with the portable `Extension` when running in the Jakarta Web Profile or Jakarta Platform where CDI Full is present.
 
-=== Jakarta Transactions Usage
+=== Jakarta Interceptors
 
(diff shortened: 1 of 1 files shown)


gavinking:
This is for #473, by the way.

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -107,16 +107,42 @@ NOTE: CDI Lite (corresponding to Jakarta Core profile) does not include a requir
...
+1. Jakarta Transactions is available,
+2. a global transaction is active on the thread of execution in which a repository operation is called, and
+3. the data source backing the repository is capable of transaction enlistment,
+
+then the repository operation must be performed within the context of the global transaction. That is, the data source resources involved in the operation must be enlisted as participants in the transaction.
+
+[NOTE]
+====
+In the Jakarta EE environment, such enlistment usually happens automatically when the repository implementation makes use of a data source which is under the management of the Jakarta EE container.
+In the case of a data source which is _not_ managed by the container, the repository implementation itself is required to ensure that such enlistment occurs.
```
njr-11, line 134: Data sources that are not manged by the container fall outside of Jakarta EE so we should avoid requirements in that area.
```suggestion
```

njr-11 (commented):
This looks great. Thanks for making this section better. I did spot one requirement that needs to be omitted because it can be interpreted to imply that data sources not managed by the container need to be supported in Jakarta EE products, which goes beyond Jakarta EE.

spec/src/main/asciidoc/jakarta-ee.adoc
gavinking: OK.

njr-11 (approved):

Issue #488, opened 2024-02-25
[clarification]: requirement to support lifecycle methods returning wrapper types and success counts


gavinking:
### Specification

Javadoc of @Insert and friends 

### I need clarification on ...

- Why support for wrapper types is even required.
- Whether lifecycle methods should return success counts instead of throwing.

### Additional information

In the Javadoc for `@Insert`, `@Update`, `@Delete`, and `@Save`, we've actually grown a bit of a combinatorial explosion of possible supported signatures.

We have to accept:

1. single enitities
2. arrays
3. `iterable`s

and we have to be able to return

1. `void`
2. `boolean`
3. `int` or `long`
4. `Boolean`, `Integer`, or `Long`
5. `Void`, in some cases
6. in the case of `@Insert`, an array 
7. in the case of `@Insert`, an `Iterable` 

Wrapper types and `Void`
---------

First, what is the point of accepting `Void` here? And why allow wrapper types? Just to increase the possibility of an NPE?

IMO, support for returning wrapper types should not be required.

"Quiet" failure
---------

The requirement to support return types other than `void` on a _lifecycle_ (as opposed to _query_) method came a bit out of the blue for me, since I would have said that such methods _should always fail with an exception_ if they can't perform the operation they've been asked to perform.

To me it seems like the operation:

```java
@Update
int update(Iterable<Book> book)
```

is really quite a dangerous API to allow, in that it's highly likely that a client will forget to check the return value of the method, in which case a failure goes unnoticed.

It's almost _always_ going to be the case that if I ask my repository to store 5 `Book`s at once, and it only succeeds in storing 4 of them, then what I want to do is fail and roll back the transaction. Indeed, being able to recover from such a failure _and just keep on working_ is something that's not necessarily even _possible_. At the very least, if I plan to keep working after such a failure, then I need to know _why_ it failed, and an `int` doesn't give me that information! Hell, very often the _database_ doesn't give me that information in a useful form.

So supporting this feels like it leads users down a bad path, and implementors down a rabbit hole of trying to understand which sorts of failures are recoverable.

I don't think this should be required in Jakarta Data 1.0.

Can we please simplify the requirements here, at least for this release?

gavinking:
P.S. note that typically when you execute an `update` or `delete` and get a zero row count back, that's due to an optimistic lock failure and it should be reported via `OptimisticLockingFailureException`.

Violation of a database-level constraint should also always be reported via an exception, in my opinion.

[...]

PR #490, opened 2024-02-25
fix requirements in Javadoc for built-in lifecycle annotations

gavinking:
See #488.

This pull request absolutely _does_ change semantics, eliminating:

- dangerous "quiet failure" which can easily lead to data corruption in the case of minor errors of program logic
- the previous "explosion" of acceptable method signatures which makes Jakarta Data more difficult to implement, while offering little additional flexibility to users
- nails down the treatment of optimistic lock acquisition failures

These changes make Jakarta Data significantly easier to implement, and lead users down a path where they are much less likely to accidentally write broken code.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Delete.java
+++ b/api/src/main/java/jakarta/data/repository/Delete.java
@@ -27,8 +27,8 @@ import java.lang.annotation.Target;
 /**
  * <p>Lifecycle annotation for repository methods which perform delete operations.</p>
  *
- * <p>The {@code Delete} annotation indicates that the annotated repository method deletes the state of one or more entities
- * from in the database. The annotated repository method usually has exactly one parameter whose type must be one of
+ * <p>The {@code Delete} annotation indicates that the annotated repository method deletes the state of one or more
+ * entities from in the database. The annotated repository method usually has exactly one parameter whose type is one of
  * the following:
  * </p>
  * <ul>
@@ -46,15 +46,18 @@ import java.lang.annotation.Target;
  *     void unpark(Car car);
  * }
  * </pre>
- * <p>Alternatively, the {@code Delete} annotation may be applied to a repository method with no parameters, indicating
- * that the annotated method deletes all instances of the primary entity type. In this case, the annotated method must
- * either be declared {@code void}, or return {@code int} or {@code long}.
- * </p>
  * <p>Deletes are performed by matching the unique identifier of the entity. If the entity is versioned, for example,
  * with {@code jakarta.persistence.Version}, the version is also checked for consistency. Attributes other than the
  * identifier and version do not need to match. If no entity with a matching identifier is found in the database, or
  * if the entity with a matching identifier does not have a matching version, the annotated method must raise
  * {@link jakarta.data.exceptions.OptimisticLockingFailureException}.
+ * </p>
+ * <p>Alternatively, the {@code Delete} annotation may be applied to a repository method with no parameters, indicating
+ * that the annotated method deletes all instances of the primary entity type. In this case, the annotated method must
+ * either be declared {@code void}, or return {@code int} or {@code long}.
+ * </p>
+ * <p> Application of the {@code Delete} annotation to a method with any other signature is not portable between Jakarta
+ * Data providers.
  * </p>
  * <p>If this annotation occurs alongside a different lifecycle annotation, the annotated repository method must raise
  * {@link UnsupportedOperationException} every time it is called. Alternatively, a Jakarta Data provider is permitted to
--- a/api/src/main/java/jakarta/data/repository/Delete.java
+++ b/api/src/main/java/jakarta/data/repository/Delete.java
@@ -25,27 +25,18 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * <p>Annotates a repository method to perform delete operations.</p>
+ * <p>Lifecycle annotation for repository methods which perform delete operations.</p>
  *
- * <p>The {@code Delete} annotation indicates that the annotated repository method requests one or more
- * entities to be removed from the database. To request deletion of specific entity instances, the annotated
- * repository method must have a single parameter whose type must be one of the following:
+ * <p>The {@code Delete} annotation indicates that the annotated repository method the state of one or more entities
+ * from in the database. The annotated repository method usually has exactly one parameter whose type must be one of
+ * the following:
  * </p>
  * <ul>
  *     <li>The entity to be deleted.</li>
  *     <li>An {@code Iterable} of entities to be deleted.</li>
  *     <li>An array of entities to be deleted.</li>
  * </ul>
- * <p>The {@code Delete} annotation can be used on a repository method that has no parameters
- * to request deletion of all entity instances of the primary entity type.</p>
- * <p>The return type of the annotated method must be {@code void}, {@code boolean}, {@code int}, {@code long},
- * or a corresponding primitive wrapper such as {@link Integer}.
- * A boolean return type indicates whether or not an entity was deleted from the database.
- * An {@code int} or {@code long} return type indicates how many entities were deleted from the database.
- * </p>
- * <p>Deletion of a given entity is performed by matching the entity's Id. If the entity is versioned (e.g.,
- * with {@code jakarta.persistence.Version}), the version is also checked for consistency during deletion.
- * Properties other than the Id and version do not need to match for deletion.
+ * <p>The annotated method must be declared {@code void}.
  * </p>
  * <p>For example, consider an interface representing a garage:</p>
  * <pre>
@@ -55,12 +46,19 @@ import java.lang.annotation.Target;
  *     void unpark(Car car);
  * }
  * </pre>
- * <p>If this annotation is combined with other operation annotations (e.g.,
- * {@code @Find}, {@code @Insert}, {@code @Query}, {@code @Update},
- * {@code @Save}), it will throw an {@link UnsupportedOperationException} as only one operation type can be specified.</p>
- * <p>If the unique identifier of a requested entity is not found in the database or its version does not match, and the return
- * type of the annotated method is {@code void} or {@code Void}, the method must
- * raise {@link jakarta.data.exceptions.OptimisticLockingFailureException}.
+ * <p>Alternatively, the {@code Delete} annotation may be applied to a repository method with no parameters, indicating
+ * that the annotated method deletes all instances of the primary entity type. In this case, the annotated method must
+ * either be declared {@code void}, or return {@code int} or {@code long}.
+ * </p>
+ * <p>Deletes are performed by matching the unique identifier of the entity. If the entity is versioned, for example,
+ * with {@code jakarta.persistence.Version}, the version is also checked for consistency. Attributes other than the
+ * identifier and version do not need to match. If no entity with a matching identifier is found in the database, or
+ * if the entity with a matching identifier does not have a matching version, the annotated method must raise
+ * {@link jakarta.data.exceptions.OptimisticLockingFailureException}.
+ * </p>
+ * <p>If this annotation occurs alongside a different lifecycle annotation, the annotated repository method must raise
+ * {@link UnsupportedOperationException} every time it is called. Alternatively, a Jakarta Data provider is permitted to
+ * reject such a method declaration at compile time.</p>
  */
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
--- a/api/src/main/java/jakarta/data/repository/Delete.java
+++ b/api/src/main/java/jakarta/data/repository/Delete.java
@@ -59,9 +59,10 @@ import java.lang.annotation.Target;
  * <p> Application of the {@code Delete} annotation to a method with any other signature is not portable between Jakarta
  * Data providers.
  * </p>
- * <p>If this annotation occurs alongside a different lifecycle annotation, the annotated repository method must raise
- * {@link UnsupportedOperationException} every time it is called. Alternatively, a Jakarta Data provider is permitted to
- * reject such a method declaration at compile time.</p>
+ * <p>Annotations such as {@code @Find}, {@code @Query}, {@code @Insert}, {@code @Update}, {@code @Delete}, and
+ * {@code @Save} are mutually-exclusive. A given method of a repository interface may have at most one {@code @Find}
+ * annotation, lifecycle annotation, or query annotation.
+ * </p
  */
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
--- a/api/src/main/java/jakarta/data/repository/Delete.java
+++ b/api/src/main/java/jakarta/data/repository/Delete.java
@@ -27,7 +27,7 @@ import java.lang.annotation.Target;
 /**
  * <p>Lifecycle annotation for repository methods which perform delete operations.</p>
  *
- * <p>The {@code Delete} annotation indicates that the annotated repository method the state of one or more entities
+ * <p>The {@code Delete} annotation indicates that the annotated repository method deletes the state of one or more entities
  * from in the database. The annotated repository method usually has exactly one parameter whose type must be one of
  * the following:
  * </p>
--- a/api/src/main/java/jakarta/data/repository/Delete.java
+++ b/api/src/main/java/jakarta/data/repository/Delete.java
@@ -28,13 +28,12 @@ import java.lang.annotation.Target;
  * <p>Lifecycle annotation for repository methods which perform delete operations.</p>
  *
  * <p>The {@code Delete} annotation indicates that the annotated repository method deletes the state of one or more
- * entities from in the database. The annotated repository method usually has exactly one parameter whose type is one of
- * the following:
+ * entities from in the database. The annotated repository method usually has exactly one parameter whose type is
+ * either:
  * </p>
  * <ul>
- *     <li>The entity to be deleted.</li>
- *     <li>An {@code Iterable} of entities to be deleted.</li>
- *     <li>An array of entities to be deleted.</li>
+ *     <li>the class of the entity to be deleted, or</li>
+ *     <li>{@code Iterable<E>} where {@code E} is the class of the entities to be deleted.</li>
  * </ul>
  * <p>The annotated method must be declared {@code void}.
  * </p>
--- a/api/src/main/java/jakarta/data/repository/Delete.java
+++ b/api/src/main/java/jakarta/data/repository/Delete.java
@@ -28,7 +28,7 @@ import java.lang.annotation.Target;
  * <p>Lifecycle annotation for repository methods which perform delete operations.</p>
  *
  * <p>The {@code Delete} annotation indicates that the annotated repository method deletes the state of one or more
- * entities from in the database.
+ * entities from the database.
  * </p>
  * <p>A {@code Delete} method might accept an instance or instances of an entity class. In this case, the method must
  * have exactly one parameter whose type is either:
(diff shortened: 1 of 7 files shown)


gavinking:
I recommend pulling this branch and reading the Javadoc in full, rather than reviewing the diffs, which are not going to be particularly readable.

otaviojava (approved):

otaviojava:
I liked having fewer options, which will make it easier to implement it.

api/src/main/java/jakarta/data/repository/Update.java
```diff
@@ -24,51 +24,21 @@
...
-
 /**
- * <p>Annotates a repository method to perform update operations.</p>
+ * <p>Lifecycle annotation for repository methods which perform update operations.</p>
  *
- * <p>The {@code Update} annotation indicates that the annotated repository method requests that one or more entities
- * be updated if found in the database. To request updates to specific entity instances, the annotated
- * repository method must have a single parameter whose type must be one of the following:
+ * <p>The {@code Update} annotation indicates that the annotated repository method updates the state of one or more
+ * entities already held in the database. The annotated repository method must have a exactly one parameter whose type
```
njr-11, line 31: This will preserve the future flexibility to use the Update annotation for query-based updates as well as allow vendors to experiment with those approaches in the mean time,
```suggestion
 * entities already held in the database. To supply entity instances to the method,
 * the annotated repository method must have a exactly one parameter whose type
```

api/src/main/java/jakarta/data/repository/Update.java
```diff
@@ -24,51 +24,21 @@
...
- * leaving out those that did not match the unique identifier and version that is in the database.
- * </p>
- * <p>Updating an entity involves modifying its existing data in the database. The method will search for the entity
- * in the database using its ID (and version, if versioned) and then update the corresponding record with the new data. After invoking
- * this method, do not continue to use the entity value that is supplied as a parameter, as it may not accurately
- * reflect the changes made during the update process.
- * </p>
- * <p>If the entity does not exist in the database or it is versioned and its version differs from the version in the database,
- * no update is made and no error is raised.
+ * <p>The annotated method must either be declared {@code void}, or have a return type that is the same as the type of
```
njr-11, line 39:
```suggestion
 * <p>When entity instances are supplied via the parameter,
 the annotated method must either be declared {@code void}, or have a return type that is the same as the type of
```

api/src/main/java/jakarta/data/repository/Update.java
```diff
@@ -77,10 +47,26 @@
...
- * {@code @Save}), it will throw an {@link UnsupportedOperationException} because only one operation type can be specified.
- * A Jakarta Data provider implementation must detect (and report) this error at compile time or at runtime.</p>
+ * <p>When the annotated method is non-{@code void}, it must return an updated entity instance for each entity instance
+ * passed as an argument. Instances returned by the annotated method must include all values that were written to the
+ * database, including all automatically generated values, updated versions, and incremented values which changed as a
+ * result of the update. The order of entities within an {@code Iterable} or array return value must match the position
+ * of entities in the argument, based on the unique identifier of the entity. After the annotated method returns, an
+ * original entity instance supplied as an argument might not accurately reflect the updated state.
+ * </p>
+ * <p>Updates are performed by matching the unique identifier of the entity. If the entity is versioned, for example,
```
njr-11, line 57:
```suggestion
 * <p>For entity instances that are supplied to the method,
 * updates are performed by matching the unique identifier of the entity. If the entity is versioned, for example,
```

api/src/main/java/jakarta/data/repository/Update.java
```diff
@@ -77,10 +47,26 @@
...
+ * with {@code jakarta.persistence.Version}, the version is also checked for consistency. Attributes other than the
+ * identifier and version do not need to match. If no entity with a matching identifier is found in the database, or
+ * if the entity with a matching identifier does not have a matching version, the annotated method must raise
+ * {@link jakarta.data.exceptions.OptimisticLockingFailureException}.
+ * </p>
+ * <p>
+ * If the database follows the BASE model, or uses an append model to write data, the annotated method behaves the same
+ * as the {@code @Insert} method.
+ * </p>
+ * <p>If this annotation occurs alongside a different lifecycle annotation, the annotated repository method must raise
```
njr-11, line 67:
```suggestion
 * <p>If this annotation occurs alongside a conflicting annotation (e.g., {@code @Insert}, {@code @Delete},
 * {@code @Find}, {@code @Query}, {@code @Save}), the annotated repository method must raise
```

njr-11 (changes requested):
This mostly looks like it should work. I corrected a few minor grammar things. found some subtle changes with the annotation conflict rules, and made sure that requirements for OptimisticLockException match up with when entities are supplied, which is currently the predominant usage, but I want to ensure we have left ourselves the right path for the query-based scenarios in the future.  Hopefully I've been able to minimally adjust for that.

api/src/main/java/jakarta/data/repository/Delete.java
```diff
@@ -55,12 +46,19 @@
...
+ * that the annotated method deletes all instances of the primary entity type. In this case, the annotated method must
+ * either be declared {@code void}, or return {@code int} or {@code long}.
+ * </p>
+ * <p>Deletes are performed by matching the unique identifier of the entity. If the entity is versioned, for example,
+ * with {@code jakarta.persistence.Version}, the version is also checked for consistency. Attributes other than the
+ * identifier and version do not need to match. If no entity with a matching identifier is found in the database, or
+ * if the entity with a matching identifier does not have a matching version, the annotated method must raise
+ * {@link jakarta.data.exceptions.OptimisticLockingFailureException}.
+ * </p>
+ * <p>If this annotation occurs alongside a different lifecycle annotation, the annotated repository method must raise
```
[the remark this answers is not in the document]
gavinking: @njr-11 I'm not a fan of this for the reason that we are begging the question of what makes two annotations "conflicting". We have not defined what that means. (And it's not enough to just give examples.)

IMO, the _specification_ and not the Javadoc of individual annotations is what really has the responsibility to say this. Otherwise we have to say it in lots of different places instead of in one place. So I almost removed these bits completely in favor of the spec addition here:

https://github.com/jakartaee/data/pull/490/commits/cac4ce1a147b96d2f9fed71617e2438fc6b433a7

So I think the issue is that you want to make clear that `@Find` can't be used together with `@Delete`. Well, if necessary the spec should say that. But when I thought about this quickly I thought that there were no method signatures which are legal for both `@Find` and `@Delete`. Perhaps that's over-optimistic.

Anyway the point is that if we need to say that `@Find` and query annotations conflict with lifecycle annotations, we should say it explicitly in the spec.

gavinking:
@njr-11 I've attempted to address your concerns in the latest pushed commits, but in a way that doesn't involve so much recitation. WDYT?

api/src/main/java/jakarta/data/repository/Delete.java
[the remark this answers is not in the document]
gavinking: (OK, I'm now proposing adding a new section to the spec to deal with this question.)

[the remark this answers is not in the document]
njr-11: Yes, go ahead and add the section to the spec.  Users typically won't read the spec, but will look at the JavaDoc of what they are using, so I do think it will help reduce user errors to call out what they cannot do.  I can definitely anticipate someone trying to annotate a method as both `@Query` and `@Delete` (or whatever type of query it is) because they think that makes sense to them.  I see what you are saying about not specifying in terms of examples.  I'll update my suggestions to fix that.  We can still call out not to combine with specific annotations.

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -62,12 +62,8 @@
...
  * </ul>
  *
- * <p>If the annotated method return type is a single instance (either through {@code Optional} or directly), but the query returns more than one element,
- * it will throw a {@link jakarta.data.exceptions.NonUniqueResultException}.</p>
- *
- * <p>If this annotation is combined with other operation annotations (e.g., {@code @Update}, {@code @Delete},
- *  {@code @Save}), it will result in an {@link UnsupportedOperationException} being thrown, as only one operation type can be specified.
- *  A Jakarta Data provider implementation must detect and report this error either at compile time or runtime.</p>
+ * <p>If the annotated method return type is a single instance (either through {@code Optional} or directly), but the
+ * query returns more than one element, it will throw a {@link jakarta.data.exceptions.NonUniqueResultException}.</p>
```
njr-11, line 66:
```suggestion
 * query returns more than one element, it will throw a {@link jakarta.data.exceptions.NonUniqueResultException}.</p>
 * <p>If this annotation occurs alongside {@code @Delete}, {@code @Find}, {@code @Insert},
 * {@code @Query}, {@code @Save}, or {@code @Update}, the annotated repository method must raise
```

api/src/main/java/jakarta/data/repository/Insert.java
```diff
@@ -25,54 +25,45 @@
 import java.lang.annotation.Target;
 
 /**
- * <p>Annotates a repository method that inserts entities.</p>
+ * <p>Lifecycle annotation for repository methods which perform insert operations.</p>
  *
- * <p>The {@code Insert} annotation indicates that the annotated repository method requests that one or more entities
- * be inserted into the database. This method must have a single parameter whose type must be one of the following:
+ * <p>The {@code Insert} annotation indicates that the annotated repository method adds the state of one or more
+ * entities to the database. The annotated repository method usually has exactly one parameter whose type is one of
```
njr-11, line 31: The more permissive language is not need for Insert.
```suggestion
 * entities to the database. The annotated repository method must have exactly one parameter whose type is one of
```

api/src/main/java/jakarta/data/repository/Update.java
```diff
@@ -24,51 +24,21 @@
...
-
 /**
- * <p>Annotates a repository method to perform update operations.</p>
+ * <p>Lifecycle annotation for repository methods which perform update operations.</p>
  *
- * <p>The {@code Update} annotation indicates that the annotated repository method requests that one or more entities
- * be updated if found in the database. To request updates to specific entity instances, the annotated
- * repository method must have a single parameter whose type must be one of the following:
+ * <p>The {@code Update} annotation indicates that the annotated repository method updates the state of one or more
+ * entities already held in the database. The annotated repository method usually has exactly one parameter whose type
```
njr-11, line 31:
```suggestion
 * entities already held in the database. To update specific entity instances,
 * the annotated repository method must have exactly one parameter whose type
```

api/src/main/java/jakarta/data/repository/Delete.java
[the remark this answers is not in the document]
gavinking: OK, if you definitely want it on the annotations as well, then fine.

gavinking:
@njr-11 Have a look now.

api/src/main/java/jakarta/data/repository/Delete.java
```diff
@@ -25,27 +25,18 @@
...
 /**
- * <p>Annotates a repository method to perform delete operations.</p>
+ * <p>Lifecycle annotation for repository methods which perform delete operations.</p>
  *
- * <p>The {@code Delete} annotation indicates that the annotated repository method requests one or more
- * entities to be removed from the database. To request deletion of specific entity instances, the annotated
- * repository method must have a single parameter whose type must be one of the following:
+ * <p>The {@code Delete} annotation indicates that the annotated repository method deletes the state of one or more
+ * entities from in the database. The annotated repository method usually has exactly one parameter whose type is one of
+ * the following:
```
njr-11, line 32:
```suggestion
 * entities from in the database. To indicate the deletion of specific entity instances,
 * the annotated repository method must have exactly one parameter whose type is one of
 * the following:
```

gavinking:
I'm now simply reciting:

```
 * <p>Annotations such as {@code @Find}, {@code @Query}, {@code @Insert}, {@code @Update}, {@code @Delete}, and
 * {@code @Save} are mutually-exclusive. A given method of a repository interface may have at most one {@code @Find}
 * annotation, lifecycle annotation, or query annotation.
```

On all the annotations.

njr-11 (commented):
The language with "usually" doesn't work for a specification. It allows for someone to write methods like,
```
@Update
List<Employee> update(List<Employee> firstGroup, List<Employee> secondGroup);
```

I updated the occurrences to switch it to "must", hopefully with better wording that will be acceptable.

I suppose it doesn't really matter so much whether we tell the user not to intermix annotations in the JavaDoc because as long as the specification clearly outlines the requirement at least the providers will enforce it so they find out at run time if they did try to intermix anything.

api/src/main/java/jakarta/data/page/PageRequest.java
```diff
@@ -415,6 +415,13 @@ static <T> PageRequest<T> ofSize(int maxPageSize) {
      */
     PageRequest<T> sortBy(Sort<T> sort1, Sort<T> sort2, Sort<T> sort3, Sort<T> sort4, Sort<T> sort5);
 
+    /**
+     * Convert this object to an equivalent {@link Limit}.
+     */
+    default Limit asLimit() {
+        return Limit.range( 1 + (page()-1) * size(), size() );
+    }
+
```
njr-11, line 424: This doesn't seem related. My guess is that it was accidentally included in this pull.
```suggestion
```

api/src/main/java/jakarta/data/repository/Insert.java
gavinking: True, good point, I will fix that.

api/src/main/java/jakarta/data/page/PageRequest.java
gavinking: Yes, it was an accident, sorry.

gavinking:
> The language with "usually" doesn't work for a specification. 

Well I mean guess I'm struggling to understand what your goal here is. I would have thought that you would want to pick between one of two approaches.

- If you want to let providers experiment and reuse the annotations for new things which are not defined in the spec, then I think "usually" is an excellent word. We're saying: _"here are some things that you're definitely allowed to do and they have well-defined semantics, but you might also be able to do some other things, and we're not going to prescribe that it's an error"_.
- Or, if you want to lock it down, because you're scared of breakage when we add new functionality down the road, then lock it down all the way. And then you can feel perfectly safe adding new well-defined interpretations in a future release, without breaking anybody.

I'm OK either way. I was perfectly happy with "must".

But I don't see how it really makes much sense to say: _"oh we let providers experiment in these one or two specific ways, but we aren't going to tell you want the semantics of these experiments are, and other experiments are verboten"_. 

As a general principle I think it's better when specs say what functionality can be relied on, rather than preventing implementations from doing things which go beyond that functionality.

> I suppose it doesn't really matter so much whether we tell the user not to intermix annotations in the JavaDoc because as long as the specification clearly outlines the requirement at least the providers will enforce it so they find out at run time if they did try to intermix anything.

Right, the only reason to put this in the Javadoc is to tell the user what they're not allowed to do. We don't need to tell them what happens when they do it.

api/src/main/java/jakarta/data/repository/Delete.java
```diff
@@ -25,27 +25,18 @@
...
 
 /**
- * <p>Annotates a repository method to perform delete operations.</p>
+ * <p>Lifecycle annotation for repository methods which perform delete operations.</p>
  *
- * <p>The {@code Delete} annotation indicates that the annotated repository method requests one or more
- * entities to be removed from the database. To request deletion of specific entity instances, the annotated
- * repository method must have a single parameter whose type must be one of the following:
+ * <p>The {@code Delete} annotation indicates that the annotated repository method deletes the state of one or more
+ * entities from in the database. The annotated repository method usually has exactly one parameter whose type is
```
njr-11, line 31: My update to get rid of "usually" ended up disappearing. Here it is again,
```suggestion
 * entities from the database. To delete specific entity instances,
 * the annotated repository method must have exactly one parameter whose type is
```

api/src/main/java/jakarta/data/repository/Insert.java
```diff
@@ -25,54 +25,47 @@
 import java.lang.annotation.Target;
 
 /**
- * <p>Annotates a repository method that inserts entities.</p>
+ * <p>Lifecycle annotation for repository methods which perform insert operations.</p>
  *
- * <p>The {@code Insert} annotation indicates that the annotated repository method requests that one or more entities
- * be inserted into the database. This method must have a single parameter whose type must be one of the following:
+ * <p>The {@code Insert} annotation indicates that the annotated repository method adds the state of one or more
+ * entities to the database. The annotated repository method usually has exactly one parameter whose type is either:
```
njr-11, line 31:
```suggestion
 * entities to the database. The annotated repository method must have exactly one parameter whose type is either:
```

njr-11 (commented):
A couple of my review comments were in an area near other changes and so they went away. Here they are again.  Both are updates to replace of the imprecise "usually" language.

gavinking:
@njr-11 I kinda feel like you're trying to sneak one of those infamous "optional features" without calling it an "optional feature". Now, I personally have no problem at all with optional features. My problem here is that it seems like it's an optional feature which doesn't come with defined semantics. :-)

api/src/main/java/jakarta/data/repository/Delete.java
gavinking: @njr-11 the trouble is I just don't know what this means. What is a "specific entity instance"?

IMO, the `delete()` method with no parameters deletes "specific entity instances". Specifically, it deletes all of them!

I don't think this phrase actually does the work you want it to do.

njr-11:
> > The language with "usually" doesn't work for a specification.
> 
> Well I mean guess I'm struggling to understand what your goal here is. I would have thought that you would want to pick between one of two approaches.

My goal here is to have complete, precise language around all usage where entities are supplied to the methods. That is what we want to specify in version 1.0. That's why I want "must" rather than "usually" when defining that approach.  It will instruct users to the exact set of valid approaches for supplying entity instances.

Technically the `Delete` annotation also has an addition usage without parameters, which is now well-defined.

And then leave the possibility of other usage open for the future.

We aren't defining any optional features, just leaving some areas to expand upon in the future.

api/src/main/java/jakarta/data/repository/Delete.java
njr-11: Yeah, I think my previous language which I've forgotten now was better than this. I gave it another try.

gavinking:
Nathan, how about this formulation:

```
 * <p>Lifecycle annotation for repository methods which perform update operations.</p>
 *
 * <p>The {@code Update} annotation indicates that the annotated repository method updates the state of one or more
 * entities already held in the database.
 * </p>
 * <p>An {@code Update} method might accept an instance or instances of an entity class. In this case, the annotated 
 * repository method should have exactly one parameter whose type is either:
 * </p>
 * <ul>
 *     <li>the class of the entity to be updated, or</li>
 *     <li>{@code Iterable<E>} or {@code E[]} where {@code E} is the class of the entities to be updated.</li>
 * </ul>
 * <p>The annotated method must either be declared {@code void}, or have a return type that is the same as the type of
 * its parameter.
 * <p>
 * All Jakarta Data providers are required to accept an {@code Update} method which conforms to this signature. 
 * Application of the {@code Update} annotation to a method with any other signature is not portable between Jakarta
 * Data providers.
 * </p>
 ```

njr-11:
> Nathan, how about this formulation:

That looks great, with one adjustment to switch "should" to "must".

gavinking:
> switch "should" to "must".

OK, fine.

gavinking:
Alright, there, let's see if we can get consensus on this approach.

njr-11 (approved):
It looks like there is one remaining comment about a grammar fix to make, but we've spent enough time on this already. I'm approving and you can include that change if you want or leave it because it will still be understood either way.

gavinking:
> It looks like there is one remaining comment about a grammar fix to make

Found it.

And I also fixed one more little typo.

otaviojava (approved):

PR #491, opened 2024-02-25
similar to #485, add convenience impls of Page and Slice

gavinking:
This is, again, to ease the job of implementors.

I hope I got the semantics right here; please check!

### DIFF
--- a/api/src/main/java/jakarta/data/page/impl/KeysetAwarePageRecord.java
+++ b/api/src/main/java/jakarta/data/page/impl/KeysetAwarePageRecord.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.page.impl;
+
+import jakarta.data.page.KeysetAwarePage;
+import jakarta.data.page.PageRequest;
+
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Record type implementing {@link KeysetAwarePage}.
+ * This may be used to simplify implementation of a repository interface.
+ *
+ * @param content The page content, that is, the query results, in order
+ * @param cursors A list of {@link PageRequest.Cursor} instances for result,
+ *                in order
+ * @param totalElements The total number of elements across all pages that
+ *                      can be requested for the query
+ * @param pageRequest The {@link PageRequest page request} for which this
+ *                    slice was obtained
+ * @param nextPageRequest A {@link PageRequest page request} for the next
+ *                        page of results
+ * @param previousPageRequest A {@link PageRequest page request} for the
+ *                            previous page of results
+ * @param <T> The type of elements on the page
+ */
+public record KeysetAwarePageRecord<T>
+        (List<T> content, List<PageRequest.Cursor> cursors, long totalElements, PageRequest<T> pageRequest,
+         PageRequest<T> nextPageRequest, PageRequest<T> previousPageRequest)
+        implements KeysetAwarePage<T> {
+
+    @Override
+    public boolean hasContent() {
+        return !content.isEmpty();
+    }
+
+    @Override
+    public int numberOfElements() {
+        return content.size();
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public <E> PageRequest<E> pageRequest(Class<E> entityClass) {
+        return (PageRequest<E>) pageRequest;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public <E> PageRequest<E> nextPageRequest(Class<E> entityClass) {
+        return (PageRequest<E>) nextPageRequest;
+    }
+
+    @Override
+    public Iterator<T> iterator() {
+        return content.iterator();
+    }
+
+    @Override
+    public PageRequest.Cursor getKeysetCursor(int index) {
+        return cursors.get(index);
+    }
+
+    @Override
+    public long totalElements() {
+        return totalElements;
+    }
+
+    @Override
+    public long totalPages() {
+        int size = pageRequest.size();
+        return (totalElements + size - 1) / size;
+    }
+}
--- a/api/src/main/java/jakarta/data/page/impl/PageRecord.java
+++ b/api/src/main/java/jakarta/data/page/impl/PageRecord.java
@@ -40,7 +40,9 @@ public record PageRecord<T>(PageRequest<T> pageRequest, List<T> content, long to
         implements Page<T> {
 
     public PageRecord(PageRequest<T> pageRequest, List<T> content, long totalElements) {
-        this( pageRequest, content, totalElements, content.size() == pageRequest.size() );
+        this( pageRequest, content, totalElements,
+                content.size() == pageRequest.size()
+                        && totalElements > pageRequest.size() * pageRequest.page() );
     }
 
     @Override
--- a/api/src/main/java/jakarta/data/page/impl/PageRecord.java
+++ b/api/src/main/java/jakarta/data/page/impl/PageRecord.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.page.impl;
+
+import jakarta.data.page.Page;
+import jakarta.data.page.PageRequest;
+import jakarta.data.page.Slice;
+
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Record type implementing {@link Slice}.
+ * This may be used to simplify implementation of a repository interface.
+ *
+ * @param pageRequest The {@link PageRequest page request} for which this
+ *                    slice was obtained
+ * @param content The page content
+ * @param totalElements The total number of elements across all pages that
+ *                      can be requested for the query
+ * @param moreResults whether there is a (nonempty) next page of results
+ * @param <T> The type of elements on the page
+ */
+public record PageRecord<T>(PageRequest<T> pageRequest, List<T> content, long totalElements, boolean moreResults)
+        implements Page<T> {
+
+    public PageRecord(PageRequest<T> pageRequest, List<T> content, long totalElements) {
+        this( pageRequest, content, totalElements, content.size() == pageRequest.size() );
+    }
+
+    @Override
+    public boolean hasContent() {
+        return !content.isEmpty();
+    }
+
+    @Override
+    public int numberOfElements() {
+        return content.size();
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public <E> PageRequest<E> pageRequest(Class<E> entityClass) {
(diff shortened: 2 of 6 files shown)


api/src/main/java/jakarta/data/page/impl/PageRecord.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Override
+    public Iterator<T> iterator() {
+        return content.iterator();
+    }
+
+    @Override
+    public long totalPages() {
+        return totalElements / pageRequest.size();
```
njr-11, line 80: With 9 results and size 10, this will return 0 total pages, which is wrong.
With 10 results and size 10, this will return 1 total page, which is correct.
With 11 results and size 10, this will return 1 total page, which is wrong.

I believe something such as the following would correct it,
```suggestion
        int size = pageRequest.size();
        return totalElements / size + (totalElements % size > 0 ? 1 : 0);
```

njr-11 (changes requested):
I added one comment for a total page computation that would need to be fixed, but then I also realized that nextPageRequest can have some different behavior if there is known to not be a next page.  I'm starting to think maybe this part of the implementation shouldn't go into the spec because it is starting to go beyond just directly returning values.

api/src/main/java/jakarta/data/page/impl/PageRecord.java
gavinking: Yeah when I wrote this I didn't realize that we were using 1-based indexing. I wrote it assuming 0-based indexing.

gavinking: Although here I guess it's just a matter of using ceiling instead of floor. That's just a mistake.

api/src/main/java/jakarta/data/page/impl/PageRecord.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    public Iterator<T> iterator() {
+        return content.iterator();
+    }
+
+    @Override
+    public long totalPages() {
+        int size = pageRequest.size();
+        return (totalElements + size - 1) / size;
```
gavinking, line 81: @njr-11 I think this fixes it, no?

api/src/main/java/jakarta/data/page/impl/PageRecord.java
gavinking: [See the updated commit.]

gavinking:
> I also realized that nextPageRequest can have some different behavior if there is known to not be a next page. 

Yes, I knew about this wrinkle which is one reason I left this in Draft, because I knew I needed to think about that and I didn't have time yesterday.

But actually it's well defined by the Javadoc:

> Returns a request for the next page, or `null` if it is known that there is no next page.

So I implemented that.

Fixed!

api/src/main/java/jakarta/data/page/impl/PageRecord.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override @SuppressWarnings("unchecked")
+    public <E> PageRequest<E> pageRequest(Class<E> entityClass) {
+        return (PageRequest<E>) pageRequest;
+    }
+
+    @Override
+    public PageRequest<T> nextPageRequest() {
+        return hasContent() ? pageRequest.next() : null;
```
gavinking, line 65: @njr-11 This implements the documented semantics.

gavinking:
Also, yes, sure, some fancier implementation could retrieve `N+1` rows from the database to be absolutely sure that there really is a next page. But I guess if you want to be fancy you can provide your own implementation of the interface.

api/src/main/java/jakarta/data/page/impl/PageRecord.java
gavinking: wait wait wait no, that was stupid.

api/src/main/java/jakarta/data/page/impl/PageRecord.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override @SuppressWarnings("unchecked")
+    public <E> PageRequest<E> pageRequest(Class<E> entityClass) {
+        return (PageRequest<E>) pageRequest;
+    }
+
+    @Override
+    public PageRequest<T> nextPageRequest() {
+        return moreResults ? pageRequest.next() : null;
```
gavinking, line 70: @njr-11 _now_ it's fixed. Sorry, I'm coding without thinking.

gavinking:
> But I guess if you want to be fancy you can provide your own implementation of the interface.

Scratch that, let's give fancy people what they want :)

gavinking:
I have added a new commit to this PR which makes the internal implementation of `Cursor` available as a `record` in the same package as the `Page` and `Slice` implementations. I don't believe there was any good reason to hide this from clients.

gavinking:
I've now also added implementations of `KeysetAwarePage` and `KeysetAwareSlice`, based on my best understanding of the semantics here.

@otaviojava @njr-11 Please take a look.

njr-11 (commented):
> I have added a new commit to this PR which makes the internal implementation of `Cursor` available as a `record` in the same package as the `Page` and `Slice` implementations. I don't believe there was any good reason to hide this from clients.

I would say it's a good practice not to externalize unless we have a good reason to believe that clients (or the provider) will need it.  It can make sense for a common implementation of Page/Slice to be shared by providers and be accessed from an impl package, but the implementation of Cursor only seems to be needed internally to Jakarta Data.  I would not want to see client code constructing cursors by accessing API in our impl package, but if we externalize it, they will be able to do that. I think we should remove the [second commit](https://github.com/jakartaee/data/pull/491/commits/3ab8a7cea667940a4f1bd613cd90d10c8785ccaa) from this pull.

gavinking:
> the implementation of Cursor only seems to be needed internally to Jakarta Data

OK if that's true I made a mistake somewhere. Let me recheck why I thought I needed one.

gavinking:
> OK if that's true I made a mistake somewhere.

Yep, OK, I see, when I saw it being instantiated directly all through the TCK tests, I guess I just assumed it was needed by implementors too.

But I just misunderstood.

I have put it back where it belongs, and changed its visibility back to package.

njr-11:
> Yep, OK, I see, when I saw it being instantiated directly all through the TCK tests, I guess I just assumed it was needed by implementors too.

The TCK tests don't directly create Cursor themselves anywhere, but the Unit Tests do a lot of that, so that is probably where you saw it.

gavinking:
> > Yep, OK, I see, when I saw it being instantiated directly all through the TCK tests, I guess I just assumed it was needed by implementors too.
> 
> The TCK tests don't directly create Cursor themselves anywhere, but the Unit Tests do a lot of that, so that is probably where you saw it.

Yeahright. That.

api/src/main/java/jakarta/data/page/KeysetCursor.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022,2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022,2023,2024 Contributors to the Eclipse Foundation
```
njr-11: Copyrights need to follow the pattern of first-year-modified, latest-year-modified
```suggestion
 * Copyright (c) 2022,2024 Contributors to the Eclipse Foundation
```

api/src/main/java/jakarta/data/page/impl/KeysetAwarePageRecord.java
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Record type implementing {@link KeysetAwarePage}.
+ * This may be used to simplify implementation of a repository interface.
+ *
+ * @param content The page content, that is, the query results, in order
+ * @param cursors A list of {@link PageRequest.Cursor} instances for result,
+ *                in order
+ * @param totalElements The total number of elements across all pages that
+ *  *                   can be requested for the query
```
njr-11, line 34:
```suggestion
 *                      can be requested for the query
```

gavinking:
> I think we should remove the [second commit](https://github.com/jakartaee/data/pull/491/commits/3ab8a7cea667940a4f1bd613cd90d10c8785ccaa) from this pull.

Done.

api/src/main/java/jakarta/data/page/KeysetCursor.java
gavinking: I ended up just dropping that commit.

api/src/main/java/jakarta/data/page/impl/KeysetAwarePageRecord.java
```diff
@@ -0,0 +1,88 @@
...
+ * @param pageRequest The {@link PageRequest page request} for which this
+ *                    slice was obtained
+ * @param nextPageRequest A {@link PageRequest page request} for the next
+ *                        page of results
+ * @param previousPageRequest A {@link PageRequest page request} for the
+ *                            previous page of results
+ * @param <T> The type of elements on the page
+ */
+public record KeysetAwarePageRecord<T>
+		(List<T> content, List<PageRequest.Cursor> cursors, long totalElements, PageRequest<T> pageRequest,
```
njr-11, line 44: Are you sure it is a good idea to precompute all of the cursors?  When I implemented this, I created each cursor only upon request, under the assumption that users will typically only want 1 or 2 cursors per page (usually for the next or previous page), and not a cursor for every result on the page.

njr-11 (commented):
Sorry for the two separate reviews.  I think the pull was updated while I was reviewing and git needed to refresh. Fortunately I was suspicious of whether it would be able to handle that and had save my final comment to the clipboard. Here it is again because it didn't seem to go through on the previous review,

gavinking:
It's a good question. At least in our implementation I believe it makes no difference because you have already done the work of retrieving that information from the server.

But if you think that other implementations might want to do it lazily, I can replace the `List` with a function ref.

gavinking:
FTR, when I first implemented this I threw away all the keys except for first and last. But then when I saw that the API seems to require that I can send back the key of any row, and @beikov told me that's useful, I changed it to retain all the keys. I already have them so...

njr-11 (approved):

gavinking:
> they can easily supply their own implementation

Yup. And furthermore it's something we can very easily add later without breaking the existing SPI.

gavinking:
Force pushed _again_ to fix the checkstyle violations.

gavinking:
> the implementation of Cursor only seems to be needed internally to Jakarta Data. I would not want to see client code constructing cursors by accessing API in our impl package

Ugh, OK, after all that, I found where it's needed. The issue is that `KeysetAwareSliceRecord` needs to be able to return a `Cursor` for each row of the result list. That's the contract of `KeysetAwareSliceRecord .getKeysetCursor()`. The implementation needs to instantiate those objects. It has nowhere else to get them from, AFAICT.

So if you don't want to make `KeysetCursor` public, I would propose adding a static method somewhere to instantiate it.

gavinking:
> So if you don't want to make KeysetCursor public, I would propose adding a static method somewhere to instantiate it.

See @2fa8ef2

gavinking:
> The implementation needs to instantiate those objects. It has nowhere else to get them from, AFAICT.

An alternative solution would be to remove `KeysetAwareSlice.getKeysetCursor(int)` for now.

njr-11:
> An alternative solution would be to remove `KeysetAwareSlice.getKeysetCursor(int)` for now.

That method is the most convenient way for a user to obtain a cursor from any starting point when they already have some results. I would say that wanting to be able to share a Cursor implementation among providers doesn't justify taking that away from users.

I like your other approach of adding `Cursor.forKeyset`,

> > So if you don't want to make KeysetCursor public, I would propose adding a static method somewhere to instantiate it.
> 
> See [@2fa8ef2](https://github.com/jakartaee/data/pull/491/commits/2fa8ef2bca5439f366bafef7e92cc4c67ffcf835)

gavinking:
>  I would say that wanting to be able to share a `Cursor` implementation among providers doesn't justify taking that away from users.

Oh, that wasn't really the idea: the idea was that as per our other discussion it would have relieved us from having to materialize `Cursor` instances for anything other than the first/last row, and by side effect providers would never _need_ to instiate a `Cursor`. 

But I wasn't arguing for it, just mapping out the solution space.

> I like your other approach of adding `Cursor.forKeyset`

Good, that solves the problem.

njr-11 (approved):
Approving - Everything looks good.

gavinking:
Folks please review https://github.com/jakartaee/data/pull/491/commits/a6b6c1c01450f512b99741223c870142d83deb7f which I just pushed, a better implementation of "more results" for both `PageRecord` and `SliceRecord`.

gavinking:
> Folks please review [a6b6c1c](https://github.com/jakartaee/data/commit/a6b6c1c01450f512b99741223c870142d83deb7f) which I just pushed, a better implementation of "more results" for both `PageRecord` and `SliceRecord`.

Oh well, Otavio just merged it, as I was typing this message.

njr-11:
> > Folks please review [a6b6c1c](https://github.com/jakartaee/data/commit/a6b6c1c01450f512b99741223c870142d83deb7f) which I just pushed, a better implementation of "more results" for both `PageRecord` and `SliceRecord`.
> 
> Oh well, Otavio just merged it, as I was typing this message.

I suppose that means he was very strongly in favor of it.
The updates look good to me.

gavinking:
Yeah it would be great except for the fact that it had a but in it (fixed in the new PR).

PR #496, opened 2024-02-28
add Cursor.elements()

gavinking:
Currently, `PageRequest.Cursor` is not even `Iterable`, which means that any time you want to read its elements you have to use an archaic indexed `for` loop. If you want to actually pass the key values as a list, you have to repackage them manually.

This adds an `elements()` method which automatically packages the elements as an unmodifiable list.

### DIFF
--- a/api/src/main/java/jakarta/data/page/KeysetCursor.java
+++ b/api/src/main/java/jakarta/data/page/KeysetCursor.java
@@ -18,6 +18,7 @@
 package jakarta.data.page;
 
 import java.util.Arrays;
+import java.util.List;
 
 /**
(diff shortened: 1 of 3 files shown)


otaviojava (approved):
I like it. 
IMHO: the  `getKeysetElement` is a candidate for removing as well.

gavinking:
> IMHO: the  getKeysetElement is a candidate for removing as well.

Yeah, I agree. Even f only because the name is a mouthful :)

njr-11 (approved):

gavinking:
Thanks folks.

PR #497, opened 2024-02-28
Update API returning Optional instead of Null

otaviojava:
# Changes

- For consistency, I've updated the return to be Optional instead of **null**.

Once we are using optional when the return might be null, such as, it would be great to keep the consistency here as well.

### DIFF
--- a/api/src/main/java/jakarta/data/page/Slice.java
+++ b/api/src/main/java/jakarta/data/page/Slice.java
@@ -20,6 +20,7 @@ package jakarta.data.page;
 import jakarta.data.Streamable;
 import jakarta.data.repository.Query;
 import java.util.List;
+import java.util.Optional;
 
 /**
  * <p>A slice contains the data that is retrieved for a page request.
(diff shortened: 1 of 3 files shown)


otaviojava:
@njr-11 don't forget of review this one.

api/src/main/java/jakarta/data/page/Slice.java
```diff
@@ -98,15 +99,15 @@ public interface Slice<T> extends Streamable<T> {
     <E> PageRequest<E> pageRequest(Class<E> entityClass);
 
     /**
-     * Returns a request for the {@link PageRequest#next() next} page, or <code>null</code> if it is known that there is no next page.
+     * Returns a request for the {@link PageRequest#next() next} page, or {@link Optional#empty()} if it is known that there is no next page.
      *
      * @return a request for the next page.
      */
-    PageRequest<T> nextPageRequest();
+    Optional<PageRequest<T>> nextPageRequest();
```
otaviojava, line 106: @njr-11 it still require this validation.

njr-11: That's only because the other pull missed the part about raising an exception from nextPageRequest/previousPageRequest instead of returning null.  I can make sure that gets included when updating the TCK to cover hasNext/hasPrevious.   Then this will follow the pattern of. java.util.Iterator and java.util.ListIterator where you must check hasNext (or hasPrevious) before requesting the next/previous

njr-11:
@otaviojava can we close this?  With other changes that have gone in, it is no longer possible to get a null from `nextPageRequest`.

PR #502, opened 2024-02-28
CrudRepository update contradicts new lifecycle  method javadoc and TCK broken

njr-11:
The requirements of the `Update` annotation are now in conflict with the requirements of the `CrudRepository` update methods, which are annotated with that annotation.  The `Update` annotation disallows the return types that these methods currently have and it has an opposite requirement for raising `OptimisticLockingFailureException` rather than indicating the non-matching condition in the method return value.  We will need to make `CrudRepository` align its own requirements to that of the `Update` annotation so that the spec doesn't contradict itself.  Also, I found another spot where the TCK is broken by the `Update` annotation changes.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022,2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022,2024 Contributors to the Eclipse Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


gavinking (approved):

api/src/main/java/jakarta/data/repository/CrudRepository.java
```diff
@@ -163,13 +166,15 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
...
+     * @return updated entity instances, in the same order as the supplied entities,
+     *         and including all automatically generated values, updated versions, and
+     *         incremented values which changed as a result of the update.
+     * @throws OptimisticLockingFailureException If any of the supplied entities is not found in the database
+     *         or has a version that differs from the version in the database.
+     * @throws NullPointerException if either the supplied {@code Iterable} is null or any element is null.
      */
     @Update
-    int updateAll(Iterable<T> entities);
+    Iterable<T> updateAll(Iterable<T> entities);
```
otaviojava, line 178: Update with the value instead int.
I love it!!!

PR #506, opened 2024-02-29
add boolean Page.hasNextPage()

gavinking:
A `Page` always knows if there are more results. But the only way to really get that information is to call `page.nextPageRequest()`. Thus, the idiom for iteration is:

```java
Page<Book> page = library.books("%", pageRequest);
...
while (page.nextPageRequest() != null) {
    page = library.books("%", slice.nextPageRequest());
    ...
}
```

which is quite nonbeautiful.

This proposal adds a `last()` method to `Page`, so that the idiom becomes:

```java
Page<Book> page = library.books("%", pageRequest);
...
while (!page.last()) {
    page = library.books("%", slice.nextPageRequest());
    ...
}
```

Now, if it were my call I would also add:

- `last()` to `Slice`, not just `Page`, and
- `first()` to `KeysetAwareSlice`

but I wanted to see how you guys really react to just this.

### DIFF
--- a/api/src/main/java/jakarta/data/page/KeysetAwareSlice.java
+++ b/api/src/main/java/jakarta/data/page/KeysetAwareSlice.java
@@ -147,7 +147,8 @@ public interface KeysetAwareSlice<T> extends Slice<T> {
 
     /**
      * Returns {@code true} when it is possible to navigate to a previous
-     * page of results.
+     * page of results or if it is necessary to request a previous page in order to
+     * determine whether there are more previous results.
      * @return {@code false} if the current page is empty or if it is known
      *         that there is not a previous page.
      */
--- a/api/src/main/java/jakarta/data/page/KeysetAwareSlice.java
+++ b/api/src/main/java/jakarta/data/page/KeysetAwareSlice.java
@@ -146,6 +146,14 @@ public interface KeysetAwareSlice<T> extends Slice<T> {
     PageRequest.Cursor getKeysetCursor(int index);
 
     /**
+     * Returns {@code true} when it is possible to navigate to a previous
+     * page of results.
+     * @return {@code false} if the current page is empty or if it is known
+     *         that there is not a previous page.
+     */
+    boolean hasPrevious();
+
+    /**
      * <p>Creates a request for the next page
      * in a forward direction from the current page. This method computes a
      * keyset cursor from the last entity of the current page and includes the
--- a/api/src/main/java/jakarta/data/page/Page.java
+++ b/api/src/main/java/jakarta/data/page/Page.java
@@ -42,14 +42,6 @@ package jakarta.data.page;
 public interface Page<T> extends Slice<T> {
 
     /**
-     * Returns {@code true} if it is known that there are more results or that it is
-     * necessary to request a next page to determine whether there are more results, so that
-     * {@link #nextPageRequest()} will definitely not return {@code null}.
-     * @return {@code false} if this is the last page of results.
-     */
-    boolean hasNext();
-
-    /**
      * Returns the total number of elements across all pages that can be requested for the query.
      * @return the total number of elements across all pages.
      */
--- a/api/src/main/java/jakarta/data/page/Page.java
+++ b/api/src/main/java/jakarta/data/page/Page.java
@@ -42,6 +42,13 @@ package jakarta.data.page;
 public interface Page<T> extends Slice<T> {
 
     /**
+     * Returns {@code true} if it is known that there are more results, so that
+     * {@link #nextPageRequest()} will definitely not return {@code null}.
+     * @return {@code false} if this the last page of results
+     */
+    boolean hasNextPage();
+
+    /**
      * Returns the total number of elements across all pages that can be requested for the query.
      * @return the total number of elements across all pages.
      */
--- a/api/src/main/java/jakarta/data/page/Page.java
+++ b/api/src/main/java/jakarta/data/page/Page.java
@@ -46,7 +46,7 @@ public interface Page<T> extends Slice<T> {
      * {@link #nextPageRequest()} will definitely not return {@code null}.
      * @return {@code false} if this the last page of results
      */
-    boolean hasNextPage();
+    boolean hasNext();
 
     /**
      * Returns the total number of elements across all pages that can be requested for the query.
--- a/api/src/main/java/jakarta/data/page/Page.java
+++ b/api/src/main/java/jakarta/data/page/Page.java
@@ -42,9 +42,10 @@ package jakarta.data.page;
 public interface Page<T> extends Slice<T> {
 
     /**
-     * Returns {@code true} if it is known that there are more results, so that
+     * Returns {@code true} if it is known that there are more results or that it is
+     * necessary to request a next page to determine whether there are more results, so that
      * {@link #nextPageRequest()} will definitely not return {@code null}.
-     * @return {@code false} if this the last page of results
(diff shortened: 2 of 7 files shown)


gavinking:
Actually `last()` is a terrible name, it sounds like you're asking for the last page. Since we're not using JavaBeans naming, I'm not sure what to call it. `hasNextPage()`, I suppose. (With the opposite polarity.)

njr-11:
I haven't looked at the code changes yet, but I like this better than the proposal to make nextPageRequest return an Optional.  Some suggestions for names are:

`hasNext()`
and
`hasPrevious()`

which will line up with the our existing method names `nextPageRequest` and `previousPageRequest`,
and also exactly match the method names that are used on `java.util.Iterator` and `java.util.ListIterator`.
This pattern lines up with the Iterator pattern of needing to check if the next exists before obtaining it.
Similarly, we can have `nextPageRequest`/`previousPageRequest` raise `NoSuchElementException` if there is not a next/previous, which will eliminate the possibility of nulls.

That said, we do need to write the documentation to allow for the possibility that a page might come back empty.  For example, if using the slice pattern where the total is not known and we currently have a full page and so return a next page request, the next page might actually come back empty because there actually weren't any more results.  This should be fine - the user just needs to be aware to allow for the possibility of the final page being empty.

gavinking:
Yes, `hasNext()` is fine by me. Shorter. Which is good.

api/src/main/java/jakarta/data/page/Page.java
```diff
@@ -41,6 +41,13 @@
  */
 public interface Page<T> extends Slice<T> {
 
+    /**
+     * Returns {@code true} if it is known that there are more results, so that
+     * {@link #nextPageRequest()} will definitely not return {@code null}.
+     * @return {@code false} if this the last page of results
```
njr-11, line 47:
```suggestion
     * Returns {@code true} if it is known that there are more results or that it is
     * necessary to request a next page to determine whether there are more results, so that
     * {@link #nextPageRequest()} will definitely not return {@code null}.
     * @return {@code false} if this is the last page of results.
```

api/src/main/java/jakarta/data/page/Page.java
```diff
@@ -41,6 +41,13 @@
...
+     */
+    boolean hasNextPage();
```
njr-11, line 49:
```suggestion
    boolean hasNext();
```

api/src/main/java/jakarta/data/page/impl/PageRecord.java
```diff
@@ -55,6 +55,11 @@ public int numberOfElements() {
         return content.size();
     }
 
+    @Override
+    public boolean hasNextPage() {
```
njr-11, line 59:
```suggestion
    public boolean hasNext() {
```

njr-11 (commented):
This should be put onto Slice rather than Page (unless we get rid of Slice as another issue discusses) because the pattern of avoiding null PageRequest applies to Slice as well.

The JavaDoc for hasNext is already well written and it just needs a minor adjustment to allow for the scenario where the next page needs to be attempted to find out whether there are more results.

If we add this, we should also include the corresponding `hasPrevious()` for keyset cursor based pagination.

gavinking:
> This should be put onto Slice rather than Page (unless we get rid of Slice as another issue discusses) because the pattern of avoiding null PageRequest applies to Slice as well.
>
> If we add this, we should also include the corresponding hasPrevious() for keyset cursor based pagination.

Yes, I know, this was just a straw man.

gavinking:
I just put `hasNext()` and `hasPrevious()` on the places they belong. Please review.

api/src/main/java/jakarta/data/page/KeysetAwareSlice.java
```diff
@@ -145,6 +145,14 @@ public interface KeysetAwareSlice<T> extends Slice<T> {
      */
     PageRequest.Cursor getKeysetCursor(int index);
 
+    /**
+     * Returns {@code true} when it is possible to navigate to a previous
+     * page of results.
```
njr-11, line 150: I know it would be an unusual usage pattern,  but it is technically possible to start from a cursor somewhere in the middle and navigate to previous pages from that, in which case, an implementation might not know whether or not there actually is a previous page.

```suggestion
     * Returns {@code true} when it is possible to navigate to a previous
     * page of results or if it is necessary to request a previous page in order to
     * determine whether there are more previous results.
```

njr-11 (approved):

api/src/main/java/jakarta/data/page/KeysetAwareSlice.java
```diff
@@ -145,6 +145,15 @@ public interface KeysetAwareSlice<T> extends Slice<T> {
...
     PageRequest.Cursor getKeysetCursor(int index);
 
+    /**
+     * Returns {@code true} when it is possible to navigate to a previous
+     * page of results or if it is necessary to request a previous page in order to
+     * determine whether there are more previous results.
+     * @return {@code false} if the current page is empty or if it is known
+     *         that there is not a previous page.
+     */
+    boolean hasPrevious();
```
otaviojava, line 155: Using false to both makes it unclear if it is empty or the database does not know the previous page.

Are we sure to proceed this way?

I mean, if I receive **false** I need to check anyway...

api/src/main/java/jakarta/data/page/Slice.java
```diff
@@ -74,6 +74,14 @@ public interface Slice<T> extends Streamable<T> {
      */
     int numberOfElements();
 
+    /**
+     * Returns {@code true} if it is known that there are more results or that it is
+     * necessary to request a next page to determine whether there are more results, so that
+     * {@link #nextPageRequest()} will definitely not return {@code null}.
+     * @return {@code false} if this is the last page of results.
+     */
+    boolean hasNext();
```
otaviojava, line 83: The same point here.

Include the UnsupportedOperationException if the database does not know.

njr-11: No -that wrecks the whole pattern.  The point is that if the provider does not know if there is going to be a next page, then we want the user to request it to find out and the user should be able to do that without having to experience exceptions.

gavinking: @otaviojava I think the idea here is that if you have a full last page of results you might not know that it is the last page. What @njr-11 is trying to permit here is that there be an "empty last page". Which I think is fine and correct. Providers can in principle use a "fancy" implementation that avoids this, but I don't think we should mandate such fanciness.

api/src/main/java/jakarta/data/page/KeysetAwareSlice.java
njr-11: > Using false to both makes it unclear if it is empty or the database does not know the previous page.
> 
> Are we sure to proceed this way?
> 
> I mean, if I receive **false** I need to check anyway...
> 
> Maybe an UnsupportedOperationException when it does not know

Otavio, are you sure you are reading this correctly?  When the value is `false` is when you should never try to request the page because it is known to not exist.   An empty page has no cursors and therefore cannot have a previous.  In the other case it is stated there is no previous page.  If you get `false` and then ask for `previousPageableRequest` anyways, you can expect to get an exception from the latter.  There is no point in trying.

otaviojava: That was my bad. Thank you for clarifying it.

PR #510, opened 2024-02-29
clarify combined sort and keyset cursors

### LINKED ISSUE #508: [clarification]: is @OrderBy really allowed in combination with key-based pagination
### Specification

@OrderBy annotation, key-based pagination

whether this actually works.

### Additional information

Correct me if I'm wrong, but I'm pretty sure this can't work:

> When multiple `OrderBy` annotations are specified on a repository method, the precedence for sorting follows the order in which the `OrderBy` annotations are specified, and after that follows any sort criteria that is supplied dynamically by `Sort` parameters, any `Order` parameter, or by a `PageRequest` parameter with sorting criteria.

I have not thought about it very hard, and perhaps I'm just being dense, but it seems to me that in key-based pagination, the key must define the complete ordering of the query. I think it breaks pagination to _prepend_ additional ordering criteria to the key. And _appending_ would be have no effect, since the key must be a unique key and determine a total order.

Is that right, or am I just being stupid?

njr-11:
fixes #508

Adds documentation to hopefully clarify that the combined sort criteria are used for the keyset cursor.
@gavinking please make suggestions to text if you have ideas to improve the wording.

### DIFF
--- a/api/src/main/java/jakarta/data/page/KeysetAwareSlice.java
+++ b/api/src/main/java/jakarta/data/page/KeysetAwareSlice.java
@@ -24,15 +24,16 @@ import jakarta.data.Sort;
  * <p>A slice of data with the ability to create a cursor from the
  * keyset of each entity in the slice.</p>
  *
- * <p>Keyset pagination is a form of pagination that aims to reduce the
+ * <p>Keyset cursor pagination is a form of pagination that aims to reduce the
  * possibility of missed or duplicate results by making the request for
  * each subsequent page relative to the observed values of entity properties
- * from the current page. This list of values is referred to as the keyset
- * and consists of the values of entity properties that are in the sort criteria
- * of the repository method. The combination of sort criteria must uniquely
- * identify each entity. The keyset values can be from the last entity on the page
- * (for pagination in a forward direction) or first entity on the page
- * (if requesting pages in a reverse direction),
+ * from the current page. This list of values is referred to as the keyset cursor
+ * and is an ordered list of the values of entity properties that are in the
+ * combined sort criteria of the repository method, in the same order of precedence.
+ * The combination of sort criteria must uniquely identify each entity.
+ * The keyset values can be from the last entity on the page
+ * (for the next page in a forward direction) or first entity on the page
+ * (if requesting the previous page),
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/data/page/KeysetAwareSlice.java
```diff
@@ -24,15 +24,16 @@
...
- * identify each entity. The keyset values can be from the last entity on the page
- * (for pagination in a forward direction) or first entity on the page
- * (if requesting pages in a reverse direction),
+ * from the current page. This list of values is referred to as the keyset cursor
+ * and is an ordered list of the values of entity properties that are in the
+ * combined sort criteria of the repository method, in the same order of precedence.
+ * The combination of sort criteria must uniquely identify each entity.
+ * The keyset values can be from the last entity on the page
+ * (for the next page in a forward direction) or first entity on the page
+ * (if requesting the previous page),
```
gavinking, line 36: This is much better!

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -1470,6 +1472,43 @@ else
...
+----
+PageRequest<Car> page2Request = Order.by(_Car.price.desc(), _Car.vin.asc())
+                                     .page(2) // cosmetic when using a cursor
+                                     .size(25)
+                                     .afterKeyset(lastCar.vehicleCondition,
+                                                  lastCar.price,
+                                                  lastCar.vin);
+KeysetAwarePage<Car> page2 = cars.find(make, model, page2Request);
+----
+
```
gavinking, line 1511: This is fine, but I guess my discomfort with almost the whole of section 4.9 is that it reads like a tutorial explaining to a user how they can use JD to do stuff. It doesn't read like a specification, and it's very difficult to distinguish the prescriptive requirements. In fairness though I think the information is there and it's probably OK.

gavinking:
P.S. I think the right way to describe the "keyset" is that it imposes a _total order_ on the result set. I see some language talking about it being a unique key, and some other language talking about it imposing a "deterministic order", but it seems to me that "total order" is the right word.

otaviojava (approved):

otaviojava (approved):

Issue #512, opened 2024-03-01
[clarification]: return types of method with built-in lifecycle annotations


gavinking:
### Specification

@Insert, @Update, @Delete, @Save

### I need clarification on ...

subtypes of `Iterable`.

### Additional information

In the module-level Javadoc, I found this statement:

> Whenever the return type is an `Iterable` subtype that is a concrete class, the class must have a public default constructor and support `Collection#addAll` or `Collection#add`.

Nothing like this is mentioned anywhere else in the spec of Javadoc, as far as I can see, so I think this is probably just obsolete.

But on the other hand, if the idea is that these lifecycle methods should accept/return _subtypes_ of `Iterable`, then I would say that, similar to #503, we should limit this to just `List`. I certainly don't think we need to allow arbitrary subtypes here.

njr-11:
> > Whenever the return type is an `Iterable` subtype that is a concrete class, the class must have a public default constructor and support `Collection#addAll` or `Collection#add`.
> 
> Nothing like this is mentioned anywhere else in the spec of Javadoc, as far as I can see, so I think this is probably just obsolete.

It will become obsolete if #511 is merged which removes the following table entry from the JavaDoc that also states the ability to use any `Collection` subtype that can be instantiated and populated according to those rules:

```
<td><code>find...By...</code></td>
 * <td><code>Collection</code> subtypes</td>
 * <td>The subtype must have a public default constructor and support <code>addAll</code> or <code>add</code></td></tr>
 ```

gavinking:
Perfect.

PR #513, opened 2024-03-01
remove an obsolete requirement

gavinking:
I suspect that this sentence is not supposed to be there.

See #512.


njr-11 (approved):
Approving under the assumption that this is paired with #511 that removes the other location where Collection subtypes are required for find method results.  Please ensure these pulls are paired so that the spec remains consistent.

otaviojava (approved):

Issue #458, opened 2024-02-20
Query Language to Jakarta Data Proposal [Vote]


otaviojava:
### As a ...

- [x] Application user/user of the configuration itself
- [x] API user (application developer)
- [ ] SPI user (container or runtime developer)
- [ ] Specification implementer

### I need to be able to ...

The Jakarta Data project aims to introduce a unified Query Language specification to facilitate seamless data querying across various databases, primarily focusing on SQL and NoSQL. The recent delay in the release of Jakarta presents an opportunity to include more features in the specification before the official release.

### Which enables me to ...

The primary goal of this proposal is to initiate a discussion and a vote regarding the inclusion of the Jakarta Data project in the Jakarta EE 1.0 release.

**Key Features:**

1. **Query Language Specification:** The Jakarta Data project will introduce a Query Language specification tailored to work across different database systems, focusing on supporting SQL and NoSQL databases.

2. **Supported Operations:** Initially, the specification will support essential query operations such as selecting, deleting, inserting, and updating data. Subsequent versions may incorporate additional query functionalities.

3. **Compatibility:** A significant challenge is ensuring compatibility between SQL and NoSQL databases. For instance, while certain operations like joins may not be universally supported across all databases, efforts will be made to accommodate such operations optionally or provide alternative approaches.

4. **Annotation Support:** The project will include annotation support, such as the `@Query` annotation, allowing developers to specify whether to use native queries from the provider or the Jakarta Data Query Language.

### Additional information

```java
@Repository
public interface Garage {

  @Query("FOR car IN cars FILTER car._key == key RETURN car", native=true)
  Optional<Car> findNative(String key);
  // Running a native query language using ArangoDB query language

  // Running a Jakarta Data query
  @Query("select * from Car where key = @key")
  Optional<Car> findDataQuery(String key);
}
```

**Scope for Discussion:**

1. **Inclusion in Jakarta EE 1.0:** Discuss whether it is appropriate to include the Jakarta Data project in the Jakarta EE 1.0 release, considering its potential benefits and impact on the ecosystem.

2. **Compatibility and Extensibility:** Evaluate the proposed Query Language specification's compatibility with different database systems and its extensibility for future enhancements.

**BNF to Start:**

```
<Query> ::= <Select> | <Delete> | <Insert> | <Update>

<Select> ::= 'select' <Fields> 'from' <Entity> <Where>? <Skip>? <Limit>? <Order>? EOF
[...]

PR #520, opened 2024-03-05
Initial specification of JDQL

gavinking:
Pushing this to collect feedback/suggestions in a disciplined way. Plenty of things still to clarify, of course.

See #458.

### DIFF
--- a/spec/src/antlr/JDQL.g4
+++ b/spec/src/antlr/JDQL.g4
@@ -1,0 +1,92 @@
+grammar JDQL;
+
+statement : select_statement | update_statement | delete_statement;
+
+select_statement : select_clause? from_clause? where_clause? orderby_clause?;
+update_statement : 'UPDATE' entity_name set_clause where_clause?;
+delete_statement : 'DELETE' from_clause where_clause?;
+
+from_clause : 'FROM' entity_name;
+
+where_clause : 'WHERE' conditional_expression;
+
+set_clause : 'SET' update_item (',' update_item)*;
+update_item : state_field_path_expression '=' (scalar_expression | 'NULL');
+
+select_clause : 'SELECT' select_list;
+select_list
+    : scalar_expression (',' scalar_expression)* // could limit to single select item and/or to state_field_path_expression
+    | aggregate_expression
+    ;
+aggregate_expression : 'COUNT' '(' '*' ')';
+
+orderby_clause : 'ORDER' 'BY' orderby_item (',' orderby_item)*;
+orderby_item : state_field_path_expression ('ASC' | 'DESC');
+
+conditional_expression
+    // highest to lowest precedence
+    : '(' conditional_expression ')'
+    | null_comparison_expression
+    | in_expression
+    | between_expression
+    | like_expression
+    | comparison_expression
+    | 'NOT' conditional_expression
+    | conditional_expression 'AND' conditional_expression
+    | conditional_expression 'OR' conditional_expression
+    ;
+
+comparison_expression : scalar_expression ('=' | '>' | '>=' | '<' | '<=' | '<>') scalar_expression;
+between_expression : scalar_expression 'NOT'? 'BETWEEN' scalar_expression 'AND' scalar_expression;
+like_expression : scalar_expression 'NOT'? 'LIKE' STRING;
+
+in_expression : state_field_path_expression 'NOT'? 'IN' '(' in_item (',' in_item)* ')';
+in_item : literal | enum_literal | input_parameter; // could simplify to just literal
+
+null_comparison_expression : state_field_path_expression 'IS' 'NOT'? 'NULL';
+
+scalar_expression
+    // highest to lowest precedence
+    : '(' scalar_expression ')'
+    | primary_expression
+    | ('+' | '-') scalar_expression
+    | scalar_expression ('*' | '/') scalar_expression
+    | scalar_expression ('+' | '-') scalar_expression
+    | scalar_expression '||' scalar_expression
+    ;
+
+primary_expression
+    : function_expression
+    | special_expression
+    | state_field_path_expression
+    | enum_literal
+    | input_parameter
+    | literal
+    ;
+
+function_expression
+    : 'ABS' '(' scalar_expression ')'
+    | 'LENGTH' '(' scalar_expression ')'
+    | 'LOWER' '(' scalar_expression ')'
+    | 'UPPER' '(' scalar_expression ')'
+    | 'LEFT' '(' scalar_expression ',' scalar_expression ')'
+    | 'RIGHT' '(' scalar_expression ',' scalar_expression ')'
+    ;
+
+special_expression
+    : 'LOCAL' 'DATE'
+    | 'LOCAL' 'DATETIME'
+    | 'TRUE'
+    | 'FALSE'
+    ;
+
+state_field_path_expression : IDENTIFIER ('.' IDENTIFIER)*;
+
+entity_name : IDENTIFIER; // no ambiguity
+
+enum_literal : IDENTIFIER; // ambiguity with state_field_path_expression resolvable semantically
+
+input_parameter : ':' IDENTIFIER | '?' INTEGER;
+
+literal : STRING | INTEGER | DOUBLE;
+
--- a/spec/src/antlr/JDQL.g4
+++ b/spec/src/antlr/JDQL.g4
@@ -15,7 +15,7 @@ update_item : state_field_path_expression '=' (scalar_expression | 'NULL');
 
 select_clause : 'SELECT' select_list;
 select_list
-    : scalar_expression (',' scalar_expression)* // could limit to single select item and/or to state_field_path_expression
+    : state_field_path_expression
     | aggregate_expression
     ;
 aggregate_expression : 'COUNT' '(' '*' ')';
--- a/spec/src/antlr/JDQL.g4
+++ b/spec/src/antlr/JDQL.g4
@@ -76,7 +76,7 @@ function_expression
 special_expression
     : 'LOCAL' 'DATE'
     | 'LOCAL' 'DATETIME'
-    | 'LOCAL' 'IME'
+    | 'LOCAL' 'TIME'
     | 'TRUE'
     | 'FALSE'
     ;
--- a/spec/src/antlr/JDQL.g4
+++ b/spec/src/antlr/JDQL.g4
@@ -76,6 +76,7 @@ function_expression
 special_expression
     : 'LOCAL' 'DATE'
     | 'LOCAL' 'DATETIME'
+    | 'LOCAL' 'IME'
     | 'TRUE'
     | 'FALSE'
     ;
--- a/spec/src/antlr/JDQL.g4
+++ b/spec/src/antlr/JDQL.g4
@@ -18,7 +18,7 @@ select_list
     : state_field_path_expression
     | aggregate_expression
     ;
-aggregate_expression : 'COUNT' '(' '*' ')';
+aggregate_expression : 'COUNT' '(' 'THIS' ')';
 
 orderby_clause : 'ORDER' 'BY' orderby_item (',' orderby_item)*;
 orderby_item : state_field_path_expression ('ASC' | 'DESC');
--- a/spec/src/main/asciidoc/query-language.asciidoc
+++ b/spec/src/main/asciidoc/query-language.asciidoc
@@ -1,0 +1,337 @@
+== Jakarta Data Query Language
+
+The Jakarta Data Query Language (JDQL) is a simple language designed to be used inside the `@Query` annotation to specify the semantics of query methods of Jakarta Data repositories. The language is in essence a subset of the widely-used Jakarta Persistence Query Language (JPQL), and thus a dialect of SQL. But, consistent with the goals of Jakarta Data, it is sufficiently limited in functionality that it is easily implementable across a wide variety of data storage technologies. Thus, the language defined in this chapter excludes features of JPQL which, while useful when the target datasource is a relational database, cannot be easily implemented on all non-relational datastores. In particular, the `from` clause of a Jakarta Data query may contain only a single entity.
+
+NOTE: A Jakarta Data provider backed by access to a relational database might choose to allow the use of a much larger subset of JPQL&mdash;or even the whole language&mdash;via the `@Query` annotation. Such extensions are not required by this specification.
+
+=== Type system
+
+Every expression in a JDQL query is assigned a Java type. An implementation of JDQL is required to support the Java types listed in <<Basic Types>>, that is: primitive types, `String`, `LocalDate`, `LocalDateTime`, `LocalTime`, and `Instant`, `java.util.UUID`, `java.math.BigInteger` and `java.math.BigDecimal`, `byte[]`, and user-defined `enum` types.
+
+NOTE: An implementation of JDQL is permitted and encouraged to support additional types. Use of such types is not guaranteed to be portable between implementations.
+
+The interpretation of an operator expression or literal expression of a given type is given by the interpretation of the equivalent expression in Java. However, the precise behavior of some queries might vary depending on the native semantics of queries on the underlying datastore. For example, numeric precision and overflow, string collation,and integer division are permitted to depart from the semantics of the Java language.
+
+NOTE: This specification should not be interpreted to mandate an inefficient implementation of query language constructs in cases where the native behavior of the database varies from Java in such minor ways. That said, portability between Jakarta Data providers is maximized when their behavior is closest to the Java language.
+
+=== Lexical structure
+
+Lexical analysis requires recognition of the following token types:
+
+- keywords (reserved identifiers),
+- regular identifiers,
+- named and ordinal parameters,
+- operators and punctuation characters,
+- literal strings, and
+- integer and decimal number literals.
+
+==== Identifiers and keywords
+
+An _identifier_ is any legal Java identifier which is not a keyword. Identifiers are case-sensitive: `hello`, `Hello`, and `HELLO` are distinct identifiers.
+
+In the JDQL grammar, identifiers are labelled with the `IDENTIFIER` token type.
+
+The following identifiers are _keywords_: `select`, `update`, `set`, `delete`, `from`, `where`, `order`, `by`, `asc`, `desc`, `not`, `and`, `or`, `between`, `like`, `in`, `null`, `local`, `true`, `false`. In addition, every reserved identifier listed in section 4.4.1 of the Jakarta Persistence specification version 3.2 is also considered a reserved identifier. Keywords and other reserved identifiers are case-insensitive: `null`, `Null`, and `NULL` are three ways to write the same keyword.
+
+NOTE: Use of a reserved identifier as a regular identifier in JDQL might be accepted by a given Jakarta Data provider, but such usage is not guaranteed to be portable between providers.
+
+==== Parameters
+
+A _named parameter_ is a legal Java identifier prefixed with the `:` character, for example, `:name`.
+
+An _ordinal parameter_ is a decimal integer prefixed with the `?` character, for example, `?1`.
+
+==== Operators and punctuation
+
+The character sequences `+`,`-`,`*`,`/`,`||`,`=`,`<`,`>`,`<>`,`&lt;=`,`>=` are _operators_.
+
+The characters `(`,`)`, and `,` are _punctuation characters_.
+
+==== String literals
+
+A _literal string_ is a character sequence quoted using the character `'`.
+
+A single literal `'` character may be included within a string literal by self-escaping it, that is, by writing `''`. For example, the string literal ``'Furry''s theorem has nothing to do with furries.'`` evaluates to the string `pass:[Furry's theorem has nothing to do with furries.]`.
+
+In the grammar, literal strings are labelled with the `STRING` token type.
+
+==== Numeric literals
+
+Numeric literals come in two flavors:
+
+- any legal Java decimal literal of type `int` or `long` is an _integer literal_, and
+- any legal Java literal of type `float` or `double` is a _decimal literal_.
+
+In the grammar, integer and decimal literals are labelled with the `INTEGER` and `DOUBLE` token types respectively.
+
+NOTE: JDQL does not require support for literals written octal or hexadecimal.
+
+==== Whitespace
+
+The characters Space, Horizontal Tab, Line Feed, Form Feed, and Carriage Return are considered whitespace characters and make no contribution to the token stream.
+
+As usual, token recognition is "greedy". Therefore, whitespace must be placed between two tokens when:
+
+- a keyword directly follows an identifier or named parameter,
+- an identifier directly follows a keyword or named parameter, or
+- a numeric literal directly follows an identifier, keyword, or parameter.
+
+=== Expressions
+
+An expression is a sequence of tokens to which a Java type can be assigned, and which evaluates to a well-defined value when the query is executed. In JDQL, expressions may be categorized as:
+
+- literals,
+- special values,
+- parameters,
+- enum literals,
+- paths,
+- function calls, and
+- operator expressions.
+
+==== Literal expressions
+
+A string, integer, or decimal literal is assigned the type it would be assigned in Java. So, for example, `'Hello'` is assigned the type `java.lang.String`, `123` is assigned the type `int`, `1e4` is assigned the type `double`, and `1.23f` is assigned the type `float`.
+
+The syntax for literal expressions is given by the `literal` grammar rule, and in the previous section titled <<Lexical structure>>.
+
+When executed, a literal expression evaluates to its literal value.
+
+==== Special values
+
+The special values `true` and `false` are assigned the type `boolean`, and evaluate to their literal values.
+
+The special values `local date` and `local datetime` are assigned the types `java.time.LocalDate` and `java.time.LocalDateTime`, and evaluate to the current date and current datetime of the database server, respectively.
+
+The syntax for special values is given by the `special_expression` grammar rule.
+
+==== Parameter expressions
+
+A parameter expression, with syntax given by `input_parameter`, is assigned the type of the repository method parameter it matches. For example, the parameter `:titlePattern` is assigned the type `java.lang.String`:
+
+[source,java]
+----
+@Query("where title like :titlePattern")
+List<Book> booksMatchingTitle(String titlePattern);
+----
+
+When executed, a parameter expression evaluates to the argument supplied to the parameter of the repository method.
+
+==== Enum literals
+
+An _enum literal expression_ is a Java identifier, with syntax specified by `enum_literal`, and may only occur as the right operand of a `set` assignment or `=`/`<>` equality comparison. It is assigned the type of the left operand of the assignment or comparison. The type must be a Java `enum` type, and the identifier must be the name of an enumerated value of the `enum` type. For example, `day <> MONDAY` is a legal comparison expression.
+
+When executed, an enum expression evaluates to the named member of the Java `enum` type.
+
+==== Path expressions
+
+A _path expression_ is a period-separated list of Java identifiers, with syntax specified by `state_field_path_expression`. Each identifier is interpreted as the name of a field of an entity or embeddable class. Each prefix of the list is assigned a Java type:
+
+- the first element of the list is assigned the type of the named field of the entity being queried, and
+- each subsequent element is assigned the type of the named field of the type assigned to the previous element.
+
+The type of the whole path expression is the type of the last element of the list. For example, `pages` is assigned the type `int`, `address` is assigned the type `org.example.Address`, and `address.street` is assigned the type `java.lang.String`.
+
+When executed, a path expression is evaluated in the context of a given record of the queried entity type, and evaluates to the value of the entity field for the given record.
+
+==== Function calls
+
+A _function call_ is the name of a JDQL function, followed by a parenthesized list of argument expressions, with syntax given by `function_expression`.
+
+- The `abs()` function is assigned the type of its numeric argument, and evaluates to the absolute value of the numeric value to which its argument evaluates. Its argument must be of numeric type.
+
+- The `length()` function is assigned the type `java.lang.Integer`, and evaluates to the length of string to which its argument evaluates. Its argument must be of type `java.lang.String`.
+
+- The `lower()` function is assigned the type `java.lang.String`, and evaluates to the lowercase form of the string to which its argument evaluates. Its argument must be of type `java.lang.String`.
+
+- The `upper()` function is assigned the type `java.lang.String`, and evaluates to the uppercase form of the string to which its argument evaluates. Its argument must be of type `java.lang.String`.
+
+- The `left()` function is assigned the type `java.lang.String`, and evaluates to a prefix of the string to which its first argument evaluates. The length of the prefix is given by the integer value to which its second argument evaluates. The first argument must be of type `java.lang.String`, and the second argument must be of integral numeric type.
+
+- The `right()` function is assigned the type `java.lang.String`, and evaluates to a suffix of the string to which its first argument evaluates. The length of the suffix is given by the integer value to which its second argument evaluates. The first argument must be of type `java.lang.String`, and the second argument must be of integral numeric type.
+
+When any argument expression of any function call evaluates to a null value, the whole function call evaluates to null.
+
+==== Operator expressions
+
+The syntax of an _operator expression_ is given by the `scalar_expression` rule. Within an operator expression, parentheses indicate grouping.
+
+The concatenation operator `||` is assigned the type `java.lang.String`. Its operand expressions must also be of type `java.lang.String`. When executed, a concatenation operator expression evaluates to a new string concatenating the strings to which its arguments evaluate.
+
+The numeric operators `+`, `-`, `*`, and `/` have the same meaning for primitive numeric types they have in Java, and operator expression involving these operators are assigned the types they would be assigned in Java.
+
+NOTE: As an exception, when the operands of `/` are both integers, a JDQL implementation is not required to interpret the operator expression as integer division if that is not the native semantics of the database. However, portability is maximized when Jakarta Data providers _do_ interpret such an expression as integer division.
+
+The four numeric operators may also be applied to an operand of wrapper type, for example, to `java.lang.Integer` or `java.lang.Double`. In this case, the operator expression is assigned a wrapper type, and evaluates to a null value when either of its operands evaluates to a null value. When both operands are non-null, the semantics are identical to the semantics of an operator expression involving the corresponding primitive types.
+
+The four numeric operators may also be applied to operands of type `java.math.BigInteger` or `java.math.BigDecimal`.
+
+A numeric operator expression is evaluated according to the native semantics of the database. In translating an operator expression to the native query language of the database, a Jakarta Data provider is encouraged, but not required, to apply reasonable transformations so that evaluation of the expression more closely mimics the semantics of the Java language.
+
+==== Numeric types and numeric type promotion
+
+The type assigned to an operator expression depends on the types of its operand expression, which need not be identical. The rules for numeric promotion are given in section 4.7 of the Jakarta Persistence specification version 3.2:
+
+> * If there is an operand of type `Double` or `double`, the expression is of type `Double`;
+> * otherwise, if there is an operand of type `Float` or `float`, the expression is of type `Float`;
+> * otherwise, if there is an operand of type `BigDecimal`, the expression is of type `BigDecimal`;
+> * otherwise, if there is an operand of type `BigInteger`, the expression is of type `BigInteger`, unless the operator is `/` (division), in which case the expression type is not defined here;
+> * otherwise, if there is an operand of type `Long` or `long`, the expression is of type `Long`, unless the operator is `/` (division), in which case the expression type is not defined here;
+> * otherwise, if there is an operand of integral type, the expression is of type `Integer`, unless the operator is `/` (division), in which case the expression type is not defined here.
+
+=== Conditional expressions
+
+A _conditional expression_ is a sequence of tokens which specifies a condition which, for a given record, might be _satisfied_ or _unsatisfied_. Unlike the scalar <<Expressions>> defined in the previous section, a conditional expression is not considered to have a well-defined type.
+
+NOTE: JPQL defines the result of a conditional expression in terms of ternary logic. JDQL does not specify that a conditional expression evaluates to well-defined value, only the effect of the conditional expression when it is used as a restriction. The "value" of a conditional expression is not considered observable by the application program.
+
+Conditional expressions may be categorized as:
+
+- `null` comparisons,
+- `in` expressions,
+- `between` expressions,
+- `like` expressions,
+- equality and inequality operator expressions, and
+- logical operator expressions.
+
+The syntax for conditional expressions is given by the `conditional_expression` rule. Within a conditional expression, parentheses indicate grouping.
+
+==== Null comparisons
+
+A `null` comparison, with syntax given by `null_comparison_expression` is satisfied when:
+
+- the `not` keyword is missing, and its operand evaluates to a null value, or
+- the `not` keyword occurs, and its operand evaluates to any non-null value.
+
+==== In expressions
+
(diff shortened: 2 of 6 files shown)


spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,337 @@
...
+
+The following identifiers are _keywords_: `select`, `update`, `set`, `delete`, `from`, `where`, `order`, `by`, `asc`, `desc`, `not`, `and`, `or`, `between`, `like`, `in`, `null`, `local`, `true`, `false`. In addition, every reserved identifier listed in section 4.4.1 of the Jakarta Persistence specification version 3.2 is also considered a reserved identifier. Keywords and other reserved identifiers are case-insensitive: `null`, `Null`, and `NULL` are three ways to write the same keyword.
+
+NOTE: Use of a reserved identifier as a regular identifier in JDQL might be accepted by a given Jakarta Data provider, but such usage is not guaranteed to be portable between providers.
+
+==== Parameters
+
+A _named parameter_ is a legal Java identifier prefixed with the `:` character, for example, `:name`.
+
+An _ordinal parameter_ is a decimal integer prefixed with the `?` character, for example, `?1`.
```
otaviojava, line 42: It would be hard inclusion of `@`.
This is more popular this use at NoSQL databases.

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,337 @@
...
+A _path expression_ is a period-separated list of Java identifiers, with syntax specified by `state_field_path_expression`. Each identifier is interpreted as the name of a field of an entity or embeddable class. Each prefix of the list is assigned a Java type:
+
+- the first element of the list is assigned the type of the named field of the entity being queried, and
+- each subsequent element is assigned the type of the named field of the type assigned to the previous element.
+
+The type of the whole path expression is the type of the last element of the list. For example, `pages` is assigned the type `int`, `address` is assigned the type `org.example.Address`, and `address.street` is assigned the type `java.lang.String`.
+
+When executed, a path expression is evaluated in the context of a given record of the queried entity type, and evaluates to the value of the entity field for the given record.
+
+==== Function calls
```
otaviojava, line 136: About those functions, there is no guarantee of this function support on the NoSQL database side.

gavinking: Yeah, so what I want to know is which functions _can_ be implemented, "typically"? Or at least, if not "typically", which can be implemented for a sufficient subset of platforms that it's worth mentioning them here/

gavinking: I mean, if the answer is simply that none of these things can typically be implemented on non-relational databases, then we should simply remove this section.

otaviojava: Typically at Document and Graph:

| Function | Supported by | Not Supported by |
|----------|--------------|------------------|
| `abs()`  | Document, Graph | Wide-column, Key-value |
| `length()`| Document, Graph | Wide-column, Key-value |
| `lower()` | Document, Graph | Wide-column, Key-value |
| `upper()` | Document, Graph | Wide-column, Key-value |
| `left()`  | Document, Graph | Wide-column, Key-value |
| `right()` | Document, Graph | Wide-column, Key-value |

But even with document, I cannot guarantee, e.g.:

- ravendb
- harperdb

otaviojava: All functions, and an UnsupportedOperationException is thrown only when the database provider does not support it.

gavinking: All the functions on that list?

otaviojava: Yes!

- abs
- length
- lower
- upper
- left
- right

spec/src/antlr/JDQL.g4
```diff
@@ -0,0 +1,92 @@
+grammar JDQL;
+
+statement : select_statement | update_statement | delete_statement;
+
+select_statement : select_clause? from_clause? where_clause? orderby_clause?;
```
otaviojava, line 5: Why we are not including the insert?

Btw: Any change of JSON inclusion, I mean for `insert`:

```
INSERT INTO mytable JSON '{ "\"myKey\"": 0, "value": 0}';
```

I know that NewSQL and some relational databases have support for it, for example, Postgresql, but I am not sure about this on JPQL.

njr-11: JPQL (at least [version 3.1](https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#bnf) of it) does not have `insert`.  In the future if we want to add the ability to insert from JSON, hopefully we can write a better API for that rather than having the user write query language for it.

otaviojava: What do you think about removing any data alteration for this version, keeping only `select`?

gavinking: > Why we are not including the `insert`?

The `insert ... values ....` form tends not to be very useful, since it competes directly with `@Insert`.

On the other hand, `insert ... select ....` is useful in very specific situations, but JPQL doesn't have it, and I doubt it can be implemented on most non-relational datastores.

HQL _does_ have both those things, but I really don't see a lot of pressure for them to be incorporated into the spec.

> Any change of JSON inclusion

I would not have thought that _literal_ JSON within the query itself was very useful. (But I might be wrong.)

What we support in Hibernate is:

- mapping a field of type `String` to a column of type `JSON`, or
- mapping an embeddable to a database column of type `JSON`.

But then the fact that this piece of an entity is stored as JSON is really just a question for the "mapping layer", and not something we need to address in the spec (given the current goals of the spec).

gavinking: > What do you think about removing any data alteration for this version, keeping only `select`?

I think this should be at least a minimal superset of the functionality you had already defined with the magic method name query stuff.

And the truth is that `update` and `delete` queries are very useful, because what they do _cannot_ be accomplished via `@Update` and `@Delete`.

spec/src/main/asciidoc/query-language.asciidoc
gavinking: Can `local date` and `local datetime` be supported on all platforms? (I'm guessing "no", but looking for confirmation.)

gavinking: Please see commit https://github.com/jakartaee/data/pull/520/commits/858686efec3b0f261d464851156e3d58fad4037c.

njr-11 (commented):
This looks great - it amounts to a very useful subset of JPQL which at least at first glances looks like it will cover everything you can do with Query by Method Name, plus some additional capability beyond that, which is good because it gives advanced users the ability to do some more powerful things with queries.  I didn't review in depth yet, especially with regard to double checking that everything is a subset of JPQL (and I hope we can rely on Jakarta Persistence spec community for that as well).  I included a few initial comments, but this is definitely headed in the right direction and it's exciting how far along it already is!

spec/src/main/asciidoc/query-language.asciidoc
graemerocher: can we mandate compilation time checking of supported functions? runtime errors are not nice

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,341 @@
...
+==== Path expressions
+
+A _path expression_ is a period-separated list of Java identifiers, with syntax specified by `state_field_path_expression`. Each identifier is interpreted as the name of a field of an entity or embeddable class. Each prefix of the list is assigned a Java type:
+
+- the first element of the list is assigned the type of the named field of the entity being queried, and
+- each subsequent element is assigned the type of the named field of the type assigned to the previous element.
+
+The type of the whole path expression is the type of the last element of the list. For example, `pages` is assigned the type `int`, `address` is assigned the type `org.example.Address`, and `address.street` is assigned the type `java.lang.String`.
+
+When executed, a path expression is evaluated in the context of a given record of the queried entity type, and evaluates to the value of the entity field for the given record.
```
graemerocher, line 134: what happens when something in the path is `null`? Do we support `null` safe dereferencing?

spec/src/antlr/JDQL.g4
```diff
@@ -0,0 +1,92 @@
...
+    | literal
+    ;
+
+function_expression
+    : 'ABS' '(' scalar_expression ')'
+    | 'LENGTH' '(' scalar_expression ')'
+    | 'LOWER' '(' scalar_expression ')'
+    | 'UPPER' '(' scalar_expression ')'
+    | 'LEFT' '(' scalar_expression ',' scalar_expression ')'
+    | 'RIGHT' '(' scalar_expression ',' scalar_expression ')'
```
graemerocher, line 73: what about `MAX`,`MIN` and `AVG`?

spec/src/antlr/JDQL.g4
```diff
@@ -0,0 +1,92 @@
...
+orderby_clause : 'ORDER' 'BY' orderby_item (',' orderby_item)*;
+orderby_item : state_field_path_expression ('ASC' | 'DESC');
+
+conditional_expression
+    // highest to lowest precedence
+    : '(' conditional_expression ')'
+    | null_comparison_expression
+    | in_expression
+    | between_expression
+    | like_expression
```
graemerocher, line 32: perhaps a statement that regex could be supported in a like expression?

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,341 @@
...
+
+==== Where clause
+
+The `where` clause, with syntax given by `where_clause`, specifies a conditional expression used to restrict the records returned, deleted, or updated by the query. Only records for which the conditional expression is satisfied are returned, deleted, or updated.
+
+The `where` clause is always optional. When it is missing, there is no restriction, and, and all records are of the queried entity type are returned, deleted, or updated.
+
+==== Select clause
+
+The `select` clause, with syntax given by `select_clause`, specifies a list of expressions which are returned by the query. Each expression in the list is evaluated for each record which satisfies the restriction imposed by the `where` clause. Alternatively, the `select` clause may contain a single `count(*)` aggregate expression, which evaluates to the number of records which satisfy the restriction.
```
graemerocher, line 298: what happens in the db doesn't support `count(*)`?

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,341 @@
...
+
+==== Set clause
+
+The `set` clause, with syntax given by `set_clause`, specifies a list of updates to fields of the queried entity. For each record which satisfies the restriction imposed by the `where` clause, and for each element of the list, the scalar expression is evaluated and assigned to the entity field identified by the path expression.
+
+==== Order clause
+
+The `order` clause (or `order by` clause), with syntax given by `orderby_clause`, specifies a lexicographic order for the query results, that is, a list of entity fields used to sort the records which satisfy the restriction imposed by the `where` clause. The keywords `asc` and `desc` specify that a given field should be sorted in ascending or descending order respectively; when neither is specified, ascending order is the default.
+
+The `order` clause is always optional. When it is missing, and when no sort criteria are given as arguments to a parameter of the repository method, the order of the query results is undefined, and may not be deterministic.
```
graemerocher, line 310: what happens when the db doesn't support `order`?

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -460,6 +460,13 @@ create table Book (
 
 NOTE: Support for entity associations is not required by this specification.
 
+==== Entity Names
+
+Each entity must be assigned an _entity name_ by the provider.
+By default, this must be the unqualified Java class name of the entity class.
```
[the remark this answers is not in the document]
gavinking: Good question: I would like to say "no", and it's the user responsibility to make sure entity names are unique. It's annoying that the facility to do that would by external to the spec.

spec/src/main/asciidoc/query-language.asciidoc
gavinking: Jakarta Data already has lots of ways to request sorting, so whatever happens in those other cases. Not sure where that's specified.

spec/src/antlr/JDQL.g4
gavinking: So this is a question I was asking myself. From what little I have seen, very many non-relational databases seem to have good support for regexes. _Traditionally_, relational databases did not, though I guess most do by now. The JPQL `like` operator is the ANSI SQL one. If we wanted to provide support for regexes, we would need to provide some syntax for that. The `REGEXP_LIKE` function seems to be the most standard thing in SQL, though I would say it's pretty ugly. (I forget what the ANSI SQL standard for this is.)

gavinking: Ah, right, so ANSI SQL has a `SIMILAR TO` operator, and its own dialect of regular expressions.

spec/src/main/asciidoc/query-language.asciidoc
gavinking: This is something I need to clarify.

spec/src/antlr/JDQL.g4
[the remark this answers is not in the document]
gavinking: Let's talk about this on the call today. 

But to summarize my view on this, it seems to me that, given:

```java
@Query("where title like '%Data%'")
List<Book> dataBooks();
```

Then it's quite trivial for the provider to transform this query to: `select b from Book as b where b.title like '%Data%'` and that this transformation is completely unambiguous[1].

And my perception is that to ask the user to write:

```java
@Query("select b from Book as b where b.title like '%Data%'")
List<Book> dataBooks();
```

is just contrary to what I perceive to be the goals of Jakarta Data. That's three times as many tokens!

[1] modulo a teensy issue with the syntax for enum literals, but that's surely solvable!

spec/src/main/asciidoc/query-language.asciidoc
gavinking: @graemerocher see 34ca9693c2bb8fce5c2b3e9c84dd69644de008f0.

spec/src/main/asciidoc/query-language.asciidoc
gavinking: Well the issue is that at the spec level, Jakarta Data allows purely-runtime implementations.

Now, as far as I can tell, we're all implementing the spec using annotation processing, and so, indeed, we can all check this at compilation time (I certainly do).

But if we were to _require_ that, then we would be prohibiting other implementation strategies.

spec/src/main/asciidoc/query-language.asciidoc
otaviojava: I guess we put the Unsupportedoperationexception, am I right, @njr-11 ?

spec/src/main/asciidoc/repository.asciidoc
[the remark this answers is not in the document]
gavinking: It's actually _really_ hard to write down something meaningful here, since the JD spec doesn't have any well-defined notion of a "persistence unit", that is, of which entities are eligible for use in query methods of a given repository.

spec/src/main/asciidoc/query-language.asciidoc
otaviojava: We don't have a guarantee for all fields in NoSQL; for example, Cassandra, when it works, will be for the primary key.
HBase does not support sorting at all; this also happens with key-value databases in the majority of scenarios.

At NoSQL, when the NoSQL database does not support I would expect the unsupportedoperationexception on those scenarios.


I mean, not only for Sorting, but even to operator.

For example, Cassandra has not support to **Or** operator.

gavinking: @graemerocher @otaviojava please review d127e7731db03a209101278ace53c2cf13933d41

spec/src/main/asciidoc/query-language.asciidoc
gavinking: @graemerocher @otaviojava please review d127e7731db03a209101278ace53c2cf13933d41

spec/src/antlr/JDQL.g4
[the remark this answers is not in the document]
otaviojava:
```java
@Repository
interface Garage extends BasicRepository<Car, Long>{
@Query("where title like '%Data%'")//it will inject the selct * from Entity as syntax sugar
List<Car> cards();
}
```

spec/src/main/asciidoc/query-language.asciidoc
njr-11: > I guess we put the Unsupportedoperationexception, am I right, @njr-11 ?

Right, based on how we treat other unsupported function in the spec, it should be `UnsupportedOperationException`.  I think the spec is actually missing this at the level of `countBy`. It specifies using the exception for keywords used in the remained of Query by Method Name that aren't supported but doesn't currently say that method name prefixes like `count` can be unsupported.

spec/src/main/asciidoc/query-language.asciidoc
gavinking: Another wrinkle here is that the apparently providers must raise `MappingException` when they can't do key-based pagination. That's not intuitive to me at all. It doesn't seem like a `MappingException` from what I can see.

njr-11: > Another wrinkle here is that the apparently providers must raise `MappingException` when they can't do key-based pagination. That's not intuitive to me at all. It doesn't seem like a `MappingException` from what I can see.

We can change that to `UnsupportedOperationException`, too.

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -324,6 +324,8 @@ The `set` clause, with syntax given by `set_clause`, specifies a list of updates
 
 The `order` clause (or `order by` clause), with syntax given by `orderby_clause`, specifies a lexicographic order for the query results, that is, a list of entity fields used to sort the records which satisfy the restriction imposed by the `where` clause. The keywords `asc` and `desc` specify that a given field should be sorted in ascending or descending order respectively; when neither is specified, ascending order is the default.
 
+By default, the order of values of a Java enumerated type is determined by the integer returned by `ordinal()`. A programming model for entity classes might allow control over the order of `enum` values. For example, Jakarta Persistence allows this via the `@Enumerated` annotation.
```
otaviojava, line 327: Are you sure that the `ordinal` should be the default?

If you include a new element or swap it, it will generate issues with data integrity.

When comparing String and `ordinal`, the unique benefit that came to my mind about ordinal is performance once it is smaller.

But, readability, data integrity, and flexibility I would prefer as String.

At the NoSQL side, we mostly use the `name`method, some samples:

- Couchbase
- Amazon DynamoDB
- RethinkDB
- Elasticsearch
- Cassandra
- Redis

njr-11: This is an unfortunate situation where neither option is ideal. Ordering on `name` allows inserting new enum values to be a compatible change, but the user has no control over the desired ordering if they don't want alphabetical. Ordering on `ordinal` gives the user that control, but doesn't allow for inserting additional enum values at a later point without migrating existing data.  Given what Otavio brought up about some of the prominent NoSQL databases having the opposite default as Jakarta Persistence, we will need to define this as being provider-specific. I think it would be fine to state what the default is for Jakarta Persistence or mention the jakarta.persistence.Enumerated annotation that documents what the default is for Jakarta Persistence.  Maybe this would be better done in the JavaDoc.  I can look into that separately.

gavinking: OK, that's fine by me.

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,364 @@
...
+=== Clauses
+
+Each JDQL statement is built from a sequence of _clauses_. The beginning of a clause is identified by a keyword: `from`, `where`, `select`, `set`, or `order`.
+
+There is a logical ordering of clauses, reflecting the order in which their effect must be computed by the datastore:
+
+1. `from`
+2. `where`,
+3. `select` or `set`,
+4. `order`.
```
njr-11, line 289: Are `select` and `order` reversed here?  In the case where `select` is reducing to an entity attribute, it seems like `order` would need to be done first to consider the various other entity attributes to order by.  For example,

`SELECT productId FROM Product ORDER BY price DESC, name ASC, productId ASC`

gavinking: In the full grammar for JPQL we have:

```
select_item ::= select_expression [[AS] result_variable]
select_expression ::=
    single_valued_path_expression |
    scalar_expression |
    aggregate_expression |
    identification_variable |
    OBJECT(identification_variable) |
    constructor_expression

...

orderby_item ::=
    state_field_path_expression |
    general_identification_variable |
    result_variable
    [ASC | DESC]
    [NULLS {FIRST | LAST}]
```

So you may order by any item in the `select` list, including complicated scalar expressions.

So yeah, it's correct.

njr-11: Good point, we need to remain consistent with JPQL even though we don't provide the additional capabilities here

gavinking:
Looks like I need to rebase this.

gavinking:
In [ebef494](https://github.com/jakartaee/data/pull/520/commits/ebef49458ce1a0d2454a4c775862e094c55df1b7) I have proposed a new subsection dealing with ordering.

spec/src/antlr/JDQL.g4
```diff
@@ -0,0 +1,92 @@
...
+    ;
+
+special_expression
+    : 'LOCAL' 'DATE'
+    | 'LOCAL' 'DATETIME'
+    | 'TRUE'
+    | 'FALSE'
+    ;
+
+state_field_path_expression : IDENTIFIER ('.' IDENTIFIER)*;
```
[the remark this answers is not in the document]
gavinking: > We will need to keep track of whether the issue/pull to make identification variables optional made/makes it into Jakarta Persistence 3.2. Hopefully it will. If it doesn't, we would need to require them as well and make them optional in the future.

I just got the go ahead from @lukasj. We are going to make this change to JPQL.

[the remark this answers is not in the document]
njr-11: > I just got the go ahead from @lukasj. We are going to make this change to JPQL.

Excellent! This will enable Jakarta Data's Query Language to omit identification variables and still be a subset of JPQL, which clears up the main concern I had with the query language.

spec/src/antlr/JDQL.g4
```diff
@@ -0,0 +1,93 @@
+grammar JDQL;
+
+statement : select_statement | update_statement | delete_statement;
+
+select_statement : select_clause? from_clause? where_clause? orderby_clause?;
```
otaviojava, line 5: @gavinking this statement has a bug because an empty query "" is valid.

We need to elaborate on this point more:

```antlr
mandatory_statement : select_clause from_clause where_clause orderby_clause
                    | select_clause from_clause where_clause
                    | select_clause from_clause orderby_clause
                    | select_clause where_clause orderby_clause
                    | from_clause where_clause orderby_clause
                    | where_clause
                    ;

select_statement : mandatory_statement
                 | select_clause
                 | from_clause
                 | where_clause
                 | orderby_clause
                 ;
```

gavinking: Well the idea here is that:

```java
@Query("")
List<Book> allBooks();
```

is a legal query with a clear interpretation.

gavinking: Yes of course. I'm not advocating it as the best way to write that code; just saying that it's consistent and "not wrong".

gavinking: [Recall also that the eventual goal of this work is to redefine other things like `@Find` _in terms of JDQL_.]

gavinking:
Moving this out of "Draft" mode, after checking with @otaviojava and @njr-11 .

Please speak up with changes that should be made before merging!

spec/src/main/asciidoc/repository.asciidoc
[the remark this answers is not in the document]
gavinking: If someone thinks we need to say something specific about the issue of name collisions, please propose some language. 

I think it's acceptable if the spec stays silent on this issue for now. If it's _not_ acceptable, I think we can fix the problem in a separate discussion/PR. It doesn't need to block the merge of this PR.

spec/src/main/asciidoc/query-language.asciidoc
gavinking: Please see a4157bf2c0b9b0add9705792da069be07c0154e3.

This should actually be `DataException`, as per our recent discussion.

spec/src/antlr/JDQL.g4
gavinking: We discussed this on a recent JD call, and there was little enthusiasm for adding regex support _in this first release_.

But @graemerocher, if this is important to you, we can rediscuss it. Perhaps you could open an issue, or propose a PR.

We don't need to do it in _this_ PR.

spec/src/main/asciidoc/query-language.asciidoc
gavinking: As discussed on a recent concall, I already added `local time`. And it seems we have a better-than-tentative consensus on the proposed list of functions.

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,404 @@
...
+
+NOTE: These functions cannot be emulated on every datastore. When a function cannot be reasonably emulated via the native query capabilities of the database, a JDQL implementation is not required to provide the function.
+
+If the JDQL implementation does not support a one of the standard functions explicitly listed above, it must throw `UnsupportedOperationException` when the function name occurs in a query. Alternatively, the Jakarta Data provider is permitted to reject a repository method declaration at compilation time if its `@Query` annotation uses an unsupported function.
+
+NOTE: On the other hand, an implementation of JDQL might provide additional built-in functions, and might even allow invocation of user-defined functions. Section 4.7 of the Jakarta Persistence specification defines a set of functions that all JPQL implementations are required to provide, including `concat`, `substring`, `trim`, `locate`, `ceiling`, `floor`, `exp`, `ln`, `mod`, `power`, `round`, `sign`, `sqrt`, `cast`, `extract`, `coalesce`, and `nullif`. JDQL implementations are encouraged to support any of these functions which are reasonably implementable.
```
njr-11, line 167: The following could be done to remain correct if the sections are renumbered in Jakarta Persistence.
```suggestion
NOTE: On the other hand, an implementation of JDQL might provide additional built-in functions, and might even allow invocation of user-defined functions. The section of the Jakarta Persistence specification titled "Scalar Expressions" defines a set of functions that all JPQL implementations are required to provide, including `concat`, `substring`, `trim`, `locate`, `ceiling`, `floor`, `exp`, `ln`, `mod`, `power`, `round`, `sign`, `sqrt`, `cast`, `extract`, `coalesce`, and `nullif`. JDQL implementations are encouraged to support any of these functions which are reasonably implementable.
```

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,404 @@
...
+
+The four numeric operators may also be applied to operands of type `java.math.BigInteger` or `java.math.BigDecimal`.
+
+A numeric operator expression is evaluated according to the native semantics of the database. In translating an operator expression to the native query language of the database, a Jakarta Data provider is encouraged, but not required, to apply reasonable transformations so that evaluation of the expression more closely mimics the semantics of the Java language.
+
+==== Numeric types and numeric type promotion
+
+The type assigned to an operator expression depends on the types of its operand expression, which need not be identical. The rules for numeric promotion are given in section 4.7 of the Jakarta Persistence specification version 3.2:
```
njr-11, line 195:
```suggestion
The type assigned to an operator expression depends on the types of its operand expression, which need not be identical. The rules for numeric promotion are given in the "Scalar Expressions" section of the Jakarta Persistence specification:
```

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,404 @@
...
+- the `not` keyword is missing, and any one of its parenthesized operands evaluates to the same value as its leftmost operand, or
+- the `not` keyword occurs, and none of its parenthesized operands evaluate to the same value as its leftmost operand.
+
+All operands must have the same type.
+
+==== Between expressions
+
+A `between` expression, with syntax given by `between_expression` is satisfied when its operands all evaluate to non-null values, and, if the `not` keyword is missing, its left operand evaluates to a value which is:
+
+- larger than or equal to the value take by its middle operand, and
```
njr-11, line 241:
```suggestion
- larger than or equal to the value taken by its middle operand, and
```

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,404 @@
...
+- smaller than or equal to the value taken by its right operand.
+
+Or, if the `not` keyword occurs, the left operand must evaluate to a value which is:
+
+- strictly smaller than to the value take by its middle operand, or
```
njr-11, line 246:
```suggestion
- strictly smaller than to the value taken by its middle operand, or
```

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,404 @@
...
+- This specification does not define an order for the sorting of Java `enum` values, which is provider-dependent. A programming model for entity classes might allow control over the order of `enum` values. For example, Jakarta Persistence allows this via the `@Enumerated` annotation.
+
+- This specification does not define an order for UUID values, which is provider-dependent.
+
+==== Logical operators
+
+The logical operators are `and`, `or`, and `not`.
+
+- An `and` operator expression is satisfied if and only if both its operands are satisfied.
+- An `or` operator expression is satisfied if and only if at least one of its operands are satisfied.
```
njr-11, line 297:
```suggestion
- An `or` operator expression is satisfied if at least one of its operands is satisfied.
```

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,404 @@
...
+----
+@Query("from Book where title like :title")
+List<Book> booksByType(String title);
+----
+
+==== Where clause
+
+The `where` clause, with syntax given by `where_clause`, specifies a conditional expression used to restrict the records returned, deleted, or updated by the query. Only records for which the conditional expression is satisfied are returned, deleted, or updated.
+
+The `where` clause is always optional. When it is missing, there is no restriction, and, and all records are of the queried entity type are returned, deleted, or updated.
```
njr-11, line 347:
```suggestion
The `where` clause is always optional. When it is missing, there is no restriction, and all records of the queried entity type are returned, deleted, or updated.
```

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,404 @@
...
+
+==== Set clause
+
+The `set` clause, with syntax given by `set_clause`, specifies a list of updates to fields of the queried entity. For each record which satisfies the restriction imposed by the `where` clause, and for each element of the list, the scalar expression is evaluated and assigned to the entity field identified by the path expression.
+
+==== Order clause
+
+The `order` clause (or `order by` clause), with syntax given by `orderby_clause`, specifies a lexicographic order for the query results, that is, a list of entity fields used to sort the records which satisfy the restriction imposed by the `where` clause. The keywords `asc` and `desc` specify that a given field should be sorted in ascending or descending order respectively; when neither is specified, ascending order is the default.
+
+The `order` clause is always optional. When it is missing, and when no sort criteria are given as arguments to a parameter of the repository method, the order of the query results is undefined, and may not be deterministic.
```
njr-11, line 369:
```suggestion
The `order` clause is always optional. When it is missing, and when no sort criteria are given as arguments to a parameter of the repository method, the order of the query results is undefined, and might not be deterministic.
```

spec/src/antlr/JDQL.g4
```diff
@@ -0,0 +1,93 @@
...
+select_list
+    : state_field_path_expression
+    | aggregate_expression
+    ;
+aggregate_expression : 'COUNT' '(' '*' ')';
+
+orderby_clause : 'ORDER' 'BY' orderby_item (',' orderby_item)*;
+orderby_item : state_field_path_expression ('ASC' | 'DESC');
+
+conditional_expression
```
njr-11, line 26: JPQL has empty_collection_comparison_expression and I think Query by Method Name has an `Empty` keyword, so we should probably include it in order to be able to transform from Query by Method Name to query language.

Similar for collection_member_expression and Collection `Contains`.

spec/src/antlr/JDQL.g4
```diff
@@ -0,0 +1,93 @@
...
+    | 'LOWER' '(' scalar_expression ')'
+    | 'UPPER' '(' scalar_expression ')'
+    | 'LEFT' '(' scalar_expression ',' scalar_expression ')'
+    | 'RIGHT' '(' scalar_expression ',' scalar_expression ')'
+    ;
+
+special_expression
+    : 'LOCAL' 'DATE'
+    | 'LOCAL' 'DATETIME'
+    | 'LOCAL' 'IME'
```
njr-11, line 79:
```suggestion
    | 'LOCAL' 'TIME'
```

njr-11 (commented):
This looks good.  I didn't find much new because a lot of it we have reviewed already.  I added a few minor comments/typo fixes.  Feel free to ignore the suggestions that aim to avoid hard coding the Jakarta Persistence specification version number and section numbers if you don't like that idea.

spec/src/antlr/JDQL.g4
gavinking: Yeah and Query by Method Name also currently has something similar to JPQL's `member of`. 

I mentioned this earlier.

The problem is that currently we don't have any sort of entity field that can be a collection, so that functionality just isn't well-defined. There's nothing for `empty` or `member of` to operate on!

Either:

1. we need to define some kind of entity field that can be a collection, or
2. simply remove that feature from Query by Method Name as well.

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -0,0 +1,404 @@
...
+- literal strings, and
+- integer and decimal number literals.
+
+==== Identifiers and keywords
+
+An _identifier_ is any legal Java identifier which is not a keyword. Identifiers are case-sensitive: `hello`, `Hello`, and `HELLO` are distinct identifiers.
+
+In the JDQL grammar, identifiers are labelled with the `IDENTIFIER` token type.
+
+The following identifiers are _keywords_: `select`, `update`, `set`, `delete`, `from`, `where`, `order`, `by`, `asc`, `desc`, `not`, `and`, `or`, `between`, `like`, `in`, `null`, `local`, `true`, `false`. In addition, every reserved identifier listed in section 4.4.1 of the Jakarta Persistence specification version 3.2 is also considered a reserved identifier. Keywords and other reserved identifiers are case-insensitive: `null`, `Null`, and `NULL` are three ways to write the same keyword.
```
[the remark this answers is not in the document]
gavinking: Mmmmm. Hrrrrm. The assumption here is that section numbers are much more likely to change than section titles, which is _probably_ true ... but I dunno, a single-word doesn't seem like a wonderful sort of reference. Not sure if this is a good idea or not.

spec/src/antlr/JDQL.g4
otaviojava: Yes, the NoSQL often includes support for multi-valued fields or embeddables, allowing for complex data types and structures directly within documents or records.

gavinking: Of course, but I was more asking what the Jakarta NoSQL _specification_ actually says about this.

njr-11: Whether to have query language for collection attributes can be decided in a separate pull.  My intent here was only to ensure we consider what we want here.

gavinking: Yup. Should we open an issue?

njr-11: Yes, let's open a separate issue.

otaviojava: But in my option, I would go to simple remove that feature from query by method name as well.
The second option.

spec/src/antlr/JDQL.g4
```diff
@@ -0,0 +1,93 @@
...
+
+set_clause : 'SET' update_item (',' update_item)*;
+update_item : state_field_path_expression '=' (scalar_expression | 'NULL');
+
+select_clause : 'SELECT' select_list;
+select_list
+    : state_field_path_expression
+    | aggregate_expression
+    ;
+aggregate_expression : 'COUNT' '(' '*' ')';
```
njr-11, line 21: `COUNT(*)` doesn't appear to be valid JPQL per the BNF that is in Jakarta Persistence 3.2 M2.
https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2-m2#bnf
Hopefully this is one of the things being added as part of making identification variables optional.

gavinking: Hrm, yeah, OK, in JPQL it would be `count(this)`. You want me to change it to that?

njr-11: Yes, we need to ensure JDQL is a subset JPQL

gavinking: Done.

spec/src/antlr/JDQL.g4
gavinking: As discussed today, 43c88751ec5adf885c1d7343ea42703b2d475aff removes `Contains` and `Empty` from QbMN, which alleviates the need for `empty` and `member of`.

njr-11 (approved):
I'm approving the pull, but we should consider if `Contains` was still useful for String attributes.  The other option here would be to remove only the part that says it applies to collections.
Also, I didn't see where the removed keywords are added to the reserved list (in case there is a need to bring them back in the future), so we will need to be sure to cover that separately.

gavinking:
> I'm approving the pull, but we should consider if `Contains` was still useful for `String` attributes. The other option here would be to remove only the part that says it applies to collections.

I wondered that, but then figured that since it's just a variation of `Like` that it wasn't really that useful. But I can put it back if necessary.

> Also, I didn't see where the removed keywords are added to the reserved list (in case there is a need to bring them back in the future), so we will need to be sure to cover that separately.

You mean `Contains` and `Empty`? You want to add a new list of "Keywords Reserved for Future Use"?

gavinking:
To be specific: `findByNameContains("hello%world")` is the same as `findByNameLike("%hello%world%")`.

njr-11:
> To be specific: `findByNameContains("hello%world")` is the same as `findByNameLike("%hello%world%")`.

It is, but it requires the repository user to supply the `%` on their parameter, which might not be very natural to their usage, and they might not even like to know about wildcard characters. In your particular example, that user wouldn't care because they were already using a `%` in the middle of their pattern.  But if they only wanted a single word or part of a word, they might not want to bother with wildcard characters on the ends of it.

njr-11:
> You mean `Contains` and `Empty`? You want to add a new list of "Keywords Reserved for Future Use"?

I don't think we need a new section. There is one in module-info here that can be used:

```
 * <h3>Reserved for Future Use</h3>
```

The important part of this is notifying the user to avoid them so that we would have the freedom to add them back if the need arises.

gavinking:
> It is, but it requires the repository user to supply the `%` on their parameter, which might not be very natural to their usage, and they might not even like to know about wildcard characters.

So if you want it back just say the word....

gavinking:
> > You mean `Contains` and `Empty`? You want to add a new list of "Keywords Reserved for Future Use"?
> 
> I don't think we need a new section. There is one in module-info here that can be used:
> 
> ```
>  * <h3>Reserved for Future Use</h3>
> ```

Ugh. It gives me the heezy-jeebys when something is _only_ mentioned in that `module-info.java`.

gavinking:
Well, see 48bea10.

njr-11 (approved):

otaviojava:
@gavinking could you fix the conflict?

gavinking:
> @gavinking could you fix the conflict?

Done.

PR #521, opened 2024-03-05
requestTotal is lost when subsequent methods are invoked

njr-11:
While I was bringing in the requestTotal updates, I spotted a bug where a value of `false` is lost when additional builder methods are invoked.  For example, if you write `PageRequest.ofSize(20).requestTotal(false).page(5)`, the `page` method reverts the requestTotal value back to its default of `true`.  This is happening because of a constructor in `Pagination` that defaults the value to `true` which all of these fluent builder-style methods are using rather than supplying the current value of requestTotal.  I added a unit test confirming this is broken, and fixed the issue by removing the extra constructor so that we force all methods to either explicitly supply or default the value as is done for all other record components.


gavinking:
Oops, I suck :-(

otaviojava (approved):

PR #523, opened 2024-03-06
Prototype a query by name method parser

starksm64:
This prototypes a library tool that parses a Query by name method name using an Antlr4 generated parser and a grammar based on the current spec "BNF Grammar for Query Methods" section.


starksm64:
This has been updated to the latest code we are using to pass the TCK with Hibernate 6.6.0

gavinking:
> This has been updated to the latest code we are using to pass the TCK with Hibernate 6.6.0

Awesome.

I hope everyone understands that this is something essentially any implementor of Jakarta Data can use to add support for QbMN.

tools/src/main/java/ee/jakarta/tck/data/tools/annp/RespositoryProcessor.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    @Override
+    public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
+        System.out.printf("RespositoryProcessor: Processing repositories, over=%s\n", roundEnv.processingOver());
+        boolean newRepos = false;
+        Set<? extends Element> repositories = roundEnv.getElementsAnnotatedWith(Repository.class);
+        for (Element repository : repositories) {
+            String provider = repository.getAnnotation(Repository.class).provider();
+            if(provider.isEmpty() || provider.equalsIgnoreCase("hibernate")) {
```
KyleAure, line 64: > I hope everyone understands that this is something essentially any implementor of Jakarta Data can use to add support for QbMN.

Would this restrict the the use case of this tool to hibernate only?

I'd be interested in a README file to help understand the general use case of this tool for a Jakarta Data provider.

gavinking: Ah yes, that should be parameterized I guess.

starksm64:
I'll add a README and a parameterization of the provider value(s). The README can have sections for providers that have made use of it to document how it is used.

Tomas-Kraus:
Well, query method by name is regular grammar, something, that should be parsed by regular expression or state machine. Why to build this whole LL(k) context-freee grammar monster which needes at least recursive descend parser?

I solved this just using AntLR 4 lexer grammar with few states.

PR #524, opened 2024-03-06
PageRecord computes incorrect moreResults when totalElements unknown

njr-11:
I noticed this warning when building the project,

```
[WARNING] Javadoc Warnings
[WARNING] /Users/njr/lgit/data/api/src/main/java/jakarta/data/page/impl/PageRecord.java:42: warning: no comment
[WARNING] public PageRecord(PageRequest<T> pageRequest, List<T> content, long totalElements) {
[WARNING] ^
[WARNING] 1 warning
```

When I looked into correcting it by adding the missing JavaDoc, I realized the documentation would need to state how the constructor computes the moreResults record component and started to do so, but then realized the logic is incorrect for the path where the `totalElements` is unknown, always computing a `false` value for `moreResults`.  I added a unit test to demonstrate the error and corrected the code.

Also, I'm fixing a minor display issue in the return types table where the alternating highlighting is out of sync for the last entry of the table, causing it to appear like the information applies to the last two entries lumped together as one.

### DIFF
--- a/api/src/main/java/jakarta/data/page/impl/PageRecord.java
+++ b/api/src/main/java/jakarta/data/page/impl/PageRecord.java
@@ -29,20 +29,39 @@ import java.util.NoSuchElementException;
  * This may be used to simplify implementation of a repository interface.
  *
  * @param pageRequest The {@link PageRequest page request} for which this
- *                    slice was obtained
+ *                    page was obtained
  * @param content The page content
  * @param totalElements The total number of elements across all pages that
- *                      can be requested for the query
+ *                      can be requested for the query. A negative value
+ *                      indicates that a total count of elements and pages
+ *                      is not available.
  * @param moreResults whether there is a (nonempty) next page of results
  * @param <T> The type of elements on the page
  */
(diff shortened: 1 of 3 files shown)


api/src/main/java/jakarta/data/page/impl/PageRecord.java
```diff
@@ -29,20 +29,39 @@
...
+     *                      that can be requested for the query. A negative
+     *                      value indicates that a total count of elements
+     *                      and pages is not available.
+     */
     public PageRecord(PageRequest<T> pageRequest, List<T> content, long totalElements) {
         this( pageRequest, content, totalElements,
                 content.size() == pageRequest.size()
-                        && totalElements > pageRequest.size() * pageRequest.page() );
+                        && (totalElements < 0
+                                || totalElements > pageRequest.size() * pageRequest.page() ));
```
gavinking, line 64: Yeah that's definitely what I intended it to say. Not quite sure how I managed to messed that up.

gavinking (approved):

PR #527, opened 2024-03-07
enum ordering behavior is not documented

njr-11:
This is split off from the Query Language issue where it was identified that we lack documentation of what behavior to expect when sorting on enum attributes. I also noticed that our JavaDoc doesn't even tell users what the basic types are (including enums) so I have included that in the main JavaDoc and mentioned the behavior for enum sorting there.

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -577,6 +577,67 @@ import java.util.List;
  * Vehicle[] searchFor(String make, String model, int year, {@code Sort<?>...} sorts);
  * </pre>
  *
+ * <h2>Basic Types</h2>
+ *
+ * <p>The following is a list of valid basic entity attribute types.
+ * These can be used as the types of repository method parameters
+ * for the respective entity attribute.</p>
+ *
+ * <table style="width: 100%">
+ * <caption><b>Basic Types for Entity Attributes</b></caption>
+ * <tr style="background-color:#ccc">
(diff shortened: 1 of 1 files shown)


mswatosh:
Adding a rendered version for other reviewers. Changes look good to me.
image

mswatosh (approved):

gavinking (approved):

mswatosh:
VS Code isn't great either
image
From a user perspective, I'm not sure how someone would find the `module-info.java` in an IDE. In theory it's linked to from the `jakarta.data` project info, but that link doesn't appear for me in my sample application in VS Code.

mswatosh:
I just did a quick search for `<tr` and `<td>` and it looks like `module-info.java` is the only file using tables, which is good, since I think the rest of the JavaDoc users will likely view in an IDE.

njr-11:
module-info is a bit different than the other JavaDoc in that its entry point from the Jakarta Data specification page https://jakarta.ee/specifications/data/1.0/ and hopefully from search engine results, so it is oriented toward display in a browser.

For this pull, I copied the format of the other tables for consistency.  When I view it from Chrome and Firefox in dark mode, it renders well, albeit with a white background, but that seems to be the case for all other JavaDoc that I view as well.

I'm glad to hear Mark's findings that none of the JavaDoc for API classes that might be accessed from an IDE have tables in them.  I wonder what other specs are doing which have tables to display and if this has already been solved.  The alternating shading of tables rows is really nice for visibility when viewed in a browser.

Issue #518, opened 2024-03-03
[clarification]: list of restrictions on repository method signatures in Javadoc of PageRequest


gavinking:
### Specification

PageRequest

### I need clarification on ...

should those restrictions be stated elsewhere.

### Additional information

In the Javadoc for `PageRequest`, we have this text:

> A repository method will fail with a DataException or a more specific subclass if
>
> - multiple PageRequest parameters are supplied to the same method.
> - PageRequest and Limit parameters are supplied to the same method.
> - a PageRequest parameter is supplied in combination with the First keyword.
> - a PageRequest parameter with sort criteria is supplied and separate Sort parameters are also supplied to the same method.
> - a PageRequest parameter with sort criteria is supplied and an Order parameter is also supplied to the same method.
the database is incapable of ordering with the requested sort criteria.

This seems to mix together two sorts of disallowed things:

- The first three conditions are things which ideally&mdash;at least in an implementation based on annotation processing&mdash;would be rejected at compile time. They form part of the programming model for repository methods.
- The last two conditions seem to be things which are only usually detectable at runtime.

Also, I wonder why this is always a `DataException`, whereas in spec section 4.5, a similar sort of problem (to the first three cases considered here) is specified to result in `UnsupportedOperationException`.

I'm thinking that this information doesn't belong on `PageRequest`, and should probably be moved to section 4.7 of the specification, which should also mention the possibility of rejecting such method signatures at compile time.

njr-11:
> Also, I wonder why this is always a `DataException`, whereas in spec section 4.5, a similar sort of problem (to the first three cases considered here) is specified to result in `UnsupportedOperationException`.

It can be difficult to decide which exception class to use sometimes. `IllegalArgumentException` would have been another possibility.  I'm fine with `UnsupportedOperationException` if other see that fitting better here, and also for consistency with elsewhere in the spec as pointed out.

> I'm thinking that this information doesn't belong on `PageRequest`, and should probably be moved to section 4.7 of the specification, which should also mention the possibility of rejecting such method signatures at compile time.

I'm fine with including it in the specification document, but I wouldn't remove it from the JavaDoc because the JavaDoc will help prevent users from attempting these error paths. Users are much more likely to read a method's JavaDoc before using the method than they are to read the spec before using a method.

gavinking:
> I wouldn't remove it from the JavaDoc because the JavaDoc will help prevent users from attempting these error paths

[...]

PR #530, opened 2024-03-08
fix up spec and javadoc requirements with respect to special parameters

gavinking:
see #518

### DIFF
--- a/api/src/main/java/jakarta/data/Limit.java
+++ b/api/src/main/java/jakarta/data/Limit.java
@@ -44,7 +44,7 @@ import jakarta.data.page.PageRequest;
  * <ul>
  * <li>more than one parameter of type {@code Limit},</li>
  * <li>a parameter of type {@code Limit} and a parameter of type
- *     {@link PageRequest}</li>, or
+ *     {@link PageRequest}, or</li>
  * <li>a parameter of type {@code Limit} in combination with the
  *     {@code First} keyword.
  * </ul>
--- a/api/src/main/java/jakarta/data/Limit.java
+++ b/api/src/main/java/jakarta/data/Limit.java
@@ -25,9 +25,11 @@ import jakarta.data.page.PageRequest;
  * can be capped to a maximum amount
  * or be limited to a given positional range.</p>
  *
- * <p>{@code Limit} is optionally specified as a parameter to a
- * repository method in one of the parameter positions after the
- * query parameters. For example,</p>
+ * <p>A query method of a repository may have a parameter of type
+ * {@code Limit} if its return type indicates that it may return multiple
+ * entities. The parameter of type {@code Limit} must occur after the
+ * method parameters representing regular parameters of the query itself.
+ * For example,</p>
  *
  * <pre>
  * Product[] findByNameLike(String namePattern, Limit limit, {@code Sort<?>...} sorts);
--- a/api/src/main/java/jakarta/data/Limit.java
+++ b/api/src/main/java/jakarta/data/Limit.java
@@ -38,14 +38,13 @@ import jakarta.data.page.PageRequest;
  * secondMostExpensive50 = products.findByNameLike(pattern, Limit.range(51, 100), Sort.desc("price"));
  * </pre>
  *
- * <p>A repository method will fail if</p>
+ * <p>A repository method may not be declared with:
  * <ul>
- * <li>multiple {@code Limit} parameters are supplied to the
- *     same method.</li>
- * <li>{@code Limit} and {@link PageRequest} parameters are supplied to the
- *     same method.</li>
- * <li>a {@code Limit} parameter is supplied in combination
- *     with the {@code First} keyword.</li>
+ * <li>more than one parameter of type {@code Limit},</li>
+ * <li>a parameter of type {@code Limit} and a parameter of type
+ *     {@link PageRequest}</li>, or
+ * <li>a parameter of type {@code Limit} in combination with the
+ *     {@code First} keyword.
  * </ul>
  *
  * @param maxResults maximum number of results for a query.
--- a/api/src/main/java/jakarta/data/Limit.java
+++ b/api/src/main/java/jakarta/data/Limit.java
@@ -26,8 +26,7 @@ import jakarta.data.page.PageRequest;
  * or be limited to a given positional range.</p>
  *
  * <p>A query method of a repository may have a parameter of type
- * {@code Limit} if its return type indicates that it may return multiple
- * entities. The parameter of type {@code Limit} must occur after the
+ * {@code Limit}. The parameter of type {@code Limit} must occur after the
  * method parameters representing regular parameters of the query itself.
  * For example,</p>
  *
--- a/api/src/main/java/jakarta/data/Order.java
+++ b/api/src/main/java/jakarta/data/Order.java
@@ -60,16 +60,16 @@ import jakarta.data.repository.Query;
  * alphabetically by last name. Employees with the same salary and last name
  * are then sorted alphabetically by first name.</p>
  *
- * <p>A repository method will fail with a
- * {@link jakarta.data.exceptions.DataException DataException}
- * or a more specific subclass if</p>
- * <ul>
- * <li>an {@code Order} parameter is
- *     specified in combination with a {@link PageRequest} parameter with
- *     {@link PageRequest#sorts()}.</li>
- * <li>the database is incapable of ordering with the requested
- *     sort criteria.</li>
- * </ul>
+ * <p>A repository method may not be declared with more than one parameter
+ * of type {@code Order}.</p>
+ *
+ * <p>A repository method throws {@link IllegalArgumentException} if it is
+ * called with an argument of type {@code PageRequest} with nonempty sort
+ * criteria, and a separate argument of type {@link Order}.</p>
+ *
+ * <p>A repository method throws {@link UnsupportedOperationException} if
+ * the database is incapable of ordering the query results using the given
+ * sort criteria.</p>
  *
  * @param <T> entity class of the attributes that are used as sort criteria.
  */
--- a/api/src/main/java/jakarta/data/Order.java
+++ b/api/src/main/java/jakarta/data/Order.java
@@ -29,8 +29,7 @@ import jakarta.data.repository.Query;
  * <p>Requests sorting on various entity attributes.</p>
  *
  * <p>A query method of a repository may have a parameter of type
- * {@code Order} if its return type indicates that it may return
- * multiple entities. The parameter of type {@code Order} must occur
+ * {@code Order}. The parameter of type {@code Order} must occur
  * after the method parameters representing regular parameters of
  * the query itself.</p>
  *
--- a/api/src/main/java/jakarta/data/Order.java
+++ b/api/src/main/java/jakarta/data/Order.java
@@ -28,11 +28,11 @@ import jakarta.data.repository.Query;
 /**
  * <p>Requests sorting on various entity attributes.</p>
  *
- * <p>{@code Order} can be optionally specified as a
- * parameter to a repository find method in any of the positions
- * that are after the query parameters, or it can be used
- * to obtain a {@link PageRequest} that is similarly
- * specified as a parameter to a repository find method.</p>
+ * <p>A query method of a repository may have a parameter of type
+ * {@code Order} if its return type indicates that it may return
+ * multiple entities. The parameter of type {@code Order} must occur
+ * after the method parameters representing regular parameters of
+ * the query itself.</p>
  *
  * <p>The {@code Order} class is useful in combination with the
  * {@link StaticMetamodel} for helping to enforce type safety of
--- a/api/src/main/java/jakarta/data/Order.java
+++ b/api/src/main/java/jakarta/data/Order.java
@@ -67,9 +67,8 @@ import jakarta.data.repository.Query;
  * called with an argument of type {@link Order} and a separate argument
  * of type {@code PageRequest} that has nonempty sort criteria.</p>
  *
- * <p>A repository method throws {@link UnsupportedOperationException} if
- * the database is incapable of ordering the query results using the given
- * sort criteria.</p>
+ * <p>A repository method throws {@link DataException} if the database
+ * is incapable of ordering the query results using the given sort criteria.</p>
  *
  * @param <T> entity class of the attributes that are used as sort criteria.
  */
--- a/api/src/main/java/jakarta/data/Order.java
+++ b/api/src/main/java/jakarta/data/Order.java
@@ -67,8 +67,9 @@ import jakarta.data.repository.Query;
  * called with an argument of type {@link Order} and a separate argument
  * of type {@code PageRequest} that has nonempty sort criteria.</p>
  *
- * <p>A repository method throws {@link DataException} if the database
- * is incapable of ordering the query results using the given sort criteria.</p>
+ * <p>A repository method throws {@link jakarta.data.exceptions.DataException}
+ * if the database is incapable of ordering the query results using the given
+ * sort criteria.</p>
  *
  * @param <T> entity class of the attributes that are used as sort criteria.
  */
--- a/api/src/main/java/jakarta/data/Order.java
+++ b/api/src/main/java/jakarta/data/Order.java
@@ -64,8 +64,8 @@ import jakarta.data.repository.Query;
  * of type {@code Order}.</p>
  *
  * <p>A repository method throws {@link IllegalArgumentException} if it is
- * called with an argument of type {@code PageRequest} with nonempty sort
- * criteria, and a separate argument of type {@link Order}.</p>
+ * called with an argument of type {@link Order} and a separate argument
+ * of type {@code PageRequest} that has nonempty sort criteria.</p>
  *
  * <p>A repository method throws {@link UnsupportedOperationException} if
  * the database is incapable of ordering the query results using the given
--- a/api/src/main/java/jakarta/data/Sort.java
+++ b/api/src/main/java/jakarta/data/Sort.java
@@ -32,8 +32,7 @@ import java.util.Objects;
  * a {@link Direction} and a property.</p>
  *
  * <p>A query method of a repository may have a parameter or parameters
- * of type {@code Sort} if its return type indicates that it may return
- * multiple entities. Parameters of type {@code Sort} must occur after
+ * of type {@code Sort}. Parameters of type {@code Sort} must occur after
  * the method parameters representing regular parameters of the query
  * itself.</p>
  *
--- a/api/src/main/java/jakarta/data/Sort.java
+++ b/api/src/main/java/jakarta/data/Sort.java
@@ -31,11 +31,14 @@ import java.util.Objects;
  * sort criteria which includes a case sensitivity request,
  * a {@link Direction} and a property.</p>
  *
- * <p>Dynamic {@code Sort} criteria can be specified when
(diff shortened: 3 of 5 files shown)


otaviojava (approved):

api/src/main/java/jakarta/data/Order.java
```diff
@@ -60,16 +60,16 @@
...
+ * <p>A repository method may not be declared with more than one parameter
+ * of type {@code Order}.</p>
+ *
+ * <p>A repository method throws {@link IllegalArgumentException} if it is
+ * called with an argument of type {@code PageRequest} with nonempty sort
+ * criteria, and a separate argument of type {@link Order}.</p>
+ *
+ * <p>A repository method throws {@link UnsupportedOperationException} if
+ * the database is incapable of ordering the query results using the given
+ * sort criteria.</p>
```
njr-11, line 72: I was trying to think of why this was originally written as DataException (which is the generic exception used by Jakarta Data) and I realized that there are cases where a Jakarta Data provider won't automatically know if some entity attribute types are valid for sort criteria until it is tried on the database, and then at that point, will only know that some error was sent back by the database/Jakarta Persistence provider/other, but not be certain of the reason for the error.  Maybe it was due to a lack of support for sorting on that column, or maybe it was due to another error condition that is completely unrelated.

We could possibly address this by writing something like this,
```
 * <p>A repository method throws {@link UnsupportedOperationException} if
 * the Jakarta Data provider is able to detect that the database is incapable of
 * ordering the query results using the given sort criteria. Otherwise, if the cause
 * of the error might be an inability to apply the sort criteria or might be something
 * else, then the repository method raises {@link DataException}.</p>
```

but I suspect that is too complicated and would be best just to stick with DataException,
```suggestion
 * <p>A repository method throws {@link DataException} if the database
 * is incapable of ordering the query results using the given sort criteria.</p>
```

api/src/main/java/jakarta/data/Sort.java
```diff
@@ -75,16 +75,14 @@
...
- * <ul>
- * <li>a {@code Sort} parameter is
- *     specified in combination with a {@link PageRequest} parameter with
- *     {@link PageRequest#sorts()}.</li>
- * <li>the database is incapable of ordering with the requested
- *     sort criteria.</li>
- * </ul>
+ * <p>A repository method throws {@link IllegalArgumentException} if it is
+ * called with an argument of type {@code PageRequest} with nonempty sort
+ * criteria, and a separate argument or arguments of type {@link Sort}.</p>
```
njr-11, line 80: similar grammar fix
```suggestion
 * <p>A repository method throws {@link IllegalArgumentException} if it is
 * called with an argument or arguments of type {@link Sort} and a separate
 * argument of type {@code PageRequest} with nonempty sort criteria.</p>
```

api/src/main/java/jakarta/data/Sort.java
```diff
@@ -75,16 +75,14 @@
...
+ *
+ * <p>A repository method throws {@link UnsupportedOperationException} if
+ * the database is incapable of ordering the query results using the given
+ * sort criteria.</p>
```
njr-11, line 84: See earlier comment
```suggestion
 * <p>A repository method throws {@link DataException} if
 * the database is incapable of ordering the query results using the given
 * sort criteria.</p>
```

api/src/main/java/jakarta/data/page/PageRequest.java
```diff
@@ -48,25 +48,24 @@
...
  * </ul>
  *
+ * <p>A repository method throws {@link IllegalArgumentException} if it
+ * is called with an argument of type {@code PageRequest} with nonempty
+ * sort criteria, and a separate argument or arguments of type {@link Sort}
+ * or {@link Order}.</p>
+ *
+ * <p>A repository method throws {@link UnsupportedOperationException} if
+ * the database is incapable of ordering the query results using the given
+ * sort criteria.</p>
```
njr-11, line 67: refer to earlier comment
```suggestion
 * <p>A repository method throws {@link DataException} if
 * the database is incapable of ordering the query results using the given
 * sort criteria.</p>
```

api/src/main/java/jakarta/data/page/PageRequest.java
```diff
@@ -48,25 +48,24 @@
  * }
  * </pre>
  *
- * <p>A repository method will fail with a
- * {@link jakarta.data.exceptions.DataException DataException} or a more
- * specific subclass if:</p>
+ * <p>A repository method may not be declared with:</p>
```
njr-11, line 51: I don't understand why this list of user errors is split where some of them are "A repository method may not be declared with:" and others are listed as IllegalArgumentException.  We should treat all of the user errors the same and require IllegalArgumentException for all of them (except for the sorting not supported one).

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -1119,38 +1119,65 @@ For relational databases, the logical operator `And` takes precedence over `Or`,
...
+A repository method must throw `UnsupportedOperationException` if it has:
+
+- more than one parameter of type `PageRequest` or `Limit`,
+- a parameter of type `PageRequest` and a parameter of type `Limit`,
+- a parameter of type `PageRequest` or `Limit`, in combination with the keyword `First`, or
+- more than one parameter of type `Order`.
+
+Alternatively, a Jakarta Data provider is permitted to reject such a repository method declaration at compile time.
+
+A repository method must throw `IllegalArgumentException` if it is called with an argument of type `PageRequest` with nonempty sort criteria, and a separate argument or arguments of type `Sort` or `Order`.
```
njr-11, line 1140: IllegalArugmentException was stated for this in the JavaDoc updates earlier in this pull.
```suggestion
A repository method must throw `IllegalArugmentException` if it has:

- more than one parameter of type `PageRequest` or `Limit`,
- a parameter of type `PageRequest` and a parameter of type `Limit`,
- a parameter of type `PageRequest` or `Limit`, in combination with the keyword `First`, or
- more than one parameter of type `Order`.
- an argument or arguments of type `Sort` or `Order` and a separate argument of type `PageRequest` with nonempty sort criteria.

Alternatively, a Jakarta Data provider is permitted to reject such a repository method declaration at compile time.
```

njr-11 (changes requested):

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -1119,38 +1119,65 @@ For relational databases, the logical operator `And` takes precedence over `Or`,
...
+
+A repository method must throw `UnsupportedOperationException` if the database is incapable of ordering the query results using the given sort criteria.
```
njr-11, line 1142: I missed this one. Refer to discussion in the earlier review comment
```suggestion
A repository method must throw `DataException` if the database is incapable of ordering the query results using the given sort criteria.
```

api/src/main/java/jakarta/data/page/PageRequest.java
gavinking: Well that was the point was trying to make earlier in the issue description: these two classes of error are _completely_ different kinds of things:

- one is an error in a method signature made by the person who wrote the repository interface, and can usually de detected at compilation time, or, at worst, at startup time, for implementations which don't do work at build time, and
- the other is an error made _by code which calls the method_, which may not even be in the same module, and which is not even in principle detectable at compilation time.

Bundling these two kinds of failure into the same list is a category error.

One of them is an error in the definition of the repository (assigned `UnsupportedOperationException` elsewhere in the spec), and one of them is an error in the arguments passed to the repository method (a classic example of an `IllegalArgumentException`).

gavinking: You're calling them both "user error", but there are two kinds of users here:

- the user of Jakarta Data (the person who creates repositories), and
- the user of the repository, who maybe doesn't know anything about Jakarta Data.

spec/src/main/asciidoc/repository.asciidoc
gavinking: I believe the list here is consistent with the list in the Javadoc for `PageRequest`.

Please compare:

```java
 * <p>A repository method may not be declared with:</p>
 * <ul>
 * <li>more than one parameter of type {@code PageRequest} or {@link Limit},
 * <li>a parameter of type {@code PageRequest} and a parameter of type
 *     {@link Limit}, or</li>
 * <li>a parameter of type {@code PageRequest} in combination with the
 *     keyword {@code First}.</li>
 * </ul>
 *
 * <p>A repository method throws {@link IllegalArgumentException} if it
 * is called with an argument of type {@code PageRequest} with nonempty
 * sort criteria, and a separate argument or arguments of type {@link Sort}
 * or {@link Order}.</p>
```

```java
A repository method must throw `UnsupportedOperationException` if it has:

- more than one parameter of type `PageRequest` or `Limit`,
- a parameter of type `PageRequest` and a parameter of type `Limit`,
- a parameter of type `PageRequest` or `Limit`, in combination with the keyword `First`, or
- more than one parameter of type `Order`.

Alternatively, a Jakarta Data provider is permitted to reject such a repository method declaration at compile time.

A repository method must throw `IllegalArgumentException` if it is called with an argument of type `PageRequest` with nonempty sort criteria, and a separate argument or arguments of type `Sort` or `Order`.
```

I believe they are saying the same thing.

api/src/main/java/jakarta/data/page/PageRequest.java
njr-11: Okay, I see what distinction you are making here. I got confused because for one category (invalid method signature) it doesn't state what the exception is and for some reason I assumed the JavaDoc said it was IllegalArgumentException as well, but I can see now you didn't write that and left it undefined in both JavaDoc and spec as to what the exception. I suppose this is fine. The user has been informed to avoid the invalid combinations of parameters, which is the part that really matters here.

gavinking: Yeah, I don't want to state that in Javadoc because the user should definitely not be writing code which depends on that exception type.

api/src/main/java/jakarta/data/Order.java
```diff
@@ -28,11 +28,11 @@
...
- * <p>{@code Order} can be optionally specified as a
- * parameter to a repository find method in any of the positions
- * that are after the query parameters, or it can be used
- * to obtain a {@link PageRequest} that is similarly
- * specified as a parameter to a repository find method.</p>
+ * <p>A query method of a repository may have a parameter of type
+ * {@code Order} if its return type indicates that it may return
+ * multiple entities. The parameter of type {@code Order} must occur
+ * after the method parameters representing regular parameters of
+ * the query itself.</p>
```
njr-11, line 35: Restricting this to multiple return type will prevent users from writing queries like this,
```
Optional<Employee> findFirstByJobTitle(String jobTitle, Order<Employee> sortBy);

highestPaid = employees.findFirstByJobTitle("Software Engineer", Order.by(
                                            _Employee.salary.desc(),
                                            _Employee.id.desc()));
```
and is contradictory to the description of NonUniqueResultException which instructs users to take this sort of approach with singular result to avoid the exception if they really want a single entity returned.

```suggestion
 * <p>A query method of a repository that performs a find operation
 * may have a parameter of type {@code Order}. The parameter of type
 * {@code Order} must occur after the method parameters representing
 * regular parameters of the query itself.</p>
```

api/src/main/java/jakarta/data/Sort.java
```diff
@@ -31,11 +31,14 @@
...
- * <p>Dynamic {@code Sort} criteria can be specified when
- * requesting a {@link PageRequest#sortBy(Sort) page} of results,
- * or can be optionally specified as
- * parameters to a repository find method in any of the positions that are after
- * the query parameters.</p>
+ * <p>A query method of a repository may have a parameter or parameters
+ * of type {@code Sort} if its return type indicates that it may return
+ * multiple entities. Parameters of type {@code Sort} must occur after
+ * the method parameters representing regular parameters of the query
+ * itself.</p>
```
njr-11, line 38:
```suggestion
 * <p>A query method of a repository that performs a find operation
 * may have a parameter or parameters of type {@code Sort}.
 * Parameters of type {@code Sort} must occur after the method
 * parameters representing regular parameters of the query itself.</p>
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -1121,7 +1121,12 @@ Refer to the Jakarta Data module JavaDoc section on "Return Types for Repository
 
 === Special Parameters for Limits, Sorting, and Pagination
 
-An <<Annotated Query methods,annotated>>, <<Parameter-based automatic query methods,parameter-based>>, or <<Query by Method Name,method name>> query method may have _special parameters_ of type `Limit`, `Order`, `Sort`, or `PageRequest`.
+An <<Annotated Query methods,annotated>>, <<Parameter-based automatic query methods,parameter-based>>, or <<Query by Method Name,method name>> query method may have _special parameters_ of type `Limit`, `Order`, `Sort`, or `PageRequest` if the method return type indicates that the method may return multiple entities, that is, if the return type is:
+
+- an array type,
+- `List` or `Stream`, or
+- `Page` or `CursoredPage`.
+
```
njr-11, line 1129:
```suggestion
An <<Annotated Query methods,annotated>>, <<Parameter-based automatic query methods,parameter-based>>, or <<Query by Method Name,method name>> query method that performs a find operation may have _special parameters_ of type `Limit`, `Order`, `Sort`, or `PageRequest`.
```

njr-11 (changes requested):

api/src/main/java/jakarta/data/Order.java
gavinking: Yeah, I mean, OK, I guess I think that such a repository method has _extremely_ weak semantics.

Find the "first" X where "first" is something completely arbitrary determined by the caller.

I can't see myself ever writing a repository method like that TBH.

gavinking: In particular, `PageRequest` is pretty aesthetically-offensive for a query which returns a single entity.

njr-11: > In particular, `PageRequest` is pretty aesthetically-offensive for a query which returns a single entity.

I didn't mean to make the comment on `PageRequest`, only `Order` and `Sort`.

gavinking: If you _really_ want this, 63686513872418b11f76ab5e39c4664b51f5b307 rolls back that change for everything except `PageRequest`.

gavinking: > I didn't mean to make the comment on `PageRequest`, only `Order` and `Sort`.

What about `Limit`?

gavinking: >  I were going to write this method, I would rather write it as

Yeah, yeah, that's the point I was trying to make. You would want the ordering to be specified statically, I believe.

> Actually, I'd rather use `@Find` and name it "highestPaid" instead

Yeah, I agree, it would be _very_ nice to be able to write:

```java
@Find @First
@OrderBy(value = _Employee.SALARY, descending = true)
@OrderBy(value = _Employee.ID)
Optional<Employee> highestPaid(String jobTitle);
```

Where `@First` would suppress the exception that you would otherwise receive by adding `limit 1` to the query, and so you would just get the top result.

At some point we should add that.

njr-11: That's a good idea separating out `@First` from `@Find` because then it could also apply to `@Query`.  Just jotting this down for future reference.

For now, having thought about this some more, I don't like how `NonUniqueResultException` is pointing at a solution that only currently works with the Query by Method Name pattern (and even there it is awkward to use) and not the other patterns.  Let's go with what you proposed in this pull to only allows these for multiple return type and then we can update the recommendation in `NonUniqueResultException` to be to use a multiple return type.  I'll locate my review comments and mark them resolved with this.

njr-11: Oops, I was too late for that. I'll just approve this and if you want to put the language back limiting to multiple, go ahead.

njr-11 (approved):

njr-11:
I did my best to resolve the merge conflict on this one so that it can be merged in advance of the cutoff for non-Query Language changes. Both pulls were rewriting the same portion of the paragraph in slightly different ways.

njr-11:
It looks like the CI has a failure on another class. I'll look into resolving that as well:
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.6.0:jar (attach-javadocs) on project jakarta.data-api: MavenReportException: Error while generating Javadoc: 
Error:  Exit code: 1
Error:  /home/runner/work/data/data/api/src/main/java/jakarta/data/Limit.java:47: error: text not allowed in <ul> element
Error:   *     {@link PageRequest}</li>, or
Error:                                 ^
Error: [ERROR] 1 error
```

api/src/main/java/jakarta/data/Limit.java
```diff
@@ -39,14 +40,13 @@
...
  * <ul>
- * <li>multiple {@code Limit} parameters are supplied to the
- *     same method.</li>
- * <li>{@code Limit} and {@link PageRequest} parameters are supplied to the
- *     same method.</li>
- * <li>a {@code Limit} parameter is supplied in combination
- *     with the {@code First} keyword.</li>
+ * <li>more than one parameter of type {@code Limit},</li>
+ * <li>a parameter of type {@code Limit} and a parameter of type
+ *     {@link PageRequest}</li>, or
```
njr-11, line 47: Needed for the CI to pass. Error is:

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.6.0:jar (attach-javadocs) on project jakarta.data-api: MavenReportException: Error while generating Javadoc: 
Error:  Exit code: 1
Error:  /home/runner/work/data/data/api/src/main/java/jakarta/data/Limit.java:47: error: text not allowed in <ul> element
Error:   *     {@link PageRequest}</li>, or
Error:                                 ^
Error: [ERROR] 1 error
```

```suggestion
 * <li>a parameter of type {@code Limit} and a parameter of type
 *     {@link PageRequest}, or</li>
```

api/src/main/java/jakarta/data/Order.java
gavinking: So, Nathan, what was your final conclusion here? 

Should I roll back https://github.com/jakartaee/data/pull/530/commits/63686513872418b11f76ab5e39c4664b51f5b307 or no?

njr-11: > So, Nathan, what was your final conclusion here?
> 
> Should I roll back [6368651](https://github.com/jakartaee/data/commit/63686513872418b11f76ab5e39c4664b51f5b307) or no?

Either way is fine with me.  This is a very minor detail and so I chose to approve and merge the pull yesterday (Otavio had already approved it) to ensure we could get the rest of your changes in before the March 8 cutoff.

gavinking: OK, well, hmmmm, I guess my preference would be to put the restriction back in, if only because it's really easy to remove a restriction like this later if we decide it was a mistake.

njr-11: > I guess my preference would be to put the restriction back in, if only because it's really easy to remove a restriction like this later if we decide it was a mistake.

That's a very good point. My only hesitation is that we passed the deadline, but this is a gray area because it could be framed as reverting a commit that shouldn't have gone in rather than adding something.

gavinking: I don't think rolling back a single commit, with the effect that there are fewer things to test and implement, is going to interfere with finishing the work on JDQL.

PR #544, opened 2024-03-14
alignment of @Find and @Delete with JDQL work

gavinking:
For #458.

In readiness for the integration of JDQL, we need to clean up this section which focussed on Query by Method Name, and said some things which were ambiguous or not quite right as applied to `@Query` and `@Find`.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/By.java
+++ b/api/src/main/java/jakarta/data/repository/By.java
@@ -27,10 +27,10 @@ import java.lang.annotation.Target;
  * <p>Annotates a parameter of a repository method, specifying a mapping to
  * a persistent field:</p>
  * <ul>
- * <li>if an explicit {@linkplain #value field name} is specified, the
- *     parameter maps to the persistent field with the specified name, or
- * <li>otherwise, if the name is missing, the parameter maps to the unique
- *     identifier field or property.
+ * <li>if a {@linkplain #value field name} is specified, the parameter maps
+ *     to the persistent field with the specified name, or
+ * <li>if the special value {@value #ID} is specified, the parameter maps
+ *     to the unique identifier field or property.
  * </ul>
  * <p>Arguments to the annotated parameter are compared to values of the
  * mapped persistent field.</p>
@@ -44,7 +44,7 @@ import java.lang.annotation.Target;
  * public interface People {
  *
  *     {@code @Find}
- *     Person findById(&#64;By String id); // maps to Person.ssn
+ *     Person findById(&#64;By(ID) String id); // maps to Person.ssn
  *
  *     {@code @Find}
  *     List&lt;Person&gt; findNamed(&#64;By("firstName") String first,
@@ -84,12 +84,19 @@ import java.lang.annotation.Target;
 public @interface By {
 
     /**
-     * The name of the persistent field mapped by the annotated parameter.
-     * Defaults to the name of the unique identifier field or property of the
-     * entity.
+     * The name of the persistent field mapped by the annotated parameter,
+     * or {@value #ID} to indicate the unique identifier field or property
+     * of the entity.
      *
-     * @return the persistent field name, or {@code ""} to indicate the unique
-     *         identifier field.
+     * @return the persistent field name, or {@value #ID} to indicate the
+     *         unique identifier field.
      */
-    String value() default "";
+    String value();
+
+    /**
+     * The special value which indicates the unique identifier field or
+     * property. The annotation {@code By(ID)} maps a parameter to the
+     * identifier.
+     */
+    String ID = "#id";
 }
--- a/api/src/main/java/jakarta/data/repository/By.java
+++ b/api/src/main/java/jakarta/data/repository/By.java
@@ -24,37 +24,72 @@ import java.lang.annotation.Target;
 
 
 /**
- * <p>Annotates a parameter of a repository method to specify the name of an
- * entity attribute against which to compare.</p>
+ * <p>Annotates a parameter of a repository method, specifying a mapping to
+ * a persistent field:</p>
+ * <ul>
+ * <li>if an explicit {@linkplain #value field name} is specified, the
+ *     parameter maps to the persistent field with the specified name, or
+ * <li>otherwise, if the name is missing, the parameter maps to the unique
+ *     identifier field or property.
+ * </ul>
+ * <p>Arguments to the annotated parameter are compared to values of the
+ * mapped persistent field.</p>
+ * <p>The field name may be a compound name like {@code address.city}.</p>
  *
- * <p>Example usage for a {@code Person} entity with attributes
- * {@code id}, {@code firstName}, and {@code lastName}:</p>
+ * <p>For example, for a {@code Person} entity with attributes {@code ssn},
+ * {@code firstName}, {@code lastName}, and {@code address} we might have:</p>
  *
  * <pre>
  * &#64;Repository
- * public interface People extends BasicRepository&lt;Person, Long&gt; {
+ * public interface People {
+ *
+ *     {@code @Find}
+ *     Person findById(&#64;By String id); // maps to Person.ssn
  *
  *     {@code @Find}
  *     List&lt;Person&gt; findNamed(&#64;By("firstName") String first,
  *                            &#64;By("lastName") String last);
+ *
+ *     {@code @Find}
+ *     Person findByCity(&#64;By("address.city") String city);
  * }
- * ...
- * found = people.findNamed(first, last);
  * </pre>
  *
  * <p>The {@code By} annotation is unnecessary when the method parameter name
  * matches the entity attribute name and the application is compiled with the
  * {@code -parameters} compiler option that makes parameter names available
  * at run time.</p>
+ *
+ * <p>Thus, when this compiler option is enabled, the previous example may be
+ * written without the use of {@code By}:</p>
+ *
+ * <pre>
+ * &#64;Repository
+ * public interface People {
+ *
+ *     {@code @Find}
+ *     Person findById(String ssn);
+ *
+ *     {@code @Find}
+ *     List&lt;Person&gt; findNamed(String firstName,
+ *                            String lastname);
+ *
+ *     {@code @Find}
+ *     Person findByCity(String address_city);
+ * }
+ * </pre>
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.PARAMETER)
 public @interface By {
 
     /**
-     * The name of the entity attribute against which to compare.
+     * The name of the persistent field mapped by the annotated parameter.
+     * Defaults to the name of the unique identifier field or property of the
+     * entity.
      *
-     * @return the entity attribute name.
+     * @return the persistent field name, or {@code ""} to indicate the unique
+     *         identifier field.
      */
-    String value();
+    String value() default "";
 }
--- a/api/src/main/java/jakarta/data/repository/Delete.java
+++ b/api/src/main/java/jakarta/data/repository/Delete.java
@@ -91,6 +91,8 @@ import java.lang.annotation.Target;
  * {@code @Save} are mutually-exclusive. A given method of a repository interface may have at most one {@code @Find}
  * annotation, lifecycle annotation, or query annotation.
  * </p>
+ *
+ * @see By
  */
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
--- a/api/src/main/java/jakarta/data/repository/Delete.java
+++ b/api/src/main/java/jakarta/data/repository/Delete.java
@@ -30,7 +30,7 @@ import java.lang.annotation.Target;
  * <p>The {@code Delete} annotation indicates that the annotated repository method deletes the state of one or more
  * entities from the database. The method must follow one of the following patterns.
  * </p>
- * <h2>Parameter for Entity Instances</h2>
+ *
  * <p>A {@code Delete} method might accept an instance or instances of an entity class. In this case, the method must
  * have exactly one parameter whose type is either:
  * </p>
@@ -56,11 +56,18 @@ import java.lang.annotation.Target;
  * if the entity with a matching identifier does not have a matching version, the annotated method must raise
  * {@link jakarta.data.exceptions.OptimisticLockingFailureException}.
  * </p>
- * <h2>Without Parameters</h2>
- * <p>Alternatively, the {@code Delete} annotation may be applied to a repository method with no parameters, indicating
- * that the annotated method deletes all instances of the primary entity type. In this case, the annotated method must
- * either be declared {@code void}, or return {@code int} or {@code long}.
- * </p>
+ *
+ * <p>Alternatively, as a special exception, the {@code Delete} annotation may be used to annotate a repository method
+ * with no a parameter of entity type:</p>
+ * <ul>
+ * <li>{@code Delete} may be applied to a repository method with no parameters, indicating that the annotated method
+ *     deletes all instances of the primary entity type. The annotated method must either be declared {@code void}, or
+ *     return {@code int} or {@code long}.</li>
+ * <li>{@code Delete} may be applied to a repository method with exactly one parameter annotated {@code By}, with a
+ *     defaulted {@linkplain By#value field name}, and with the same type as the unique identifier field or property of
+ *     the primary entity of the repository, indicating that the annotated method deletes the entity instance with the
+ *     given unique identifier. The annotated method must be declared {@code void}.
+ * </ul>
  * <p>Annotations such as {@code @Find}, {@code @Query}, {@code @Insert}, {@code @Update}, {@code @Delete}, and
  * {@code @Save} are mutually-exclusive. A given method of a repository interface may have at most one {@code @Find}
  * annotation, lifecycle annotation, or query annotation.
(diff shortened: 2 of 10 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -460,53 +460,68 @@ create table Book (
...
 |`Sort.asc("address_zipCode")`
 
-|`@By` or `@OrderBy` annotation value
+|`@By` or `@OrderBy` | Annotation value
 |`.` or `_`
 |`@Find List<Order> find(@By("address.zipCode") int zip);`
 |===
 
-For a given entity property name, delimiter usage must be consistent. Either the delimiter must always be used within the entity property name to delimit subcomponents or the delimiter must never be used within the entity property name. Except in the case of `@Query` where the delimiter is required by the query language, delimiters can be omitted entirely from an entity property name when it is unnecessary to disambiguate the entity property to which the name refers.
+NOTE: Application programmers are strongly encouraged to follow Java's camel case naming standard for fields of entities, relations, and embeddable classes, avoiding underscores in field names. The resolution algorithm for persistent field identification relies on the use of underscore as a delimiter. Adhering to the camel case naming convention ensures consistency and eliminates ambiguity.
```
njr-11, line 505: This requirement should not have been moved to a section that is only for Query by Method Name. This rule is needed to prevent mixed usage like
```
@Find
@OrderBy("location_mailingAddress.zipCode")
List<Business> find(@By("location.mailingAddress_zipCode") int zip);
```
or
```
Sort.asc("location_mailingAddress.zipCode");
```

```suggestion
For a given entity property name in which a delimiter is used, the same delimiter must be used between every pair of persistent field names within the compound name.

NOTE: Application programmers are strongly encouraged to follow Java's camel case naming standard for fields of entities, relations, and embeddable classes, avoiding underscores in field names. The resolution algorithm for persistent field identification relies on the use of underscore as a delimiter. Adhering to the camel case naming convention ensures consistency and eliminates ambiguity.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -460,53 +460,68 @@ create table Book (
...
+
+
+==== Persistent Field Names in Query by Method Name
 
-The resolution of properties involves the following steps:
+For _Query by Method Name_, the use of delimiters within a compound name is optional. Delimiters may be omitted entirely from a compound name when they are not needed to disambiguate the persistent field to which the name refers. But for a given entity property name, delimiter usage must be consistent: either the delimiter must be used between every pair of persistent field names within the compound name, or it must not occur within the compound name.
```
njr-11, line 510: This can be taken out of this section because it belongs above (see prior comment) where it applies to the other patterns as well as Query by Method Name

```suggestion
For _Query by Method Name_, the use of delimiters within a compound name is optional. Delimiters may be omitted entirely from a compound name when they are not needed to disambiguate the persistent field to which the name refers.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -460,53 +460,68 @@ create table Book (
...
 
-4. *Resolution Outcome*: If the framework successfully identifies a property in the domain class or along the specified traversal path that matches the extracted property name, it uses that property.
+3. Otherwise, if no match is found among the fields of the entity, the extracted name is matched against the fields of entity classes and embedded classes reachable from the entity class, interpreting the extracted name as a compound name, as outlined in the previous section, both with and without the optional delimiter. If the compound name assigned to a persistent field matches the extracted name, also interpreted as a compound name, and ignoring case, then the extracted name resolves to that field.
 
-When `@Query` is used, the Jakarta Data specification defers to the rules of the query language on whether the delimiter is required and whether the case must match.
+4. If no matching persistent field is found in either of the previous steps, the provider is permitted to reject the query method, or to throw `UnsupportedOperationException` when the method is called.
```
njr-11, line 520:
```suggestion
4. If no matching persistent field is found in either of the previous steps, the provider is permitted to reject the query method or to throw `UnsupportedOperationException` when the method is called.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -557,10 +572,6 @@ The `Order` entity has an `addressZipCode` field, as well as an `address` field
 List<Order> findByAddress_zipcode(int zipCode);
 ----
 
-WARNING: Define entity properties following standard Java naming conventions for camel case, using underscore only as the last resort.
-
-In all places where entity property names can be specified other than `@Query`, `Id` is an alias for the entity property that is designated as the id. Entity property names that are used in _Query by Method Name_ must not contain reserved words.
```
njr-11, line 562: This section already had an exception for `Query` so it should have been fine.
Removing this will mean you can no longer write the following when the unique identifier attribute is `productIdNum`:

```
@Find
Optional<Product> get(String id);
```

njr-11 (commented):
Mostly looks good, but I'm unsure whether it was intentional to take away the `id` alias behavior from Parameter-based Automatic Query methods and other usage that isn't Query by Method Name.

gavinking:
> Mostly looks good, but I'm unsure whether it was intentional to take away the `id` alias behavior from Parameter-based Automatic Query methods and other usage that isn't Query by Method Name.

Yeah so at first I thought that was just a mistake, since `Id` with initial uppercase doesn't make sense as a parameter name, nor as a query path element. (It should be `id`, if anything.)

But then I realized why you want it (for generic super type repositories) and now I'm thinking about what is the best way to put it back in.

spec/src/main/asciidoc/repository.asciidoc
gavinking: Yeah but `id` != `Id`.

gavinking:
The issue is that once you reserve `id` to mean the `@Id` field, you prevent people from using `id` as a non-`@Id` field. Which is a bit of a pain. In our product we work around that issue by letting you write `id(entity)` in functional form but that doesn't quite work for `@Find` methods.

gavinking:
The solution space for `@Find` methods includes things like:

```java
@Find T findById(@ById K id)
```

and even just:

```java
@Find T findById(K _id)
```

njr-11:
> The issue is that once you reserve `id` to mean the `@Id` field, you prevent people from using `id` as a non-`@Id` field. Which is a bit of a pain. In our product we work around that issue by letting you write `id(entity)` in functional form but that doesn't quite work for `@Find` methods.

I guess I saw it as an advantage that a user wouldn't be able to put an `id` field on an entity and use it for something other than the `@Id`.  That seems like a confusing, bad practice.  Do users really want to do that?

gavinking:
> Do users really want to do that?

After > 20 years of this, I'm still surprised by all the things users want to do :)

But I think in this case it makes sense. If we are going to have `id` automatically mean the identifier, we may as well get rid of the `@Id` annotation and tell them to just name the identifier field `id`. Because they already can't use that name for anything else.

So that's not the idea of the annotation-based model. The idea of the annotation-based model is that we attach semantics to annotations, not to naming conventions.

njr-11:
> But then I realized why you want it (for generic super type repositories) and now I'm thinking about what is the best way to put it back in.

In the future, I could see the remaining Query by Method Name methods of `BasicRepository` being compatibly made into valid Parameter-based Automatic Query Methods by adding an `ID="id"` constant and adding `@By(ID)`,

This one would already be possible:
```
    @Find
    Optional<T> findById(@By(ID) K id);
```

Others would be possible in the future if we were to allow things like,

```
    @Delete
    void deleteById(@By(ID) K id);
```

It would make the built-in repositories more explicit and less magic.

gavinking:
All that said, I think it would probably be fine to punt on this for now and say that if the entity has no persistent field named `id`, then `id` refers to the identifier field. (In both `@Find` methods and in JDQL.)

gavinking:
> > But then I realized why you want it (for generic super type repositories) and now I'm thinking about what is the best way to put it back in.
> 
> In the future, I could see the remaining Query by Method Name methods of `BasicRepository` being compatibly made into valid Parameter-based Automatic Query Methods by adding an `ID="id"` constant and adding `@By(ID)`,

That's also an option, I suppose. But I guess I think it would be just as easy, and more elegant, to just add a `@ById` annotation. It's literally 6 lines of code.

> It would make the built-in repositories more explicit and less magic.

I _definitely_ want to see that happen ASAP.

gavinking:
> > But I guess I think it would be just as easy, and more elegant, to just add a `@ById` annotation. It's literally 6 lines of code.
> 
> Alternatively, an empty `@By` annotation could mean "by id"

Y'know, I don't actually hate this as much as I thought I would when I was typing the comment. It's growing on me.

gavinking:
Alright, so I've been letting this stew for a bit, and here's the range of options I can see:

1] Reserve `id`
------------
The field name `id` always means the id field. It's not the end of the world, but all I know is that this is what I did in the _very_ early days of Hibernate and we quickly moved away from it, and replaced it with a function. JPA has never treated the name `id` as special.

2] Reserve `_id` or `Id` or `ID` or `@id`
------------
We could very much reduce the possibility of collision by choosing something that's does not conform to Java's naming convention, or is not a legal Java identifier. So you could write:

```java
@Find 
T findById(@By("_id") K id);
```
```java
@Delete 
void deleteById(@By("_id") K id);
```
```java
@Query("where _id = :id")
Entity findById(String id);
```
This works. But it looks and feels like a workaround. It looks like something we will want to "fix" later.

3] Introduce `@ById`
---------
On the other hand, a new annotation is unambiguous, less verbose, and doesn't look like a wart.

```java
@Find 
T findById(@ById K id);
```
```java
@Delete 
void deleteById(@ById K id);
```
The only thing to say against this option is that it doesn't work for JDQL. But (a) the natural thing in JDQL would be an `id()` function, like in HQL, and (b) we can add that later.

4] Just use `@By`
---------
Alternatively, we could simply make the `value` member of `@By` optional, and let them write:

```java
@Find 
T findById(@By K id);
```
```java
@Delete 
void deleteById(@By K id);
```
This is actually ... well, perfectly fine, AFAICT. If we went down this path, I don't think it's something we would feel the need to "fix" in the future.

It's easy for me to write up either of the last two options.

gavinking:
> 4] Just use `@By`

I just pushed two commits which show what this would look like. The impact is very minimal. Let's talk it through on the call later today.

njr-11:
The scenario with `BasicRepository` methods is one aspect of this. The other piece of this is having consistent behavior across Query by Method Name and Parameter-based Automatic Query Methods.  It should be easy for a repository to convert from Query by Method Name to Parameter-based Automatic Query Methods without impacting behavior that repository users see.

For example, these methods ought to be interchangeable:

```
// Query by Method Name
List<Car> findByMakeAndModel(String make, String model, Sort<?>... sortBy);
```

```
// Parameter-based Automatic Query Method (with the compiler option enabled)
@Find
List<Car> findByMakeAndModel(String make, String model, Sort<?>... sortBy);
```

Currently, they are interchangeable (assuming we fix the mistake where we didn't explicitly say that language around `Id` is meant to be case-insensitive). The repository user can write the following and not care which pattern is used:

```
List<Car> found = cars.findByMakeAndModel("Honda", "Civic",
                                          Sort.desc("year"),
                                          Sort.desc("price"),
                                          Sort.asc("id"));
```

But if we change the spec so that `id` is only an alias when Query by Method Name is used, then the above will have different behavior depending on which pattern is used.  That is confusing and puts obstacles in the way of a repository converting from Query by Method Name to Parameter-based Automatic Query Methods.

Id should either be an alias in both Query by Method Name and Parameter-based Automatic Query Methods or it should be an alias in neither.

Also, I thought some more about my previous question,

> I guess I saw it as an advantage that a user wouldn't be able to put an `id` field on an entity and use it for something other than the `@Id`. That seems like a confusing, bad practice. Do users really want to do that?

I stated that wrongly. The user will define the entity according to what the spec says is valid. The problem would be if they are stuck with an `ID` column in the database that is not the unique identifier.  But for that, they can rely on the entity model,

```
@Column(name="ID")
String myFieldName;
```

gavinking:
Commits 493f59397555df3c07fdebcd3a1531b851c1e774, ba214b5add0106ffd46e78c28f5dc148536d3072, and c1f31e3 remove the special-casing of "id" everywhere, as discussed.

Still TODO: define `@Delete` as a built-in automatic query annotation. This requires *slightly* generalizing section 4.3.

api/src/main/java/jakarta/data/repository/By.java
```diff
@@ -24,37 +24,79 @@
...
+     *         unique identifier field.
      */
     String value();
+
+    /**
+     * The special value which indicates the unique identifier field or
+     * property. The annotation {@code By(ID)} maps a parameter to the
+     * identifier.
+     */
+    String ID = "#id";
```
otaviojava, line 101: At NoSQL, we usually have the "_id" as the key word to ID. Maybe we can use the same nomenclature here.

gavinking: Well I think `_id` is natural, but it's still a legal Java identifier.

gavinking:
> Still TODO: define `@Delete` as a built-in automatic query annotation. This requires _slightly_ generalizing section 4.3.

Done in https://github.com/jakartaee/data/pull/544/commits/fe0c78fc41b4ff42bee04dbc8e6ef10cc6e02898.

@otaviojava @njr-11 I believe this PR now correctly reflects the outcome of our conversation on Friday.

And I think it's a very nice improvement to the specification, that removes some roadblocks to alignment between the query facilities.

api/src/main/java/jakarta/data/repository/By.java
njr-11: Right, the point is to use a value that not a valid Java identifier so that we do not interfere with a user who wants to have an entity attribute that is named `id` or `_id` or `ID` or similar which is not the unique identifier.  I think `#id` is a good choice because it will be meaningful if anyone happens to output the value. For the most part, the value doesn't matter other than ensuring it is not a valid Java identifier.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -591,8 +591,6 @@ For each field of the metamodel class that is to be initialized by a Jakarta Dat
 - The name of the field, ignoring case, must match the name of an entity attribute, with the `_` character in the field name delimiting the attribute names of hierarchical structures or relationships, such as embedded classes.
 - The value of the field must be uninitialized or `null`.
 
-If the entity class has a unique identifier that is represented by a single entity attribute, then the application can also include a field that follows the above criteria, except that it has a field name of `id` or `ID`.
```
njr-11, line 585: This is one of the places. I believe it also needs to be removed from StaticMetamodel JavaDoc:

```
 * <p>Additionally, a field that meets the above criteria except for the name and is named
 * {@code id} or {@code ID} must be assigned by the Jakarta Data provider for the unique
 * identifier entity attribute if a single entity attribute represents the unique identifier.</p>
```

gavinking: Right, missed that one. Thanks, fixed.

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -26,13 +26,21 @@
...
  * <p>The {@code Find} annotation indicates that the annotated repository method executes a query to retrieve entities
- * based on specified parameters. The method parameters must match the types and names of the corresponding fields of
- * the entity being queried. There is no specific naming convention for methods annotated with {@code @Find}; they may
- * be named arbitrarily, and their names do not carry any specific semantic meaning. The method return type identifies
- * the entity type returned by the query.
+ * based on its parameters and on the arguments assigned to its parameters. The method return type identifies the entity
+ * type returned by the query. Each parameter of the annotated method must either:
+ * </p>
+ * <ul>
+ * <li>have exactly the same type and name as a persistent field or property of the entity class, or</li>
```
njr-11, line 36:
```suggestion
 * <li>have exactly the same type and name (which can be the parameter name or a
 *     name assigned by the parameter's {@link By} annotation) as a persistent field
 *     or property of the entity class, or</li>
```

api/src/main/java/jakarta/data/repository/Delete.java
```diff
@@ -56,11 +59,34 @@
...
- * <p>Alternatively, the {@code Delete} annotation may be applied to a repository method with no parameters, indicating
- * that the annotated method deletes all instances of the primary entity type. In this case, the annotated method must
- * either be declared {@code void}, or return {@code int} or {@code long}.
+ *
+ * <p>Alternatively, the {@code Delete} annotation may be used to annotate a repository method with no parameter of
+ * entity type. Then the repository method is interpreted as a parameter-based automatic query method. The entity type
+ * to be deleted is the primary entity type of the repository. The method return type must be {@code void}, {@code int},
+ * or {@code long}. Every parameter of the annotated method must have exactly the same type and name as a persistent
+ * field or property of the entity class. Parameters of type {@code Sort}, {@code Order}, {@code Limit}, and
+ * {@code PageRequest} are prohibited.
```
njr-11, line 68:
```suggestion
 * or {@code long}. Every parameter of the annotated method must have exactly the same type and name
 * (which can be the parameter name or a name assigned by the parameter's {@link By} annotation) as a
 * persistent field or property of the entity class. Parameters of type {@code Sort}, {@code Order},
 * {@code Limit}, and {@code PageRequest} are prohibited.
```

njr-11 (approved):
This looks great. I added one suggestion for clarifying that `By` can be used for Find and Delete so that it isn't interpreted as only the parameter name.

gavinking:
> This looks great. I added one suggestion for clarifying that `By` can be used for Find and Delete so that it isn't interpreted as only the parameter name.

Yeah I agree; I was _this_ close to writing that myself. But lemme do those changes locally because suggestions result in uneven-length lines in the javadoc.

njr-11 (approved):
Looks good. Thanks!

PR #551, opened 2024-03-17
Update the Javadoc of @Query after merge of JDQL

gavinking:
This was now very out of date, and I forgot about it in #520.

Also caught some more instances of 'run time' -> 'runtime', 'life cycle' -> 'lifecycle'.

### DIFF
--- a/api/src/main/java/jakarta/data/page/CursoredPage.java
+++ b/api/src/main/java/jakarta/data/page/CursoredPage.java
@@ -120,7 +120,7 @@ import java.util.NoSuchElementException;
  * <p>For example:</p>
  *
  * <pre>
- * &#64;Query("SELECT o FROM Customer o WHERE (o.ordersPlaced &gt;= ?1 OR o.totalSpent &gt;= ?2)")
+ * &#64;Query("WHERE (ordersPlaced &gt;= ?1 OR totalSpent &gt;= ?2)")
  * &#64;OrderBy("zipcode")
  * &#64;OrderBy("birthYear")
  * &#64;OrderBy("id")
--- a/api/src/main/java/jakarta/data/page/CursoredPage.java
+++ b/api/src/main/java/jakarta/data/page/CursoredPage.java
@@ -120,8 +120,8 @@ import java.util.NoSuchElementException;
  * {@code @OrderBy("zipcode")}
  * {@code @OrderBy("birthYear")}
  * {@code @OrderBy("id")}
- * {@code CursoredPage<Customer> getTopBuyers(int minOrders, float minSpent,
- *                                     {@code PageRequest<Customer> pageRequest);}
+ * {@code CursoredPage<Customer>} getTopBuyers(int minOrders, float minSpent,
+ *                                     {@code PageRequest<Customer>} pageRequest);
  * </pre>
  *
  * <p>Only queries which return entities may be used with cursor-based pagination
--- a/api/src/main/java/jakarta/data/page/CursoredPage.java
+++ b/api/src/main/java/jakarta/data/page/CursoredPage.java
@@ -102,6 +102,28 @@ import java.util.NoSuchElementException;
  * existing entities are modified or if an entity is re-added with different
  * sort criteria after having previously been removed.</p>
  *
+ * <h2>Cursor-based Pagination with {@code @Query}</h2>
+ *
+ * <p>Cursor-based pagination involves generating and appending additional
+ * restrictions involving the key fields to the {@code WHERE} clause of the
+ * query. For this to be possible, a user-provided JDQL or JPQL query must
+ * end with a {@code WHERE} clause to which additional conditions may be
+ * appended.</p>
+ *
+ * <p>Sorting criteria must be specified independently of the user-provided
+ * query, either via the {@link OrderBy} annotation or, or by passing
+ * {@link Sort} criteria within the {@linkplain PageRequest#sorts() page
+ * request}. For example:</p>
+ *
+ * <pre>
+ * {@code @Query("WHERE ordersPlaced >= ?1 OR totalSpent >= ?2")}
+ * {@code @OrderBy("zipcode")}
+ * {@code @OrderBy("birthYear")}
+ * {@code @OrderBy("id")}
+ * {@code CursoredPage<Customer> getTopBuyers(int minOrders, float minSpent,
+ *                                     {@code PageRequest<Customer> pageRequest);}
+ * </pre>
+ *
  * <p>Only queries which return entities may be used with cursor-based pagination
  * because cursors are created from the entity attribute values that
  * form the unique key.</p>
--- a/api/src/main/java/jakarta/data/page/CursoredPage.java
+++ b/api/src/main/java/jakarta/data/page/CursoredPage.java
@@ -102,32 +102,6 @@ import java.util.NoSuchElementException;
  * existing entities are modified or if an entity is re-added with different
  * sort criteria after having previously been removed.</p>
  *
- * <h2>Cursor-based Pagination with {@code @Query}</h2>
- *
- * <p>Cursor-based pagination involves generating and appending additional
- * restrictions involving the key fields to the {@code WHERE} clause of the
- * query. For this to be possible, a user-provided JDQL or JPQL query must
- * end with a {@code WHERE} clause to which additional conditions may be
- * appended without otherwise changing the semantics of the query:</p>
- * <ul>
- * <li>The entire conditional expression of the {@code WHERE} clause must
- *     be enclosed in parentheses.
- * <li>Sorting criteria must be specified independently of the user-provided
- *     query, either via the {@link OrderBy} annotation or, or by passing
- *     {@link Sort} criteria within the {@linkplain PageRequest#sorts() page
- *     request}.
- * </ul>
- * <p>For example:</p>
- *
- * <pre>
- * &#64;Query("WHERE (ordersPlaced &gt;= ?1 OR totalSpent &gt;= ?2)")
- * &#64;OrderBy("zipcode")
- * &#64;OrderBy("birthYear")
- * &#64;OrderBy("id")
- * CursoredPage&lt;Customer&gt; getTopBuyers(int minOrders, float minSpent,
- *                                         {@code PageRequest<Customer>} pageRequest);
- * </pre>
- *
  * <p>Only queries which return entities may be used with cursor-based pagination
  * because cursors are created from the entity attribute values that
  * form the unique key.</p>
--- a/api/src/main/java/jakarta/data/repository/Query.java
+++ b/api/src/main/java/jakarta/data/repository/Query.java
@@ -20,6 +20,7 @@ package jakarta.data.repository;
 import jakarta.data.Sort;
 import jakarta.data.page.CursoredPage;
 import jakarta.data.page.Page;
+import jakarta.data.page.PageRequest;
 
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
@@ -27,29 +28,49 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * <p>Annotates a repository method to specify a query string such as SQL, JPQL, Cypher etc. to execute.</p>
+ * <p>Annotates a repository method as a query method, specifying a query written in Jakarta Data Query Language (JDQL)
+ * or in Jakarta Persistence Query Language (JPQL). A Jakarta Data provider is not required to support the complete JPQL
+ * language, which targets relational data stores.</p>
  *
- * <p>Jakarta Data providers for relational databases must support
- * JPQL queries if backed by a Jakarta Persistence provider,
- * and otherwise must support SQL queries.</p>
+ * <p>The required {@link #value} member specifies the JDQL or JPQL query as a string. When {@linkplain Page pagination}
+ * is used, the {@link #count} member optionally specifies a query which returns the total number of elements satisfying
+ * the query.</p>
  *
- * <h2>Parameters</h2>
+ * <p>For {@code select} statements, the return type of the query method must be consistent with the type returned by
+ * the query. For {@code update} or {@code delete} statements, it must be {@code void}, {@code int} or {@code long}.</p>
  *
- * <p>The query language can used named parameters or positional parameters.</p>
+ * <p>Compared to SQL, JDQL allows an abbreviated syntax for {@code select} statements:</p>
+ * <ul>
+ * <li>The {@code from} clause is optional in JDQL. When it is missing, the queried entity is determined by the return
+ *     type of the repository method, or, if the return type is not an entity type, by the primary entity type of the
+ *     repository.</li>
+ * <li>The {@code select} clause is optional in both JDQL and JPQL. When it is missing, the query returns the queried
+ *     entity.</li>
+ * </ul>
  *
- * <p><b>Named parameters</b> are referred to by name within the query language.
- * The {@link Param} annotation annotates method parameters to bind them to a named parameter name.
- * The {@code Param} annotation is unnecessary for named parameters when the method parameter name
- * matches the query language named parameter name and the application is compiled with the
- * {@code -parameters} compiler option that makes parameter names available
- * at runtime. When the {@code Param} annotation is not used, the Jakarta Data provider must
- * interpret the query by scanning for the delimiter that is used for positional parameters.
- * If the delimiter appears for another purpose in a query that requires named parameters,
- * it might be necessary for the application to explicitly define the {@code Param} in order to
- * disambiguate.</p>
+ * <p>A query might involve:</p>
+ * <ul>
+ * <li>named parameters of form {@code :name} where the labels {@code name} are legal Java identifiers, or </li>
+ * <li>ordinal parameters of form {@code ?n} where the labels {@code n} are sequential positive integers starting
+ *     from {@code 1}.</li>
+ * </ul>
+ * <p>A given query may not mix named and ordinal parameters.</p>
  *
- * <p><b>Positional parameters</b> are referred to by a number that corresponds to the
- * numerical position, starting with 1, of the repository method parameters.</p>
+ * <p>Each parameter of an annotated query method must either:</p>
+ * <ul>
+ * <li>have exactly the same name (the parameter name in the Java source, or a name assigned by {@link Param @Param})
+ *     and type as a named parameter of the query,</li>
+ * <li>have exactly the same type and position within the parameter list of the method as a positional parameter of the
+ *     query, or</li>
+ * <li>be of type {@code Limit}, {@code Order}, {@code PageRequest}, or {@code Sort}.</li>
+ * </ul>
+ *
+ * <p>The {@link Param} annotation associates a method parameter with a named parameter. The {@code Param} annotation is
+ * unnecessary when the method parameter name matches the name of a named parameter and the application is compiled with
+ * the {@code -parameters} compiler option making parameter names available at runtime.</p>
+ *
+ * <p>A method parameter is associated with an ordinal parameter by its position in the method parameter list. The first
+ * parameter of the method is associated with the ordinal parameter {@code ?1}.</p>
  *
  * <p>For example,</p>
  *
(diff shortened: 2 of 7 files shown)


api/src/main/java/module-info.java
```diff
@@ -149,7 +149,7 @@
  *     &#64;OrderBy("address.zipCode")
  *     List&lt;Purchase&gt; findByAddressZipCodeIn(List&lt;Integer&gt; zipCodes);
  *
- *     &#64;Query("SELECT o FROM Purchase o WHERE o.address.zipCode=?1")
+ *     &#64;Query("WHERE address.zipCode = ?1")
```
otaviojava, line 152: IMHO: this syntax only using "where" is my favorite part of the language.

api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -20,49 +20,78 @@
...
+ *     // JDQL with positional parameters
+ *     {@code @Query("where firstName = ?1 and lastName = ?2")}
+ *     {@code List<Person>} byName(String first, String last);
+ *
+ *     // JDQL with a named parameter
+ *     {@code @Query("where firstName || ' ' || lastName like :pattern")}
+ *     {@code List<Person>} byName(String pattern);
+ *
+ *     // JPQL using a positional parameter
+ *     {@code @Query("from Person where extract(year from birthdate = ?1)")}
```
njr-11, line 90:
```suggestion
 *     {@code @Query("from Person where (extract(year from birthdate) = ?1)")}
```

api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -94,30 +135,30 @@
...
-     * and supplies other forms of sorting (such as {@link Sort}) to that method,
-     * then it is the responsibility of the application to compose the query in
-     * such a way that an {@code ORDER BY} clause (or query language equivalent)
-     * can be validly appended. The Jakarta Data provider is not expected to
-     * parse query language that is provided by the application.</p>
+     * <p>If the annotated repository method accepts other forms of sorting
+     * (such as a parameter of type {@link Sort}), it is the responsibility
+     * of the application programmer to compose the query so that an
+     * {@code ORDER BY} clause can be validly appended to the text of the
+     * query.</p>
```
njr-11, line 145: This part was lost in the rewrite:
```suggestion
     * query. The Jakarta Data provider is not expected to parse query language
     * that is provided by the application.</p>
```

api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -94,30 +135,30 @@
...
+     * methods returning {@link Page} or {@link CursoredPage}.</p>
      *
-     * <p>The default value of empty string indicates that no counting query
-     * is provided. A counting query is unnecessary when pagination is
-     * performed with {@link jakarta.data.page.PageRequest#withoutTotal()} or
-     * when pagination is not used at all.</p>
+     * <p>The additional query is optional. It is not used when pagination
+     * is performed with {@link PageRequest#withoutTotal()} or when the
+     * repository method returns a type other than {@link Page} or
+     * {@link CursoredPage}.</p>
```
njr-11, line 161: Clarify when it is required,
```suggestion
     * <p>The additional query is optional in most cases. It is required when
     * the return type is {@link Page} or {@link CursoredPage} and
     * pagination is performed with {@link PageRequest#requestTotal()}
     * of {@code true} and the primary query includes a `SELECT` or
     * `ORDER BY` clause.</p>
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -1483,7 +1483,7 @@ Here is an example where an application uses `@Query` to provide a partial query
 ----
 @Repository
 public interface CustomerRepository extends BasicRepository<Customer, Long> {
-  @Query("SELECT o FROM Customer o WHERE (o.totalSpent / o.totalPurchases > ?1)")
+  @Query("WHERE (o.totalSpent / o.totalPurchases > ?1)")
```
njr-11, line 1486:
```suggestion
  @Query("WHERE (totalSpent / totalPurchases > ?1)")
```

api/src/main/java/jakarta/data/page/CursoredPage.java
```diff
@@ -120,7 +120,7 @@
  * <p>For example:</p>
  *
  * <pre>
- * &#64;Query("SELECT o FROM Customer o WHERE (o.ordersPlaced &gt;= ?1 OR o.totalSpent &gt;= ?2)")
+ * &#64;Query("WHERE ordersPlaced &gt;= ?1 OR totalSpent &gt;= ?2")
```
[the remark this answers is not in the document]
gavinking: Oh, yes, it is still there:

> Cursor-based pagination involves generating and appending additional restrictions involving the key fields to the WHERE clause of the query. For this to be possible, a user-provided JDQL or JPQL query must end with a WHERE clause to which additional conditions may be appended without otherwise changing the semantics of the query:

I thought it was gone.

[And, FTR, I think it *should* go. We don't have subqueries in JDQL, so the task of inserting parens in between `where` and `order by` is just completely trivial.]

api/src/main/java/jakarta/data/repository/Query.java
gavinking: I intentionally removed it. Because a Jakarta Data provider _does_ parse JDQL. Now, perhaps it passes it on to a JPQL implementation to parse, but in that case the JPQL implementation is part of the Jakarta Data provider.

api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -20,49 +20,78 @@
...
+ *     entity.</li>
+ * </ul>
+ *
+ * <p>A query might involve:</p>
+ * <ul>
+ * <li>named parameters of form {@code :name} where the labels {@code name} are legal Java identifiers, or </li>
+ * <li>ordinal parameters of form {@code ?n} where the labels {@code n} are sequential positive integers starting
+ *     from {@code 1}.</li>
+ * </ul>
+ * <p>A given query may not mix named and ordinal parameters.</p>
```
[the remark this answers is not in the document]
gavinking: This means the same thing? I mean, I don't care but....

api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -20,49 +20,78 @@
...
+ * <p>For {@code select} statements, the return type of the query method must be consistent with the type returned by
+ * the query. For {@code update} or {@code delete} statements, it must be {@code void}, {@code int} or {@code long}.</p>
+ *
+ * <p>Compared to SQL, JDQL allows an abbreviated syntax for {@code select} statements:</p>
+ * <ul>
+ * <li>The {@code from} clause is optional in JDQL. When it is missing, the queried entity is determined by the return
+ *     type of the repository method, or, if the return type is not an entity type, by the primary entity type of the
+ *     repository.</li>
+ * <li>The {@code select} clause is optional in both JDQL and JPQL. When it is missing, the query returns the queried
+ *     entity.</li>
```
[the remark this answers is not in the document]
gavinking: Yes, as I told you guys, my pull request for this was already merged. The current spec states that:

>> The SELECT clause is optional. A query with a missing SELECT clause is interpreted as if it had the following single-item SELECT clause: `select this`, where this is the implicit identification variable.

spec/src/main/asciidoc/repository.asciidoc
gavinking: I mean this is just really hostile to the user. Why would you want to make the user do this every time?

api/src/main/java/jakarta/data/repository/Query.java
gavinking: I would remove this requirement too. It's completely trivial for the Jakarta Data provider to replace everything before `from` with `select count(this)`.

gavinking:
@njr-11 These restrictions you're worried about made perfect sense when `@Query` was a generic annotation that could accommodate an open-ended list of languages: SQL, JPQL, etc.

But now the language is JDQL, which is syntactically trivial. Those restrictions no longer make any sense.

api/src/main/java/jakarta/data/repository/Query.java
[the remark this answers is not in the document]
njr-11: > This means the same thing? I mean, I don't care but....

Unfortunately there are a lot of people who read/use the word "may" as meaning "might" in common language so we are told to generally avoid it in specifications and use "must" to avoid confusion.

If someone does read it as "might" in this case, they will consider the statement to be informational rather than a requirement.

api/src/main/java/jakarta/data/repository/Query.java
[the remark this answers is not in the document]
njr-11: Thanks!  I must have missed the detail about SELECT becoming optional.  I'll resolve the conversation now.

spec/src/main/asciidoc/repository.asciidoc
njr-11: > I mean this is just really hostile to the user. Why would you want to make the user do this every time?

Did this comment end up in the wrong place?  It is attached to my correction of query language where the indentification variable was incompletely removed.

gavinking: Oh I misunderstood. I thought it was also about parentheses.

JDQL does not have identification variables, and they're also now optional in JPQL.

gavinking: Fixed.

njr-11: > JDQL does not have identification variables, and they're also now optional in JPQL.

The problem is that the change halfway removed them from a particular query.  The query used to look like this:

```
  @Query("SELECT o FROM Customer o WHERE (o.totalSpent / o.totalPurchases > ?1)")
```

You removed the part that defines the variable `o`, but not its usage:
```
@Query("WHERE (o.totalSpent / o.totalPurchases > ?1)")
```

So my proposed change is that we need to also remove `o.` from the rest of the query.

gavinking:
> But now the language is JDQL, which is syntactically trivial. Those restrictions no longer make any sense.

In 6041a62 and 0c81a33 I hereby propose simply remove them. :)

njr-11:
> > But now the language is JDQL, which is syntactically trivial. Those restrictions no longer make any sense.
> 
> In [6041a62](https://github.com/jakartaee/data/commit/6041a623ee82250d023a1f7444d3d61665842328) and [0c81a33](https://github.com/jakartaee/data/commit/0c81a3371d890f16aff9922a0e392751060ea3bf) I hereby propose simply remove them. :)

What we have written is that the language can be JDQL or JPQL.  If a Jakarta Persistence-backed Jakarta Data provider is going to handle both, then the additional restrictions do matter.

api/src/main/java/jakarta/data/repository/Query.java
gavinking: In fact, the separate `count()` query is just no longer needed. That's one advantage of having our own query language!

api/src/main/java/jakarta/data/repository/Query.java
gavinking: Fixed.

gavinking:
> If a Jakarta Persistence-backed Jakarta Data provider is going to handle both, then the additional restrictions do matter.

Sure, but I really don't think it's placing an undue burden on a Jakarta Data provider _which is backed by Jakarta Persistence_ to be able to understand JPQL _if and only if that Jakarta Data provider decides to also support arbitrary JPQL and pagination_.

The spec *does not* place any requirement on such a provider to be able to support pagination with arbitrary JPQL, so if the provider thinks that being able to parse JPQL is a burden, then it is between the provider and the user to define how such extensions work.

All the spec requires is that the provider be able to handle JDQL.

gavinking:
To be clear, what we say about this is:

> A Jakarta Data provider is not required to support the complete JPQL language, which targets relational data stores.

and in the spec:

> A Jakarta Data provider backed by access to a relational database might choose to allow the use of a much larger subset of JPQL&mdash;or even the whole language&mdash;via the `@Query` annotation. Such extensions are not required by this specification.


So we do not need to explicitly place limitations on _JPQL_ support.

gavinking:
@njr-11 Please see fa379ad, which makes this explicit.

gavinking:
Force pushed to clean up commit history.

njr-11 (changes requested):
> In fact, the separate `count()` query is just no longer needed. That's one advantage of having our own query language!

I can certainly see the appeal of being able to get rid of the `count` query now that one can be assembled by replacing the `SELECT` clause and removing the `ORDER BY` clause from the primary query if using JDQL.  I need to give this some more thought about where we end up with JPQL.  It is certain that many users will opt for JPQL given that they are able to accomplish more with it, and we will want to be sure they aren't broken or consider themselves broken because there is no standard assuring them that pagination will still work.

gavinking:
> I need to give this some more thought about where we end up with JPQL.

FTR, the thing you would need to do with JPQL is replace everything between `select` and `from` with `select count(1) from`. This is easy in pure JPQL, since JPQL does not allow subqueries in the `select` clause.

Now, on the other hand, some providers, [i.e. Hibernate, but not, I believe, EclipseLink], _do_ allow subqueries in `select`, and that makes the problem a bit more difficult. But that's not part of the spec and for multiple reasons not something you would need to implement support for.

gavinking:
Note also, that Hibernate has a method that does this for you: `SelectionQuery.getResultCount()`. 

EclipseLink might also have something like that.

gavinking:
> FTR, the thing you would need to do with JPQL is replace everything between `select` and `from` with `select count(1) from`.

Oh yes, and remove `order by`, as you say. Also easy, since subqueries don't have `order by` clauses.

api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -71,18 +97,34 @@
...
+ * {@link Page#nextPageRequest(Class)}, explicitly specifying the entity class, to obtain the next page of results. For
+ * example,</p>
  *
  * <pre>
  * {@code Page<String>} page2 = people.namesOfLength(5, 10, page1.nextPageRequest(Person.class));
  * </pre>
  *
+ * <p>A Jakarta Data provider is permitted to support features of JPQL which go beyond the subset required by JDQL, even
+ * when the provider does not provide a complete implementation of JPQL. A provider which does support such extensions
+ * to JDQL might place restrictions on whether or how they may be used together with cursor-based pagination.</p>
```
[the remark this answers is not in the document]
njr-11: There are two different scenarios in which query manipulation is needed: appending conditions to a WHERE clause  and computing a count query. The former is for cursor-based pagination only. The latter is for both types of pagination. When I eventually find time to review this, we will need to ensure that it is possible to achieve both without restrictions or that we have stated whatever restrictions might remain.

api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -18,51 +18,77 @@
...
- * {@code -parameters} compiler option that makes parameter names available
- * at runtime. When the {@code Param} annotation is not used, the Jakarta Data provider must
- * interpret the query by scanning for the delimiter that is used for positional parameters.
- * If the delimiter appears for another purpose in a query that requires named parameters,
- * it might be necessary for the application to explicitly define the {@code Param} in order to
- * disambiguate.</p>
- *
- * <p><b>Positional parameters</b> are referred to by a number that corresponds to the
- * numerical position, starting with 1, of the repository method parameters.</p>
+ * <p>Annotates a repository method as a query method, specifying a query written in Jakarta Data Query Language (JDQL)
```
otaviojava, line 30: That might lead to a complex discussion, but using the self-continued principle and using JPQL as a sample, we are once again able to cross-database specification.

```
/**
 * <p>Annotates a repository method as a query method, specifying a query written in Jakarta Data Query Language (JDQL)
 * as the minimum requirement. This encourages the use of JDQL for cross-database compatibility. Additionally, a Jakarta Data
 * provider has the freedom to support extensions to the language, including features from the Jakarta Persistence Query Language (JPQL).
 * While providers are not required to support the complete JPQL language, which primarily targets relational data stores, they are
 * encouraged to provide as broad a support as possible to enhance the flexibility and power of data access queries.</p>
 */
```

gavinking: There's nothing wrong with what you just wrote, just not sure it's quite the right place to be saying things like that.

gavinking: OK I think I understand what you're looking for. Just not quite sure how to say it in a way which isn't out of place in the doc for that annotation. (It's the kind of thing that fits more comfortably in the spec or in `module-info`.

api/src/main/java/jakarta/data/page/CursoredPage.java
```diff
@@ -102,32 +102,6 @@
  * existing entities are modified or if an entity is re-added with different
  * sort criteria after having previously been removed.</p>
  *
```
njr-11, line 104: I am willing to lift the restriction requiring the WHERE clause to be enclosed in parenthesis if we keep the other restriction disallowing the query to end with other clauses like GROUP BY, HAVING, or ORDER BY.
Here is the old text of the section updated according to that:

```suggestion
 *
 * <h2>Cursor-based Pagination with {@code @Query}</h2>
 *
 * <p>Cursor-based pagination involves generating and appending additional
 * restrictions involving the key fields to the {@code WHERE} clause of the
 * query. For this to be possible, a user-provided JDQL or JPQL query must
 * end with a {@code WHERE} clause to which additional conditions may be
 * appended.</p>
 *
 * <p>Sorting criteria must be specified independently of the user-provided
 * query, either via the {@link OrderBy} annotation or, or by passing
 * {@link Sort} criteria within the {@linkplain PageRequest#sorts() page
 * request}. For example:</p>
 *
 * <pre>
 * {@code @Query("WHERE (ordersPlaced >= ?1 OR totalSpent >= ?2)")}
 * {@code @OrderBy("zipcode")}
 * {@code @OrderBy("birthYear")}
 * {@code @OrderBy("id")}
 * {@code CursoredPage<Customer> getTopBuyers(int minOrders, float minSpent,
 *                                     {@code PageRequest<Customer> pageRequest);}
 * </pre>
 *
```

njr-11 (changes requested):
I think it will be acceptable to remove the counting query and have it be generated from the primary query and also to lift the restriction to have parenthesis around WHERE.  The other restriction should not be removed.

api/src/main/java/jakarta/data/page/CursoredPage.java
gavinking: OK, that's fine by me.

gavinking: >    {@code @Query("WHERE (ordersPlaced >= ?1 OR totalSpent >= ?2)")}

But we can get rid of these parens, right?

njr-11: Yes, those parenthesis are optional now. I'll take them out of the example.

api/src/main/java/jakarta/data/repository/Query.java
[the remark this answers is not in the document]
gavinking: @njr-11 do you want this language or would you prefer to drop it?

[I don't care either way.]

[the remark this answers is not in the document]
njr-11: I don't think we need this. We already added the restriction I was looking for to ensure cursor-based pagination works well.

api/src/main/java/jakarta/data/repository/Query.java
gavinking: @otaviojava please see 628abf4. Is that sorta approximately like what you would like to see?

api/src/main/java/jakarta/data/repository/Query.java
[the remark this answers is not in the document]
gavinking: OK I guess I will drop it then.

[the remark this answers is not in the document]
gavinking: Done.

gavinking:
@njr-11 @otaviojava are there any remaining problems with this or is it OK to merge?

njr-11 (approved):
> @njr-11 @otaviojava are there any remaining problems with this or is it OK to merge?

I was only waiting to review the outcome of the conversation between you and Otavio, but the pull looks find as is so I can approve it.  If more commits do get added, I'll try to remember to watch out for that and review them.

gavinking:
Awesome!!

So according to me, JDQL is done and ~crossed off the list~.

PR #552, opened 2024-03-17
Keep record style at the CursoredPage

otaviojava:


gavinking (approved):

api/src/main/java/jakarta/data/page/CursoredPage.java
```diff
@@ -154,7 +154,7 @@ public interface CursoredPage<T> extends Page<T> {
      * @param index position (0 is first) of a result on the page.
      * @return cursor for key values at the specified position.
      */
-    PageRequest.Cursor getCursor(int index);
+    PageRequest.Cursor cursor(int index);
```
njr-11, line 157: This is not a record style accessor. It accepts a parameter value.

njr-11 (commented):
I noticed this is already merged, but please see my review comment,

api/src/main/java/jakarta/data/page/CursoredPage.java
otaviojava: That is a good point, but it still accesses info.

IMHO: It should keep without "get" as the other ones.

PR #554, opened 2024-03-18
change Iterable<E> to List<E> in lifecycle method signatures

gavinking:
This is way more friendly to both users and implementors.

`Iterable` is an absolute PITA to work with, both for users, when a method returns `Iterable`, and for implementors, when a method accepts it. And there are no interesting types which implement `Iterable` but not `Collection`.

Now, I considered making the substitution `Iterable` -> `Collection`, but really all that allows is `Set`s. And here, in lifecycle methods, `Set` actually has the wrong semantics, because the returned values are correlated with input values by their order.

So `List` is really the right thing to use here!

### DIFF
--- a/api/src/main/java/jakarta/data/repository/BasicRepository.java
+++ b/api/src/main/java/jakarta/data/repository/BasicRepository.java
@@ -21,6 +21,7 @@ import jakarta.data.exceptions.OptimisticLockingFailureException;
 import jakarta.data.page.Page;
 import jakarta.data.page.PageRequest;
 
+import java.util.List;
 import java.util.Optional;
 import java.util.stream.Stream;
 
@@ -131,7 +132,7 @@ public interface BasicRepository<T, K> extends DataRepository<T, K> {
      * @throws NullPointerException If either the iterable is null or any element is null.
      */
     @Save
-    <S extends T> Iterable<S> saveAll(Iterable<S> entities);
+    <S extends T> List<S> saveAll(List<S> entities);
 
     /**
      * Retrieves an entity by its Id.
@@ -232,6 +233,6 @@ public interface BasicRepository<T, K> extends DataRepository<T, K> {
      * @throws NullPointerException If the iterable is {@code null} or contains {@code null} elements.
      */
     @Delete
-    void deleteAll(Iterable<? extends T> entities);
+    void deleteAll(List<? extends T> entities);
 
 }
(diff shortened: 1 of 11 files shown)


njr-11 (approved):
`List` will be much more convenient for users than `Iterable` for accessing and using the results.  In addition to the changes you made, there is also an occurrence of `Iterable` that needs updating to `List` in module-info.

```
 * <li>the class of the entity, or</li>
 * <li>{@code Iterable<E>} or {@code E[]} where {@code E} is the class of the
 *     entities.</li>
 * </ul>
```

gavinking:
>  there is also an occurrence of Iterable that needs updating to List in module-info.

Thanks, fixed!

gavinking:
@otaviojava any objections to this?

otaviojava:
> @otaviojava any objections to this?

Nope, merged.

api/src/main/java/jakarta/data/repository/BasicRepository.java
```diff
@@ -131,7 +132,7 @@ public interface BasicRepository<T, K> extends DataRepository<T, K> {
      * @throws NullPointerException If either the iterable is null or any element is null.
      */
     @Save
-    <S extends T> Iterable<S> saveAll(Iterable<S> entities);
+    <S extends T> List<S> saveAll(List<S> entities);
```
otaviojava, line 135: This might be late, but could we keep the parameter as Iterable?
Thus, it can work with any input that we wish.

gavinking: Well, I mean, we _could_, but why? What's the concrete use case? 

The only use case I can see is that you want to pass a `Set` here. But `Set` just has the wrong semantics, because it's in general _not_ ordered, and so its elements can't be correlated with the elements in the resulting `List`.

And apart from `Set`, I just don't know any other type of thing that's `Iterable` that you might have a reasonable expectation of wanting to pass in here.

gavinking: And for another thing, I just want simple rules here: "use arrays or `List`s."

Not: "you can use `Iterable` here in this place but it has to be a `List` here in this other place".

Nor do I want to leave the user with this vague feeling of uncertainty about whether they should have used `List` or `Collection` or `Iterable`. In this case, I think it's better to just decide for them.

otaviojava: Yes, it makes sense.
Thus, never mind.

gavinking: cool.

PR #562, opened 2024-03-19
three minor changes

gavinking:
These changes were included in PRs #551 and #553. I believe them to be uncontroversial. And since we're still discussing those other PRs, I have separated these changes out so that they are not blocked.


njr-11 (approved):
Good idea splitting out parts that everyone agrees on to make progress.

Issue #559, opened 2024-03-19
[clarification]: Clarification on Handling Embedded Types in Metamodel Generation


otaviojava:
### Specification

https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/repository.asciidoc#type-safe-access-to-entity-attributes

### I need clarification on ...

Hello,

I've been exploring the metamodel specifications and have encountered a scenario requiring further clarification regarding handling embedded types during generation. Specifically, I am concerned about the disambiguation mechanism (if any) for embedded types used across different entities.

To illustrate, consider the following Java classes:

```java
@Entity
public class Person {
    @Id
    private String id;
    @Column
    private String name;
    @Column
    private Address address; // Embedded Address
}

@Entity
public class Order {
    @Id
    private UUID id;
    @Column
    private String name;
    @Column
    private Address delivery; // Another instance of Embedded Address
}

@Embedded
public class Address {
    private String city;
}
```

In the example above, both `Person` and `Order` entities contain an embedded `Address` type. My question revolves around how the metamodel generation process distinguishes between these two instances of the `Address` class. Is there a specific mechanism to handle such cases of embedded types to avoid ambiguity? What best practices should be followed to ensure that the metamodel remains clear and consistent when dealing with embedded types across multiple entities?

gavinking:
Not sure I see where ambiguity could arise. What are you thinking of here?

otaviojava:
Given the same entities of the previous sample:

How do you generate this metadata class?

Mainly because:

```java
var name = _Person.address.city.name():// it must generate "address.city"
var name2 = _Order.delivery.city.name();// it must generate "delivery.city"
```

I mean, if it were a constant, it wouldn't get the proper path:

```java
var name = _Person.Address.city.name():// it must generate "address.city"
var name2 = _Order.Address.city.name();// it must generate "delivery.city"
```

My point is that our spec covers only the basic attributes, but what happens if it is not a basic attribute?

gavinking:
This is also covered in the spec by this bullet point:

> • The name of the field, ignoring case, must match the name of an entity attribute, with the _ character in the field name delimiting the attribute names of hierarchical structures or relationships, such as embedded classes.

gavinking:
[Note that it's a little bit hard to find that, since it's in section 3.1.9.1, but the natural place to look for it is section 3.1.9.4. But it's there.]

gavinking:
@otaviojava yes, I believe so, best check with @njr-11.

[...]

PR #565, opened 2024-03-20
Small fix at genetic staticmetamodel

otaviojava:
Ref: https://github.com/jakartaee/data/issues/559


njr-11 (approved):
As a bug fix to a line of a code example that is contradictory to the rest of the specification, this should be an acceptable change to make during the extension period even though it is not related to the Query Language feature.  It does not alter the API or requirements of the specification in any way and has no impact on the TCK or on implementations.

PR #575, opened 2024-03-21
clarify that user-defined generic repository supertypes are not supported

gavinking:
At least not in Jakarta Data 1.0.


njr-11 (approved):

gavinking:
Well, OK, so I found the method `Types.asMemberOf()`, which let me implement support for user-written generic repository supertypes, I believe robustly, using the compiler's own logic for type argument substitution. So for any implementation based on annotation processors, the requirement which is relaxed by this pull request is actually not anywhere near as much of a difficultly as I feared.

However, we don't require the use of annotation processing by implementors of Jakarta Data, and so I guess there's still a concern for implementations _not_ based on annotation processing.

Anyway, I'll let ya'll decide what to do with this. I guess I'm OK with making or not making the change.

otaviojava (approved):

Issue #576, opened 2024-03-21
[clarification]: query by method name method parameters


gavinking:
### Specification

Query by Method Name

### I need clarification on ...

The relationship between Query by Method Name operators and the parameters of the repository method.

### Additional information

If we're going to keep this as a required part of the spec [now that we have JDQL, `@Query`,  and `@Find` properly nailed down, I'm ready to have a probably-difficult conversation about that], then we at least need to make sure that this facility is fully-specified.

Reviewing this, it does seem to be mostly sorta OK, but:

- the allowed return types are defined only in `module-info`, in a very sketchy format, and not in the spec itself, and
- the allowed parameter types, and their relationship to the query itself are, AFICS, completely unspecified. This is less-than-trivial when we have operators like `In` and `Between`.

So there is at least some work to do here.

Also I'm also not sure I understand how to interpret shorthands like `OrderBy____(Asc|Desc)*(Asc |Desc)`. Actually that might just be an error.

njr-11:
> Also I'm also not sure I understand how to interpret shorthands like `OrderBy____(Asc|Desc)*(Asc |Desc)`. Actually that might just be an error

I remember needing to ask what that meant as well when I wanted to write valid examples of it to put in module-info.  If it doesn't make sense to the authors of the spec, it won't make sense to users either and we should fix it.

I included at least part of an attempt at that in #577

njr-11:
> * the allowed return types are defined only in `module-info`, in a very sketchy format, and not in the spec itself, and

The JavaDoc in module-info is intended as a quick reference for application developers and often relies on tables and examples that you can copy from to try to visually get the point across.  If the information is missing from the specification document, we should make sure to add it there.

njr-11:
> * the allowed parameter types, and their relationship to the query itself are, AFICS, completely unspecified. This is less-than-trivial when we have operators like `In` and `Between`.

The table of Query by Method Name keywords reserved for the predicate in module-info was supposed to get across some of that information in a general sense.  Here is the information that is currently there for `Between`,

Keyword: `Between`

Applies to: numeric, strings, time

Description: Requires that the entity's attribute value be within the range specified by two parameters, inclusive of the parameters. The minimum is listed first, then the maximum.

Example: `findByAgeBetween(minAge, maxAge)`


One deficiency here is that we never define what "numeric", "strings", and "time" actually mean.  We should be more precise and do that.

The description should probably say "...inclusive of the parameter values..."

[...]

PR #577, opened 2024-03-21
fix unclear spec doc for the OrderBy keyword

njr-11:
This aims to at least partially fix unclear specification language identified in #576 around the OrderBy keyword.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -936,7 +936,7 @@ The parsing of query method names follows a specific format:
 - The method name consists of the subject, the predicate, and optionally the order clause.
 - The subject begins with the action (such as `find` or `delete`) and is optionally followed by an expression (for example, `First10`), followed by any number of other characters, followed by `By`.
 - The predicate defines the query's condition or filtering criteria, where multiple conditions are delimited by `And` or `Or`. For example, `PriceLessThanAndNameLike`.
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -934,7 +934,7 @@ The parsing of query method names follows a specific format:
 - The method name consists of the subject, the predicate, and optionally the order clause.
 - The subject begins with the action (such as `find` or `delete`) and is optionally followed by an expression (for example, `First10`), followed by any number of other characters, followed by `By`.
 - The predicate defines the query's condition or filtering criteria, where multiple conditions are delimited by `And` or `Or`. For example, `PriceLessThanAndNameLike`.
-- The order clause, which is optional, begins with `OrderBy` and consists of an ordered collection of entity attributes by which to sort results, delimited by `Asc` or `Desc` to specify the sort direction of the preceding attribute.
+- The order clause, which is optional, begins with `OrderBy` and is followed by one or more ordered pairings consisting of entity attribute name and direction of sorting (`Asc` or `Desc`). For example, `OrderByPriceDescIdAsc`. Ordering of results is performed by sorting the first entity attribute according to its respective direction, where results that sort to the same position are subsequently sorted by the second entity attribute according to its respective direction, and so forth.
```
otaviojava, line 937: But, when we don't put `Asc` it is asc implicitly.

It also happens in SQL, and we share the same behavior in NoSQL:

```SQL
SELECT * FROM Products ORDER BY Price;
```

It would be great if you reproduce it here as well.

njr-11: That's a good point. The BNF does allow an optimization where the direction can be omitted on the final property and we need to define the meaning of that in the spec.  I added commit https://github.com/jakartaee/data/pull/577/commits/487098a35eabc36dfb8631dd88ffb91a17782b1a to address this.

otaviojava (approved):

PR #584, opened 2024-03-23
Revert "chore(deps): Bump org.apache.maven.plugins:maven-gpg-plugin from 3.1.0 to 3.2.1"

mswatosh:
Reverts jakartaee/data#581

This is preventing the M4 release.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -282,7 +282,7 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-gpg-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


KyleAure:
I think we can fix this on the build side.  Let's hold off on reverting until I can test this.

KyleAure:
Fixed the build no need to revert.

mswatosh:
I figured that was probably the case, but didn't want to hold anything up. Thanks for getting it fixed on the build side

otaviojava:
Thanks @KyleAure

PR #585, opened 2024-03-23
split repository.asciidoc into multiple files

gavinking:
This one file `repository.asciidoc` contained almost the whole spec, so I moved some of that material into an unused file that already existed, and some into a new file.

I and split chapter 3 into three sections, instead of hanging everything off section 3.1.


spec/src/main/asciidoc/entity.asciidoc
```diff
@@ -0,0 +1,528 @@
...
+    title varchar,
+    category varchar
+)
+----
+
+NOTE: Support for entity associations is not required by this specification.
+
+=== Entity Names and Persistent Field Names
+
+Entities and their persistent fields may be referenced by name in the query language defined in <<Jakarta Data Query Language>>.
```
gavinking, line 326: Note that this one-sentence preamble to section 3.2 is the only change to the actual words.

otaviojava (approved):

gavinking:
@njr-11 do you have any objection to making this change, before it starts running into conflicts?

njr-11:
> @njr-11 do you have any objection to making this change, before it starts running into conflicts?

I forgot about this one. I remember seeing it and not wanting to spend time comparing the files to ensure nothing was left out, so I set it aside for later.  If you or another reviewer has already verified that, I'm fine with it.

gavinking:
Well, the difference in linecount should be 13+2+2-1 = 16 lines (a copyright header, two includes, the one added line of text, minus a heading. And 669-653 = 16, so that checks out.

mswatosh:
I double checked in a diff tool and confirmed nothing is left out. Just the change from Repository to Architecture, the new text mentioned above, and the heading level changes to split up chapter 3.

mswatosh (approved):

njr-11 (approved):
I copied the moved sections into a separate file and compared them to the original. It looks like everything is still there. Approving.

gavinking:
Gracias, caballeros.

PR #586, opened 2024-03-23
compile with -parameters

gavinking:
We should compile the library with `-parameters`, to avoid seeing things like `arg0` in code generated from the built-in repository supertypes.

I don't use Maven, and so I don't know how to make this work, but we should do something like it.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -70,4 +70,19 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.13.0</version>
+                <configuration>
+                    <compilerArgs>
+                        <arg>-parameters</arg>
+                    </compilerArgs>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
 </project>


gavinking:
@KyleAure @otaviojava Do you guys how to do this so it works?

pom.xml
```diff
@@ -136,6 +136,11 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>${maven.compile.version}</version>
+                <configuration>
+                    <compilerArgs>
+                        <arg>-parameters</arg>
```
KyleAure, line 141: @gavinking  Yes this is how you would provide an argument to the javac compiler using maven.  

I still don't necessarily understand what the `-parameters` argument will achieve in terms of generated code.  As far as I can tell we don't package any generated code within the API so wouldn't this argument only be necessary during code generation for users? 

If we do need to use this argument however I would rather it be just in the `api/pom.xml` as we have no need to use this for the TCK jar.

KyleAure (changes requested):

pom.xml
gavinking: The issue is that I as a user can inherit from one of the repository supertypes, e.g. `CrudRepository`, and then the Jakarta Data provider needs to generate an implementation of its methods. The generated code is prettier if it has the original parameter names like `entity`, `pageRequest`, `entities`,and  `id` instead of `arg0`. 

It's not essential, and of course, and of course there are ways I could special-case this stuff.

But IMO [rant mode on] this compiler option should be enabled by default, and the arguments that these guys use to justify _not_ enabling it are lame, literally just "oh, it might not be backward compatible for old code but I can't give you a single concrete example of how or why or what might break". Of course, the only code that could possibly break is code that depends on method parameters being named `arg0`, `arg1`, etc. And in our case, since everything is new, there's no code like that.

KyleAure: Okay thanks for the explanation :)
Actually I forgot that `tck/pom.xml` already compiles with `-parameters`.
In that case, I don't see any issue.

KyleAure (approved):

pom.xml
gavinking: Yeah, but I tried out that change, and it didn't seem to actually work. So I think something's missing.

otaviojava (approved):

starksm64:
What is missing is that the api jar also needs to be compiled as tck tests extend from jakarta.data.repository.BasicRepository for example and this is where the missing parameter names are coming from.

gavinking:
OK, I figured it out. I need to set this in a new section in `api/pom.xml`.

It does not work to set it in the root pom.

Anyway I've tested it and it appears to be working perfectly now.

@KyleAure Could you please review this?

KyleAure (approved):
LGTM

gavinking:
Great, thanks.

PR #591, opened 2024-03-25
change Iterable -> List in TCK & remove requirement that repo be backed by EntityManager

gavinking:
1. change Iterable -> List in TCK to comply with recent change to the spec
2. remove requirement that repo be backed by `EntityManager`, which was never required

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/Catalog.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/Catalog.java
@@ -54,7 +54,7 @@ public interface Catalog extends DataRepository<Product, String> {
     Product[] modifyMultiple(Product... products);
 
     @Delete
-    boolean remove(Product product);
+    void remove(Product product);
 
     @Delete
     void removeMultiple(Product... products);
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/Catalog.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/Catalog.java
@@ -33,8 +33,6 @@ import jakarta.data.repository.Query;
 import jakarta.data.repository.Repository;
 import jakarta.data.repository.Save;
 import jakarta.data.repository.Update;
-import jakarta.persistence.EntityManager;
-import jakarta.persistence.TypedQuery;
 import ee.jakarta.tck.data.standalone.persistence.Product.Department;
 
 @Repository
@@ -92,22 +90,22 @@ public interface Catalog extends DataRepository<Product, String> {
 
     List<Product> findByProductNumLike(String productNum);
 
-    EntityManager getEntityManager();
-
-    default double sumPrices(Department... departments) {
-        StringBuilder jpql = new StringBuilder("SELECT SUM(o.price) FROM Product o");
-        for (int d = 1; d <= departments.length; d++) {
-            jpql.append(d == 1 ? " WHERE " : " OR ");
-            jpql.append('?').append(d).append(" MEMBER OF o.departments");
-        }
-
-        EntityManager em = getEntityManager();
-        TypedQuery<Double> query = em.createQuery(jpql.toString(), Double.class);
-        for (int d = 1; d <= departments.length; d++) {
-            query.setParameter(d, departments[d - 1]);
-        }
-        return query.getSingleResult();
-    }
+//    EntityManager getEntityManager();
+//
+//    default double sumPrices(Department... departments) {
+//        StringBuilder jpql = new StringBuilder("SELECT SUM(o.price) FROM Product o");
+//        for (int d = 1; d <= departments.length; d++) {
+//            jpql.append(d == 1 ? " WHERE " : " OR ");
+//            jpql.append('?').append(d).append(" MEMBER OF o.departments");
+//        }
+//
+//        EntityManager em = getEntityManager();
+//        TypedQuery<Double> query = em.createQuery(jpql.toString(), Double.class);
+//        for (int d = 1; d <= departments.length; d++) {
+//            query.setParameter(d, departments[d - 1]);
+//        }
+//        return query.getSingleResult();
+//    }
 
     @Query("SELECT o FROM Product o WHERE (:rate * o.price <= :max AND :rate * o.price >= :min) ORDER BY o.name")
     Stream<Product> withTaxBetween(@Param("min") double mininunTaxAmount,
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/PersistenceEntityTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/PersistenceEntityTests.java
@@ -110,8 +110,8 @@ public class PersistenceEntityTests {
         catalog.save(Product.of("socks", 5.99, "TEST-PROD-87", Department.CLOTHING));
         catalog.save(Product.of("volleyball", 10.99, "TEST-PROD-89", Department.SPORTING_GOODS));
 
-        assertEquals(385.95, catalog.sumPrices(Department.CLOTHING, Department.SPORTING_GOODS), 0.001);
-        assertEquals(794.09, catalog.sumPrices(Department.CLOTHING, Department.APPLIANCES), 0.001);
+//        assertEquals(385.95, catalog.sumPrices(Department.CLOTHING, Department.SPORTING_GOODS), 0.001);
+//        assertEquals(794.09, catalog.sumPrices(Department.CLOTHING, Department.APPLIANCES), 0.001);
     }
 
     @Assertion(id = "133", strategy = "Use a repository method findByIdBetween where the entity's Id attribute is named something other than id.")
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/AsciiCharacters.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/AsciiCharacters.java
@@ -80,7 +80,7 @@ public interface AsciiCharacters extends DataRepository<AsciiCharacter, Long>, I
     }
 
     @Save
-    Iterable<AsciiCharacter> saveAll(Iterable<AsciiCharacter> characters);
+    List<AsciiCharacter> saveAll(List<AsciiCharacter> characters);
 
     @Query("SELECT COUNT(THIS) WHERE numericValue <= 97 AND numericValue >= 74")
     int twentyFour();
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/ReadOnlyRepository.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/ReadOnlyRepository.java
@@ -15,6 +15,7 @@
  */
 package ee.jakarta.tck.data.framework.read.only;
 
+import java.util.List;
 import java.util.Optional;
 import java.util.stream.Stream;
 
@@ -31,7 +32,7 @@ public interface ReadOnlyRepository<T, K> extends DataRepository<T, K>{
     // WRITE - default method
     // Necessary for pre-population
     @Save
-    <S extends T> Iterable<S> saveAll(Iterable<S> entities);
+    <S extends T> List<S> saveAll(List<S> entities);
 
     // READ - default methods
     Optional<T> findById(K id);
@@ -40,7 +41,7 @@ public interface ReadOnlyRepository<T, K> extends DataRepository<T, K>{
 
     Stream<T> findAll();
 
-    Stream<T> findByIdIn(Iterable<K> ids);
+    Stream<T> findByIdIn(List<K> ids);
 
     long count();
 


tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/PersistenceEntityTests.java
```diff
@@ -110,8 +110,8 @@ public void testEntityManager() {
         catalog.save(Product.of("socks", 5.99, "TEST-PROD-87", Department.CLOTHING));
         catalog.save(Product.of("volleyball", 10.99, "TEST-PROD-89", Department.SPORTING_GOODS));
 
-        assertEquals(385.95, catalog.sumPrices(Department.CLOTHING, Department.SPORTING_GOODS), 0.001);
-        assertEquals(794.09, catalog.sumPrices(Department.CLOTHING, Department.APPLIANCES), 0.001);
+//        assertEquals(385.95, catalog.sumPrices(Department.CLOTHING, Department.SPORTING_GOODS), 0.001);
```
otaviojava, line 113: Should we remove this one instead?

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/Catalog.java
```diff
@@ -92,22 +90,22 @@ public interface Catalog extends DataRepository<Product, String> {
...
+//            jpql.append('?').append(d).append(" MEMBER OF o.departments");
+//        }
+//
+//        EntityManager em = getEntityManager();
+//        TypedQuery<Double> query = em.createQuery(jpql.toString(), Double.class);
+//        for (int d = 1; d <= departments.length; d++) {
+//            query.setParameter(d, departments[d - 1]);
+//        }
+//        return query.getSingleResult();
+//    }
```
njr-11, line 108:
```suggestion
    EntityManager getEntityManager();

    default double sumPrices(Department... departments) {
        StringBuilder jpql = new StringBuilder("SELECT SUM(o.price) FROM Product o");
        for (int d = 1; d <= departments.length; d++) {
            jpql.append(d == 1 ? " WHERE " : " OR ");
            jpql.append('?').append(d).append(" MEMBER OF o.departments");
        }

        EntityManager em = getEntityManager();
        TypedQuery<Double> query = em.createQuery(jpql.toString(), Double.class);
        for (int d = 1; d <= departments.length; d++) {
            query.setParameter(d, departments[d - 1]);
        }
        return query.getSingleResult();
    }
```

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/Catalog.java
```diff
@@ -33,8 +33,6 @@
 import jakarta.data.repository.Repository;
 import jakarta.data.repository.Save;
 import jakarta.data.repository.Update;
```
njr-11, line 35:
```suggestion
import jakarta.data.repository.Update;
import jakarta.persistence.EntityManager;
import jakarta.persistence.TypedQuery;
```

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/PersistenceEntityTests.java
```diff
@@ -110,8 +110,8 @@ public void testEntityManager() {
...
+//        assertEquals(794.09, catalog.sumPrices(Department.CLOTHING, Department.APPLIANCES), 0.001);
```
njr-11, line 114:
```suggestion
            assertEquals(794.09, catalog.sumPrices(Department.CLOTHING, Department.APPLIANCES), 0.001);
        }
```

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/PersistenceEntityTests.java
njr-11, line 113:
```suggestion
        boolean isSupported;
        try {
            assertNotNull(catalog.getEntityManager());
            isSupported = true;
        } catch (UnsupportedOperationException x) {
            isSupported = false;
        }

        if (isSupported) {
            assertEquals(385.95, catalog.sumPrices(Department.CLOTHING, Department.SPORTING_GOODS), 0.001);
```

njr-11 (changes requested):
The commit with changing Iterable to List makes sense.
See comment regarding the EntityManager resource accessor method. Instead of losing coverage, it can be written in a way that tests requirements of the spec whether EntityManager can be provided or not.

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/PersistenceEntityTests.java
njr-11: > The consequence of that is that the TCK has absolutely no business testing anything related to `EntityManager`.

The TCK has the obligation to test what is defined in the specification. Most of the language around Jakarta Persistence and EntityManager was removed from the specification, so it would be incorrect for the TCK to test behavior that depends on the usage or non-usage of an entity manager (whether stateful or stateless) for the implementation of repository insert/update/delete/find and so forth.  However, where `EntityManager` remains in the spec is where it states that a resource accessor method can be defined to return `EntityManager` and the Jakarta Data provider might give you one if it can or otherwise raise UnsupportedOperationException. The right course of action for testing resource accessor methods is to test that one of those two outcomes happens, which is what I have put into the suggestions in this pull.  There ought to be TCK tests for other two types defined in the spec as well: `javax.sql.DataSource` and `java.sql.Connection`.  Those might still need to be added.

gavinking: That's not quite right. The only thing we say about this (in 4.10) is that:

> A repository method that does not fit any of the above patterns and is not handled as a vendor-specific extension to the specification must either result in an error at build time or raise `UnsupportedOperationException` at runtime.

In this case, our implementation takes the first approach, and rejects this as an error at compilation time. (Which is friendlier to the user.)

Which means we reject that _whole_ repository interface.

gavinking: This suggestion assumes that the provider is not implemented to reject the code at compilation time. But the spec explicitly allows the provider to do that.

njr-11: > In this case, our implementation takes the first approach, and rejects this as an error at compilation time. (Which is friendlier to the user.)
> 
> Which means we reject that _whole_ repository interface.

I overlooked that detail when reading it.  Yes, it's friendlier to the typical user writing an application to be used with a single Jakarta Data provider.  But not to a user like the TCK that builds an application to run against multiple Jakarta Data providers and comes already compiled.  I'm copying @KyleAure on this for ideas because I expect there are other places like this in the TCK where exceptions are trapped at run time to find out whether a capability is available. We will need a more general solution to this. Probably a way to specify the information to the TCK when running it, but it seems like this could grow out of control when I think about NoSQL restrictions for different databases, which I don't think we've fully obtained from Otavio yet.

njr-11: @gavinking for general scenarios like this in the TCK, given that the TCK has already been successfully compiled and built, would it be acceptable to leave the methods on the repository if the TCK can guarantee it will never invoke them based on information that it is supplied with about what it can and cannot do?  It seems like this might be the simplest way to address these rather than isolating every permutation into its own repository.

gavinking: Not sure. On our side it's just this one method, I believe. And this isn't a case of it being a well-defined optional feature, exactly; it's rather that the provider itself has freedom to choose what the return type should be. You could solve the problem by a source-processing step in the TCK, to replace that return type with something specified by the provider.

But for NoSQL I don't have a good understanding. There it's a matter of a lot of things being well-defined but optional. I always wondered what you were planning to do about that.

gavinking: I guess my kneejerk reaction is simply that the TCK doesn't need to test things which are optional.

njr-11: > I guess my kneejerk reaction is simply that the TCK doesn't need to test things which are optional.

The argument from Jakarta EE would be that we aren't supposed to have optional capability in the spec at all.  In other cases, such as NoSQL limitations, the way we have been getting around this to to say that we are defining requirements for Jakarta Data implementations to externalize the capabilities of the underlying data store: by rejecting the behavior where the data store doesn't have it and providing the behavior that it is it capable of.  When we started writing the TCK, UnsupportedOperationException was required to be raised at run time, which made these things easily testable.

It seems to me like we have a gap in the spec because a user can compile in any way they like to get around the compile time errors and then attempt to use what they built on an implementation that relied on detecting errors only at compile time. Implementations will need to confront this possibility. Maybe we should adjust the specification language to say that UnsupportedOperationException must be raised if attempted at run time and that Jakarta Data providers can also reject the error earlier at compile time if they choose.  Would that work?

njr-11: Actually there would be worse errors than just the lack of the exception because other parts would be missing, so never mind that idea.

KyleAure: @gavinking @njr-11 
For your implementation, if at build time you produce an error for the method `EntityManager getEntityManager();` what is your behavior? 

1. Do you fail to install the application all together?
2. Do you fail to produce an implementation of `Catalog` and thus leave the injection point `@Inject Catalog catalog;` null?

I think the specification needs to be updated to clarify what should happen as a result of a `build time error` in this case. 
My vote would be for (2) since the spec running outside of a Jakarta EE platform doesn't have the option to fail installation.

I can work on updating the spec / TCK based on which behavior we agree on.

gavinking: I mean, it's just a regular compilation error at development time. So just like with any other error in user-written code, the code is rejected by the Java compiler, end of story. In the old language of Java EE, there's nothing to "deploy" to the "server", because the build failed and there was no artifact generated.

I'm don't think I understand what you guys are trying to get at here.

gavinking: > a user can compile in any way they like to get around the compile time errors and then attempt to use what they built on an implementation

I mean, sure, if they disable the annotation processor, then they can compile their interfaces, and then when they attempt to "deploy" their program, CDI will complain in a well-defined way about the unsatisfied injection points. This will happen when the application starts up (in WildFly) or at build time (in Quarkus).

It's just like with any other interface you try to inject but don't supply an implementing bean.

KyleAure: > In the old language of Java EE, there's nothing to "deploy" to the "server", because the build failed and there was no artifact generated.

Ahhh thanks for the explanation.  I'm still learning and was misinterpreting how an `Annotation Processor` works.  I figured the annotation processor you were speaking about hooked in at "build time" and was somehow hooked into the CDI lifecycle.  I didn't realize that there are annotation processors that actually ran at compile time.  Learn something new every day. 

Hmm I guess that does make it hard to write a TCK at all since I assume the annotation processor will complain at compile time and never end up running the TCK what-so-ever.  But at the same time the TCK has to test these error paths.  🤷🏼  I'm not sure what the solution is here.
FYI @njr-11 @gavinking

njr-11: After investigating usage of exceptions in the TCK somewhat, it appears that many cases are for exceptions that can be run time-only, although a number were trapping for the wrong exception type after it was switched and I opened a separate pull to fix those.  Given that, we might be able to address the others on a more individual basis.

Options that I can see here, none of which are very nice:

- Move the resource accessor method test to a test application of its own and add some switch for the TCK to be told whether or not to attempt it.
- Require resource accessor methods to raise `UnsupportedOperationException` at run time when incapable of the requested return type.
- Remove resource accessor methods from Jakarta Data entirely and then we don't need tests.
- Leave resource accessor methods in Jakarta Data, but have no tests for it.

gavinking: What about my suggestion: do a simple replacement in the source code.

njr-11: > What about my suggestion: do a simple replacement in the source code.

If that can be made to fit into the TCK somehow, then yes, that's fine.  Currently the TCK both has EntityManager as a repository method return type and also obtains the EntityManager and uses it.  So we would need to replace both the usage as well as the method return type.  I suppose the test could be simplified to not use the EntityManager at all and just assert a non-null value when it isn't replaced.

gavinking: > I suppose the test could be simplified to not use the EntityManager at all and just assert a non-null value when it isn't replaced.

That's what I would do.

gavinking:
Force pushed to resolve conflict.

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/PersistenceEntityTests.java
gavinking: I took a stab at implementing this using https://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html but because I'm a complete retard at Maven I failed. But I believe it to be possible.

njr-11 (approved):
The work on how to fix the testing for resource accessor methods can be done separately.  The test code is currently wrong to assume it will always be possible to obtain EntityManager and there are corrections to other areas in this pull as well, so I'll approve it so we can merge it an make the TCK valid.  Another pull can be used to solve the resource accessor method testing.

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/PersistenceEntityTests.java
KyleAure: I can help out just not sure what we are attempting to do here.  Document/create a maven task that will replace the Catalog.class prior to compile time?

PR #592, opened 2024-03-25
TCK tests for literals in query language

njr-11:
Add TCK tests for literals defined by the Jakarta Data Query Language.

While testing this, I noticed that we did not comply with [Jakarta Persistence 3.2 M2 specification]( https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2-m2#literals) regarding how enum literals and so I fixed that as well.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/AsciiCharacters.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/AsciiCharacters.java
@@ -27,6 +27,7 @@ import jakarta.data.page.PageRequest;
 import jakarta.data.repository.By;
 import jakarta.data.repository.DataRepository;
 import jakarta.data.repository.Find;
+import jakarta.data.repository.Query;
 import jakarta.data.repository.Repository;
 import jakarta.data.repository.Save;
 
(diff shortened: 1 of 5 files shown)


gavinking:
> While testing this, I noticed that we did not comply with [Jakarta Persistence 3.2 M2 specification](https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2-m2#literals) regarding how enum literals and so I fixed that as well.

Ah, nice catch, I missed that (Hibernate does not require the class name).

gavinking:
Then the grammar needs to change too:

    enum_literal : IDENTIFIER ('.' IDENTIFIER)*;

njr-11:
> Then the grammar needs to change too:
> 
> ```
> enum_literal : IDENTIFIER ('.' IDENTIFIER)*;
> ```

Thanks for pointing that out.  I added it to [this commit](https://github.com/jakartaee/data/pull/592/commits/42b281b1eac28912dedb1e71b48ccf4bb57eb872).

gavinking (approved):

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/AsciiCharacters.java
```diff
@@ -70,11 +71,17 @@ Stream<AsciiCharacter> findByHexadecimalIgnoreCaseBetweenAndHexadecimalNotIn(Str
 
     Optional<AsciiCharacter> findFirstByHexadecimalStartsWithAndIsControlOrderByIdAsc(String firstHexDigit, boolean isControlChar);
 
+    @Query("WHERE hexadecimal <> ' ORDER BY isn''t a keyword when inside a literal' AND hexadecimal IN ('4a', '4b', '4c', ?1)")
```
otaviojava, line 74: If it is verified at build time, it might not compile.

njr-11: > If it is verified at build time, it might not compile.

Can you explain why and where the error is? As best I can see, the query language used is consistent with what is defined in the spec here. A user can write this query and they will expect it to work, per the spec.

PR #593, opened 2024-03-26
TCK has outdated expectation for exception type

njr-11:
While investigating exceptions that are trapped in the TCK, I found a bunch of places where the TCK is trapping for `MappingException` that should have been updated when the specification was switched to `UnsupportedOperationException`. Fortunately, these are all cases where the requirement for the exception is at run time only because it requires finding out information from the database.  Otherwise we would have a problem testing the behavior at all in these cases.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/PositiveIntegers.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/PositiveIntegers.java
@@ -41,7 +41,7 @@ public interface PositiveIntegers extends BasicRepository<NaturalNumber, Long> {
 
     boolean existsByIdGreaterThan(Long number);
 
-    CursoredPage<NaturalNumber> findByFloorOfSquareRootNotAndIdLessThanOrderByBitsRequiredDesc(long excludeSqrt,
+    CursoredPage<NaturalNumber> findByFloorOfSquareRootNotAndIdLessThanOrderByNumBitsRequiredDesc(long excludeSqrt,
                                                                                                   long eclusiveMax,
                                                                                                   PageRequest<NaturalNumber> pagination);
 
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
@@ -703,16 +703,18 @@ public class EntityTests {
     public void testFirstCursoredPageOf8AndNextPages() {
         // The query for this test returns 1-15,25-32 in the following order:
 
-        // 25, 26, 27, 28, 29, 30, 31, 32 square root rounds down to 4
-        // 9, 10, 11, 12, 13, 14, 15 square root rounds down to 3
-        // 4, 5, 6, 7, 8 square root rounds down to 2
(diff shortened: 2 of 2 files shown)


tck/src/main/java/ee/jakarta/tck/data/framework/read/only/PositiveIntegers.java
```diff
@@ -41,7 +41,7 @@ public interface PositiveIntegers extends BasicRepository<NaturalNumber, Long> {
 
     boolean existsByIdGreaterThan(Long number);
 
-    CursoredPage<NaturalNumber> findByFloorOfSquareRootNotAndIdLessThanOrderByBitsRequiredDesc(long excludeSqrt,
+    CursoredPage<NaturalNumber> findByFloorOfSquareRootNotAndIdLessThanOrderByNumBitsRequiredDesc(long excludeSqrt,
```
njr-11, line 44: Query by Method Name was using incorrect field name that did not match the entity.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
```diff
@@ -704,51 +703,53 @@ public void testFindPage() {
...
-        // 25, 26, 27, 28, 29, 30, 31, 32 square root rounds down to 4
-        // 9, 10, 11, 12, 13, 14, 15 square root rounds down to 3
-        // 4, 5, 6, 7, 8 square root rounds down to 2
-        // 1, 2, 3 square root rounds down to 1
+        // 32 requires 6 bits
+        // 25, 26, 27, 28, 29, 30, 31 requires 5 bits
+        // 8, 9, 10, 11, 12, 13, 14, 15 requires 4 bits
+        // 4, 5, 6, 7, 8 requires 3 bits
+        // 2, 3 requires 2 bits
+        // 1 requires 1 bit
```
njr-11, line 711: This test case was trying to assert ordering by square root descending but it actually asks for sorting by number of bits required descending.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
```diff
@@ -966,17 +967,17 @@ public void testCursoredPageOf7FromCursor() {
...
+        } catch (UnsupportedOperationException x) {
+            // Test passes: Jakarta Data providers must raise UnsupportedOperationException when the database
             // is not capable of cursor-based pagination.
             return;
         }
 
         assertEquals(Arrays.toString(new Long[] { 10L, 9L, // 4 bits required, square root rounds down to 3
                                                   7L, 6L, 5L, 4L, // 3 bits required, square root rounds down to 2
-                                                  3L, 2L // 2 bits required, square root rounds down to 1
+                                                  3L // 2 bits required, square root rounds down to 1
```
njr-11, line 980: This test case asks for pages of size 7, but mistakenly tries to assert that this page will have 8 elements.

njr-11 (commented):
When running the TCK, I found that MappingException has hiding a few bugs in the TCK.  See the following comments for explanations of what those errors are.

KyleAure (approved):
LGTM

PR #599, opened 2024-03-26
fully incorporate id(this) into JDQL

gavinking:
A proposal, in case we want this in 1.0.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/BasicRepository.java
+++ b/api/src/main/java/jakarta/data/repository/BasicRepository.java
@@ -149,6 +149,19 @@ public interface BasicRepository<T, K> extends DataRepository<T, K> {
     Optional<T> findById(@By(ID) K id);
 
     /**
+     * Retrieves the entities with the given unique identifiers.
+     * The returned list of entities is sorted by unique identifier
+     * in ascending order. If any of the unique identifiers is not
+     * found in the database, the returned list will be smaller
+     * than the supplied list of unique identifiers.
+     *
+     * @param ids a list of unique identifiers.
+     * @return the entities with the given unique identifiers.
+     */
+    @Query("where id(this) in :ids order by id(this)")
+    List<T> findByIdIn(@Param("ids") List<K> ids);
+
+    /**
      * Retrieves all persistent entities of the specified type from the database.
      *
      * @return a stream of all entities; will never be {@code null}.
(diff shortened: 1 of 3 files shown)


otaviojava (approved):

njr-11 (changes requested):

api/src/main/java/jakarta/data/repository/BasicRepository.java
```diff
@@ -146,6 +146,15 @@ public interface BasicRepository<T, K> extends DataRepository<T, K> {
...
     Optional<T> findById(@By(ID) K id);
 
+    /**
+     * Retrieves the entities with the given unique identifiers.
+     *
+     * @param ids a list of unique identifiers.
+     * @return the entities with the given unique identifiers.
+     */
+    @Query("where id(this) in :ids order by id(this)")
+    List<T> findByIdIn(@By(ID) List<K> ids);
```
otaviojava, line 156: Why put this method back again?
We agree to remove it.

njr-11 (approved):

api/src/main/java/jakarta/data/repository/BasicRepository.java
gavinking: Well, we decided to remove it because we didn't have a good way to express it in any of the available query facilities, not because we hated it and thought it was useless.

I think it's kinda, well, at least medium-useful. Not a _critical_ thing to have, but not a wart either.

otaviojava (approved):

otaviojava:
@gavinking could you rebase this one?

gavinking:
I will, but we have not decided to do it yet.

otaviojava:
@gavinking can you rebase this one, please?

gavinking:
> @gavinking can you rebase this one, please?

Yep, done.

api/src/main/java/jakarta/data/repository/BasicRepository.java
```diff
@@ -148,6 +148,19 @@ public interface BasicRepository<T, K> extends DataRepository<T, K> {
...
+     * in ascending order. If any of the unique identifiers is not
+     * found in the database, the returned list will be smaller
+     * than the supplied list of unique identifiers.
+     *
+     * @param ids a list of unique identifiers.
+     * @return the entities with the given unique identifiers.
+     */
+    @Query("where id(this) in :ids order by id(this)")
+    List<T> findByIdIn(@Param("ids") List<K> ids);
+
```
njr-11, line 163: The rest of this PR looks good, but I don't think we can add a new method to the existing BasicRepository interface now that we have released 1.0.  This would be a breaking change and would require at least an update of the major version.

```suggestion
```

gavinking: OK

njr-11 (approved):

PR #604, opened 2024-03-28
clarify that the type of "count(this)" is Long

gavinking:
and that it collapses the result set to one result


gavinking:
Note that this is for consistency with the section titled "Aggregate Functions in the SELECT Clause" of the JPA spec.

otaviojava (approved):

njr-11 (approved):

PR #605, opened 2024-03-28
char[] -> Character[] in TCK query method return type

gavinking:
The return type of these queries in JPQL is implicitly, by definitely, `List<Character>`, and Jakarta Data does not require conversion from `Character[]` to `char[]`.


njr-11 (approved):
It's fine to update the TCK, but I could see users easily overlooking this like I did and trying the same sort of thing when selecting a single entity attribute, although probably with `int` or `long` fields rather than `char`.

gavinking:
> It's fine to update the TCK, but I could see users easily overlooking this like I did and trying the same sort of thing when selecting a single entity attribute, although probably with `int` or `long` fields rather than `char`.

That's why I reject it at compile time :)

PR #606, opened 2024-03-29
Add Order.by(List)

gavinking:
These are convenience methods, but useful ones.

### DIFF
--- a/api/src/main/java/jakarta/data/Order.java
+++ b/api/src/main/java/jakarta/data/Order.java
@@ -104,6 +104,19 @@ public class Order<T> implements Iterable<Sort<? super T>> {
     }
 
     /**
+     * <p>Defines a list of {@link Sort} criteria, ordered from highest precedence
+     * to lowest precedence.</p>
+     *
+     * @param <T>   entity class of the attributes that are used as sort criteria.
+     * @param sorts sort criteria to use, ordered from highest precedence to lowest precedence.
+     * @return a new instance indicating the order of precedence for sort criteria.
+     *         This method never returns {@code null}.
+     */
+    public static <T> Order<T> by(List<Sort<? super T>> sorts) {
+        return new Order<T>(List.copyOf(sorts));
+    }
+
+    /**
      * The instances of {@link Sort} belonging to this {@code Order}.
      *
      * @return the instances of {@link Sort}, from highest precedence to lowest precedence.


api/src/main/java/jakarta/data/page/Pagination.java
```diff
@@ -131,6 +131,16 @@ public PageRequest<T> previous() {
...
 
+    @Override
+    public PageRequest<T> nextWithKey(Object... key) {
+        return new Pagination<>(page + 1, this.size, this.sorts, Mode.OFFSET, Cursor.forKey(key), requestTotal);
+    }
+
+    @Override
+    public PageRequest<T> previousWithKey(Object... key) {
+        return page()<=1 ? null : new Pagination<>(page - 1, this.size, this.sorts, Mode.OFFSET, Cursor.forKey(key), requestTotal);
+    }
```
njr-11, line 142: Because cursor-based pagination allows for updates to the data which can cause new entries to appear, I think we have language in the spec that says that a previous page to page 1 should also be numbered as 1.
The JavaDoc to the new method you proposed doesn't state anything about the possibility of a null result.
```suggestion
    public PageRequest<T> previousWithKey(Object... key) {
        return new Pagination<>(page > 1 ? page - 1 : 1, this.size, this.sorts, Mode.OFFSET, Cursor.forKey(key), requestTotal);
    }
```

api/src/main/java/jakarta/data/page/PageRequest.java
```diff
@@ -344,6 +344,32 @@ static <T> PageRequest<T> ofSize(int maxPageSize) {
      */
     PageRequest<T> previous();
 
+    /**
+     * <p>Returns a {@code PageRequest} requesting the next page
+     * if using cursor-based pagination. The given key must be the
+     * key of the last result on this page.</p>
+     *
+     * <p>Note that traversal of pages is usually done via
+     * {@link CursoredPage#nextPageRequest()}</p>
```
njr-11, line 353: I don't understand the benefit of adding the nextWithKey/previousWithKey methods.  Their own JavaDoc seems to imply that it should typically by done another way, and I agree with that.  Why are these needed?  I'd rather not add methods for a non-preferred way of doing things because it obscures the more useful methods.

njr-11 (changes requested):
`Order.by(List<Sort>)` seems fine, but I'm unclear on the benefit of the other methods

api/src/main/java/jakarta/data/Order.java
```diff
@@ -100,7 +100,20 @@ private Order(List<Sort<? super T>> sorts) {
...
+     * <p>Defines a list of {@link Sort} criteria, ordered from highest precedence
+     * to lowest precedence.</p>
+     *
+     * @param <T>   entity class of the attributes that are used as sort criteria.
+     * @param sorts sort criteria to use, ordered from highest precedence to lowest precedence.
+     * @return a new instance indicating the order of precedence for sort criteria.
+     *         This method never returns {@code null}.
+     */
+    public static <T> Order<T> by(List<Sort<? super T>> sorts) {
+        return new Order<>(List.copyOf(sorts));
```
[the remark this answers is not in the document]
gavinking: That is ... a bit weird, since IntelliJ complained it was redundant.

api/src/main/java/jakarta/data/page/PageRequest.java
gavinking: Because the only other way to do it is this disgusting monstrosity:

```java
			var _page =
					PageRequest.of(NaturalNumber.class)
							.sortBy((Iterable) pagination.sorts())
							.size(pagination.size())
							.page(pagination.page() + 1);
			return new CursoredPageRecord(_results.getResultList(), _cursors, _totalResults, pagination,
					_results.isLastPage() ? null : _page.afterKey(_results.getNextPage().getKey().toArray()),
					_results.isFirstPage() ? null : _page.beforeKey(_results.getPreviousPage().getKey().toArray()));
```

That's from my generated code, of course, but if I needed to do this, I guarantee you that at least some user will need to do it.

gavinking: I mean, the other option is to kill the `UnsupportedOperationException` that `Page.next()` throws, which is anyway of dubious value.

api/src/main/java/jakarta/data/page/Pagination.java
gavinking: >  I think we have language in the spec that says that a previous page to page 1 should also be numbered as 1.

OK, I missed that, but I think it's fine.

api/src/main/java/jakarta/data/Order.java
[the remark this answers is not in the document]
gavinking: So, confirmed, IntelliJ is wrong. Fixed.

api/src/main/java/jakarta/data/page/PageRequest.java
njr-11: If I'm reading your generated code correctly (I'm not sure what the types of some of the variables are and some of those methods don't exist in Jakarta Data), I think you could do this:
```java
long page = pagination.page();
return new CursoredPageRecord<NaturalNumber>(_results.getResultList(), _cursors, _totalResults, pagination,
                _results.isLastPage() ? null : pagination.page(page + 1).afterCursor(_cursors.get(_cursors.size() - 1)),
                _results.isFirstPage() ? null : pagination.page(page > 1 ? (page - 1) : 1).beforeCursor(_cursors.get(0)));
```

gavinking: Ah, thanks, let me try that.

gavinking: [So it's at least a _little_ weird that `nextPage()` throws when `page(page+1)` doesn't 😆]

gavinking: It's not _that_ bad. They actually just get a request for the next page by offset, with no cursor in it. So they don't get cursor-based pagination, but the _do_ get a perfectly legit next page, I think.

Right?

gavinking: >  I think you could do this

This works perfectly, thanks.

gavinking:
> I'm unclear on the benefit of the other methods

I have now dropped the commits which added those two methods.

njr-11 (approved):
Looks good now.

PR #607, opened 2024-03-29
TCK tests for JDQL UPDATE and DELETE

njr-11:
Add TCK tests for JDQL UPDATE and DELETE with and without a WHERE clause.
One of the tests uses a repository that handles multiple entity types where there is no primary entity type, relying on the FROM clause to identify the desired entity.  This also adds an additional entity to the TCK so that we can have some coverage for float, double, and UUID.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/MultipleEntityRepo.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/MultipleEntityRepo.java
@@ -1,0 +1,55 @@
+/**
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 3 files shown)


otaviojava (approved):

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/MultipleEntityRepo.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * A repository that performs operations on different types of entities.
+ */
+@Repository
+public interface MultipleEntityRepo { // Do not add a primary entity type.
+
+    // Methods for Box entity:
+
+    @Insert
+    Box[] addAll(Box... boxes);
```
gavinking, line 34: We've never said explicitly that varargs is allowed here. 

I'm OK to start requiring support for it, but if support is required we should mention that in the Javadoc and spec.

gavinking (approved):

PR #614, opened 2024-03-30
clarify that variadic parameters are allowed for lifecycle methods

gavinking:
Since the TCK is now testing a variadic lifecycle method, we had better bless that in the spec.


otaviojava (approved):

njr-11 (approved):
For some reason I thought we already had language allowing variable arguments parameters to life cycle methods when I wrote that TCK test, but I just checked and it is not there.  I probably confused it with allowing variable arguments Sort.  In any case, this looks good and will cover it.

PR #616, opened 2024-04-01
Fix readme with a proper repository

otaviojava:


README.adoc
```diff
@@ -121,11 +121,11 @@ public interface BookRepository extends BasicRepository<Book, UUID> {
 
   // Find books with titles matching a specific pattern
   @Query("where title like :titlePattern")
-  List<Book> booksMatchingTitle(String titlePattern);
+  List<Book> booksMatchingTitle(@Param("titlePattern") String titlePattern);
 
   // Select books by a specific author and sort them by title
   @Query("where author.name = :author order by title")
-  List<Book> findByAuthorSortedByTitle(String author);
+  List<Book> findByAuthorSortedByTitle(@Param("author") String author);
```
gavinking, line 128: Do we really need to add this `@Param` annotations?

gavinking: I mean, I guess I think that the readme here should be showing examples of what we think Jakarta Data repositories are most likely to look like in practice.

It seems to me that anyone using a provider based on reflection is going to enable `-parameters` and save themselves the effort of typing out most of those `@By` and `@Param` annotations. And anyone using a provider based on annotation processing doesn't need to do even that.

otaviojava:
@gavinking @njr-11 ?

njr-11 (approved):

PR #620, opened 2024-04-01
TCK test for ID(THIS)

njr-11:
If #599 goes in, here is a TCK test to cover ID(THIS) in various clauses where it is allowed.


KyleAure (approved):
This looks good.  If the dependent change gets merged.

PR #627, opened 2024-04-03
sync requirements for sorting with JPQL

gavinking:
this is a traditional restriction on the 'order by' clause

see section 4.10 of latest JPA 3.2 specification

### DIFF
--- a/spec/src/main/asciidoc/query-language.asciidoc
+++ b/spec/src/main/asciidoc/query-language.asciidoc
@@ -366,6 +366,8 @@ The `set` clause, with syntax given by `set_clause`, specifies a list of updates
 
 The `order` clause (or `order by` clause), with syntax given by `orderby_clause`, specifies a lexicographic order for the query results, that is, a list of entity fields used to sort the records which satisfy the restriction imposed by the `where` clause. The keywords `asc` and `desc` specify that a given field should be sorted in ascending or descending order respectively; when neither is specified, ascending order is the default.
 
+NOTE: An implementation of JDQL is not required to support sorting by entity fields which are not returned by the query. If a query returns the queried entity, then any field of the queried entity may occur in the `order` clause. Otherwise, if the query has is an explicit `select` clause, a provider might require that a field which occurs in the `order` also occurs in the `select`.
+
 Entity fields occurring earlier in the `order by` clause take precedence. That is, a field occurring later in the `order by` clause is only used to resolve "ties" between records which cannot be unambiguously ordered using only earlier fields.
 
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -366,6 +366,8 @@ The `set` clause, with syntax given by `set_clause`, specifies a list of updates
 
 The `order` clause (or `order by` clause), with syntax given by `orderby_clause`, specifies a lexicographic order for the query results, that is, a list of entity fields used to sort the records which satisfy the restriction imposed by the `where` clause. The keywords `asc` and `desc` specify that a given field should be sorted in ascending or descending order respectively; when neither is specified, ascending order is the default.
 
+NOTE: An implementation of JDQL is not required to support sorting by entity fields which are not returned by the query. If a query returns the queried entity, then any field of the queried entity may occur in the `order` clause. Otherwise, if the query has is an explicit `select` clause, a provider might require that a field which occurs in the `order` also occurs in the `select`.
```
njr-11, line 369: Restrict the language only to sortable fields. Otherwise, it looks like we are declaring all entity fields to be sortable.
```suggestion
NOTE: An implementation of JDQL is not required to support sorting by entity fields which are not returned by the query. If a query returns the queried entity, then any sortable field of the queried entity may occur in the `order` clause. Otherwise, if the query has an explicit `select` clause, a provider might require that a field which occurs in the `order` also occurs in the `select`.
```

njr-11 (commented):
This clears up an issue that I tried to raise while reviewing the JDQL pull.  I remember asking about the order of evaluation of the clauses and how SELECT could be done before ORDER BY when SELECT might restrict the available entity attributes such that some aren't available for ORDER BY to sort on. What you added causes the section that I was asking about to make more sense now.

spec/src/main/asciidoc/query-language.asciidoc
gavinking: Applied.

njr-11 (approved):

gavinking:
> This clears up an issue that I tried to raise while reviewing the JDQL pull. I remember asking about the order of evaluation of the clauses and how SELECT could be done before ORDER BY when SELECT might restrict the available entity attributes such that some aren't available for ORDER BY to sort on. What you added in this pull causes the section that I was asking about to make more sense now.

OK, cool, yes, I remember the discussion. And yeah, strictly-speaking you're right that if projection happens before sorting then you can't sort by columns which are projected away. In actually practice I ignore this restriction all the time with no ill effects. But here we should be more careful.

PR #628, opened 2024-04-03
remove TCK test for removed features of query by method name

gavinking:
we removed the `Contains` and `Empty` operators from query by method name


njr-11 (commented):
I agree with removing these because they aren't in the spec anymore, but we also need to remove the test methods `testContainsInCollection` and `testEmpty` that invoke these methods.

gavinking:
Ooops.🤭

njr-11 (approved):

PR #632, opened 2024-04-03
Extract method query in a single asciidoc

otaviojava:
# Changes

- Create a single page at method-query.asciidoc


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -264,304 +264,7 @@ If a method of a repository interface has more than one such annotation, the ann
...
-
-===== Logical Operator Precedence
-
-For relational databases, the logical operator `And` takes precedence over `Or`, meaning that `And` is evaluated on conditions before `Or` when both are specified on the same method. For other database types, the precedence is limited to the capabilities of the database. For example, some graph databases are limited to precedence in traversal order.
-
-==== Return Types
-
-Refer to the Jakarta Data module Javadoc section on "Return Types for Repository Methods" for a listing of valid return types for methods using Query by Method Name.
-
-=== Special Parameters for Limits, Sorting, and Pagination
```
njr-11, line 500: This section applies to various patterns in Jakarta Data and is not unique to Query by Method Name.  It should remain in the main document instead of being moved.

njr-11: To be clear, I'm specifically referring to the section starting from 
```
=== Special Parameters for Limits, Sorting, and Pagination
```
Github also shows some prior lines on my comment, but I am not referring to those preceding lines.

otaviojava: @njr-11 This section is on the repository

Screenshot 2024-04-03 at 17 44 34

otaviojava:
> This moves Query by Method Name spec doc to a new file from the development side. I think Gavin also wants it in a separate PDF, but this is a good step in that direction.

We can do a PDF from a single file, there is no issue on it.

PR #636, opened 2024-04-03
separate PDF document for Query by Method Name

gavinking:
Follows up on @otaviojava's PR #632 which moved this material to a separate file.

This PR compiles that material to a new file, and adds language explaining the relationship between the two documents.

I also moved a section of chapter 3 which was specific to Query by Method Name to the new file.

See issue #637.


spec/src/main/asciidoc/method-query.asciidoc
```diff
@@ -1,6 +1,67 @@
-=== Query by Method Name
-
-The Query by method mechanism allows for creating query commands by naming convention.
+// Copyright (c) 2022, 2023 Contributors to the Eclipse Foundation
```
njr-11, line 1:
```suggestion
// Copyright (c) 2022, 2024 Contributors to the Eclipse Foundation
```

spec/src/main/asciidoc/method-query.asciidoc
```diff
@@ -1,6 +1,67 @@
...
+ifndef::sourcedir[:sourcedir: ../../main/java]
+
+:sectnums!:
+include::chapters/license/license-efsl.adoc[]
+
+:sectnums:
+
+== Introduction
+
+_Query by Method Name_ is a query language suitable for embedding in the names of methods written in a programming language such as Java.
```
njr-11, line 46:
```suggestion
_Query by Method Name_ is a query language suitable for embedding in the names of methods written in Java.
```

spec/src/main/asciidoc/method-query.asciidoc
```diff
@@ -1,6 +1,67 @@
...
+As such, its syntax is limited to the use of legal identifier characters, and so the text of a query may not contain whitespace, punctuation characters, numeric operators, nor comparison operators.
```
njr-11, line 47: minor grammar update
```suggestion
As such, its syntax is limited to the use of legal identifier characters, so the text of a query must neither contain whitespace, nor punctuation characters, nor numeric operators, nor comparison operators.
```

spec/pom.xml
```diff
@@ -68,28 +68,53 @@
...
+                    <execution>
+                        <id>generate-pdf-doc-method-query</id>
+                        <phase>generate-resources</phase>
+                        <goals>
+                            <goal>process-asciidoc</goal>
+                        </goals>
+                        <configuration>
+                            <sourceDocumentName>method-query.asciidoc</sourceDocumentName>
+                            <backend>pdf</backend>
+                            <outputFile>${gen-doc-dir}/method-query-${project.version}.pdf</outputFile>
```
njr-11, line 85: Given that the file name is likely to be externally visible, we should give some thought to what to call it, such as whether to prefix it with jakarta or jakarta-data and whether to make it more explicitly named after Query by Method Name, such as query-by-method-name. But if we do either of these, the name would get longer. I don't know what is best so I thought I'd point it out.

gavinking: Yes. I was not sure either, and I wanted to see whachall thought about it.

spec/src/main/asciidoc/method-query.asciidoc
```diff
@@ -1,6 +1,67 @@
...
+
+:sectnums!:
+include::chapters/license/license-efsl.adoc[]
+
+:sectnums:
+
+== Introduction
+
+_Query by Method Name_ is a query language suitable for embedding in the names of methods written in Java.
+As such, its syntax is limited to the use of legal identifier characters, so the text of a query must neither contain whitespace, nor punctuation characters, nor numeric operators, nor comparison operators.
```
gavinking, line 47:
```suggestion
As such, its syntax is limited to the use of legal identifier characters, so the text of a query must contain neither whitespace, nor punctuation characters, nor numeric operators, nor comparison operators.
```

spec/pom.xml
gavinking: `jakarta-method-name-query.pdf`?

gavinking: OK, with the thumbs up from Nathan, I've made that change to the file name.

njr-11 (approved):

PR #642, opened 2024-04-04
fix type of query by method name parameter in TCK test

gavinking:
There is no requirement in the spec that providers convert long to int in repository method parameters.


njr-11 (approved):
It looks like the test was mistakenly using the wrong type. The entity has `int`:
```
    private int numericValue;
```

KyleAure (approved):

PR #645, opened 2024-04-04
copy information from module-info to spec

gavinking:
The information about Query by Method Name return types should be in the spec (by value, not by reference).

I have also taken the liberty to remove some weird statements about Jakarta Persistence being limited to `Integer.MAX_VALUE`. To be clear, the type of the `count()` function in JPQL is `long`.

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -636,7 +636,7 @@ import jakarta.data.repository.Update;
  * <tr style="vertical-align: top; background-color:#eee"><td>{@code countBy...}</td>
  * <td>{@code long},
  * <br>{@code int}</td>
- * <td>Jakarta Persistence providers limit the maximum to {@code Integer.MAX_VALUE}</td></tr>
+ * <td></td></tr>
  *
  * <tr style="vertical-align: top"><td>{@code deleteBy...},
  * <br>{@code updateBy...}</td>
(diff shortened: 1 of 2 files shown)


api/src/main/java/module-info.java
```diff
@@ -636,15 +636,15 @@
  * <tr style="vertical-align: top; background-color:#eee"><td>{@code countBy...}</td>
  * <td>{@code long},
  * <br>{@code int}</td>
- * <td>Jakarta Persistence providers limit the maximum to {@code Integer.MAX_VALUE}</td></tr>
```
njr-11, line 639: I know I wrote these statements at one point, and I thought I remember doing so in response to limitations that I found in the Jakarta Persistence spec but I cannot find it now, and in any case, it doesn't make any sense to me why the language was on methods that are returning counts rather than find methods that return lists of results, which seems the more likely place to encounter limits.

njr-11 (approved):

api/src/main/java/module-info.java
gavinking: Ah, right, well it's true that a Java `List` or array can only have `MAX_INTEGER` elements, but that's not specific to Jakarta Persistence. Anyway, that many entities in memory will definitely blow up your computer.

otaviojava (approved):

PR #644, opened 2024-04-04
There should be an EnityQualifier type of annotation to mark custom entity defining annotations

otaviojava:
Fix: https://github.com/jakartaee/data/issues/638

PR #647, opened 2024-04-04
EntityDefining was added without updating TCK

njr-11:
#644 adds the requirement for custom entity defining annotations to be annotated with EntityDefining, but didn't update the one in the TCK.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/common/cdi/PersonEntity.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/common/cdi/PersonEntity.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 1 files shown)


gavinking (approved):

gavinking:
Wait, I don't understand how this test is supposed to work.

`Person` is declared with a custom entity-defining annotation that no provider supports.

Therefore every implementation should just ignore the `Directory` repository and generate no implementation for it.

That's fine. But then why is `Directory` actually used all over the place in the TCK?

njr-11:
> That's fine. But then why is `Directory` actually used all over the place in the TCK?

`Directory` is not used all over the place in the TCK. It is only used in the two ExtensionTests (one for core profile, one for web profile), which are the tests that use the fake Jakarta Data providers supplied by the TCK that are supposed to handle this entity type.  I didn't check whether those test Data providers are properly handling that entity. Hopefully they are. If not, the tests won't work and we will need to fix them.

gavinking:
OK. Thanks.

njr-11:
It looks to me like they are cheating and relying on `@Repository(provider =` to be assigned the `Directory` repository,
```
```

That means the coverage here isn't as good as it should be.  We should leave this in place because it's also a valid scenario and create a second repository for `Person` that omits `provider = ...` and instead relies on the type of Entity annotation.

KyleAure (approved):

KyleAure:
Merging with failed CI build - this branch was not re-based after the compile time errors were fixed.

KyleAure:
Verified build passed after merge: https://github.com/jakartaee/data/actions/runs/8571764122

PR #649, opened 2024-04-04
Fix QBN issue in CDI tests

### LINKED ISSUE #640: [TCK Challenge]: ExtensionTests#testDataProviderWithBuildCompatibleExtension conflicts with QBN
### Specification

spec/src/main/asciidoc/repository.asciidoc

### Assertion

A valid Query by Method Name exists but the TCK is expecting something different

### TCK Version

SNAPSHOT (Locally built)

HIbernate

### Challenge Scenario

Claims that a test assertion conflicts with the specification.

### Full Description

The `ExtensionTests#testDataProviderWithBuildCompatibleExtension` test is expecting that it can call the injected ee.jakarta.tck.data.common.cdi.Directory#findFirstNameByIdInOrderByAgeDesc() method and obtain a list of `Person.firstName` values as though the repository was annotated with:

```java
    @Query("select firstName from Person where id in ?1 order by age desc")
    List<String> findFirstNameByIdInOrderByAgeDesc(List<Long> ids);
}
```

while `findFirstNameByIdInOrderByAgeDesc` is a valid QBN method that parses to:
find-expression: findFirst
ignored-text: Name
predicate: id IN 
order-clause: OrderBy age Desc

which results in an interface annotated like:
```java
public interface Directory$ extends Directory {
            @Override
            @Query("where id in ?1 order by '' limit 1")
            @OrderBy(value="age", descending = true)
            public java.util.List<java.lang.String> findFirstNameByIdInOrderByAgeDesc (java.util.List<java.lang.Long> ids);
}
```

but the return type should be type of `Person` rather than `firstName` values.

### Additional Context

_No response_

KyleAure:
- fixes #640 
- Verifies that another Jakarta Data Provider does not attempt to implement the Dictonary repository
- Fixes the Signature Tests

~DRAFT - signatures need to be updated with jakarta.data.spi package, but a compile time failure was merged.~

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/common/cdi/Directory.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/common/cdi/Directory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023, 2024 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -31,5 +31,5 @@ public interface Directory extends DataRepository<Person, Long> {
     //Self referencing provider so that this repository can be used for both WEB and CORE tests
     public static final String PERSON_PROVIDER = "PERSON_PROVIDER";
     
-    List<String> findFirstNameByIdInOrderByAgeDesc(List<Long> ids);
+    List<String> findLastNameByIdInOrderByAgeDesc(List<Long> ids);
 }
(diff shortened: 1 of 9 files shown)


tck/src/main/java/ee/jakarta/tck/data/common/cdi/Directory.java
```diff
@@ -31,5 +35,12 @@ public interface Directory extends DataRepository<Person, Long> {
     //Self referencing provider so that this repository can be used for both WEB and CORE tests
     public static final String PERSON_PROVIDER = "PERSON_PROVIDER";
     
-    List<String> findFirstNameByIdInOrderByAgeDesc(List<Long> ids);
+    List<String> findLastNameByIdInOrderByAgeDesc(List<Long> ids);
```
gavinking, line 38: I think this is still wrong, isn't it? The problem wasn't that `First` was ambiguous, the problem was that QbMN simply does not have a projection operator.

KyleAure: Sorry had to re-read the spec to understand completely.  Updated the test so it should follow the query by method name BNF

gavinking:
Nice, that looks better, thanks.

njr-11 (approved):

PR #654, opened 2024-04-04
Fixing subtitle for Extension html file

### LINKED ISSUE #653: [Bug]: The HTML for the separate Query by Method Name doc looks like it is the Jakarta Data spec rather than part of it
### Specification Version

SNAPSHOT (Locally built)

### Bug report

When I view the HTML version of the Query by Method Name document, it has the appearance that that document is the Jakarta Data spec rather than a part of it.  See the following image, which titles it as "Jakarta Data".

image

### Additional information

_No response_

mswatosh:
fixes #653

Sets the title for the html version of the new Extension file to:
`Jakarta Data: Query by Method Name Extension`

I had to switch the Specification to use a separate variable, to prevent the specification in the Extension file from using the doctitle.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -14,7 +14,7 @@ body:
       required: true
     attributes:
       label: Specification Version
-      description: What version of the Concurrency Spec are you running using?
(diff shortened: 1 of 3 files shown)


gavinking (approved):
Yup, look good, thanks!

PR #639, opened 2024-04-04
Remove Sort at PageRequest

otaviojava:
# Changes

The goal is to define single responsibility at PageRequest

PR #655, opened 2024-04-04
specification document is out of sync with #639

njr-11:
This is another pull to fix the halfway completed work that was done in #639 where Sort was decoupled from PageRequest.  Examples and language throughout the specification document were not updated to match the API changes.  This pull attempts to add that.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -145,8 +145,8 @@ For example, using a named parameter:
 
 [source,java]
 ----
-@Query("where title like :title order by title")
-Page<Book> booksByTitle(String title, PageRequest<Book> pageRequest);
+@Query("where title like :title order by title asc, id asc")
+Page<Book> booksByTitle(String title, PageRequest pageRequest);
 ----
 
 [source,java]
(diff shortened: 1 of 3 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -346,32 +344,36 @@ Sort criteria are provided statically for a repository method by using the `Orde
...
 The following examples work through scenarios where static and dynamic sort criteria are provided to the same method.
 
 [source,java]
 ----
-// Sorts first by type. When type is the same, applies the PageRequest's sort criteria
-Page<User> findByNameStartsWithOrderByType(String namePrefix, PageRequest<User> pagination);
+// Sorts first by type. When type is the same, applies the Order's sort criteria
+Page<User> findByNameStartsWithOrderByType(String namePrefix,
+                                           PageRequest pagination,
+                                           Order<User> sorts);
```
gavinking, line 358: We don't have to do it in this PR, but it would be better IMO, to rewrite the examples in this section to use something other than Query by Method Name.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -611,21 +608,22 @@ Here is an example where an application uses `@Query` to provide a partial query
...
 [source,java]
 ----
-PageRequest<Customer> pageRequest =
-        Order.by(_Customer.yearBorn.desc(),
-                 _Customer.name.asc(),
-                 _Customer.id.asc())
-             .pageSize(25);
+Order<Customer> order = Order.by(_Customer.yearBorn.desc(),
+                                 _Customer.name.asc(),
+                                 _Customer.id.asc());
```
gavinking, line 623: This is like, super-nitpicky and peevey, but at least to my eyes, this is much prettier to look at:
```suggestion
Order<Customer> order = 
        Order.by(_Customer.yearBorn.desc(),
                 _Customer.name.asc(),
                 _Customer.id.asc());
```
But it's up to you.

gavinking (approved):

spec/src/main/asciidoc/repository.asciidoc
njr-11: Yes, that would be nice, but it is definitely out of scope for this PR. It could even be done as a compatible change post v1.0 when we add some good alternatives.

Issue #657, opened 2024-04-05
[TCK Challenge]: ee.jakarta.tck.data.web.validation.Rectangles has two bad methods


starksm64:
### Specification

@Delete, spec/src/main/asciidoc/method-query.asciidoc

### Assertion

Automatic query methods, QBN findBy...

### TCK Version

SNAPSHOT (Locally built)

### Implementation being tested

HIbernate

### Challenge Scenario

Something else

### Full Description

The `ee.jakarta.tck.data.web.validation.Rectangles` repository interface has what looks to be two invalid methods:

```java
@Repository
public interface Rectangles extends DataRepository<Rectangle, String> {
...
    void deleteAllBy();

    @Size(min = 0, max = 3)
    List<Rectangle> findAllByOrderByIdAsc();
}
```

### Additional Context

The `deleteAllBy` method is not a QBN method, and its usage in the TCK implies it should be annotated with `@Delete` to function as an automatic query method that deletes all entities in the repo.

The `findAllByOrderByIdAsc` method fails to parse as a QBN method because there is no predicate between the By and OrderBy, and the current documented QBN grammar does not allow for an empty predicate. It does not seem like there is an approriate QBN method for the given TCK usage to attempt to return all entities and order them by id. It seems like the correct method would be:
```java
    @Size(min = 0, max = 3)
    @Find
    @OrderBy(value = "id")
    List<Rectangle> findAllByOrderByIdAsc();
```

### Is there an existing challenge for this?

- [X] I have searched the existing issues

gavinking:
I wonder if this is just an error in the grammar. Perhaps it was supposed to be:

```
<query-method> ::= <subject> [<predicate>] [<order-clause>]
<subject> ::= (<action> | "find" <find-expression>) [<ignored-text>] "By"
<action> ::= "find" | "delete" | "update" | "count" | "exists"
```

I also wonder why the grammar isn't:

```
<query-method> ::= <subject> [<ignored-text>] <restriction> [<order-clause>]
<subject> ::= <action> | "find" <find-expression>
<restriction> ::= "All" | "By" <predicate>
<action> ::= "find" | "delete" | "update" | "count" | "exists"
```

Then sensible things like `findAllOrderByIdAsc()` and `deleteAll()` would be legal, but insensible things like `deleteAllBy()`, `countBy(), and `findAllBy()` would not.

gavinking:
[...]

PR #658, opened 2024-04-05
improvements to Query by Method Name grammar

gavinking:
In light of #657 I took a close look at this and I propose the following improvements:

1. only `find` queries can have sorting
2. introduce `All` keyword to allow queries with no restriction, resolving #657
3. simplify `OrderBy` to disallow "weird" things
4. rewrite the grammar using a more readable (ANTLR-style) syntax

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -362,8 +362,8 @@ import java.util.Set;
  * must have a method name that begins with one of the following prefixes and
  * must not include the {@code @Find} annotation, {@code @Query} annotation, or
  * any lifecycle annotations on the method or any data access related annotations
- * on the method parameters. Query conditions
- * are determined by the portion of the method name following the {@code By} keyword.</p>
+ * on the method parameters. Query conditions are determined by the portion of
+ * the method name following the {@code By} keyword.</p>
  *
  * <table id="methodNamePrefixes" style="width: 100%">
  * <caption><b>Query By Method Name</b></caption>
@@ -373,25 +373,21 @@ import java.util.Set;
  * <td style="vertical-align: top; width: 65%"><b>Example</b></td>
  * </tr>
  *
- * <tr style="vertical-align: top"><td>{@code countBy}</td>
+ * <tr style="vertical-align: top"><td>{@code count}</td>
  * <td>counts the number of entities</td>
  * <td>{@code countByAgeGreaterThanEqual(ageLimit)}</td></tr>
  *
- * <tr style="vertical-align: top; background-color:#eee"><td>{@code deleteBy}</td>
+ * <tr style="vertical-align: top; background-color:#eee"><td>{@code delete}</td>
  * <td>for delete operations</td>
  * <td>{@code deleteByStatus("DISCONTINUED")}</td></tr>
  *
- * <tr style="vertical-align: top"><td>{@code existsBy}</td>
+ * <tr style="vertical-align: top"><td>{@code exists}</td>
  * <td>for determining existence</td>
  * <td>{@code existsByYearHiredAndWageLessThan(2022, 60000)}</td></tr>
  *
- * <tr style="vertical-align: top; background-color:#eee"><td>{@code find...By}</td>
+ * <tr style="vertical-align: top; background-color:#eee"><td>{@code find}</td>
  * <td>for find operations</td>
  * <td>{@code findByHeightBetween(minHeight, maxHeight)}</td></tr>
- *
- * <tr style="vertical-align: top"><td>{@code updateBy}</td>
- * <td>for simple update operations</td>
- * <td>{@code updateByIdSetModifiedOnAddPrice(productId, now, 10.0)}</td></tr>
  * </table>
  *
  * <p>The conditions are defined by the portion of the repository method name
@@ -601,7 +597,7 @@ import java.util.Set;
  * {@code WithQuarter}, {@code WithSecond}, {@code WithWeek}, {@code WithYear}.
  * </p>
  * <p>
- * Reserved for find...By and count...By: {@code Distinct}.
+ * Reserved for {@code find} and {@code count}: {@code Distinct}.
  * </p>
  * <p>
  * Reserved for updates: {@code Add}, {@code Divide}, {@code Multiply}, {@code Set}, {@code Subtract}.
@@ -625,10 +621,7 @@ import java.util.Set;
  * <h3>Return types for Query by Method Name</h3>
  *
  * <p>The following is a table of valid return types.
- * The <b>Method</b> column shows name patterns for Query by Method Name.
- * For example, to identify the valid return types for a method,
- * {@code findByName}, that accepts a {@code PageRequest} parameter,
- * refer to the row for {@code find...By...(..., PageRequest)}.</p>
+ * The <b>Method</b> column shows name patterns for Query by Method Name.</p>
  *
  * <table style="width: 100%">
  * <caption><b>Return Types for Query by Method Name</b></caption>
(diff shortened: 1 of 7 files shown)


gavinking:
>  I just looked through Spring documentation and don't see an All keyword, but I might not be looking in the right place.

So I went and tried it out. And the answer is kinda murky.

Most Spring repositories have `findAll()` and `deleteAll()` as well as `findAllById()` and `deleteAllById()` because they inherit `CrudRepository` which declares those methods. There are various flavors (overrides and overloads) of these methods defined throughout their zoo of subinterfaces as well. So a Spring user never needs to explicitly write one of those methods.

[Strangely, their `countAll()` is called just plain `count()`.]

But the example code I'm looking at uses `findAll` and `readAll` and `streamAll` in lots of other places as well. I'm not sure if that actually changes any semantics; it's likely just decoration.

On the other hand, Spring rejects a method name like `findAllOrderByNameAsc`, and, quite bizarrely, IMO, wants me to write `findAllByOrderByNameAsc`. 

Anyway, I want to point out that our situation is a bit different. Jakarta Data users are _not_ required to inherit `BasicRepository`, and anyway `BasicRepository ` doesn't have a `deleteAll()` method. So in our case I can imagine a user migrating from Spring who _would_ like to write a `findAll()` or `deleteAll()` method. 

Yes, sure, they can do it with `@Find` or `@Delete`, but, I dunno, the whole QbMN facility just looks _incomplete_ to me if there's no way to express the _simplest possible query_.

spec/src/main/asciidoc/method-query.asciidoc
```diff
@@ -104,35 +104,52 @@ Query methods allow developers to create database queries using method naming co
...
-<property> ::= <identifier> | <identifier> "_" <property>
-<identifier> ::= <word>
-<positive-integer> ::= <digit> { <digit> }
-<order-clause> ::= "OrderBy" { <order-item> } ( <order-item> | <property> )
-<order-item> ::= <property> ("Asc" | "Desc")
+query : find | action
+find : "find" specification order?
+action : ("delete" | "update" | "count" | "exists") restriction
+specification : "All" ignoredText? | limit? ignoredText? "By" predicate
+restriction : "All" ignoredText? | ignoredText? "By" predicate
```
njr-11, line 111: The following is simpler, more flexible, and does not require adding `All` as a keyword. Instead, `All` becomes part of the ignored text. The restriction clause beginning with "By" becomes optional. When omitted it means no restrictions, which for prefixes other than `exists` has the meaning of all, so you can have `count()` or `countAll()` but not `countBy()` because the latter needs at least one condition, `countByNameLike(pattern)`. When using the `exists` prefix, the repository writer can name it more meaningfully, such as `existsAny()` or `existsOne()`, to reflect what it actually does. This avoids having the confusing `existsAll()` method name be something that is recommended by the grammar.

```suggestion
find : "find" limit? ignoredText? restriction? order?
action : ("delete" | "update" | "count" | "exists") ignoredText? restriction?
restriction : "By" predicate
```

njr-11 (commented):
> the whole QbMN facility just looks _incomplete_ to me if there's no way to express the _simplest possible query_.

Yes, I think you are right.  I proposed a change to your improvements to the grammar that gets this working without making `All` into a keyword.

spec/src/main/asciidoc/method-query.asciidoc
gavinking: The main reason I did not do it that way was I was a bit uncomfortable with allowing just plain `List<E> find()` and `void delete()` without any other qualification. But I guess it's OK. It simplifies the grammar. And this is an extension anyway.

But, wait, another question is: why is `update` an allowed action in this grammar?? I don't think we actually have update queries.

spec/src/main/asciidoc/method-query.asciidoc
```diff
@@ -104,35 +104,51 @@ Query methods allow developers to create database queries using method naming co
...
-<condition> ::= <property> ["IgnoreCase"] ["Not"] [<operator>]
-<operator> ::= "Contains" | "EndsWith" | "StartsWith" | "LessThan"| "LessThanEqual" | "GreaterThan" | "GreaterThanEqual" | "Between" | "Like" | "In" | "Null" | "True" | "False"
-<property> ::= <identifier> | <identifier> "_" <property>
-<identifier> ::= <word>
-<positive-integer> ::= <digit> { <digit> }
-<order-clause> ::= "OrderBy" { <order-item> } ( <order-item> | <property> )
-<order-item> ::= <property> ("Asc" | "Desc")
+query : find | action
+find : "find" limit? ignoredText? restriction? order?
+action : ("delete" | "update" | "count" | "exists") ignoredText? restriction?
```
gavinking, line 109:
```suggestion
action : ("delete" | "count" | "exists") ignoredText? restriction?
```

spec/src/main/asciidoc/method-query.asciidoc
```diff
@@ -104,35 +104,51 @@ Query methods allow developers to create database queries using method naming co
...
-- `<ignored-text>`: Optional text that does not contain "By".
-- `<predicate>`: Represents the query's condition or filtering criteria, which can include multiple conditions separated by "And" or "Or."
-- `<condition>`: Specifies a property and an operator for the condition.
-- `<operator>`: Defines the operator for the condition, like "Between" or "LessThan." When absent, equality is implied.
-- `<property>`: Represents a property name, which can include underscores for nested properties.
-- `<identifier>`: Represents a word (e.g., property names, action names, etc.).
-- `<positive-integer>`: Represents a whole number greater than zero.
-- `<order-clause>`: Specifies the optional order clause, starting with "OrderBy" and followed by one or more order items.
-- `<order-item>`: Represents an ordered collection of entity attributes by which to sort results, including an optional "Asc" or "Desc" to specify the sort direction.
+- `query`: May be a `find` query, or an `update`, `delete`, `count`, or `exists` operation.
```
gavinking, line 137:
```suggestion
- `query`: May be a `find` query, or a `delete`, `count`, or `exists` operation.
```

spec/src/main/asciidoc/method-query.asciidoc
gavinking: I have removed `update`, and also updated the text to reflect Nathan's change to the grammar.

gavinking:
I have removed a couple of other references to `update` methods.

spec/src/main/asciidoc/method-query.asciidoc
njr-11: > But, wait, another question is: why is `update` an allowed action in this grammar?? I don't think we actually have update queries.

A long time ago the possibility of update was being investigated, but it turned out not to be something that existing Query by Method Name solutions had.  I suppose this was a leftover remnant of that.  Thanks for spotting it!

api/src/main/java/module-info.java
```diff
@@ -638,39 +631,37 @@
...
+ * <tr style="vertical-align: top; background-color:#eee"><td>{@code count}</td>
  * <td>{@code long},
  * <br>{@code int}</td>
  * <td></td></tr>
  *
- * <tr style="vertical-align: top"><td>{@code deleteBy...},
- * <br>{@code updateBy...}</td>
+ * <tr style="vertical-align: top"><td>{@code delete}</td>
  * <td>{@code void},
- * <br>{@code boolean},
```
njr-11, line 649: A boolean return value for delete seems convenient for a query that can only possibly delete one entity, such as deletion by unique identifier, so it would be nice to keep it unless it is causing trouble or inconsistency with JDQL delete, in which case I suppose Jakarta Data providers could allow it on their own rather than as a required type.

njr-11 (approved):

api/src/main/java/module-info.java
gavinking: So, what do we say for other kinds of delete?

In ch5, for JDQL:

> A `delete` statement ... deletes each record which satisfies the restriction imposed by the `where` clause, and returns the number of deleted records to the client.

In the javadoc for `@Delete`:

> The method return type must be `void`, `int`, or `long`.

In the javadoc for `@Query`:

> For `update` or `delete` statements, it must be `void`, `int` or `long`.

So I think we should stay consistent here.

(As usual, it doesn't mean implementations can't experiment with allowing other things. We can always add more flexibility later.)

PR #662, opened 2024-04-05
fix  {@code } inside <pre>

### LINKED ISSUE #661: [Bug]: use of {@code } inside <pre>
### Specification Version

1.0.0-M3

### Bug report

The latest update to IntelliJ renders `{@code }` inside `<pre>` even worse than before, so we need to fix all the javadoc.

### Additional information

_No response_

gavinking:
fixes #661

### DIFF
--- a/api/src/main/java/jakarta/data/Sort.java
+++ b/api/src/main/java/jakarta/data/Sort.java
@@ -41,7 +41,7 @@ import java.util.Objects;
  * <pre>
  * Employee[] findByYearHired(int yearHired, Limit maxResults, Sort&lt;?&gt;... sortBy);
  * ...
- * highestPaidNewHires = employees.findByYearHired(Year.now(),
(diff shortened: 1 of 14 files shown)


njr-11 (approved):
Approving, but it's sad to see our work of switching to `{@code}` to make the source more readable being undone.

gavinking:
We could use `@snippet`.

gavinking:
(I'm not sure if you've seen how it renders in the latest IntelliJ update, so I'm attaching a screenshot.)
Screenshot 2024-04-05 at 6 56 31 PM

njr-11:
> We could use `@snippet`.

Yes, let's plan to do that next version or whatever version we don't need to still support Java 17.
For now, let's just get this merged. It is already potentially conflicting with other JavaDoc fixes I've made locally, although I think one of them with PageRequest you already covered under this.

gavinking:
> For now, let's just get this merged. 

OK I will merge it now.

PR #665, opened 2024-04-05
Fix more javadoc broken by the PageRequest/Sort decoupling

njr-11:
Another pull to fix example code that was broken by PageRequest/Sort decoupling.
and I also spotted a few places where we have bad example code that isn't allowing for a consistent ordering when using pagination.

### DIFF
--- a/api/src/main/java/jakarta/data/page/CursoredPage.java
+++ b/api/src/main/java/jakarta/data/page/CursoredPage.java
@@ -120,7 +120,7 @@ import java.util.NoSuchElementException;
  * &#64;OrderBy("birthYear")
  * &#64;OrderBy("id")
  * CursoredPage&lt;Customer&gt; getTopBuyers(int minOrders, float minSpent,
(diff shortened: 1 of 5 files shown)


api/src/main/java/jakarta/data/page/CursoredPage.java
```diff
@@ -120,7 +120,7 @@
  * &#64;OrderBy("birthYear")
  * &#64;OrderBy("id")
  * CursoredPage&lt;Customer&gt; getTopBuyers(int minOrders, float minSpent,
- *                                     PageRequest pageRequest, Order&lt;Customer&gt;} sort);
+ *                                     PageRequest pageRequest);
```
njr-11, line 123: In this case, we overdid it and added a `Order` parameter that doesn't make sense -- it is already ordered by its unique identifier, making any lesser priority sorting unreachable.

gavinking (approved):

gavinking:
I think we're good to go ahead and merge this, no?

PR #666, opened 2024-04-05
Static metamodel examples show bad coding practices

njr-11:
I noticed this while reviewing the PageRequest/Sort decoupling, but it is not actually related so I opened a separate PR for it.

The examples we give for static metamodel classes in the javadoc have some bad practices.  In one example, we are initializing all the fields and unnecessarily marking them `volatile` rather than `final`, which results in unsafe code that could accidentally be overwritten at run time with no benefit. I don't want anyone copy/pasting from that - either when they write an annotation processor or for direct use.

In another example, we aren't even using our own constants: `new SortableAttributeRecord&lt;&gt;("ssn");`
We already defined an `SSN` constant for the `"ssn"` value only a few lines above, and we should be using it here.

Also, these could be a lot cleaner with implicit `public static final` if using an interface class rather than a concrete class.

This pull makes the above changes.

One improvement that I didn't make, because it would go beyond being a bug fix and require an API change is to switch
`SortableAttribute<Person> ssn = new SortableAttributeRecord<>(SSN);` // uses class in impl package
to
`SortableAttribute<Person> ssn = SortableAttribute.of(SSN);`
so that we could avoid externalizing the impl package entirely. That would be something to think about for future.

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
+++ b/api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
@@ -35,6 +35,7 @@ import jakarta.data.Sort;
  *
  * <p>Jakarta Data defines the following conventions for static metamodel classes:</p>
  * <ul>
+ * <li>The metamodel class can be an interface or concrete class.</li>
  * <li>The name of the static metamodel class should consist of underscore ({@code _})
  *     followed by the entity class name.</li>
  * <li>Fields of type {@code String} should be named with all upper case.</li>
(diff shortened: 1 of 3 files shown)


gavinking:
The spec talks about an entity having a "metamodel class". If we want to change it to an interface, I don't see naively anything wrong with that, but we had better say it in the spec.

njr-11:
> The spec talks about an entity having a "metamodel class". If we want to change it to an interface, I don't see naively anything wrong with that, but we had better say it in the spec.

That's a good point. The spec needs clarification here so that it is understood whether concrete classes, abstract classes, interface classes (or even record classes or enum classes?) are permitted for the metamodel. Unless I'm missing more specific language in the spec, the word class currently leaves it ambiguous.

gavinking:
> Unless I'm missing more specific language in the spec, the word class currently leaves it ambiguous.

To me a "class" is a `class` and an `interface` is not a class.

In JPA they have always been classes, but that's because back then interfaces couldn't have static members.

But now we have a clean slate here and I think you're right that we should say that for Jakarta Data they are interfaces.

njr-11:
Added clarification in https://github.com/jakartaee/data/pull/666/commits/621383eb3396d4e7f3c5dee54959877e9a440854

gavinking (approved):

otaviojava:
I am fine with both once it is okay with Jakarta Persistence. It is acceptable to me.

PR #667, opened 2024-04-05
Add test case for custom entity types and fix query methods for validation

### LINKED ISSUE #657: [TCK Challenge]: ee.jakarta.tck.data.web.validation.Rectangles has two bad methods

### LINKED ISSUE #660: Missing test case for custom entity types
It looks to me like they are cheating and relying on `@Repository(provider =` to be assigned the `Directory` repository,
```
```

That means the coverage here isn't as good as it should be.  We should leave this in place because it's also a valid scenario and create a second repository for `Person` that omits `provider = ...` and instead relies on the type of Entity annotation.

_Originally posted by @njr-11 in https://github.com/jakartaee/data/issues/647#issuecomment-2038298851_

KyleAure:
- Fixes #657 
- Fixes #660


tck/src/main/java/ee/jakarta/tck/data/common/cdi/Person.java
```diff
@@ -17,11 +17,7 @@
 
 /**
  * A test entity that will be persisted to a repository.
- * Uses the custom {@code @PersonEntity} annotation.
- * 
- * @see ee.jakarta.tck.data.common.cdi.PersonEntity
  */
-@PersonEntity
 public class Person {
```
njr-11, line 21: Should we put the NoSQL and/or Persitence `@Entity` onto this to help test that Jakarta Data providers against which the TCK is run do not try to handle it because its repository specifically asks for the TCK mock Data provider `provider = Directory.PERSON_PROVIDER`?

KyleAure:
Resolved merge conflicts.

mswatosh (approved):

njr-11 (approved):

PR #669, opened 2024-04-07
improve presentation of info on Query by Method Name

### LINKED ISSUE #576: [clarification]: query by method name method parameters

gavinking:
Fixes #576.

### DIFF
--- a/spec/src/main/asciidoc/method-query.asciidoc
+++ b/spec/src/main/asciidoc/method-query.asciidoc
@@ -57,6 +57,8 @@ Therefore, these alternative approaches are strongly preferred for newly-written
 
 A Jakarta Data provider is required to support the Query by Method Name extension in Jakarta Data 1.0.
 
+NOTE: A Jakarta Data provider backed by a key-value or wide-column datastore is not required to support Query by Method Name.
+
 WARNING: This functionality is considered deprecated, and the requirement that a provider support the Query by Method Name extension will be removed in a future version of Jakarta Data.
 
 == Query by Method Name
@@ -82,8 +84,8 @@ The parsing of query method names follows a specific format:
 
 - The method name consists of the subject, the predicate, and optionally the order clause.
 - The subject begins with the action (such as `find` or `delete`) and is optionally followed by an expression (for example, `First10`), followed by any number of other characters, followed by `By`.
-- The predicate defines the query's condition or filtering criteria, where multiple conditions are delimited by `And` or `Or`. For example, `PriceLessThanAndNameLike`. Each condition consists of a property name, optionally followed by `IgnoreCase` (for text properties), optionally followed by `Not`, optionally followed by an operator such as `LessThan` or `Like`. Absence of the operator implies the equality condition.
-- The order clause, which is optional, begins with `OrderBy` and is followed by one or more ordered pairings consisting of entity attribute name and direction of sorting (`Asc` or `Desc`). The direction can be omitted on the final property, in which case `Asc` is implied. Ordering of results is performed by sorting the first entity attribute according to its respective direction, where results that sort to the same position are subsequently sorted by the second entity attribute according to its respective direction, and so forth.
+- The predicate defines the condition or filtering criteria, where multiple conditions are delimited by `And` or `Or`. For example, `PriceLessThanAndNameLike`. Each condition consists of a property name, optionally followed by `IgnoreCase` (for text properties), optionally followed by `Not`, optionally followed by an operator such as `LessThan` or `Like`. Absence of the operator implies the equality condition.
+- The order clause, which is optional, begins with `OrderBy` and is followed by one or more ordered pairings consisting of entity attribute name and direction of sorting (`Asc` or `Desc`). The direction can be omitted if there is only one property, in which case `Asc` is implied. Ordering of results is performed by sorting the first entity attribute according to its respective direction, where results that sort to the same position are subsequently sorted by the second entity attribute according to its respective direction, and so forth.
 - The method name is formed by combining the subject, predicate, and order clause, in that order.
 
 NOTE: This specification uses the terms subject and predicate in a way that aligns with industry terminology rather than how they are defined in English grammar.
@@ -92,11 +94,11 @@ Queries can also handle entities with relation attributes by specifying the rela
 
 Example query methods:
 
-- `findByName(String name)`: Find entities by the 'name' property.
-- `findByAgeGreaterThan(int age)`: Find entities where 'age' is greater than the specified value.
-- `findByAuthorName(String authorName)`: Find entities by the 'authorName' property of a related entity.
-- `findByCategoryNameAndPriceLessThan(String categoryName, double price)`: Find entities by 'categoryName' and 'price' properties, applying an 'And' condition.
-- `findByNameLikeOrderByPriceDescIdAsc`: Find entities by matching the 'name' property against a pattern, sorting the results by 'price' in descending order, and sorting results with the same 'price' by the 'id' in ascending order.
+- `findByName(String name)` -- find entities by the `name` property.
+- `findByAgeGreaterThan(int age)` -- find entities where `age` is greater than the specified value.
+- `findByAuthorName(String authorName)` -- find entities by the `authorName` property of a related entity.
+- `findByCategoryNameAndPriceLessThan(String categoryName, double price)` -- find entities by `categoryName` and `price` properties, applying an `And` condition.
+- `findByNameLikeOrderByPriceDescIdAsc` -- find entities by matching the `name` property against a pattern, sorting the results by `price` in descending order, and sorting results with the same `price` by the `id` in ascending order.
 
 === BNF Grammar for Query Methods
 
@@ -132,6 +134,8 @@ order : "OrderBy" (property | orderItem+)
 orderItem : property ("Asc" | "Desc")
 ----
 
+Quoted names are considered case-sensitive keywords.
+
 Explanation of the BNF elements:
 
 - `query`: May be a `find` query, or a `delete`, `count`, or `exists` operation.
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/method-query.asciidoc
```diff
@@ -57,6 +57,8 @@ Therefore, these alternative approaches are strongly preferred for newly-written
 
 A Jakarta Data provider is required to support the Query by Method Name extension in Jakarta Data 1.0.
 
+NOTE: A Jakarta Data provider backed by a key-value or wide-column datastore is not required to support Query by Method Name.
```
njr-11, line 60: This is probably more a question for @otaviojava , but this seems wrong. Surely these database types must be capable of at least some subset of it, otherwise how can they be supported at all in Jakarta Data?

spec/src/main/asciidoc/method-query.asciidoc
```diff
@@ -151,152 +155,139 @@ Explanation of the BNF elements:
...
-====
-Jakarta Data implementations must support the following list of Query by Method Name keywords, except where indicated for a database type. A repository method must raise `java.lang.UnsupportedOperationException` or a more specific subclass of the exception if the database does not provide the requested functionality.
+An implementation of Query by Method Name must support the following keywords.
 
-[cols="15,~,~,25"]
+.Query by Method Name Keywords
+[cols="15,~,~"]
 |===
-| Keyword |Description | Method signature Sample | Not Required For
+| Keyword | Description | Method signature example
```
njr-11, line 183: I don't agree with getting rid of the place to document the NoSQL imitations.

spec/src/main/asciidoc/method-query.asciidoc
```diff
@@ -151,152 +155,139 @@ Explanation of the BNF elements:
...
 |Returns `true` if at least one record satisfies the restriction or if there is at least one record in the database when there is no restriction.
-|Key-value, Wide-Column
 |===
 
-[NOTE]
-====
-The "Not Required For" column indicates the database types for which the respective keyword is not required or applicable.
-====
-Jakarta Data implementations must support the following list of Query by Method Name keywords, except where indicated for a database type. A repository method must raise `java.lang.UnsupportedOperationException` or a more specific subclass of the exception if the database does not provide the requested functionality.
+An implementation of Query by Method Name must support the following keywords.
```
njr-11, line 178: Don't get rid of this. It is what allows for NoSQL implementations that are incapable of a subset of keywords, such as `Or`. The TCK relies on this for accurate testing.

njr-11 (changes requested):

spec/src/main/asciidoc/method-query.asciidoc
gavinking: Well that's what we have in the table right now. Every action is marked "Not required" for these kinds of databases. It seemed strange to me too.

spec/src/main/asciidoc/method-query.asciidoc
gavinking: I think this a good change. There was only one kind of limitation here, which is much more compactly written in a single sentence than taking up that extra space in the table.

spec/src/main/asciidoc/method-query.asciidoc
gavinking: I didn't get rid of anything. I moved it.

gavinking:
@njr-11 I recommend compiling the PDF and reading it from top to bottom, the diff is probably misleading.

spec/src/main/asciidoc/method-query.asciidoc
otaviojava: The `key-value` can only support any operation related to the key, with some exceptions. Otherwise, it will return an `UnsupportedOperationException`; the same generally happens with most wide-columns, such as Hbase, ScyllaDB, and Cassandra. 

This message was sent because all possible key operations were converted at the BasicRepository. Therefore, we decided to proceed in this direction.

⚠️ Cassandra and ScyllaDB can support if including the secondary index, but, it it not like

spec/src/main/asciidoc/method-query.asciidoc
otaviojava: We have already informed that the database that does not support it will throw an UnsupportedOperationException.

spec/src/main/asciidoc/method-query.asciidoc
gavinking: > Don't get rid of this. It is what allows for NoSQL implementations that are incapable of a subset of keywords, such as Or. 

FTR, `Or` is not on the list of things which are only supported by a subset of databases. It was listed as 

> Not required for: Key-value, Wide-Column

But so was `findBy`, so that wasn't a meaningful additional restriction.

I hope you see how the presentation in terms of tables was making this more confusing.

gavinking: > Not required for: Key-value, Wide-Column

Indeed, this was listed for _every single keyword_.

spec/src/main/asciidoc/method-query.asciidoc
njr-11: > all possible key operations were converted at the BasicRepository. Therefore, we decided to proceed in this direction.

It's probably worth pointing out that since then a number of operations have been removed from `BasicRepository`, including some that were id-based, so that might not be true anymore.  In any case I would say it seems wrong because Jakarta Data does not force the use of `BasicRepository`.

gavinking: OK, so then we need to sort out what are the real requirements here.

I'm not sure that such discussions should block this PR, since that's a longer conversation, which will be facilitated by having the current requirements presented in a better/clearer way.

gavinking: OTOH, I guess I would say that if key-value databases only support key-based operations, then what is the value in having QbMN for those databases? They can already do everything possible with `@Delete` and `@Find`.

spec/src/main/asciidoc/method-query.asciidoc
njr-11: > I think this a good change. There was only one kind of limitation here, which is much more compactly written in a single sentence than taking up that extra space in the table.

I think the compactness is an inferior way to convey the information to the user that will be more difficult for them to understand, even more so as more restrictions are identified with NoSQL database types that need to be documented.

njr-11 (approved):
Some of the changes in this pull are nice, some I don't like and think make the spec more difficult to understand. Much of it is a question of preference. Given that so much of it is about NoSQL, I'll let Otavio make the call on these things.

gavinking:
And I don't see how it's more difficult to understand (or easy to miss) the following admonitions:

Screenshot 2024-04-08 at 6 16 24 PM
Screenshot 2024-04-08 at 6 16 35 PM

njr-11 (changes requested):
> And I don't see how it's more difficult to understand (or easy to miss) the following admonitions:

That may be so, but I just realized when look at those examples, you have also changed the requirement allowing these patterns to be rejected at compile time, which breaks the TCK, so now I've got to reverse my approval on this one.

I thought the idea here was that these sort of methods, unlike some of the others, being dependent on what a database can do, would be something to be determined at run time based on which type and version of the database is observed to be present.

gavinking:
A Jakarta Data provider knows what kind of datastore it targets, and it knows the capabilities of that datastore. There's no good reason it can't reject a repository method at compile time if the repository method makes use of capabilities that the provider/datastore doesn't have.

> I thought the idea here was that these sort of methods, unlike some of the others, being dependent on what a database can do, would be something to be determined at run time based on which type and version of the database is observed to be present.

I don't see why you would think that.

> breaks the TCK

So we fix it. Horses go before carts.

njr-11:
> > breaks the TCK
> 
> So we fix it. Horses go before carts.

Unfortunately, this is a type of problem that we haven't been able to find a solution for in the TCK. Failing at compile time breaks the whole TCK application. The only way we got around that last time was removing TCK coverage for a single scenario and rationalizing that it did not impact other usage because we could continue trapping for UnsupportedOperationException in the other database type/version category of places -- but that's changed under this PR.  I'm not willing to put the TCK into an unrecoverable state at this point in the release cycle.  I don't even see how the UnsupportedOperationException impacts your implementation here. This is all NoSQL stuff that would be raising it. If we find a solution next version, the spec can be compatibly updated (changes to exception paths are allowed) to allow compile time failure then. This is a case where there is no clear benefit to making the change but obvious problems with allowing it in.

gavinking:
> I don't even see how the `UnsupportedOperationException` impacts your implementation here. 

It doesn't impact _my_ implementation at all. But @otaviojava has an annotation-processor-based implementation which targets NoSQL.

gavinking:
Anyway, fine, I'll back out that change for now, but in future the TCK is going to need well-defined profiles for each kind of database.

njr-11 (approved):

gavinking:
> I'll back out that change for now

Done by 2e66b2635e5f3ac798c07d53727ff20caf0370af.

gavinking:
@otaviojava?

spec/src/main/asciidoc/method-query.asciidoc
otaviojava: That is my opinion as well.
It does not make sense to key-value database this feature.

otaviojava (approved):

PR #671, opened 2024-04-07
fix incorrect implementation of PageRequest.page()

gavinking:
this operation should not propagate the `Cursor` value

### DIFF
--- a/api/src/main/java/jakarta/data/page/Pagination.java
+++ b/api/src/main/java/jakarta/data/page/Pagination.java
@@ -22,7 +22,7 @@ import java.util.Optional;
 /**
  * Built-in implementation of PageRequest.
  */
-record Pagination(long page, int size, Mode mode, Cursor type, boolean requestTotal) implements PageRequest {
+record Pagination(long page, int size, Mode mode, Cursor keys, boolean requestTotal) implements PageRequest {
 
     Pagination {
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/data/page/Pagination.java
```diff
@@ -97,20 +97,20 @@ public String toString() {
...
+        if (keys != null) {
+            s.append(", ").append(keys.size()).append(" keys");
         }
         return s.append("}").toString();
     }
 
     @Override
     public PageRequest page(long pageNumber) {
-        return new Pagination(pageNumber, size, mode, type, requestTotal);
+        return new Pagination(pageNumber, size, Mode.OFFSET, null, requestTotal);
```
gavinking, line 108: Here is the important change.

gavinking:
Actually I'm just now realizing that there's a huge confusion about what these copying mutators actually mean for this class.

Currently, a `PageRequest` mixes together "navigational" methods, and also builder-type methods. That's actually really bad.

We need to split some of these operations onto a `PageRequestBuilder`, so that it's obvious which operations are capable of producing a `PageRequest` in an inconsistent state.

gavinking:
Closing this in favor of #672 because there is a much deeper problem here.

PR #674, opened 2024-04-08
next() and previous() are confusing the purpose of PageRequest

### LINKED ISSUE #672: [Bug]: PageRequest doesn't know whether it's a Builder
### Specification Version

1.0.0-M3

### Bug report

`PageRequest` has multiple operations capable of producing an object in an inconsistent state. The most dangerous of these is `page(long)`, which looks harmless enough but actually results in something completely broken if you call it an a `PageRequest` with a `Cursor`. And there's no warning of this in the Javadoc, even. On the other hand, it also has operations like `next()` which are written to be called by the application. 

It is in no way obvious to the user that these operations `page(long)` and `next()` are so different in nature. So much so that even I have tripped up twice over the difference and I am far from a casual user of this API.

We need to remove these dangerous operations to a `Builder`-type object, or do what I suggested a while ago, and expose the constructor of `Pagination` to Jakarta Data implementations and simply remove the "dangerous" methods. (This is the simplest solution.)

### Additional information

_No response_

njr-11:
Covers some of the issues behind #672 by removing next() and previous() from PageRequest. Those methods are causing confusion about the purpose of PageRequest.  PageRequest is already a builder that collects the configuration for requesting a next page.  It is out of place for it to have operations on it that relate to navigation.  It is not even capable of performing those operations correctly as evidenced by its own JavaDoc which states those operations fail and don't work with cursor-based pagination. They aren't even safe for offset-based pagination because they will keep telling you there are more and more next pages to request with ever increasing page numbers even when no such pages exist.  The next() method is highly likely to lead some users to write code with an infinite loop.  The correct way to perform page traversal is with Page.hasNext()/hasPrevious() and Page.nextPageRequest() and previousPageRequest(). These methods properly account for both types of pagination and are built to inform the user when there cannot be a next or previous page, and, if the user neglects to check if there can be a next/previous page, to fail instead of looping infinitely.

### DIFF
--- a/api/src/main/java/jakarta/data/page/CursoredPage.java
+++ b/api/src/main/java/jakarta/data/page/CursoredPage.java
@@ -192,10 +192,9 @@ public interface CursoredPage<T> extends Page<T> {
      * when matching entities are added prior to the first page and the
      * previous page is requested) by assigning a page number of {@code 1}
      * to such pages. This means that there can be multiple consecutive pages
-     * numbered {@code 1} and that
-     * {@code currentPage.previousPageRequest().next().page()}
-     * cannot be relied upon to return a page number that is equal to the
-     * current page number.</p>
+     * numbered {@code 1} and that navigating to the previous page and then
+     * forward again cannot be relied upon to return a page number that is
+     * equal to the current page number.</p>
      *
      * @return pagination information for requesting the previous page.
      * @throws NoSuchElementException if the current page is empty or if
--- a/api/src/main/java/jakarta/data/page/Page.java
+++ b/api/src/main/java/jakarta/data/page/Page.java
@@ -109,8 +109,8 @@ public interface Page<T> extends Iterable<T> {
 
 
     /**
-     * Returns a request for the {@linkplain PageRequest#next() next} page if
-     * {@link #hasNext()} indicates there might be a next page.
+     * Returns a request for the next page if {@link #hasNext()} indicates there
+     * might be a next page.
      *
      * @return a request for the next page.
      * @throws NoSuchElementException if it is known that there is no next page.
@@ -122,8 +122,8 @@ public interface Page<T> extends Iterable<T> {
 
 
     /**
-     * <p>Returns a request for the {@link PageRequest#previous() previous} page,
-     * if {@link #hasPrevious()} indicates there might be a previous page.</p>
+     * <p>Returns a request for the previous page, if {@link #hasPrevious()}
+     * indicates there might be a previous page.</p>
      *
      * @return a request for the previous page.
      * @throws NoSuchElementException if it is known that there is no previous page.
--- a/api/src/main/java/jakarta/data/page/PageRequest.java
+++ b/api/src/main/java/jakarta/data/page/PageRequest.java
@@ -197,40 +197,6 @@ public interface PageRequest {
      */
     boolean requestTotal();
 
-
-    /**
-     * <p>Returns the {@code PageRequest} requesting the next page if
-     * using offset pagination.</p>
-     *
-     * <p>If using cursor-based pagination, traversal of pages must only be done
-     * via the {@link CursoredPage#nextPageRequest()},
-     * {@link CursoredPage#previousPageRequest()}, or
-     * {@linkplain CursoredPage#cursor(int) cursor},
-     * not with this method.</p>
-     *
-     * @return The next PageRequest.
-     * @throws UnsupportedOperationException if this {@code PageRequest}
-     *         has a {@link PageRequest.Cursor Cursor}.
-     */
-    PageRequest next();
-
-    /**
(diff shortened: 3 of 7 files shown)


njr-11:
> I like this approach, but it only solves some of the issues.

That's a fair point. I updated the description to state that it addresses some of the issues from 672 rather than claiming to fix it entirely.

gavinking:
I'm fine with removing this method, but I think we need to sit down and rethink this from first principles.

otaviojava:
> I'm fine with removing these methods, but I think we need to sit down and rethink this from first principles.

I like this approach.

Thus, can we move this PR forward?

otaviojava (approved):

otaviojava:
> I'm thinking, for example, that we could remove all (copy-on-write) mutating methods from `PageRequest`, and instead expose constructors for a `PageRequestRecord` or whatever as part of an SPI package.
> 
> But that leaves disagreement over whether a `PageRequest` must be constructed in a consistent state, with a page number that agrees with the cursor. If Nathan insists that this is unacceptable, then we will need two different kinds of `PageRequest`, one with only a page number (an `OffsetBasedPageRequest`), and one with a cursor (a `CursoredPageRequest`) but no page number.

The specialization is what I do have in my mind.

otaviojava:
> Can we move the discussion back to 672? This current PR is only mean to address the portion with next()/previous() which I was hoping to make progress on first because it will be helpful to have those out of the way when solving the rest of the problem.

I agree, @gavinking can we merge this one and return back this discussion to 672?

gavinking:
I guess I would:

1. remove `next()`, `previous()`, `page(long)`, `size(int)`, `afterKey/Cursor()` ,`beforeKey/Cursor()` as instance methods,
2. introduce a different way to construct a `PageRequest` at a given `Cursor`/`Key` value (either static method or constructors), preferably in an SPI package, and
3. mandate that callers who construct a `PageRequest` based on a `Cursor` or `Key` supply a meaningful/consistent page number.

But if 3 is unacceptable, then I would go with Otavio's suggestion.

gavinking:
> remove `next()`, `previous()`

Well, actually, no, on second thoughts, _these_ operations are actually fine. They _do_ return instances in perfectly consistent states.

> I agree, @gavinking can we merge this one and return back this discussion to 672?

Well the thing is that it depends how we ultimately decide to resolve this. We don't necessarily need to remove `next()` and `previous()`. If we assume that page numbers are always meaningful, even when curser-based pagination is in use, then these operations are OK. It's `page(long)` that's the really broken one.

We only need to remove them if we want to do the split as you suggest.

otaviojava:
> > remove `next()`, `previous()`
> 
> Well, actually, no, on second thoughts, _these_ operations are actually fine. They _do_ return instances in perfectly consistent states.
> 
> > I agree, @gavinking can we merge this one and return back this discussion to 672?
> 
> Well the thing is that it depends how we ultimately decide to resolve this. We don't necessarily need to remove `next()` and `previous()`. If we assume that page numbers are always meaningful, even when curser-based pagination is in use, then these operations are OK. It's `page(long)` that's the really broken one.
> 
> We only need to remove them if we want to do the split as you suggest.

Hum, I got your point.

I will create a PR using this one as a base.

njr-11:
> > remove `next()`, `previous()`
> 
> Well, actually, no, on second thoughts, _these_ operations are actually fine. They _do_ return instances in perfectly consistent states.

No, they don't return instances when cursor-based pagination is requested,

>     * @throws UnsupportedOperationException if this {@code PageRequest}
>     *         has a {@link PageRequest.Cursor Cursor}.
>     */
>    PageRequest next();

And while they do return real instances when offset pagination is requested, those instances are misleading and bad programming practices. You'll notice that in this PR we don't have a single example in the spec recommending their use, nor does the TCK use them at all. The JavaDoc doesn't have a single example using them either, and has only a single reference that mentioned them in relation to a pattern with unexpected behavior.  This is a case where the validity of thing doesn't automatically make it a good idea or something that is useful.  These methods are at best a distraction and at worse will mislead users into writing bad code.

gavinking:
> No, they don't return instances when cursor-based pagination is requested

They return perfectly consistent offset-based `PageRequest`s. They're not *cursor-based*, that's true, but they're nonbroken.

gavinking:
But OK, fine, I'm not going to stand in the way of removing them, I'm just saying it doesn't really get us any closer to solving the issues I actually care about here.

otaviojava:
> But OK, fine, I'm not going to stand in the way of removing them, I'm just saying it doesn't really get us any closer to solving the issues I actually care about here.

We agree with that:

I created this work here:

https://github.com/jakartaee/data/pull/675

gavinking:
I believe that this can be closed, since these changes were rolled into #679.

PR #675, opened 2024-04-08
Create an option with specilization

otaviojava:
# Changes

This change breaks the API with specialization for both offset and cursor.

* Creates two specializations for PageRequest
* Update Page and CursoredPage to return it respective specialization
* Create a Slice interface to get the common behavior of both pagination structures.

⚠️ the CI is broken; please let me know if I can proceed with this idea.


api/src/main/java/jakarta/data/page/Slice.java
```diff
@@ -0,0 +1,85 @@
+package jakarta.data.page;
+
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+public interface Slice<T> extends Iterable<T> {
```
gavinking, line 8: Please let's not bring back `Slice`. What do we need it for?

api/src/main/java/jakarta/data/page/Pagination.java
```diff
@@ -22,7 +22,7 @@
 /**
  * Built-in implementation of PageRequest.
  */
-record Pagination(long page, int size, Mode mode, Cursor type, boolean requestTotal) implements PageRequest {
+record Pagination(long page, int size, Mode mode, Cursor type, boolean requestTotal) implements PageRequest, CursoredPageRequest, OffsetBasedPageRequest {
```
gavinking, line 25: I would split this into `CursoredPageRequestRecord` and `OffsetBasedPageRequestRecord`.

api/src/main/java/jakarta/data/page/OffsetBasedPageRequest.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * <p>Creates a new page request with the same pagination information,
+     * but with the specified page number.</p>
+     *
+     * @param pageNumber The page number
+     * @return a new instance of {@code PageRequest}.
+     *         This method never returns {@code null}.
+     */
+    PageRequest page(long pageNumber);
```
gavinking, line 31: I would also pull down `size(int)` if this is the route we're going.

api/src/main/java/jakarta/data/page/Slice.java
njr-11: I think Otavio is proposing to get rid of page numbers entirely with cursor-based pagination.  I think that would be a mistake.  While some users of cursor-based pagination will not want or care about page numbers, other users of cursor-based pagination will.  The spec is currently very lazily written with respect to page numbers when using cursor-based pagination, which is my fault, and says not to rely on them.  Given recent discussion, it sounds like we will need to add more guarantees in the spec to ensure they can be useful when they are wanted.  Getting rid of them completely doesn't seem like a good idea.

api/src/main/java/jakarta/data/page/OffsetBasedPageRequest.java
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.page;
+
+public interface OffsetBasedPageRequest extends PageRequest {
```
gavinking, line 20: I would go with just `OffsetPageRequest`.

gavinking:
So this looks OK.

We need to weigh the value of:

> Creates two specializations for `PageRequest`

I guess you're kinda selling me on the idea, though I was sorta trying to avoid it in order to minimize the blast radius.

The question is: are the operations introduced by the subtypes really critical things we need to expose to the user. I guess `int page()` is needed.

api/src/main/java/jakarta/data/page/Slice.java
gavinking: > I think Otavio is proposing to get rid of page numbers entirely with cursor-based pagination. I think that would be a mistake. 

Yeah I guess it's not what I would have done.

But if we have page numbers, they have to be consistent, at least in the absence of phantom reads.

otaviojava: I need a common interface for both OffSetPage and CursoredPage.

I will use Page as the main interface instead of Slice.

api/src/main/java/jakarta/data/page/Pagination.java
otaviojava: I like this.

api/src/main/java/jakarta/data/page/Slice.java
otaviojava: @gavinking @njr-11 I returned the method, and I also renamed it to OffSetPage

njr-11 (changes requested):
Otavio, I think you might be moving ahead too quickly with implementation here before we have figured out what we are trying to design or solve here. We should go back to the issue that Gavin opened and get the discussion to a point where we all have a clear understanding of the problem and what we want the solution to be.  That should include some examples of how we see the API being used under a proposed solution vs how it is currently used.  I don't want you to have to expend effort on a solution that we might not all be on board with.

otaviojava:
@njr-11 I still working on the Gavin problem.

```java
        var offsetPageRequest = PageRequest.ofPage(10);
        long page = offsetPageRequest.page();//it does not have cursor
        var cursoredPageRequest = PageRequest.ofPage(10).afterKey(1, 2, 3);
        PageRequest.Cursor cursor = cursoredPageRequest.cursor();//it does not have page number
```

I only updated the pagination to have this specialization as well, otherwise, CursoredPage breaks:

```java

        OffSetPage<String> offsetPage = //from repository;
        CursoredPage<String> cursoredPage = //from repository;;
        OffsetPageRequest offsetPageRequest1 = offsetPage.nextPageRequest();
        CursoredPageRequest cursoredPageRequest1 = cursoredPage.nextPageRequest();

```

PR #678, opened 2024-04-08
give cursor TCK tests a meaningful page number rather than defaulting to 1

### LINKED ISSUE #670: [TCK Challenge]: testCursoredPageOf7FromCursor() and testCursoredPageWithoutTotalOf9FromCursor()
### Specification

`PageRequest.afterKey()`

### Assertion

line 977 of `EntityTests`

### TCK Version

1.0.0-M3

Hibernate

### Challenge Scenario

Claims that a test is not portable or depends on a particular implementation.

### Full Description

The assertion here makes little sense and depends on an inconsistently constructed `PageRequest`. Thus, it appears to depend on implementation details of a particular product.

### Additional Context

The following line:

```java
PageRequest middle7 = PageRequest.ofSize(7) .afterKey((short) 5, 5L, 26L)
```

produces a `PageRequest` whose internal state is inconsistent with the data in the database. It manifestly represents a "middle" page, but has a page number of `1`.

Thus, later assertions can fail (in particular the one at line 977).

Now, sure, a certain implementation of Jakarta Data might pass these assertions _by luck_, but nothing in the specification or API documentation requires that an implementation be this lucky.

njr-11:
fixes #670

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
@@ -941,7 +941,7 @@ public class EntityTests {
         //                                                                                  ^^^^^ next page ^^^^
 
         Order<NaturalNumber> order = Order.by(Sort.asc("floorOfSquareRoot"), Sort.desc("id"));
-        PageRequest middle7 = PageRequest.ofSize(7)
+        PageRequest middle7 = PageRequest.ofPage(4).size(7)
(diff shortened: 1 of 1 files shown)


gavinking:
OK, perfect!

Now we're down to one lonely TCK test failure on our end. That last one is something weird that's going to require debugging, but I doubt it's a bug in our implementation.

gavinking (approved):

gavinking:
With this change, we now pass the TCK:
Screenshot 2024-04-09 at 4 00 31 AM

gavinking:
Wait, I might have celebrated too soon. I think there's more tests that Scott has not wired in yet.

PR #679, opened 2024-04-08
PageRequest.page(num) method can cause wrong usage

### LINKED ISSUE #672: [Bug]: PageRequest doesn't know whether it's a Builder

njr-11:
fixes #672

This adds onto the change to remove next()/previous() (first commit) and also removes page(number) from PageRequest (second commit).  All of these methods can mislead applications into writing code that causes wrong behavior.

### DIFF
--- a/api/src/main/java/jakarta/data/Order.java
+++ b/api/src/main/java/jakarta/data/Order.java
@@ -40,7 +40,7 @@ import jakarta.data.repository.OrderBy;
  * Page&lt;Employee>&gt; findByYearHired(int year, PageRequest pageRequest, Order&lt;Employee&gt;);
  * ...
  * page1 = employees.findByYearHired(Year.now(),
- *                                   PageRequest.ofSize(10).page(1),
+ *                                   PageRequest.ofSize(10),
  *                                   Order.by(_Employee.salary.desc(),
  *                                            _Employee.lastName.asc(),
  *                                            _Employee.firstName.asc()));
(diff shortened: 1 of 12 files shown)


otaviojava:
As I said, I like this as the first step, but it only fixes some problems.

The specialization will solve it entirely, but I know it has an impact.

njr-11:
> As I said, I like this as the first step, but it only fixes some problems.
> 
> The specialization will solve it entirely, but I know it has an impact.

It is not clear to me what problem the specialization solves.  I asked Gavin for more clarification under #672 and came up with this to address what he identified.

otaviojava:
In term of impact, I prefer this one

gavinking:
OK, this looks much better.

gavinking (approved):

PR #684, opened 2024-04-10
JavaDoc needs to align with recent changes to method name grammar

njr-11:
The javadoc needs a few updates to align with the recent grammar changes that made `By` optional and disallowed its use in the absence of any conditions (for example, `deleteBy()`). Also, it looks like the javadoc doesn't make any mention of the ability to place ignored text in certain parts of the method name.  This PR attempts to get the javadoc more lined up with the grammar and the description of it within the specification document.  There should be no changes to specification requirements in this PR, only making the javadoc consistent with what the specification already requires.

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -358,12 +358,14 @@ import java.util.Set;
  *
  * <h2>Query by Method Name</h2>
  *
- * <p>Repository methods following the <em>Query by Method Name</em> pattern
- * must have a method name that begins with one of the following prefixes and
- * must not include the {@code @Find} annotation, {@code @Query} annotation, or
- * any lifecycle annotations on the method or any data access related annotations
- * on the method parameters. Query conditions are determined by the portion of
- * the method name following the {@code By} keyword.</p>
+ * <p>The <em>Query by Method Name</em> pattern translates the name of the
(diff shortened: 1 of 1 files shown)


api/src/main/java/module-info.java
```diff
@@ -387,19 +389,35 @@
...
+ * followed by the {@code IgnoreCase} keyword (for text properties), optionally
+ * followed by the {@code Not} keyword, optionally followed by a condition operator
+ * keyword such as {@code StartsWith}. The equality condition is implied when no
+ * condition operator keyword is present. Most of the condition operations, such as
+ * {@code Like} or {@code LessThan}, correspond to a single method parameter. The
+ * exceptions to this rule are {@code Between}, which corresponds to two method
+ * parameters, and {@code Null}, {@code True}, and {@code False}, which require
+ * no method parameters.</p>
+ *
+ * <p>A {@code find} query can optionally end with an <em>order clause</em>,
```
gavinking, line 415:
```suggestion
 * <p>A {@code find} query can optionally end with an <em>order</em>,
```

We're not using the term "clause" anywhere else in the description of QbMN.

gavinking (approved):

gavinking:
I think we can merge this one, no?

PR #685, opened 2024-04-10
more improvements to the presentation of the QbMN feature

gavinking:
There should be no semantic changes here.

### DIFF
--- a/spec/src/main/asciidoc/method-query.asciidoc
+++ b/spec/src/main/asciidoc/method-query.asciidoc
@@ -48,23 +48,6 @@ As such, its syntax is limited to the use of legal identifier characters, so the
 
 Jakarta Data 1.0 offers a Query by Method Name facility as an extension to the specification, providing a migration path for existing applications written for repository frameworks which offer similar functionality.
 
-The functionality described here overlaps significantly with both:
-
-- parameter-based automatic query methods, that is, `@Find` and `@Delete`, and
-- annotated query methods, that is `@Query` and Jakarta Data Query Language or Jakarta Persistence Query Language.
-
-Therefore, these alternative approaches are strongly preferred for newly-written code.
-
-A Jakarta Data provider is required to support the Query by Method Name extension in Jakarta Data 1.0.
-
-NOTE: A Jakarta Data provider backed by a key-value or wide-column datastore is not required to support Query by Method Name.
-
-WARNING: This functionality is considered deprecated, and the requirement that a provider support the Query by Method Name extension will be removed in a future version of Jakarta Data.
-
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/method-query.asciidoc
```diff
@@ -63,46 +78,60 @@ WARNING: This functionality is considered deprecated, and the requirement that a
...
+- an optional <<Limits,limit>>,
+- a optional <<Restrictions,restriction>>, and
+- an optional <<Orders,order>>.
 
-  @OrderBy("price")
-  List<Product> findByNameLike(String namePattern);
+=== Limits
 
-  List<Product> findByNameLikeAndPriceLessThanOrderByPriceDesc(String namePattern, float priceBelow);
+A `find` query may have a _limit_, for example, `First` or `First10`. Other actions may not be combined with a limit.
```
njr-11, line 92:
```suggestion
A `find` query may have a _limit_, for example, `First` or `First10`. Other actions must not be combined with a limit.
```

spec/src/main/asciidoc/method-query.asciidoc
```diff
@@ -63,46 +78,60 @@ WARNING: This functionality is considered deprecated, and the requirement that a
...
-- The method name is formed by combining the subject, predicate, and order clause, in that order.
+The conditions belonging to the restriction determine the parameters of the method, as specified below in <<Query by Method Name Conditions>>.
 
-NOTE: This specification uses the terms subject and predicate in a way that aligns with industry terminology rather than how they are defined in English grammar.
+=== Orders
 
-Queries can also handle entities with relation attributes by specifying the relationship using dot notation, with the dot converted to underscore so that it is a valid character within the method name.
+A `find` query may have an _order_. The _order_ specifies how records should be sorted. It is formed by concatenating `OrderBy` with one or more ordered pairs of an entity attribute name and a direction of sorting, `Asc` or `Desc`. The direction may be omitted if there is only one property, in which case `Asc` is implied.
```
njr-11, line 113:
```suggestion
A `find` query may have an _order_. The _order_ specifies how records must be sorted. It is formed by concatenating `OrderBy` with one or more ordered pairs of an entity attribute name and a direction of sorting, `Asc` or `Desc`. The direction may be omitted if there is only one property, in which case `Asc` is implied.
```

njr-11 (approved):
This looks good. I identified a few places to switch "may" or "should" to "must".  It looks like we were both working on Query by Method Name documentation at the exact same time here, but fortunately all of your updates are in the spec document and all of mine are in the javadoc so we managed to completely avoid merge conflicts.

PR #686, opened 2024-04-11
beforeKey and afterKey were missed in some examples

njr-11:
I should have spotted these code examples in spec/javadoc that need updating due to the replacement of the PageRequest.beforeKey and afterKey methods, but somehow missed them. These methods don't exist anymore and should be switched over to the beforeCursor/afterCursor methods. Also, I added a page number to these examples to match up with updates that Gavin wanted to a TCK test when requesting a middle page with cursor pagination.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -544,8 +544,9 @@ Or you can obtain the next (or previous) page relative to a known entity,
 [source,java]
 ----
 Customer c = ...
-PageRequest p = PageRequest.ofSize(50)
-                           .afterKey(c.lastName, c.firstName, c.id);
+PageRequest p = PageRequest.ofPage(10)
+                           .size(50)
+                           .afterCursor(Cursor.forKey(c.lastName, c.firstName, c.id));
 page = customers.findByZipcodeOrderByLastNameAscFirstNameAscIdAsc(55902, p);
 ----
 
@@ -654,8 +655,8 @@ Obtaining the next 10 products that cost $50.00 or more:
 float priceMidpoint = 50.0f;
 Order<Product> order = Order.by(_Product.price.asc(),
                                 _Product.id.asc());
-PageRequest pageRequest = PageRequest.ofSize(10)
-             .afterKey(priceMidpoint, 0L);
+PageRequest pageRequest = PageRequest.ofPage(5).size(10)
+             .afterCursor(Cursor.forKey(priceMidpoint, 0L));
 CursoredPage<Product> moreExpensive =
         products.findByNameLike(pattern, pageRequest, order);
 ----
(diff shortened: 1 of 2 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -539,8 +539,9 @@ Or you can obtain the next (or previous) page relative to a known entity,
 [source,java]
 ----
 Customer c = ...
-PageRequest p = PageRequest.ofSize(50)
-                           .afterKey(c.lastName, c.firstName, c.id);
+PageRequest p = PageRequest.ofPage(10)
```
otaviojava, line 542:
```suggestion
PageRequest p = PageRequest.ofSize(50)
````

File: spec/src/main/asciidoc/repository.asciidoc [hunk 2]
```diff
@@ -653,8 +654,9 @@ Order<Product> order =
         Order.by(_Product.price.asc(),
                  _Product.id.asc());
 PageRequest pageRequest =
-        PageRequest.ofSize(10)
-                   .afterKey(priceMidpoint, 0L);
+        PageRequest.ofPage(5)
```
otaviojava, line 657: The same is true here; maybe I did not get your point, or I am missing some details.

But once we are talking about the cursor, we should not define the page number.

spec/src/main/asciidoc/repository.asciidoc
njr-11: Your suggestion would result in duplicating the size,
```
PageRequest p = PageRequest.ofSize(50).size(50).afterCursor...
```
so we shouldn't add this change exactly as proposed but could still remove the page if we decide to.

gavinking (approved):

spec/src/main/asciidoc/repository.asciidoc
gavinking: > But once we are talking about the cursor, we should not define the page number.

I disagree. If this thing always carries a page number, the page number should have a sensible value.

(We have not even defined a magical value like `-1` to mean "the page number is nonsense".)

spec/src/main/asciidoc/repository.asciidoc
otaviojava: Nope, remove the size.

My question is: in this demo with the cursor, do we need the page?

gavinking: > do we need the page?

Yes you need it because backward navigation from the (approximate) first page works differently from backward navigation from any other page. So setting `page` to `1` when you are not at the first page results in weird behavior.

spec/src/main/asciidoc/repository.asciidoc
otaviojava: @gavinking Sorry, but I don't follow you.

My point is that the demo focuses on a cursor that won't use the page number. Thus, why do we need to put something like this:

```java
 PageRequest.ofSize(10).afterCursor(Cursor.forKey(priceMidpoint, 0L));
```

⚠️ I am not talking about putting a magical number, but I am going to the point and putting information that the cursor requires, such as the cursor and the size, skipping the page.

spec/src/main/asciidoc/repository.asciidoc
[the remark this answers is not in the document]
otaviojava: Thank you for the explanation.

spec/src/main/asciidoc/repository.asciidoc
gavinking: When that `PageRequest` is ultimately used to construct a `CursoredPageRecord`, it matters whether it represents a "first" page. 

- If it _does_ then previous page navigation should result in an offset-based `PageRequest`. 
- It it _doesn't_, then previous page navigation should result in a cursor-based `PageRequest`.

Completely apart from that, `page()` is a public member of `PageRequest`, exposed to the application programmer, and shouldn't return an incorrect value.

PR #687, opened 2024-04-11
update sigtest to 2.2 everywhere

gavinking:
I think but am not certain that this is needed.


KyleAure (changes requested):
Thanks for opening this.  I was going to update the TCK docs when I got back from dev nexus.
Need to also do the replacement `org.netbeans.tools` -> `jakarta.tck

gavinking:
> There are some other areas in the tck-dist where this shows up.

Ah, OK, got 'em. Cheers.

KyleAure (approved):

PR #688, opened 2024-04-11
fix bug in TCK testVersionedInsertUpdateDelete()

gavinking:
If `modify()` increments the version number before throwing, the `delete()` call will succeed instead of failing.

A general rule of JPA is that one must discard entity instances after receiving an error, since there is simply no way to know if they are in sync with the database.

This test disobeys that rule, and reuses an instance after a thrown exception. Whether the version number of this instance has been incremented before the exception is thrown is 100% provider dependent and JPA makes absolutely no guarantees.

The simplest fix is to just set the version number back to the initial version. A better fix would be to not reuse the instance.


njr-11 (approved):
Thanks for spotting this

PR #696, opened 2024-04-12
Extract Constant at PageRequest

otaviojava:
# Changes

- Extract constant and avoid magic number at the PageRequest
- Show the default value and the default page number explicitly.
- Also point where the first page number is, helping, because in some vendors the first page starts with "zero".

### DIFF
--- a/api/src/main/java/jakarta/data/page/PageRequest.java
+++ b/api/src/main/java/jakarta/data/page/PageRequest.java
@@ -69,6 +69,9 @@ import java.util.Optional;
 public interface PageRequest {
 
 
+    int SIZE = 10;
+    int FIRST_PAGE = 1;
+
     /**
      * Creates a new page request with the given page number and with a default size of 10.
      *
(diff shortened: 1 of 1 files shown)


njr-11 (commented):
In thinking about where a user would take advantage of these constants, I cannot see a good use case for `FIRST_PAGE`.

It would be pointless to write `PageRequest.ofPage(PageRequest.FIRST_PAGE).size(20)` when you can already write the more concise `PageRequest.ofSize(20)` to get the same thing.

Similarly, `PageRequest.ofPage(PageRequest.FIRST_PAGE, 20, true)` is more concisely written as `PageRequest.ofSize(20)` and `PageRequest.ofPage(PageRequest.FIRST_PAGE, 20, true)` is more concisely written as `PageRequest.ofSize(20).withoutTotal()`.

`FIRST_PAGE` is unwanted on static before/afterCursor.

I can't see anywhere that it is a good practice to use it.

There might be some use cases where `DEFAULT_SIZE` would be useful, but I would expect that typically users will want to explicitly specify their own maximum page size rather than receiving a default, and if they do want a default, the typical pattern will be similar to page number where it is more concisely written with something like, `PageRequest.ofPage(4)`.

PR #707, opened 2024-04-17
Use web archive instead of java archive for standalone

### LINKED ISSUE #705: [TCK Challenge]: ee.jakarta.tck.data.standalone.persistence.PersistenceEntityTests in non-standalone mode
### Specification

tck guide

### Assertion

tck setup issue

### TCK Version

SNAPSHOT (Locally built)

WildFly

### Challenge Scenario

Claims that a test is not portable or depends on a particular implementation.

### Full Description

When running in a non-standalone mode in some EE profile, the standalone tests are still deployed as JARs rather than WARs. There is no universal support for this especially when EE components like entities are being deployed. Most of the tests just happen to work because WildFly can deploy them as service jars, but deploying the PersistenceEntityTests does not. This needs a persistence.xml descriptor and WildFly will not process a bare jar as a potential persistence unit deployment.

### Additional Context

The deploy method on PersistenceEntityTests should only return a JavaArchive when running in standalone mode. In any non-standalone mode it should return a WebArchive. It is possible that all @Standalone annotated tests should use this approach.

KyleAure:
Fixes #705 

- Packages standalone tests as .war instead of .jar.  When running in SE mode we never deploy or run any artifacts from the deployment.  These should always be deployed as a .war otherwise the test code gets added as a Library in the .war constructed by Arquillian (did not realize this when writing these tests). 

Below is before/after this change: 
Screenshot 2024-04-17 at 10 15 54
Screenshot 2024-04-17 at 10 08 32

- Also removed example tests that already have real equivalents
- Fixed some typos


gavinking:
@starksm64 WDYT?

gavinking (approved):
Scott says it looks good.

Issue #709, opened 2024-04-18
[TCK Challenge]: ExtensionTests


starksm64:
### Specification

https://github.com/jakartaee/data/blob/main/tck/src/main/java/ee/jakarta/tck/data/core/cdi/ExtensionTests.java

### Assertion

CDI beans, custom entity annotation

### TCK Version

SNAPSHOT (Locally built)

### Implementation being tested

HIbernate / Weld

### Challenge Scenario

Claims that a test is not portable or depends on a particular implementation.

### Full Description

We are running into two problems with the ExtensionTests, both ee.jakarta.tck.data.core.cdi and ee.jakarta.tck.data.web.cdi.
The first issue is that the `ee.jakarta.tck.data.common.cdi.Directory` has a custom provider and so Hibernate does not process it to generate an implementation that is a CDI bean. Instead there is a `ee.jakarta.tck.data.common.cdi.DirectoryRepository` implementation, but it does not have a bean defining annotation, and so is not picked up for processing by the BuildCompatibleExtensionImpl. When Hibernate generates a repository bean it annotates it with a `@RequestScoped` annotation. The `DirectoryRepository` should have a bean defining annotation or a beans.xml descriptor that identifies this as a CDI bean.

The second problem relates to the `ee.jakarta.tck.data.common.cdi.AddressBook` repository interface. This is using a custom `EntityDefining` annotation, and Hibernate does not generate a repository bean for this as expected. However, the `ee.jakarta.tck.data.common.cdi.AddressRepository` is also missing a bean defining annotation.

If I add `@RequestScoped` annotations to both repository beans, we see all tests passing.

### Is there an existing challenge for this?

- [X] I have searched the existing issues

gavinking:
So I spent like 2 hours trying to understand all this last night, and it's all still  very murky to me.

Apparently, there are two competing ways that these beans are supposed to get registered by a CDI extension. Either:

1. by `ExtensionImpl` registering a `DirectoryRepositoryProducer` which creates a `DirectoryRepository`, or
2. by `BuildCompatibleExtensionImpl` registering the `DirectoryRepository` bean directly.

But as a special twist, neither of the extensions just registers the bean. Instead, the extensions want to be notified of the existence of the repository _interface_ via a `ProcessAnnotatedType` event, or via a `@Enhancement` callback. I can only guess at the motivation behind this: I speculate that what it's doing is attempting to simulate how the OpenLiberty implementation of Jakarta Data works.

[...]

PR #711, opened 2024-04-18
Further refactoring of CDI tests

KyleAure:
Related #709

At this point we are testing CDI and the integration with Jakarta Data, but we don't actually need to implement the repositories for our goals.

I have simplified the CDI tests down for what we actually want to test from the Spec which is: 
1. Jakarta Data providers can co-exist based on the `provider` attribute of `@Repository`
2. Jakarta Data providers can co-exist based on the `@EntityDefining` annotation from `@Repository` via the Entity.

I am still new to CDI and still learning so I think I was a bit heavy handed before.


gavinking (approved):
LGTM

mswatosh (approved):

Issue #732, opened 2024-05-08
[clarification]: Can I mix parameters in the Query?


otaviojava:
### Specification

https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/query-language.asciidoc#parameters

### I need clarification on ...

The spec does not clarify whether we can combine the original with the named parameter.

I did not find any combination on the internet or in our specs, so I assume they cannot mix.

Based on my understanding, if there is an ordinal, all parameters should be ordinal. This is my current interpretation of the situation.

Does it make sense?

gavinking:
Ah, this is a very good point. Nice catch.

So, in our implementation we *do* let you mix named and positional parameters.

However, the JPA spec says:

> Positional and named parameters must not be mixed in a single query.

(Section 4.7.4 of JPA 3.2.)

So we should add that same restriction to the spec for JDQL.

otaviojava:
> Ah, this is a very good point. Nice catch.
> 
> So, in our implementation we _do_ let you mix named and positional parameters.
> 
> However, the JPA spec says:
> 
> > Positional and named parameters must not be mixed in a single query.
> 
> (Section 4.7.4 of JPA 3.2.)
> 
> So we should add that same restriction to the spec for JDQL.

Ok, I will open a PR with this information.

PR #733, opened 2024-05-08
Include a note about no mix types of query in a single one

otaviojava:
Changes

# Fixes https://github.com/jakartaee/data/issues/732


njr-11 (approved):
This is consistent with the `@Query` JavaDoc, which states:
```
 * <p>A given query must not mix named and ordinal parameters.</p>
```

gavinking:
@otaviojava this is the wrong section, that section is supposed to be the lexical structure.

gavinking:
You could put it in 5.3.3, for example.

otaviojava:
> You could put it in 5.3.3, for example.

done

PR #745, opened 2024-05-23
Update TCK license

KyleAure:
Missed updating the TCK License in the last round of mentor reviews.


tck-dist/src/main/asciidoc/data-tck-reference-guide.adoc
```diff
@@ -20,7 +20,7 @@
 :JavaVersion1: 17
 :JavaVersion2: 21
 
-:license: Eclipse Foundation Technology Compatibility Kit License - v1.1
+:license: Eclipse Foundation Technology Compatibility Kit License
```
mswatosh, line 23: Why is the v1.1 being removed here if we're switching to v1.1?

mswatosh (commented):
The assembly.xml should be updated to point to the adoc instead of txt: https://github.com/jakartaee/data/blob/28debff7af0d6e1b08b013ddbde1df3869520d24/tck-dist/src/main/assembly/assembly.xml#L31

tck-dist/src/main/asciidoc/data-tck-reference-guide.adoc
mswatosh: (Unless the intent is to remove the versioning altogether, which would be fine. Seems like the rest of the repository uses EFTL and not EFTL v1.1)

KyleAure: In the docs we provide a link to the license.  The link is static and will always point to the most recent license.

PR #771, opened 2024-06-14
Verify behavior based on Database type

### LINKED ISSUE #768: [Bug]: Contains and EndsWith should not be mandatory, but the TCK makes it mandatory.
### Specification Version

1.0.0

### Bug report

By specification:

> An implementation of Query by Method Name backed by a document or graph database is not required to support Contains, EndsWith, StartsWith, Like, IgnoreCase, In, or Null. A repository method must raise java.lang.UnsupportedOperationException or a more specific subclass of the exception if the database does not provide the requested functionality.

Ref: https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/method-query.asciidoc#27-query-by-method-name-conditions

However, when I ran at the TCK, it shows it as an error:

```
[ERROR]   MyEntityTests>EntityTests.testContainsInString:419 » UnsupportedOperation Contains is not supported in Eclipse JNoSQL method query
[ERROR]   MyEntityTests>EntityTests.testFindFirst3:617 » UnsupportedOperation EndsWith is not supported in Eclipse JNoSQL method query
```

### Additional information

_No response_

KyleAure:
Fixes #768 

- Added two new properties:
  - jakarta.tck.database.type - So that we can write test logic based on database type, such as, column or graph
  - jakarta.tck.database.name - so that we can write test logic based on database name, such as, oracle or mongo
    - This is for future consideration and currently isn't used. 
- Fixed some spacing issues
- Only read system properties once instead of every time they are used.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/utilities/DatabaseType.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/utilities/DatabaseType.java
@@ -21,7 +21,7 @@ import java.util.Arrays;
  * This enum represents the configured DatabaseType based on the {@link TestProperty} databaseType
  */
 public enum DatabaseType {
-    UNKNOWN(Integer.MAX_VALUE), //No database type was configured
+    OTHER(Integer.MAX_VALUE), //No database type was configured
     RELATIONAL(100),
     GRAPH(50),
     DOCUMENT(40),
@@ -36,7 +36,7 @@ public enum DatabaseType {
     }
     
     public static DatabaseType valueOfIgnoreCase(String value) {
-        return Arrays.stream(DatabaseType.values()).filter(type -> type.name().equalsIgnoreCase(value)).findAny().orElse(DatabaseType.UNKNOWN);
+        return Arrays.stream(DatabaseType.values()).filter(type -> type.name().equalsIgnoreCase(value)).findAny().orElse(DatabaseType.OTHER);
     }
     
     public boolean isKeywordSupportAtOrBelow(DatabaseType benchmark) {
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/utilities/DatabaseType.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/utilities/DatabaseType.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+package ee.jakarta.tck.data.framework.utilities;
+
+import java.util.Arrays;
+
+/**
+ * This enum represents the configured DatabaseType based on the {@link TestProperty} databaseType
+ */
+public enum DatabaseType {
+    UNKNOWN, //No database type was configured
+    RELATIONAL,
+    COLUMN,
+    DOCUMENT,
+    GRAPH,
+    KEY_VALUE;
+    
+    public static DatabaseType valueOfIgnoreCase(String value) {
+        return Arrays.stream(DatabaseType.values()).filter(type -> type.name().equalsIgnoreCase(value)).findAny().orElse(DatabaseType.UNKNOWN);
+    }
+}
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/utilities/DatabaseType.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/utilities/DatabaseType.java
@@ -21,14 +21,25 @@ import java.util.Arrays;
  * This enum represents the configured DatabaseType based on the {@link TestProperty} databaseType
  */
 public enum DatabaseType {
-    UNKNOWN, //No database type was configured
-    RELATIONAL,
-    COLUMN,
-    DOCUMENT,
-    GRAPH,
-    KEY_VALUE;
+    UNKNOWN(Integer.MAX_VALUE), //No database type was configured
+    RELATIONAL(100),
+    GRAPH(50),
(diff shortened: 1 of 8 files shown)


njr-11 (approved):

tck/src/main/java/ee/jakarta/tck/data/framework/utilities/DatabaseType.java
```diff
@@ -0,0 +1,34 @@
...
+package ee.jakarta.tck.data.framework.utilities;
+
+import java.util.Arrays;
+
+/**
+ * This enum represents the configured DatabaseType based on the {@link TestProperty} databaseType
+ */
+public enum DatabaseType {
+    UNKNOWN, //No database type was configured
+    RELATIONAL,
```
otaviojava, line 25: NoSQL is huge!
The main challenge is not having specification

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
```diff
@@ -614,7 +627,17 @@ public void testFindFirst() {
     @Assertion(id = "133",
                strategy = "Use a repository method with findFirst3By that returns the first 3 results.")
     public void testFindFirst3() {
-        AsciiCharacter[] found = characters.findFirst3ByNumericValueGreaterThanEqualAndHexadecimalEndsWith(40, "4", Sort.asc("numericValue"));
+        AsciiCharacter[] found;
+        
+        try {
+        found = characters.findFirst3ByNumericValueGreaterThanEqualAndHexadecimalEndsWith(40, "4", Sort.asc("numericValue"));
```
otaviojava, line 633: When we talk about query flexibility, Documents, and graphs allow similar queries to those in the relational world.

In this case, none of NoSQL might not support those.

Screenshot 2024-06-23 at 07 03 54


There is a huge universe of NoSQL databases, but doing a ballparking of those:

***Key-value**: It will only support key-based operations using equals and in. Anything else, ignore.
* **Wild-column** or column-family: Most keywords might work, but it needs to check the field itself. 


| Keyword         | Graph                                | Document                              | Column                                     | Key-value |
|-----------------|--------------------------------------|---------------------------------------|--------------------------------------------|-----------|
| `Between`       | Yes                                  | Not guaranteed (varies with vendor)   | Yes, but using ID or index field (Cassandra) | Nope      |
| `Contains`      | Not guaranteed (varies with vendor)  | Not guaranteed (varies with vendor)   | Not guaranteed (varies with vendor)        | Nope      |
| `EndsWith`      | Not guaranteed (varies with vendor)  | Not guaranteed (varies with vendor)   | Not guaranteed (varies with vendor)        | Nope      |
| `LessThan`      | Yes                                  | Yes                                   | Yes                                        | Nope      |
| `GreaterThan`   | Yes                                  | Yes                                   | Yes                                        | Nope      |
| `LessThanEqual` | Yes                                  | Yes                                   | Yes                                        | Nope      |
| `GreaterThanEqual` | Yes                               | Yes                                   | Yes                                        | Nope      |
| `Like`          | Not guaranteed (varies with vendor)  | Not guaranteed (varies with vendor)   | Not guaranteed (varies with vendor)        | Nope      |
| `IgnoreCase`    | Not guaranteed (varies with vendor)  | Not guaranteed (varies with vendor)   | Not guaranteed (varies with vendor)        | Nope      |
| `In`            | Yes                                  | Yes                                   | Yes, but using ID or index field (Cassandra) | Yes (ID or key only) |
| `Null`          | Not guaranteed (varies with vendor)  | Not guaranteed (varies with vendor)   | Not guaranteed (varies with vendor)        | Nope      |
| `StartsWith`    | Not guaranteed (varies with vendor)  | Not guaranteed (varies with vendor)   | Not guaranteed (varies with vendor)        | Nope (there is Redis where you can do regex in the key, but it only by the key: [Redis KEYS](https://redis.io/docs/latest/commands/keys/)) |
| `True`          | Yes                                  | Yes                                   | Yes, but using ID or index field (Cassandra) | Nope      |
| `False`         | Yes                                  | Yes                                   | Yes, but using ID or index field (Cassandra) | Nope      |
| `And`         | Yes                                  | Yes                                   | Yes, but using ID or index field (Cassandra) | Nope      |
| `Or`         | Yes                                  | Yes              | Nope | Nope      |
| Sort Operation         | Yes                                  | Yes              | Not guaranteed (varies with vendor and the field) | Nope      |

KyleAure:
@otaviojava  I added a commit to add some flexibility based on the table you provided.
Would something like that work?

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
otaviojava: ⚠️ IMHO: key-value database should only support the insert, update, delete and save operations.
The queries such as Jakarta Data Query Language and method by query those does not make sense to support it.

tck/src/main/java/ee/jakarta/tck/data/framework/utilities/DatabaseType.java
```diff
@@ -21,14 +21,25 @@
  * This enum represents the configured DatabaseType based on the {@link TestProperty} databaseType
  */
 public enum DatabaseType {
-    UNKNOWN, //No database type was configured
-    RELATIONAL,
-    COLUMN,
-    DOCUMENT,
-    GRAPH,
-    KEY_VALUE;
+    UNKNOWN(Integer.MAX_VALUE), //No database type was configured
```
otaviojava, line 24: I believe an UNKNOWN would be the data storage that won't fit on the previous ones, right?

For example, we are not talking about a time series database, such as [InfluxDB](https://www.influxdata.com/time-series-database/) or an Object database, such as [Eclipse Store](https://eclipsestore.io/).

In those cases, I believe it will fit better with a key-value database or don't have this option.
Make it mandatory or throw an exception.

KyleAure: I don't think this needs to be mandatory.  If you don't configure the TCK with a database type then we assume your database supports all features tested.

The TCK cannot be written to perfectly run against every database, it is meant for certifying implementations.  If your Object database supports a feature list closer to a Key-Value database then run the TCK in that mode.

otaviojava: Maybe define a relational database as default and remove this `UNKNOWN.`

otaviojava: OTHER is fine.

KyleAure: ✅  Done

njr-11 (approved):

otaviojava (approved):

PR #779, opened 2024-06-28
TCK cannot require GT, GTE, LT, LTE for Key-Value databases

njr-11:
Per @otaviojava 's comment in https://github.com/jakartaee/data/pull/771/files#r1649951754 , Key-Value databases are not capable of GreaterThan, GreaterThanEqual, LessThan, and LessThanEqual, whereas other database types are able to perform them.  This PR updates TCK tests accordingly.

Additional TCK updates are also needed per the table, but this PR is only attempting to address this subset. 

Even with just this subset, I can already see that a large amount of test coverage is lost, so in the future there will also need to be an effort to ensure the TCK has coverage for various aspects of the spec in terms of capability that is possible for all database types.  We did ask for all of this sort of NoSQL database capability information before starting to write the TCK so that it could be written accordingly, but unfortunately it was never provided until now after the release of 1.0.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
@@ -377,8 +377,20 @@ public class EntityTests {
     @Assertion(id = "133", strategy = "Request a Slice higher than the final Slice, expecting an empty Slice with 0 results.")
     public void testBeyondFinalSlice() {
         PageRequest sixth = PageRequest.ofPage(6).size(5).withoutTotal();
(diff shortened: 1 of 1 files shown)


otaviojava (approved):

otaviojava (approved):

gavinking:
Folks, it's also important that we capture this in the spec itself.

Currently section 5.4.5 does not have any exception documenting this.

njr-11:
> Folks, it's also important that we capture this in the spec itself.
> 
> Currently section 5.4.5 does not have any exception documenting this.

Yes, definitely. This and the rest of the table from Otavio all need to be written into the spec. I'll open an issue for that.

PR #783, opened 2024-07-01
TCK cannot require Contains, EndsWith, IgnoreCase, Like, Null, StartsWith for Document and Graph

### LINKED ISSUE #781: [clarification]: IgnoreCase should be optional, but it is mandatory by TCK.
### Specification

https://github.com/jakartaee/data/blob/main/spec/src/main/asciidoc/method-query.asciidoc

> An implementation of Query by Method Name backed by a document or graph database is not required to support Contains, EndsWith, StartsWith, Like, IgnoreCase, In, or Null. A repository method must raise java.lang.UnsupportedOperationException or a more specific subclass of the exception if the database does not provide the requested functionality.

Method:
- testSingleEntity

### Additional information

_No response_

njr-11:
fixes #781

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
@@ -468,11 +468,19 @@ public class EntityTests {
         assertEquals("7f", del.getHexadecimal());
         assertEquals(true, del.isControl());
 
-        AsciiCharacter j = characters.findByHexadecimalIgnoreCase("6A");
-        assertEquals("6a", j.getHexadecimal());
-        assertEquals('j', j.getThisCharacter());
-        assertEquals(106, j.getNumericValue());
-        assertEquals(false, j.isControl());
+        try {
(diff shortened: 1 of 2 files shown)


otaviojava (approved):

otaviojava:
@KyleAure @njr-11 I am still facing issues using the newest version, e.g.:

https://github.com/eclipse/jnosql-extensions/blob/main/jnosql-data-tck-runner/pom.xml#L175-L186

And I still have a problem with the:

```
testContainsInString
```

And similar.

otaviojava:
Furthermore, what do you think if we simplify to only RELATIONAL and NoSQL databases?

njr-11:
> Furthermore, what do you think if we simplify to only RELATIONAL and NoSQL databases?

As we currently have it, there are a large number of cases where vendors can skip TCK test assertions when run against Key-Value and a much lesser number assertions that are skipped for Graph and Document.  If you simplify to NoSQL, but still want Key-Value to pass, then we would need to start allowing Graph and Document to skip a large number of assertions, meaning the TCK would not be enforcing the specification requirements that it should be for those database types, and vendors would be able to certify implementations that pass on NoSQL even though they don't do what the spec says to.  We should not make a change like that.

PR #796, opened 2024-07-11
README.adoc - maven best practices

lprimak:
Updated README to conform to maven best practices :)


KyleAure (approved):
LGTM - no need to install unless user wants to use a local SNAPSHOT version of the API in their application which is uncommon.

otaviojava:
Thanks @lprimak

PR #804, opened 2024-07-27
Tck test at the method testBeyondFinalPage

otaviojava:


tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
```diff
@@ -394,8 +394,16 @@ public void testBeyondFinalPage() {
...
-        assertEquals(5L, page.totalPages());
+        try {
+            assertEquals(43L, page.totalElements());
+            assertEquals(5L, page.totalPages());
+        } catch (UnsupportedOperationException x) {
+            // Some NoSQL databases lack the ability to count the total results
+            // and therefore cannot support a return type of Page.
+            // Column and Key-Value databases might not be capable of sorting.
+            // Key-Value databases might not be capable of Between.
+            return;
```
njr-11, line 405: Don't allow relational databases to raise UnsupportedOperationException,
```suggestion
            if (type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH)) {
                // Some NoSQL databases lack the ability to count the total results
            } else {
                throw x;
            }
```

njr-11 (changes requested):
All of the changes in the additional commit have the same problem that I identified in the first review, that they are no longer requiring totals to work for relational databases.  I have added suggestions to fix all of them.  Also, I noticed that some of them were going beyond allowing the totalElement/totalPages to be unsupported and also adding other operations.  I have the code suggestions restoring those assertions as well.

njr-11 (approved):

PR #805, opened 2024-07-31
Disable Complex Arithmetic Updates Test in Jakarta Data TCK

otaviojava:
The test for arithmetic updates (e.g., `UPDATE Box SET length = length + ?1, width = width - ?1, height = height * ?2`) should be disabled in the Jakarta Data TCK due to the inconsistent support for such operations across different NoSQL databases. NoSQL databases like MongoDB, Cassandra, Redis, Couchbase, and DynamoDB have varying capabilities and syntax, leading to potential interoperability issues and making it difficult to implement a standardized and reliable test.

Additionally, NoSQL databases often struggle with arithmetic operations involving parentheses in update statements, further complicating the implementation of these tests. Disabling this test will help maintain the consistency and reliability of the Jakarta Data TCK.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
@@ -2396,7 +2396,12 @@ public class EntityTests {
             assertEquals(120, b3.height); // increased by factor of 2
         }
 
-        assertEquals(3, shared.removeAll());
+        var removeAllResult = shared.removeAll();
+
+        if (!type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH)) {
+            //We don't have any guarantee of NoSQL, mainly on eventual consistency databases
+            assertEquals(3, removeAllResult);
+        }
 
         TestPropertyUtility.waitForEventualConsistency();
 


njr-11 (changes requested):
We can exclude arithmetic in updates for NoSQL databases that aren't capable, but the changes here are too broad and allow many different operations to raise UnsuportedOperationException.  See #806 for a more precise solution that only excludes the operations with arithmetic in update, and doesn't lose other testing or allow UnsuportedOperationException in other parts of the test that should still need to work.

otaviojava:
> We can exclude arithmetic in updates for NoSQL databases that aren't capable, but the changes here are too broad and allow many different operations to raise UnsuportedOperationException. See #806 for a more precise solution that only excludes the operations with arithmetic in update, and doesn't lose other testing or allow UnsuportedOperationException in other parts of the test that should still need to work.

@njr-11 Once the method has the scenario to test with parenthesis, which I cannot do, why should I have a partial test working?

> "This method also tests the addition, subtraction, and multiplication operators."

What we can do is break this test into two smaller ones.

What I do have in mind: As soon as I finish with these common tests, I would love to create a test that only makes sense to NoSQL, mainly because we are handled with two different databases, and it is natural to have different behaviors sometimes.


What do you think?

njr-11:
> @njr-11 Once the method has the scenario to test with parenthesis, which I cannot do, why should I have a partial test working?

To preserve TCK coverage for the aspects of the spec (such as JDQL without a WHERE clause) that NoSQL databases are still required to be capable of.

> What we can do is break this test into two smaller ones.
> 
> What I do have in mind: As soon as I finish with these common tests, I would love to create a test that only makes sense to NoSQL, mainly because we are handled with two different databases, and it is natural to have different behaviors sometimes.
> 
> What do you think?

The same applies to all of these exclusions for NoSQL non-capabilities that are being discovered after the 1.0 release. It's a good idea to have improvements like this in a 1.1 or 2.0 release. I think I have mentioned this before. But we are working on 1.0.1 now where it is inappropriate to be doing any refactors or even introducing new tests.  In 1.0.1, we should make only the necessary exclusions to accommodate the capabilities of NoSQL and not go beyond it by unnecessarily removing other assertions and coverage that NoSQL still needs to support.

PR #816, opened 2024-08-07
Enable PMD at Jakarta Data API

otaviojava:


njr-11 (changes requested):
These are mostly trivial changes that would normally be fine. However, because we are working on 1.0.1 which should be TCK updates only, we should not be editing the source code or spec yet.

Even small changes like this come with risks.  An example of that is the addition of `Array.copyOf` to `PageRequestCursor(Object... key)`, which is used by API method `Cursor.forKey(Object... key)`.  Given that the user is supplying the parameters as varargs, it is extremely unlikely to see a non-varargs array passed in instead, but nonetheless a user could do that.  And if a user does so, and then modifies it afterward but before otherwise using the PageRequest, the user could be relying on a behavior that we would break by making this change.  To emphasize, this is extremely unlikely.  Given how unlikely it is for someone to pass a non varargs array to our varargs API method, a good argument could be made that we ought not to go out of our way and incur a performance penalty (copying arrays) for something that a user is really only ever going to break if they are trying to. It might be better just to add a line to the JavaDoc of `Cursor.forKey` saying if you pass a non-varargs array in as the varargs parameter, don't modify it afterward.

otaviojava:
@njr-11, once we merge, could we move this one forward?

njr-11:
> @njr-11, once we merge, could we move this one forward?

Yes, once we merge the updates to switch to v1.1, we can start making code updates again, although we still need to watch out to avoid making breaking changes.

otaviojava:
> > @njr-11, once we merge, could we move this one forward?
> 
> Yes, once we merge the updates to switch to v1.1, we can start making code updates again, although we still need to watch out to avoid making breaking changes.

I did not change any API or any breaking changes; this only includes PMD and checkstyle.

otaviojava:
> > I did not change any API or any breaking changes; this only includes PMD and checkstyle.
> 
> See my prior [review comment](https://github.com/jakartaee/data/pull/816#pullrequestreview-2227900607) . One of the updates is technically a breaking change in behavior, even though it is unlikely applications would use it that way.

Thanks, @njr-11. Done; could you check it, please?

njr-11 (approved):
Looks good now. Thanks for updating.

PR #831, opened 2024-08-26
TCK cannot expect sorting on enum

### LINKED ISSUE #830: [TCK]:  Inconsistent Enum Persistence Between Jakarta Persistence and Jakarta NoSQL Affects Query Behavior in TCK Tests
### Specification Version

1.0.0

### Bug report

There is an inconsistency in how enums are persisted by default in Jakarta Persistence (JPA) and Jakarta NoSQL, which significantly impacts queries and the related TCK (Technology Compatibility Kit) tests that involve fields of enum type.

### **Issue Overview:**

- **Jakarta Persistence (JPA):**
  - By default, JPA persists enums as ordinal values (integers) in SQL databases. This means the position of the enum in its declaration (e.g., `0` for the first constant, `1` for the second, etc.) is stored in the database.
  
- **Jakarta NoSQL:**
  - In contrast, Jakarta NoSQL persists enums as their string names by default in NoSQL databases like MongoDB. This means that the actual name of the enum constant (e.g., `"ONE"`, `"PRIME"`, `"COMPOSITE"`) is stored.

### **Impact on Query Behavior:**

This difference in storage strategies leads to inconsistencies in query results between SQL and NoSQL databases when dealing with enum fields. Specifically:

- **Sorting Behavior:**
  - In SQL, sorting by an enum field will follow the ordinal (numerical) order. For example, `ONE (0)`, `PRIME (1)`, `COMPOSITE (2)`.
  - In NoSQL, sorting by the same enum field will follow lexicographical (alphabetical) order. For example, `"COMPOSITE"`, `"ONE"`, `"PRIME"`.

- **Filtering and Equality Checks:**
  - Comparisons and equality checks against enum fields will behave differently depending on whether the field is stored as an ordinal or a string, potentially leading to unexpected results in TCK tests.

# Solution

That is a classic case for test specialization; thus, I could create tests to verify it on the NoSQL side; we already have a package for NoSQL. The goal could be to use the `NoSQLEntityTests` for this.

njr-11:
Fixes #830

This corrects the error in the TCK where it expects sorting on an enum attribute to be done in a particular way, which [per the spec is vendor specific](https://github.com/jakartaee/data/issues/830#issuecomment-2308476456).

Because this fix is needed in the 1.0.1 maintenance release, it should be as minimal as possible to fix the TCK, which is achieved by adding an entity attribute for the ordinal value and switching a few places in the TCK to sort on that value instead of the enum.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/NaturalNumber.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/NaturalNumber.java
@@ -39,6 +39,7 @@ public class NaturalNumber implements Serializable {
     // Sorting on enum types is vendor-specific in Jakarta Data.
     // Use numTypeOrdinal for sorting instead.
     @jakarta.nosql.Column
+    @jakarta.persistence.Enumerated(jakarta.persistence.EnumType.STRING)
     private NumberType numType; // enum of ONE | PRIME | COMPOSITE
 
     @jakarta.nosql.Column
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/NaturalNumber.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/NaturalNumber.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -36,8 +36,13 @@ public class NaturalNumber implements Serializable {
     @jakarta.nosql.Column
     private Short numBitsRequired;
 
+    // Sorting on enum types is vendor-specific in Jakarta Data.
+    // Use numTypeOrdinal for sorting instead.
     @jakarta.nosql.Column
     private NumberType numType; // enum of ONE | PRIME | COMPOSITE
+
+    @jakarta.nosql.Column
+    private int numTypeOrdinal; // ordinal value of numType
 
     @jakarta.nosql.Column
     private long floorOfSquareRoot;
@@ -72,6 +77,14 @@ public class NaturalNumber implements Serializable {
 
     public void setNumType(NumberType numType) {
         this.numType = numType;
+    }
+
+    public int getNumTypeOrdinal() {
+        return numTypeOrdinal;
+    }
+
+    public void setNumTypeOrdinal(int value) {
+        numTypeOrdinal = value;
     }
 
     public long getFloorOfSquareRoot() {
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/NaturalNumbers.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/NaturalNumbers.java
@@ -45,9 +45,9 @@ public interface NaturalNumbers extends BasicRepository<NaturalNumber, Long>, Id
     CursoredPage<NaturalNumber> findByFloorOfSquareRootOrderByIdAsc(long sqrtFloor,
                                                                     PageRequest pagination);
 
-    Stream<NaturalNumber> findByIdBetweenOrderByNumTypeAsc(long minimum,
-                                                           long maximum,
-                                                           Order<NaturalNumber> sorts);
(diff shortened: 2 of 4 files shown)


tck/src/main/java/ee/jakarta/tck/data/framework/read/only/NaturalNumber.java
```diff
@@ -36,9 +36,14 @@ public enum NumberType {
     @jakarta.nosql.Column
     private Short numBitsRequired;
 
+    // Sorting on enum types is vendor-specific in Jakarta Data.
+    // Use numTypeOrdinal for sorting instead.
```
otaviojava, line 40: If you are already changing this entity why not put ` @Enumerated(STRING)`

for semantic and that sounds better.

njr-11: > If you are already changing this entity why not put ` @Enumerated(STRING)`
> 
> for semantic and that sounds better.

If you like, I can add that too, but it isn't a solution on its own and doesn't fix the problem. Without the other changes I made here, the test would remain broken for providers that are neither JPA nor NoSQL.

otaviojava: @njr-11 I know, but once you can change the tests, I recommend using String and updating the tests based on the String instead of int.

njr-11: > @njr-11 I know, but once you can change the tests, I recommend using String and updating the tests based on the String instead of int.

I added Enumerated(STRING) for you under https://github.com/jakartaee/data/pull/831/commits/ba6854459e26d5b50e81a9f93ac98d7b3f23b571
but it is wrong to redo test assertions like you are suggesting in a .0.1 maintenance release.  The scope of a fix in a maintenance should be limited to fixing the problem only, not refactoring tests or switching things around that aren't required to fix the bug.

otaviojava: > > @njr-11 I know, but once you can change the tests, I recommend using String and updating the tests based on the String instead of int.
> 
> I added Enumerated(STRING) for you under [ba68544](https://github.com/jakartaee/data/commit/ba6854459e26d5b50e81a9f93ac98d7b3f23b571) but it is wrong to redo test assertions like you are suggesting in a .0.1 maintenance release. The scope of a fix in a maintenance should be limited to fixing the problem only, not refactoring tests or switching things around that aren't required to fix the bug.

I agreed, and that is why I proposed to create those tests as a NoSQL extension:

https://github.com/jakartaee/data/issues/830#issuecomment-2308480914

njr-11: > I agreed, and that is why I proposed to create those tests as a NoSQL extension:
> 
> [#830 (comment)](https://github.com/jakartaee/data/issues/830#issuecomment-2308480914)

Adding some NoSQL specific tests (and also some Jakarta Persistence specific ones) in this are is fine for the 1.1 release.  It would be out of place for the 1.0.1 maintenance release and is not a substitute for fixing the bug that exists in 1.0 / 1.0.1 which this PR addresses.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/NaturalNumber.java
```diff
@@ -36,9 +36,15 @@ public enum NumberType {
...
     @jakarta.nosql.Column
+    @jakarta.persistence.Enumerated(jakarta.persistence.EnumType.STRING)
     private NumberType numType; // enum of ONE | PRIME | COMPOSITE
 
+    @jakarta.nosql.Column
+    private int numTypeOrdinal; // ordinal value of numType
```
otaviojava, line 46: @njr-11 you changed to String, but you still using this one here: numTypeOrdinal.

Shall we remove this one and update the tests to work with the enum as String?

njr-11: > @njr-11 you changed to String, but you still using this one here: numTypeOrdinal.
> 
> Shall we remove this one and update the tests to work with the enum as String?

To repeat what I said earlier, doing that "isn't a solution on its own and doesn't fix the problem. Without the other changes I made here, the test would remain broken for providers that are neither JPA nor NoSQL."

otaviojava (approved):

PR #838, opened 2024-08-29
test: fix testUpdateQueryWithWhereClause method

otaviojava:


otaviojava:
@njr-11 I forget to update this one:

https://github.com/jakartaee/data/pull/828

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
```diff
@@ -2442,8 +2442,10 @@ public void testUpdateQueryWithWhereClause() {
             // Ensure there is no data left over from other tests:
             shared.deleteIfPositive();
         } catch (UnsupportedOperationException x) {
-            if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
-                return; // Key-Value databases might not be capable of And.
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH) && TestProperty.delay.isSet()) {
+                // NoSQL databases with eventual consistency might not be capable
+                // of counting removed entities.
+                // Use alternative approach for ensuring no data is present:
```
njr-11, line 2448: Restore the skip for key-value databases that are not using eventual consistency:
```suggestion
            if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
                return; // Key-Value databases might not be capable of And.
            } else if (type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH) && TestProperty.delay.isSet()) {
                // NoSQL databases with eventual consistency might not be capable
                // of counting removed entities.
                // Use alternative approach for ensuring no data is present:
```

njr-11 (commented):
It looks like one of the skips (covering key-value without eventual consistency) was lost, so I added a suggestion to put it back.

Also, I see this line later in the test which is another occurrence that will need to be covered:
```
        assertEquals(2, shared.deleteIfPositive());
```

Also, I think we need to add a method to the MultipleEntityRepo returning void,
```
    @Query("DELETE FROM Coordinate WHERE x > 0.0d AND y > 0.0f")
    void deleteIfPositive_EventualConsistency();
```
which can be used within the catch where it says:
`// Use alternative approach for ensuring no data is present:`

because it doesn't currently do anything, and then subsequent assertions check that the data was deleted.

otaviojava:
@njr-11 done

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
```diff
@@ -2444,6 +2444,10 @@ public void testUpdateQueryWithWhereClause() {
         } catch (UnsupportedOperationException x) {
             if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
                 return; // Key-Value databases might not be capable of And.
+            } else if (type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH) && TestProperty.delay.isSet()) {
+                // NoSQL databases with eventual consistency might not be capable
+                // of counting removed entities.
+                // Use alternative approach for ensuring no data is present:
```
njr-11, line 2450:
```suggestion
                // Use alternative approach for ensuring no data is present:
                shared.deleteIfPositiveWithoutReturnRecords();
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
```diff
@@ -2477,8 +2481,19 @@ public void testUpdateQueryWithWhereClause() {
...
+        try {
         assertEquals(2, shared.deleteIfPositive());
-
+        } catch (UnsupportedOperationException x) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
+                return; // Key-Value databases might not be capable of And.
+            } else if (type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH) && TestProperty.delay.isSet()) {
+                // NoSQL databases with eventual consistency might not be capable
+                // of counting removed entities.
+                // Use alternative approach for ensuring no data is present:
```
njr-11, line 2492:
```suggestion
                // Use alternative approach for ensuring no data is present:
                shared.deleteIfPositiveWithoutReturnRecords();
```

njr-11 (commented):
Looks good, but we also need to add the line to use the new method in the two places where it says "Use alternative approach..."

njr-11 (approved):

PR #839, opened 2024-08-29
It skips the pagination to false when it NoSQL databases

otaviojava:


tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
```diff
@@ -945,7 +945,10 @@ public void testFindPage() {
         assertEquals(List.of(82L, 81L),
                      page3.stream().map(NaturalNumber::getId).collect(Collectors.toList()));
 
-        assertEquals(false, page3.hasNext());
+        if (type.isKeywordSupportAtOrBelow(DatabaseType.RELATIONAL)) {
+            //skip this test for NoSQL databases
+            assertEquals(false, page3.hasNext());
+        }
```
njr-11, line 951: Note that, despite the comment you added, your change here excludes the assertion for DatabaseType=OTHER and causes it to run for Relational and all types of NoSQL.

Regardless, it should be pointed out that according to specification JavaDoc, Page.next returns `{@code false} if this is the last page of results.` which is clearly the case here regardless of the database type because the current page has only 2 out of a maximum of 7 elements.  I cannot see why we should skip enforcing the specification requirement for any database type.

Issue #857, opened 2024-10-02
Parameter-based Automatic Queries as a full alternative to Query by Method Name


njr-11:
The parameter-based Automatic Queries pattern already provides an alternative to much of what Query by Method Name offers for the repository author to define queries upfront, but some parts are lacking which we should address.  Applications ought to be able to choose to fully utilize this pattern and not need to rely on Query by Method Name in places.

The most notable piece that is lacking is the ability to request conditions other than equality. One way to address that would be with a single `Is` annotation that accepts an enumerated value for the condition to apply.  For example,

```java
@Find
@OrderBy(_Product.PRICE)
@OrderBy(_Product.NAME)
@OrderBy(_Product.ID)
Page<Product> search(@By(_Product.NAME) @Is(Like) String pattern,
                     @By(_Product.PRICE) @Is(LessThan) float max,
                     PageRequest pageRequest);
```

The above directly corresponds to the following Query by Method Name method that could currently be used:

```
Page<Product> findByNameLikeAndPriceLessThanOrderByPriceAscNameAscIdAsc(String pattern,
                                                                        float max,
                                                                        PageRequest pageRequest);
```

Note that everything that is needed for the former is already in the Jakarta Data specification except for the mechanism for specifying the condition to be something other than equals.  So the enhancement here has a direct equivalence to the Query by Method Name and achieves it with a minor addition that fits right in to an existing specification pattern.

There has been some confusion over how this differs from other new capability like Restrictions and new Special parameter types (Pattern/Range) with which the repository user can customize different types of conditions.  These are two very different usage patterns.  In some cases, the nature of the application involves a repository user needing the flexibility to compute different conditions at run time, such as in response to an end user defining different search conditions.  But in other cases, the repository author wishes to provide specific capability and to externalize it to the repository user in the most simplistic and direct way, eliminating as much misuse and error conditions as possible. 

When the repository author has a set of requirements that can be encoded into the repository method definition in advance, they ought to be able to do so, in order to abstract the complexity away from the repository user, who can then use the repository in the simplest way, without concern for complexities that the repository author has already taken care of.

A brief example of this is a repository for a library that wants to identify overdue books in order to remind the borrower to return them. The repository author might write the following method, encoding the LESS_THAN condition for the dueDate to identify overdue books, and appropriately naming the method so that the repository user knows exactly what it is for,

```java
@Find
@OrderBy(_Book.DUE_DATE)
@OrderBy(_Book.BORROWER)
[...]

Issue #829, opened 2024-08-22
Ability to define query restrictions dynamically


njr-11:
Replaces #825

Users will sometimes need to define queries dynamically instead of statically in response to what they see happen at run time. For example, a user specifies various customizable search options and in response, the application dynamically adds restrictions to a query to filter for the matching results.

An application ought to be able to define the same kind of query restrictions dynamically that it can already define statically (with patterns like Query by Method Name).

Query restrictions should make use of the Jakarta Data Static Metamodel and generally be agnostic to the database as much as possible, other than the same sort of limitations that Query By Method Name and the other static patterns have on which operations certain classes of NoSQL databases are incapable of.

The Jakarta Data Static Metamodel will be expanded to represent conditions that are possible per attribute type.  These conditions can be supplied as an extra repository method parameter, similar to what is currently possible with Order.by/Sorts.

Example with one static and one dynamic restriction:

```
Page<Product> page1 = products.search(namePattern,
                                      _Product.price.lessThan(100.0),
                                      Order.by(_Product.price.desc(),
                                               _Product.id.asc()));
```

For multiple dynamic restrictions,

One possible way [suggested by Otavio in 825](https://github.com/jakartaee/data/issues/825#issuecomment-2304071462)
```
List<Product> list = products.search(
                         namePattern,
                         _Product.price.greaterThan(min)
                             .and(_Product.price.lessThan(max))
                             .and(_Product.color.equal(BLUE)
                                 .or(_Product.color.equal(GREEN))),
                         Order.by(_Product.name.asc(),
                                  _Product.id.asc()));
```

Another possible way (slightly more wordy, possibly more readable)
```
List<Product> list = products.search(
                         namePattern,
                         Require.all(_Product.price.greaterThan(min),
                                     _Product.price.lessThan(max),
                                     Require.any(_Product.color.equal(BLUE),
                                                 _Product.color.equal(GREEN))),
                         Order.by(_Product.name.asc(),
                                  _Product.id.asc()));
```

Or we could think of another way to cover and/or/not. That part can be figured out later. The basic idea here will be to add the restrictions to the static metamodel and decide how type safe to make them. For example, requiring `Number` for numeric attributes, or possibly more precise, or just Object for everything, ...

gavinking:
```java
                         _Product.price.greaterThan(min)
                             .and(_Product.price.lessThan(max))
                             .and(_Product.color.equal(BLUE)
                                 .or(_Product.color.equal(GREEN))),
```

[...]

PR #867, opened 2024-10-10
Update PageRequest.java: fix small JavaDoc issue

hohwille:
just stumbled across a `false` vs. `true` in your JavaDoc and thought I simply create a PR.

### DIFF
--- a/api/src/main/java/jakarta/data/page/PageRequest.java
+++ b/api/src/main/java/jakarta/data/page/PageRequest.java
@@ -31,7 +31,7 @@ import java.util.Optional;
  * <p>A query method of a repository may have a parameter of type
  * {@code PageRequest} if its return type indicates that it may return
  * multiple entities, that is, if its return type is an array type,
- * {@code List}, {@code Stream}, {@link Page}, or {@link CursoredPage}.
+ * {@link List}, {@code Stream}, {@link Page}, or {@link CursoredPage}.
  * The parameter of type {@code PageRequest} must occur after the method
  * parameters representing regular parameters of the query itself. For
  * example:</p>
@@ -67,7 +67,6 @@ import java.util.Optional;
  * </ul>
  */
 public interface PageRequest {
-
 
     /**
      * Creates a new page request with the given page number and with a default size of 10.
@@ -243,7 +242,6 @@ public interface PageRequest {
      */
     PageRequest size(int maxPageSize);
 
-
     /**
      * Returns an otherwise-equivalent page request with
      * {@link #requestTotal()} set to {@code false}, so that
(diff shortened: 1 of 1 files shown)


hohwille:
Just saw your contributing guidelines:
https://github.com/jakartaee/data/blob/f7fbd095443c43d9c3a62ed75ac196f405ac8c1d/CONTRIBUTING.adoc#developer-certificate-of-origin---dco

Feel free to close for this small change, probably not worth it. You can also replace the inverted boolean literal yourself...
In case I ever create another PR, I will follow your rules...

hohwille:
p.s.: when an existing thing out in the wild is later considered for standardization things could be challenged more.
JPA is so full of flaws that have been blindly taken over from hibernate.
Jakarta data is on the same way with the "magic query from method name" approach adopted from spring-data.

All just my very personal opinion - feel free to ignore ;)

api/src/main/java/jakarta/data/page/PageRequest.java
```diff
@@ -31,7 +31,7 @@
  * <p>A query method of a repository may have a parameter of type
  * {@code PageRequest} if its return type indicates that it may return
  * multiple entities, that is, if its return type is an array type,
- * {@code List}, {@code Stream}, {@link Page}, or {@link CursoredPage}.
+ * {@link List}, {@code Stream}, {@link Page}, or {@link CursoredPage}.
```
otaviojava, line 34: Maybe moving all of those tags to `link`

njr-11:
@hohwille thanks for your feedback.

> Just saw your contributing guidelines:
> https://github.com/jakartaee/data/blob/f7fbd095443c43d9c3a62ed75ac196f405ac8c1d/CONTRIBUTING.adoc#developer-certificate-of-origin---dco

This is standard process for all Jakarta EE projects under Eclipse. If you are able to, I'd encourage you to sign the Eclipse Contributor Agreement so that we can merge your PR.

> Jakarta data is on the same way with the "magic query from method name" approach

The Query by Method Name does offer some compatibility with existing solutions, but looking forward to the next version, we are planning some alternative patterns (such as #857 and #829).

> And having findAll by default is also a huge anti-pattern

I remember there was a review of BasicRepository methods prior to version 1.0, due to which several were omitted. I don't recall why findAll made it through.  While it could be useful in some specific scenarios, in general, it is as you said, and anti-pattern. I would have rather seen users opt in to it than get it by default.

njr-11:
Thanks again for spotting this mistake and I'm sorry it didn't work out to sign the Eclipse Contributor Agreement in order to take your contribution directly.  I'm closing this because the fix was added by another participant under #959

Issue #373, opened 2023-11-23
Lifecycle CDI Event


hantsy:
Like the JPA event listener and do side effect by annotations `@Save`, `@Merge`, etc(till now I do not know well about the purpose of these annotations).

Spring Data provides a collection of `XXXCallback` when converting/saving/updating/delete an entity object.

In Jakarta Data, we can introduce CDI Event instead, make it compatible with all implementation modules.

It is more freely for developers.

gavinking:
> In Jakarta Data, we can introduce CDI Event instead, make it compatible with all implementation modules.

It's an appealing idea to use the CDI event bus for this, actually. Definitely worth considering in future.

[But I doubt it's a priority for 1.0.]

gavinking:
I think I'll take a crack at implementing this tonight, make sure it actually works...

gavinking:
> I think I'll take a crack at implementing this tonight, make sure it actually works...

Just implemented it. Works beautifully. Love it.

Issue #374, opened 2023-11-23
Introduce Aggregate Root, Entity, ValueObject, DomainEvent abstracts


hantsy:
In Eric's DDD, a `Repository` is tightly tied with Aggregate Root, and used to maintain the lifecycle of a Aggregate Root. And a Aggregate Root could be consist of Entity, Value Objects, and can handle domain events. 

In a JPA, it is easy to maintain the lifecycle like relation between PurchaseOrder and OrderItems, but switching to Jdbc, it should apple changes to the whole Aggregate when calling a repository save/update etc.

gavinking:
We discussed this issue quickly on the Jakarta Data call yesterday.

It's my opinion&mdash;and I believe that the others agree&mdash;that:

1. Jakarta Data 1.1 provides sufficiently-powerful primitives to express almost any DDD-style Repository that a user could possibly wish to write, but that
2. the specification itself should not directly encode particular patterns or ideologies which are embraced by only a small sliver of the Jakarta community.

The idea of "design patterns" is that they're patterns, not APIs.

On the call we agreed that this issue can now be closed.

hantsy:
We can not avoid the patterns in the real world application development.

If we do not embrace the concept of DDD, here the Data Repository is just a DAO in the classic J2EE Design Patterns.

In an AggregateRoot, we hope can maintain the data easily in a root graph.

gavinking:
> We cannot avoid the patterns in the real world application development.

I mean, sure, but if a pattern requires cannot be implemented without a the use of a specific purpose-designed framework, then it's not a very good pattern.

> If we do not embrace the concept of DDD, here the Data Repository is just a DAO in the classic J2EE Design Patterns.

OK, if you want to use the "Repository" to mean "Repository in DDD", I can't stop you. *I'm* not going to use the word like that, but nothing requires us to agree on definitions of words. 

So yes, concretely, I agree that "Repository" and "DAO" are in fact synonyms.

> In an AggregateRoot, we hope can maintain the data easily in a root graph.

Nothing is stopping you from doing that.

But, just in case you're interested, here's an explanation as to why "aggregates" are a terrible idea when working with relational (as opposed to document or hierarchical) databases.

<img width="835" alt="Image" src="https://github.com/user-attachments/assets/666daec3-29d6-4c2f-afe1-667d91b0828a" />

gavinking:
[...]

PR #876, opened 2024-10-24
first rough proposal for events

gavinking:
See #373. [Note that the first to propose this was @hantsy.]

Here's a rough and fairly minimal idea for what this could look like. 

Note that here there is an `Event` class for each kind of event, instead of a single `LifecycleEvent` with `@Before @Update` and `@After @Insert` qualifiers on the observer side. The reason for this is that we don't have a dependency on CDI, so I can't use the `@Qualifier` annotation to declare event qualifier types.

An event observer would look like:

```java
void afterUpdate(@Observes PostUpdateEvent<Book> event) { ... }
```

~I would love to be able to filter by entity type, but I don't think CDI allows filtering events by type argument. (At least it didn't in 1.0, though that was a long time ago.)~

~As a workaround for that, we could have `@Before @Update @EntityType(Book.class)` as qualifiers, but, again, that would pick up a dependency on CDI.~

### DIFF
--- a/api/src/main/java/jakarta/data/event/LifecycleEvent.java
+++ b/api/src/main/java/jakarta/data/event/LifecycleEvent.java
@@ -1,0 +1,29 @@
+package jakarta.data.event;
+
+/**
+ * <p>Abstract supertype of events relating to lifecycle methods.</p>
+ * <p>In Jakarta EE, a bean may observe such events via CDI:</p>
+ * <pre>
+ * void onInsertBook(&#64;Observes PostInsertEvent&lt;Book&gt; bookInsertion) {
+ *     Book book = bookInsertion.entity();
+ *     ...
+ * }
+ * </pre>
+ *
+ * @param <E> the entity type
+ */
+public abstract class LifecycleEvent<E> {
+    private final E entity;
+
+    public LifecycleEvent(E entity) {
+        this.entity = entity;
+    }
+
+    /**
+     * The entity instance which was passed as an argument to
+     * the lifecycle method.
+     */
+    public E entity() {
+        return entity;
+    }
+}
--- a/api/src/main/java/jakarta/data/event/LifecycleEvent.java
+++ b/api/src/main/java/jakarta/data/event/LifecycleEvent.java
@@ -37,8 +37,32 @@ public abstract class LifecycleEvent<E> {
     }
 
     /**
-     * The entity instance which was passed as an argument to
-     * the lifecycle method.
+     * The entity which is being processed by the lifecycle method.
+     * <ul>
+     * <li>For a {@code Pre} event, this is always the instance
+     *     which was passed as an argument to the lifecycle
+     *     method, and its state reflects the state of the
+     *     entity before execution of the lifecycle method.</li>
+     * <li>For a {@code Post} event, it may or may not be
+     *     identical to the object passed as an argument
+     *     to the lifecycle method, and it may or may not be
+     *     identical to the instance returned by the lifecycle
+     *     method, if any. If the state of the entity changes
+     *     as a result of execution of the lifecycle method,
+     *     those changes may or may not be reflected in the
+     *     entity returned by this method.</li>
+     * </ul>
+     * <p>
+     * Thus, a portable application should not assume that the
+     * state of the entity in a {@code Post} event faithfully
+     * reflects the current state of the corresponding record
+     * in the database.
+     * <p>
+     * A portable application must not mutate the state of the
+     * entity instance returned by this method. If the state
+     * of the entity instance is mutated during while event
+     * listeners are being notified, then the resulting
+     * behavior is undefined and unportable.
      */
     public E entity() {
         return entity;
--- a/api/src/main/java/jakarta/data/event/LifecycleEvent.java
+++ b/api/src/main/java/jakarta/data/event/LifecycleEvent.java
@@ -1,3 +1,20 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
 package jakarta.data.event;
 
 /**
--- a/api/src/main/java/jakarta/data/event/LifecycleEvent.java
+++ b/api/src/main/java/jakarta/data/event/LifecycleEvent.java
@@ -26,6 +26,25 @@ package jakarta.data.event;
  *     ...
  * }
  * </pre>
+ * <p>As usual for a CDI event, an observer of a {@code LifecycleEvent}
+ * is notified synchronously and immediately by default. An observer may
+ * elect to receive notifications during a phase of the transaction
+ * completion cycle by explicitly specifying a {@code TransactionPhase},
+ * for example:</p>
+ * <ul>
+ * <li>{@code @Observes(during=BEFORE_COMPLETION)} to be notified just
+ *     before transaction completion, or</li>
+ * <li>{@code @Observes(during=AFTER_SUCCESS)} to be notified after
+ *     successful completion of the transaction.</li>
+ * </ul>
+ * <p>An observer may choose to be notified asynchronously using
+ * {@code @ObservesAsync}. However, the mutable state held by the
+ * {@link #entity} is not in general safe for concurrent access,
+ * and so portable applications must not use {@code @ObservesAsync}
+ * to observe a {@code LifecycleEvent}. If the state of an entity is
+ * accessed from an asynchronous observer method for a lifecycle
+ * event, the resulting behavior is undefined and unportable.</p>
+ *
  *
  * @param <E> the entity type
  */
@@ -60,9 +79,9 @@ public abstract class LifecycleEvent<E> {
      * <p>
      * A portable application must not mutate the state of the
      * entity instance returned by this method. If the state
-     * of the entity instance is mutated during while event
-     * listeners are being notified, then the resulting
-     * behavior is undefined and unportable.
+     * of the entity instance is mutated while event listeners
+     * are being notified, the resulting behavior is undefined
+     * and unportable.
      */
     public E entity() {
         return entity;
--- a/api/src/main/java/jakarta/data/event/PostDeleteEvent.java
+++ b/api/src/main/java/jakarta/data/event/PostDeleteEvent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/api/src/main/java/jakarta/data/event/PostDeleteEvent.java
+++ b/api/src/main/java/jakarta/data/event/PostDeleteEvent.java
@@ -18,8 +18,8 @@
 package jakarta.data.event;
 
 /**
- * An event that occurs when an {@link jakarta.data.repository.Delete}
- * lifecycle method is called, and after each record is deleted from the datastore.
+ * An event that occurs when a {@link jakarta.data.repository.Delete}
+ * lifecycle method is called, after each record is deleted from the datastore.
  *
  * @param <E> the entity type
  */
--- a/api/src/main/java/jakarta/data/event/PostDeleteEvent.java
+++ b/api/src/main/java/jakarta/data/event/PostDeleteEvent.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.event;
+
+/**
+ * An event that occurs when an {@link jakarta.data.repository.Delete}
+ * method is called, and after the record was deleted from the datastore.
+ *
+ * @param <E> the entity type
+ */
+public class PostDeleteEvent<E> extends LifecycleEvent<E> {
+    public PostDeleteEvent(E entity) {
+        super(entity);
+    }
+}
--- a/api/src/main/java/jakarta/data/event/PostDeleteEvent.java
+++ b/api/src/main/java/jakarta/data/event/PostDeleteEvent.java
@@ -19,7 +19,7 @@ package jakarta.data.event;
 
 /**
  * An event that occurs when an {@link jakarta.data.repository.Delete}
- * method is called, and after the record was deleted from the datastore.
+ * lifecycle method is called, and after each record is deleted from the datastore.
  *
  * @param <E> the entity type
  */
--- a/api/src/main/java/jakarta/data/event/PostInsertEvent.java
+++ b/api/src/main/java/jakarta/data/event/PostInsertEvent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/api/src/main/java/jakarta/data/event/PostInsertEvent.java
+++ b/api/src/main/java/jakarta/data/event/PostInsertEvent.java
@@ -19,7 +19,7 @@ package jakarta.data.event;
 
 /**
  * An event that occurs when an {@link jakarta.data.repository.Insert}
- * method is called, and after the record was inserted into the datastore.
+ * lifecycle method is called, and after each record is inserted into the datastore.
  *
  * @param <E> the entity type
  */
--- a/api/src/main/java/jakarta/data/event/PostInsertEvent.java
+++ b/api/src/main/java/jakarta/data/event/PostInsertEvent.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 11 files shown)


gavinking:
> I would love to be able to filter by entity type, but I don't think CDI allows filtering events by type argument.

An intriguing possibility here would be to generate event types into the static metamodel, so that you could write:

```java
void afterUpdate(@Observes _Book.PostUpdateEvent event) { ... }
```

gavinking:
> I don't think CDI allows filtering events by type argument. (At least it didn't in 1.0, though that was a long time ago.)

I take that back. Apparently they relaxed this restriction, and you can now fire a parameterized event as long as the type argument is _somehow_ resolvable, for example:

```java
    @Inject
    Event<MyEvent<String>> event;
```

Or:

```java
        beanManager.getEvent()
                .select(new TypeLiteral<MyEvent<String>>() {})
                .fire(new MyEvent<>(title));
```

So in fact we can use a parameterized event type, e.g. `PostUpdateEvent<Book>`.

api/src/main/java/jakarta/data/event/PostUpdateEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.event;
+
+/**
+ * An event that occurs when an {@link jakarta.data.repository.Update}
+ * method is called, and after the entity was updated in the datastore.
```
njr-11, line 22:
```suggestion
 * lifecycle method is called, and after each entity is updated in the datastore.
```

api/src/main/java/jakarta/data/event/PostUpdateEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @param <E> the entity type
+ */
+public class PostUpdateEvent<E> extends LifecycleEvent<E> {
+    public PostUpdateEvent(E entity) {
+        super(entity);
+    }
```
njr-11, line 29: if switching to interface,
```suggestion
public record PostUpdateEvent<E>(E entity) implements LifecycleEvent<E> {
```

api/src/main/java/jakarta/data/event/PreDeleteEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.event;
+
+/**
+ * An event that occurs when an {@link jakarta.data.repository.Delete}
+ * method is called, but before the record is deleted from the datastore.
```
njr-11, line 22:
```suggestion
 * lifecycle method is called, but before each record is deleted from the datastore.
```

api/src/main/java/jakarta/data/event/PreDeleteEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @param <E> the entity type
+ */
+public class PreDeleteEvent<E> extends LifecycleEvent<E> {
+    public PreDeleteEvent(E entity) {
+        super(entity);
+    }
```
njr-11, line 29: if switching to interface,
```suggestion
public record PreDeleteEvent<E>(E entity) implements LifecycleEvent<E> {
```

api/src/main/java/jakarta/data/event/PreInsertEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.event;
+
+/**
+ * An event that occurs when an {@link jakarta.data.repository.Insert}
+ * method is called, but before the record is inserted into the datastore.
```
njr-11, line 22:
```suggestion
 * lifecycle method is called, but before each record is inserted into the datastore.
```

api/src/main/java/jakarta/data/event/PreInsertEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @param <E> the entity type
+ */
+public class PreInsertEvent<E> extends LifecycleEvent<E> {
+    public PreInsertEvent(E entity) {
+        super(entity);
+    }
```
njr-11, line 29: if switching to interface,
```suggestion
public record PreInsertEvent<E>(E entity) implements LifecycleEvent<E> {
```

api/src/main/java/jakarta/data/event/PreUpdateEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.event;
+
+/**
+ * An event that occurs when an {@link jakarta.data.repository.Update}
+ * method is called, but before the entity is updated in the datastore.
```
njr-11, line 22:
```suggestion
 * lifecycle method is called, but before each entity is updated in the datastore.
```

api/src/main/java/jakarta/data/event/PreUpdateEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @param <E> the entity type
+ */
+public class PreUpdateEvent<E> extends LifecycleEvent<E> {
+    public PreUpdateEvent(E entity) {
+        super(entity);
+    }
```
njr-11, line 29: If switching to interface,
```suggestion
public record PreUpdateEvent<E>(E entity) implements LifecycleEvent<E> {
```

api/src/main/java/jakarta/data/repository/Delete.java
```diff
@@ -59,6 +59,11 @@
  * if the entity with a matching identifier does not have a matching version, the annotated method must raise
  * {@link jakarta.data.exceptions.OptimisticLockingFailureException}.
  * </p>
+ * <p>
+ * An event of type {@link jakarta.data.event.PreDeleteEvent} must be raised by the annotated method before the record
+ * is deleted. An event of type {@link jakarta.data.event.PostDeleteEvent} must be raised by the annotated method after
+ * the record was successfully deleted.
```
njr-11, line 65:
```suggestion
 * An event of type {@link jakarta.data.event.PreDeleteEvent} must be raised by the annotated lifecycle
 * method before each record is deleted. An event of type {@link jakarta.data.event.PostDeleteEvent}
 * must be raised by the annotated lifecycle method after each record is successfully deleted.
```

api/src/main/java/jakarta/data/repository/Insert.java
```diff
@@ -66,6 +66,11 @@
  * then the annotated method must raise {@link jakarta.data.exceptions.EntityExistsException}.
  * If the database follows the BASE model, or uses an append model to write data, this exception is not thrown.
  * </p>
+ * <p>
+ * An event of type {@link jakarta.data.event.PreInsertEvent} must be raised by the annotated method before the record
+ * is inserted. An event of type {@link jakarta.data.event.PostInsertEvent} must be raised by the annotated method after
+ * the record was successfully inserted.
```
njr-11, line 72:
```suggestion
 * An event of type {@link jakarta.data.event.PreInsertEvent} must be raised by the annotated lifecycle
 * method before each record is inserted. An event of type {@link jakarta.data.event.PostInsertEvent}
 * must be raised by the annotated lifecycle method after each record is successfully inserted.
```

api/src/main/java/jakarta/data/repository/Update.java
```diff
@@ -67,6 +67,11 @@
  * If the database follows the BASE model, or uses an append model to write data, the annotated method behaves the same
  * as the {@code @Insert} method.
  * </p>
+ * <p>
+ * An event of type {@link jakarta.data.event.PreUpdateEvent} must be raised by the annotated method before the record
+ * is updated. An event of type {@link jakarta.data.event.PostUpdateEvent} must be raised by the annotated method after
+ * the record was successfully updated.
```
njr-11, line 73:
```suggestion
 * An event of type {@link jakarta.data.event.PreUpdateEvent} must be raised by the annotated lifecycle
 * method before each record is updated. An event of type {@link jakarta.data.event.PostUpdateEvent}
 * must be raised by the annotated lifecycle method after each record is successfully updated.
```

njr-11 (changes requested):
This is nicely written and I like the change to use the parameterized entity type rather than Object.

I made some minor suggestions and proposed switching the Abstract supertype for the event into an interface.

I addition to the inline review comments, there are a few other things we will need to consider.

One important item is clarifying what the user is (or is not) allowed to do with the entity instance that they get within the event. Are they allowed to modify it? And if so, do those modifications get persisted to the datastore?

The entity value returned on the Post events is a difficult question to consider.  I can certainly see why you wrote it to state the value passed in the lifecycle method, because that makes the most sense for PostDeleteEvent (to avoid extra fetch of values) and is consistent with the Delete lifecycle method which only supports a void return value.
For the PostInsertEvent and PostUpdateEvent, I wonder if some implementations based on a JPA provider might inadvertently end up with the entity value updated to include an automatically generated id and autoincremented version, which would no longer match the passed in entity value.  And having that information might be useful to the user as well, so I could see a case for defining these differently, but then they would be inconsistent.  Not really sure what is best here, but I thought I would at least bring it up.

gavinking:
> One important item is clarifying what the user is (or is not) allowed to do with the entity instance that they get within the event. Are they allowed to modify it? And if so, do those modifications get persisted to the datastore?

We can look at section 3.6.2 of the JPA spec for inspiration here.

> For the PostInsertEvent and PostUpdateEvent, I wonder if some implementations based on a JPA provider might inadvertently end up with the entity value updated to include an automatically generated id and autoincremented version, which would no longer match the passed in entity value. 

Yeah good point. Hrrm. Not sure.

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -115,6 +117,11 @@ This fragment shows how the application might request injection of a `CarReposit
 
 This integration between CDI and Jakarta Data allows for seamless management of repository instances within Jakarta EE applications.
 
+==== CDI Events
```
hantsy, line 120: We should differentiate these lifecycle events from other events, such as Domain Event, see: https://github.com/jakartaee/data/issues/374

gavinking: Honestly I don't really see what the difference is. How is a "Domain Event" different from a lifecycle event for an entity that you're choosing call an "aggregate"?

hantsy: Technically, there is no difference.

gavinking: Right, that's what I thought.

njr-11:
> > For the PostInsertEvent and PostUpdateEvent, I wonder if some implementations based on a JPA provider might inadvertently end up with the entity value updated to include an automatically generated id and autoincremented version, which would no longer match the passed in entity value.
> 
> Yeah good point. Hrrm. Not sure.

I tried this out on the EclipseLink JPA provider and observed it to be doing the following:

After invoking `persist`, the AutoGenerated Id value was updated on the passed-in entity, but the Version value was not incremented.  After also invoking `flush`, then it updated the Version value on that entity instance.

After invoking `merge`, the Version value was not immediately incremented on the passed-in entity.  As in the former case, after also invoking `flush`, it then updated the Version value on the passed-in entity instance.

Note that Jakarta Data's CrudRepository.insert has language like, "After invoking this method, do not continue to use the instance that is supplied as a parameter. This method makes no guarantees about the state of the instance that is supplied as a parameter."  All of this makes me uncomfortable about supplying that same instance via the PostInsert/PostUpdate events. But I can also see why we don't want to require it to supply an updated entity instance. That might require some implementations to do extra fetching just for the sake of publishing events that the application might not even care about.  Leaving out the entity from the event makes the Post event not useful because you cannot correlate it with the Pre event.  I suppose one option would be to supply only the Id value on the Post event.  Not really sure what would be best.

gavinking:
> Not really sure what would be best.

I mean with things like this, there's always the option of saying "portable applications should not rely on .... ".

api/src/main/java/jakarta/data/event/LifecycleEvent.java
```diff
@@ -0,0 +1,46 @@
...
+public abstract class LifecycleEvent<E> {
+    private final E entity;
+
+    public LifecycleEvent(E entity) {
+        this.entity = entity;
+    }
+
+    /**
+     * The entity instance which was passed as an argument to
+     * the lifecycle method.
```
njr-11, line 41:
```suggestion
     * <p>The entity instance which was passed as an argument to
     * the lifecycle method.<p>
     *
     * <p>Portable applications must not mutate the event's entity instance and
     * must not rely on the state of the entity instance outside the scope of the
     * {@link PreDeleteEvent}, {@link PreInsertEvent}, or {@link PreUpdateEvent}.</p>
```

njr-11 (changes requested):
In addition to the suggested updated, the other item we discussed during today's Jakarta Data spec meeting was making the Post events not include the entity (and alternatively maybe include the Pre event instead if correlation is needed).

To do this, LifecycleEvent will probably need to be split into LifecyclePreEvent and LifecyclePostEvent.

njr-11:
I suppose another alternative is to have the Post event be defined to include the same entity instance as the corresponding Pre event, with the restriction that portable applications must not rely on its state.  In this case, the event subscriber would be able to portably perform a == comparison to correlate the events.

gavinking:
Well they can also use the id.

njr-11:
> Well they can also use the id.

Yes, for update, delete, and insert that doesn't use autogenerated.

gavinking:
> insert that doesn't use autogenerated

Ah yeah, sure, you're right.

gavinking:
I have taken a stab at the semantics of the entity in `Post` events.

WDYT?

api/src/main/java/jakarta/data/event/LifecycleEvent.java
```diff
@@ -0,0 +1,70 @@
...
+     * Thus, a portable application should not assume that the
+     * state of the entity in a {@code Post} event faithfully
+     * reflects the current state of the corresponding record
+     * in the database.
+     * <p>
+     * A portable application must not mutate the state of the
+     * entity instance returned by this method. If the state
+     * of the entity instance is mutated during while event
+     * listeners are being notified, then the resulting
+     * behavior is undefined and unportable.
```
njr-11, line 65: Correcting a minor typo:
```suggestion
     * entity instance returned by this method. If the state
     * of the entity instance is mutated by event
     * listeners, then the resulting
     * behavior is undefined and unportable.
```

api/src/main/java/jakarta/data/event/PostDeleteEvent.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
```
njr-11, line 2:
```suggestion
 * Copyright (c) 2024 Contributors to the Eclipse Foundation
```

api/src/main/java/jakarta/data/event/PostDeleteEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.event;
+
+/**
+ * An event that occurs when an {@link jakarta.data.repository.Delete}
+ * lifecycle method is called, and after each record is deleted from the datastore.
```
njr-11, line 22: Adjust wording to avoid possible misinterpretation that this post event occurs both when Delete is called and after a record is deleted.
```suggestion
 * An event that occurs when a {@link jakarta.data.repository.Delete}
 * lifecycle method is called, after each record is deleted from the datastore.
```

api/src/main/java/jakarta/data/event/PostInsertEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.event;
+
+/**
+ * An event that occurs when an {@link jakarta.data.repository.Insert}
+ * lifecycle method is called, and after each record is inserted into the datastore.
```
njr-11, line 22:
```suggestion
 * An event that occurs when an {@link jakarta.data.repository.Insert}
 * lifecycle method is called, after each record is inserted into the datastore.
```

api/src/main/java/jakarta/data/event/PostInsertEvent.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
```
njr-11, line 2:
```suggestion
 * Copyright (c) 2024 Contributors to the Eclipse Foundation
```

api/src/main/java/jakarta/data/event/PostUpdateEvent.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
```
njr-11, line 2:
```suggestion
 * Copyright (c) 2024 Contributors to the Eclipse Foundation
```

api/src/main/java/jakarta/data/event/PostUpdateEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.event;
+
+/**
+ * An event that occurs when an {@link jakarta.data.repository.Update}
+ * lifecycle method is called, and after each entity is updated in the datastore.
```
njr-11, line 22:
```suggestion
 * An event that occurs when an {@link jakarta.data.repository.Update}
 * lifecycle method is called, after each entity is updated in the datastore.
```

api/src/main/java/jakarta/data/event/PreDeleteEvent.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
```
njr-11, line 2:
```suggestion
 * Copyright (c) 2024 Contributors to the Eclipse Foundation
```

njr-11 (changes requested):
Looks good.  I added some minor typo corrections and copyright corrections.

njr-11:
I should also point out that if the application is using transactions and rolls back a successful life cycle operation, the post event will have been sent even though the change is not committed to the database. For example,

```java
tran.begin();
books.update(modifiedBook); // Pre and Post events are sent here
...
tran.rollback();
```

Just pointing it out in case it is a concern.

hantsy:
Why not use `sealed` interface and `record` class for the events. This feature should be included after data 1.0, we can consider new Java features.

gavinking:
@hantsy theres no motivation for treating these as a sum type.

1. Client code doesn't switch  over them, and
2. anyway, it's an open universe here: the set of lifecycle method annotations is extensible (and we will extend it!) and so therefor is the set of lifecycle event types.

If we made this a sum type, then we would break client code whenever we add a new lifecycle method type.

P.S. this is for Data 1.1, so it's still targeting the same version of the Java platform.

gavinking:
> back in 2008, I anticipated that we would run into this issue sometime in 2024

Ok, I guess that sounds pretty implausible to you guys.

What really happened is that I just went back in time and added `TransactionPhase` to CDI 1.0 when I saw Nathan's comment.

njr-11 (approved):
> What really happened is that I just went back in time and added `TransactionPhase` to CDI 1.0 when I saw Nathan's comment.

However it was that you got `TransactionPhase` to already be in CDI, nice work!  It fits perfectly with this.  And good idea getting it into the LifecycleEvent Javadoc, which will help users know they can use it.

Everything I pointed out for this PR has been addressed, and I'm approving it now.

KyleAure (approved):

otaviojava:
For the first step, it is great, IMHO, I will merge, and we can refactor this code or the documentation with time.

gavinking:
Excellent, thanks folks!

PR #877, opened 2024-10-25
Adding sample of descending property to OrderBy Javadoc

mswatosh:
I noticed the class javadoc for `@OrderBy` doesn't mention or show an example of the `descending` property. I'm not sure if it's the case for other IDE's but in VSCode, the javadoc shown when hovering over an annotation doesn't include the properties, so without a description in the class javadoc, the availability of the `descending` property may not be obvious.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/OrderBy.java
+++ b/api/src/main/java/jakarta/data/repository/OrderBy.java
@@ -59,7 +59,7 @@ import java.lang.annotation.Target;
  *     {@code true}.</li>
  * </ul>
  * 
- * The default sort is ascending. The {@code descending} property can be
+ * The default sort order is ascending. The {@code descending} member can be
  * used to specify the sort direction.
  * <pre>
  * &#64;OrderBy(value = "price", descending = true)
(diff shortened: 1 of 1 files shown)


njr-11 (approved):
Thanks for contributing the improvement!

api/src/main/java/jakarta/data/repository/OrderBy.java
```diff
@@ -58,6 +58,13 @@
  * <li>ascending order for boolean values places {@code false} before
  *     {@code true}.</li>
  * </ul>
+ * 
+ * The default sort is ascending. The {@code descending} property can be
```
gavinking, line 62:
```suggestion
 * The default sort order is ascending. The {@code descending} member can be
```

api/src/main/java/jakarta/data/repository/OrderBy.java
```diff
@@ -58,6 +58,13 @@
...
+ * used to specify the sort direction.
```
gavinking, line 63:
```suggestion
 * used to specify the sort direction.</p>
```

gavinking (approved):
Thanks.

mswatosh:
The build didn't like the `</p>` change, but I think I understand what you were trying to fix and corrected it here: [06ffea1](https://github.com/jakartaee/data/pull/877/commits/06ffea14a8ce9dec0757c3a3be7d638664a15157)

PR #883, opened 2024-10-30
Javadoc example for By annotation has return type that does not make sense

njr-11:
An example in the By annotation Javadoc ought to be using a return type that can handle multiple results because the query can product multiple results.


mswatosh (approved):
Good catch 👍

gavinking (approved):

PR #884, opened 2024-10-30
Inconsistent method signature in pagination example

njr-11:
Javadoc has a section about pagination which has an example method

```
* Product[] findByNameLike(String pattern, PageRequest pageRequest, Order&lt;Product&gt; order);
```

that is supposed to be used for the following example usage pattern,

```
 * page1 = products.findByNameLikeAndPriceBetween(
 *                 namePattern, minPrice, maxPrice, page1Request,
 *                 Order.by(Sort.desc("price"), Sort.asc("id"));
```

The two are inconsistent. The methods have different names and accept a different number of parameters.

This PR updates them to match so that the example will make more sense.

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -798,7 +798,11 @@ import java.util.Set;
  * For example,</p>
  *
  * <pre>
- * Product[] findByNameLike(String pattern, PageRequest pageRequest, Order&lt;Product&gt; order);
+ * Product[] findByNameLikeAndPriceBetween(String pattern,
+ *                                         float minPrice,
+ *                                         float maxPrice,
+ *                                         PageRequest pageRequest,
+ *                                         Order&lt;Product&gt; order);
  *
  * ...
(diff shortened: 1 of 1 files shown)


api/src/main/java/module-info.java
```diff
@@ -798,7 +798,11 @@
  * For example,</p>
  *
  * <pre>
- * Product[] findByNameLike(String pattern, PageRequest pageRequest, Order&lt;Product&gt; order);
+ * Product[] findByNameLikeAndPriceBetween(String pattern,
+ *                                         float minPrice,
+ *                                         float maxPrice,
+ *                                         PageRequest pageRequest,
+ *                                         Order&lt;Product&gt; order);
```
njr-11, line 805: FYI - Here is the example in the Javadoc that this is supposed to correspond to:

```
 * page1 = products.findByNameLikeAndPriceBetween(
 *                 namePattern, minPrice, maxPrice, page1Request,
 *                 Order.by(Sort.desc("price"), Sort.asc("id"));
```

PR #892, opened 2024-11-05
other types of comparisons for parameter based automatic query

njr-11:
Here is what a solution to #857 might look like that enables comparisons other than equality for parameter based automatic query, making that pattern more complete and better able to match what Query by Method Name can do.  It is very concise and is all isolated to a single annotation in which the comparison types are specified by enum constants that read very naturally when used.

This allows us to switch basic code examples from Query by Method Name over to this more explicit pattern that does not seem like magic or require already having familiarity with query language.  I did so for several examples under this PR, although it might be best to view them in generated JavaDoc rather than source due to all of the escaping needed for `@`, `<`, `>` characters.

I didn't go through and comprehensively switch everything over because this approach will likely be debated for a while.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Is.java
+++ b/api/src/main/java/jakarta/data/repository/Is.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2024,2025 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.repository;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import jakarta.data.constraint.AtLeast;
+import jakarta.data.constraint.AtMost;
+import jakarta.data.constraint.Constraint;
+import jakarta.data.constraint.EqualTo;
+import jakarta.data.constraint.GreaterThan;
+import jakarta.data.constraint.In;
+import jakarta.data.constraint.LessThan;
+import jakarta.data.constraint.Like;
+import jakarta.data.constraint.NotEqualTo;
+import jakarta.data.constraint.NotIn;
+import jakarta.data.constraint.NotLike;
+
+/**
+ * <p>Annotates a parameter of a repository {@link Find} or {@link Delete}
+ * method, indicating how an entity attribute is compared with the parameter's
+ * value.</p>
+ *
+ * <p>The {@code @Is} annotation's {@link #value()} supplies the type of
+ * comparison as a subtype of {@link jakarta.data.constraint Constraint}.</p>
+ *
+ * <p>The {@link By} annotation must annotate the same parameter to indicate
+ * the entity attribute name, or otherwise, if the {@code -parameters} compile
+ * option is enabled, the persistent field is inferred by matching the name of
+ * the method parameter.</p>
+ *
+ * <p>For example,</p>
+ *
+ * <pre>
+ * &#64;Repository
+ * public interface Products extends CrudRepository&lt;Product, Long&gt; {
+ *
+ *     // Find Product entities where the price attribute is less than a maximum value.
+ *     &#64;Find
+ *     List&lt;Product&gt; pricedBelow(&#64;By(_Product.PRICE) &#64;Is(LessThan.class) float max);
+ *
+ *     // Find a page of Product entities where the name field matches a pattern.
+ *     &#64;Find
+ *     Page&lt;Product&gt; search(&#64;By(_Product.NAME) &#64;Is(Like.class) String pattern,
+ *                          PageRequest pagination,
+ *                          Order&lt;Product&gt; order);
+ *
+ *     // Remove Product entities with any of the unique identifiers listed.
+ *     &#64;Delete
+ *     void remove(&#64;By(ID) &#64;Is(In.class) List&lt;Long&gt; productIds);
+ * }
+ * </pre>
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.PARAMETER)
+public @interface Is {
+
+    /**
+     * <p>A subtype of {@link jakarta.data.constraint Constraint} that
+     * indicates how the entity attribute is compared with a value.</p>
+     *
+     * <p>The constraint subtype must have a static builder method that accepts
+     * as its only parameter a value compatible with the type (or if primitive,
+     * a wrapper for the type) of the repository method parameter to which the
+     * {@code @Is} annotation is applied. The repository method parameter type
+     * must also be consistent with the respective entity attribute type. This
+     * list indicates the constraint subtypes that can be used and links to the
+     * applicable builder method for each:</p>
+     *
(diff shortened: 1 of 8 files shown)


gavinking:
Honestly this isn't as bad as I had feared in terms of blowing up the API surface.

Nevertheless, I think we should seriously consider the alternative proposed by @beikov where `Pattern` is a class, and you can flexibly instantiate it, for example:

- `Pattern.literal("literal string").ignoreCase()`
- `Pattern.of("some%pattern")`
- `Pattern.of("some*pattern").wildcard('*')`

And for `>` and `<` I would just add a `Range` class:

- `Range.from(0)`
- `Range.to(100)`
- `Range.between('a', 'z')`

But if you really want typed upper and lower bounds, we could have `From` and `To` (or `Max`/`Min` or `UpperBound`/`LowerBound`, whatever) classes as well.

It seems to me that this just fits much more comfortably into the whole framework we've already established. It's consistent with how we handle `Limit`s, `Sort`s, `Restriction`s, etc, etc. And in certain ways it's more flexible.

You could write:

```java
List<Book> byTitleAndDate(Pattern title, Range <LocalDate> publicationDate);
```

And call it like this:

```java
var books = bookshop.byTitleAndDate(Pattern.of("Jakarta %")).ignoreCase(), 
                                    Range.between(LocalDate.now().minusYears(3), LocalDate.now()))
```

gavinking:
Ooooh, I also just realized that `Pattern` and `Range` might work rather nicely with `Restriction`s, in the sense that they simplify the number of distinct operations. Consider:

```java
_Book.title.like(Pattern.of("some%pattern").ignoreCase())
```

We now don't have to worry about distinct `like()` and `likeIgnoreCase()` operations.

Similarly:

```java
_Book.pages.in(Range.from(100))
_Book.publicationDate.in(Range.between(LocalDate.now().minusYears(3), LocalDate.now()))
```

I love the composability of this, and the potential for future extensibility.

And it nicely brings together and unifies the two things into one facility.

njr-11:
This PR/issue is aiming for a statically defined query alternative to Query by Method Name by filling in some gaps in parameter-based automatic methods.  The two comments above, which are good ideas to explore, are for dynamically / at-runtime composed queries, which we do also want to cover in Jakarta Data, but I think we should discuss them in the issue(s) for dynamic.

gavinking:
> This PR/issue is aiming for a statically defined query alternative to Query by Method Name by filling in some gaps in parameter-based automatic methods. The two comments above, which are good ideas to explore, are for dynamically / at-runtime composed queries, which we do also want to cover in Jakarta Data, but I think we should discuss them in the issue(s) for dynamic.

I understand that. But I guess I'm implicitly suggesting that this is not a very useful distinction to make _in the APIs_, and is leading us into fragmentation and Perl-like TMTOWTDI. What I've been looking for is solutions which don't involve fragmentation, but rather _unification_ of things we already have with things that we want to add.

njr-11:
This does unify with what we already have. It fits in perfectly with parameter-based automatic query and fills in a gaping hole where that pattern currently only offers the static pattern for equality comparisons and no static way of defining other useful comparisons like Query by Method Name does.

njr-11:
Per discussion in issue #857, I removed the Op enumeration, and made the constants in to Strings on the Is annotation. This allows for vendor extensions. It also makes the spec-provided options even more self documenting and obvious to the user.

api/src/main/java/jakarta/data/repository/Is.java
```diff
@@ -0,0 +1,103 @@
...
+ *     &#64;Delete
+ *     void remove(&#64;By(ID) &#64;Is(IN) List&lt;Long&gt; productIds);
+ * }
+ * </pre>
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.PARAMETER)
+public @interface Is {
+    // TODO add JavaDoc with examples to these
+    String EQUAL                = "EQUAL";
```
otaviojava, line 57: I understand the power of flexibility of `String`; however, a typo can be a disaster; plus, the enum can show the available options by default.


In this case, I still prefer the enum; if a vendor wants to externalize this behavior, it can easily create its annotation.

njr-11: I changed this to make it more agreeable to Gavin.  My own preference was for enum which I do still prefer, but I don't think it is a big deal one way or the other and would be happy with either.  If we did leave it this way, I'm not t all worried about typos because the user doesn't write these, they use the constants we define in the spec (or that a vendor defines in their API).  And yes, having a vendor extend behavior with an annotation of their seems perfectly reasonable to me, but I don't think Gavin is going to be on board with that and he was the one who brought up not being extensible as a drawback.  In any case, if the two of you can agree on one way or the other (enum or String constant), I'll switch it to that because I really am fine with either.

otaviojava: I would vote to keep an enum or reuse the `By` annotation to give the provider more flexibility.


https://github.com/jakartaee/data/issues/857#issuecomment-2461965944

njr-11: That's fine, I put the enum back for now. I was only anticipating that Gavin would prefer it with String constants to enable vendor extensions without an annotation, but I'm not actually sure of that.  This is also now updated with the shorter `ANY_CASE`, and I added in the full list of possible enum constants, which I realize might need to be pared down later to gain agreement on getting this in.

otaviojava:
@njr-11 could you rebase this one?

api/src/main/java/jakarta/data/repository/Is.java
gavinking: Just to be clear, I'm still a -1 on the whole approach. I think it's a path we shouldn't go down.

njr-11:
> @njr-11 could you rebase this one?

Yes, thanks for pointing out it needed a merge conflict resolved. It is updated now.

> Just to be clear, I'm still a -1 on the whole approach. I think it's a path we shouldn't go down.

Understood - this is an area where we have not reached agreement and will need more discussion. In the meantime, I have been trying to keep the PR up-to-date so that everyone has a clear understanding of it.

gavinking:
> Understood - this is an area where we have not reached agreement and will need more discussion. 

It would be good to talk it over on the next call. Unfortunately I have not been able to make it to any recent Data calls due to other commitments.

> In the meantime, I have been trying to keep the PR up-to-date so that everyone has a clear understanding of it.

Yes, fine, makes sense.

njr-11:
rebased and updated to reuse the existing `Operator` enum

api/src/main/java/jakarta/data/metamodel/restrict/Operator.java
```diff
@@ -27,14 +27,23 @@ public enum Operator {
...
     NOT_EQUAL,
     NOT_IN,
-    NOT_LIKE;
+    NOT_LIKE,
+    NOT_PREFIXED,
+    NOT_SUBSTRINGED,
+    NOT_SUFFIXED,
+    PREFIXED,
+    SUBSTRINGED,
+    SUFFIXED;
```
gavinking, line 36: FWIW I would not add these unless we were also going to add equivalent operations to JDQL.

njr-11: > FWIW I would not add these unless we were also going to add equivalent operations to JDQL.

In JDQL, these are `LEFT` and `RIGHT` (and substring/contains comes from using `LIKE` in a particular way - that one is less direct).  I copied them over from the `Is` annotation to avoid losing track of them.  If this does go in, there will need to be discussion over what the exact list should be.

api/src/main/java/jakarta/data/repository/Is.java
```diff
@@ -0,0 +1,120 @@
...
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.PARAMETER)
+public @interface Is {
+
+    /**
+     * <p>A subtype of {@link jakarta.data.constraint Constraint} that
+     * indicates how the entity attribute is compared with a value.</p>
+     *
+     * <p>The constraint subtype must have a static builder method that accepts
```
gavinking: I would stay away from the undefined term "builder method". Just saying: "a static method" is enough, I think.

njr-11: Sure, it's updated now.

Issue #460, opened 2024-02-20
Proposal for Criteria Feature in Jakarta Data[Vote] 


otaviojava:
### As a ...

- [X] Application user/user of the configuration itself
- [X] API user (application developer)
- [ ] SPI user (container or runtime developer)
- [ ] Specification implementer

### I need to be able to ...

Expanding upon the capabilities of the Jakarta Data Query Language ([#458](https://github.com/jakartaee/data/issues/458)), this proposal introduces the Criteria feature to enable programmatically constructing queries within the Jakarta Data framework. The Criteria API provides developers with an object-oriented approach to building query criteria, offering greater flexibility and control over data retrieval operations.

### Which enables me to ...

The primary goal of this proposal is to initiate a discussion and a vote regarding the inclusion of the Criteria feature in the Jakarta Data project.

**Key Features:**

1. **Programmatic Query Construction:** The Criteria API allows developers to programmatically build query criteria using a fluent and object-oriented interface, eliminating the need for raw queries.

2. **Filtering and Logical Conditions:** Developers can apply various filtration rules and logical conditions to the criteria query object, enabling precise data retrieval based on specific criteria.

3. **Integration with Jakarta Data Query Language:** The Criteria feature seamlessly integrates with the Jakarta Data Query Language, providing an additional method for constructing queries within the framework.

4. **Integration with Metadata:** Criteria queries can be combined with metadata-driven queries, offering developers multiple options for constructing queries based on their requirements.

5. **Integration with Repository:** Introducing a new interface, `CriteriaRepository,` that extends the existing `PageableRepository,` and enables seamless integration of criteria-based queries with repository pattern implementations.

### Additional information

**Example Usage:**

```java
// Combined with Metadata
List<Car> cars = template
    .select(Car.class)
    .where(_Car.city.eq("Rome").and(_Car.active.isTrue()))
    .orderBy(_Car.name.asc())
    .result();

// Combined with Repository
@Repository
public interface CardEntityRepository extends PageableRepository<Car, String>, CriteriaRepository<Car> {

    default Page<Car> findAll(String city, Pageable pageable) {
        Specification<Car> criteria = (root, query, criteriaBuilder) -> {
            List<Predicate> predicates = new ArrayList<>();
            predicates.add(criteriaBuilder.equal(root.get("city"), city));
            return criteriaBuilder.and(predicates.toArray(new Predicate[0]));
        };
        return findAll(criteria, pageable);
    }
}

// Pure Implementation
[...]

PR #894, opened 2024-11-05
Add Criteria-Based Filtering

otaviojava:
This PR introduces enhanced attribute-based querying capabilities to Jakarta Data, consolidating query construction into Attribute, SortableAttribute, and TextAttribute interfaces with corresponding implementations. These updates provide:

* A streamlined and type-safe approach to filtering and sorting.
* Flexible criteria-based filtering with support for equality, null checks, range queries, and pattern-based matching.
* Inline declarative filtering and sorting, simplifying query construction.

```java

@Repository
public interface Library {
    List<Book> find(CompositeRestriction<Book> restriction, Sort<Book> order);
}

// Usage in a service or query:
CompositeRestriction<Book> restriction = CompositeRestriction.all(
    _Book.title.like(Pattern.prefixed("Jakarta")),
    _Book.edition.equal(1),
    _Book.publicationDate.between(pastDate, LocalDate.now())
);

Sort<Book> order = _Book.title.asc();

// Executing the query
List<Book> results = library.find(restriction, order);
```

```java
Criteria nameCriteria = _Book.title.like("Data%");
Criteria dateCriteria = _Book.publicationDate.between(pastDate, LocalDate.now());
List<Book> books = bookRepository.findByCriteria(nameCriteria, dateCriteria);
```


otaviojava:
I see both working together:

- https://github.com/jakartaee/data/issues/829
- https://github.com/jakartaee/data/issues/460

Instead of having one API for each other,

api/src/main/java/jakarta/data/Criteria.java
```diff
@@ -0,0 +1,67 @@
...
+     * Creates a basic criteria condition without negation.
+     */
+    public static Criteria where(String field, FilterType operator, Object value) {
+        return new Criteria(field, operator, value, false);
+    }
+
+    /**
+     * Creates a criteria condition with negation.
+     */
+    public static Criteria not(String field, FilterType operator, Object value) {
```
otaviojava, line 50: I could include more factory methods here

api/src/main/java/jakarta/data/FilterType.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+package jakarta.data;
+
+
+/**
+ * Enum representing common filter types used to compare, match, or restrict values in Jakarta Data queries.
+ * The {@code FilterType} enum defines a set of operators to specify how values should be compared or matched
+ * in data filtering, providing flexibility for various querying needs across different contexts.
+ */
+public enum FilterType {
```
otaviojava, line 26: I put looking only for NoSQL, but for sure, we could include SQL and put as UnsupportedException when it does not work on NoSQL databases.

api/src/main/java/jakarta/data/metamodel/FilterableAttribute.java
```diff
@@ -0,0 +1,118 @@
...
+ */
+public interface FilterableAttribute<T, V> extends Attribute<T> {
+
+    /**
+     * Creates an equality condition for the attribute, matching the specified value.
+     *
+     * @param value the value to compare with the attribute.
+     * @return a criteria condition representing "attribute = value".
+     */
+    Criteria equal(V value);
```
otaviojava, line 51: Here, I am not sure if it should be `eq` instead of `equal`

api/src/main/java/jakarta/data/repository/CriteriaRepository.java
```diff
@@ -0,0 +1,83 @@
...
+ * List<DriverLicense> allResults = licenses.findByCriteria(
+ *     _DriverLicense.license.like("ABC%"),
+ *     _DriverLicense.expiry.greaterThan(LocalDate.now())
+ * );
+ * </pre>
+ *
+ * @param <T> the type of the primary entity class of the repository.
+ * @param <K> the type of the unique identifier field or property of the primary entity.
+ */
+public interface CriteriaRepository<T, K> extends DataRepository<T, K> {
```
otaviojava, line 56: About this new repository, I'm just wondering if it makes sense.

api/src/main/java/jakarta/data/metamodel/FilterableAttribute.java
```diff
@@ -0,0 +1,118 @@
...
+ * Criteria authorCriteria = _Book.author.equal("John Doe");
+ *
+ * // Apply criteria in a repository query
+ * repo.books(List.of(titleCriteria, dateCriteria, authorCriteria));
+ * }</pre>
+ *
+ * @param <T> the type of the entity class in the static metamodel.
+ * @param <V> the type of the attribute's value.
+ */
+public interface FilterableAttribute<T, V> extends Attribute<T> {
```
otaviojava, line 43: I would love to have more specializations, such as Boolean, with `isTrue` and `isFalse`methods on `String` `contains` and `like` and so on.

gavinking:
Some comments:

1. The `Criteria` type should be parameterized by the entity type, for type safety.
2. I prefer `Restriction` and `Operator` to `Criteria`  and `FilterType`. In particular, "criteria" is a really badly overloaded word, and potentially very confusing here.
3. I would make `not()` or `negated()` a method of `Restriction`/`Criteria`, not a separate static factory method.
4. Unless I've missed it, the binary logical operations are missing from this proposal. I suppose they should follow the pattern we already established with `Sort` and `Order`.
5. I believe you're going to have warnings if you put varargs on an interface, so it's better if those operations on `CriteriaRepository` accept a single conjunction of restrictions, instead of multiple restrictions.
6. I'm still extremely unexcited by the idea of embedding an expression language in annotations: `@Is(value=LIKE, not=true)` doesn't float my boat at all.

Apart from those comments this is pretty much along the lines of what I was hoping for.

njr-11:
Yes, this is generally the right ideas.

>     1. The `Criteria` type should be parameterized by the entity type, for type safety.

agree

>     2. I prefer `Restriction` and `Operator` to `Criteria`  and `FilterType`. In particular, "criteria" is a really badly overloaded word, and potentially very confusing here.

agree (I actually have some of this coded up locally using the name Restriction, but hadn't finished it yet)
 
>     3. I would make `not()` or `negated()` a method of `Restriction`/`Criteria`, not a separate static factory method.

I've been going back and forth on how best to handle not/negate. There are other options as well beyond what is in this PR. I need to think about this some more.

>     4. Unless I've missed it, the binary logical operations are missing from this proposal. I suppose they should follow the pattern we already established with `Sort` and `Order`.

yes

>     5. I believe you're going to have warnings if you put varargs on an interface, so it's better if those operations on `CriteriaRepository` accept a single conjunction of restrictions, instead of multiple restrictions.

First, I don't think we need a CriteriaRepository. But the same issue would arise regardless when a user composes a method with multiple restrictions.  The way I wrote it up (and which I've also tried locally) is to have Restrict.all(Restriction...) and Restrict.any(Restriction...) which basically covers item 4 above (ignoring that I'm still considering options for item 3 on not/negate)

>     6. I'm still extremely unexcited by the idea of embedding an expression language in annotations: `@Is(value=LIKE, not=true)` doesn't float my boat at all.

I'm very much in favor of `@Is` and really want to include it.  However, for the ideal usability it should only ever have a single field being the enumerated value, so it would be `@Is(NOT_LIKE)` which reads much better.  See PR #892 which I had already written up for it.

api/src/main/java/jakarta/data/metamodel/FilterableAttribute.java
gavinking: JPQL doesn't currently have `is true` and `is false` operators. I think you're supposed to write `= true` and `= false`. So here you would just write `_Entity.booleanField.equals(true)`, for example.

JPQL *does* have `is null`, which looks like it's missing.

gavinking:
> The way I wrote it up (and which I've also tried locally) is to have `Restrict.all(Restriction...)` and `Restrict.any(Restriction...)` which basically covers item 4 above

Yup, that's exactly what I meant by following the precedent already established by `Sort`/`Order`.

api/src/main/java/jakarta/data/metamodel/FilterableAttribute.java
otaviojava: I mean, by abstractions.

```java
_Car.available.isTrue();
```

The metadata will do it properly as you said.

otaviojava:
> 1. The `Criteria` type should be parameterized by the entity type, for type safety.
Done

> 2. I prefer `Restriction` and `Operator` to `Criteria`  and `FilterType`. In particular, "criteria" is a really badly overloaded word, and potentially very confusing here.

Done
> 3. I would make `not()` or `negated()` a method of `Restriction`/`Criteria`, not a separate static factory method.

Do you have any suggestion of how?

> 4. Unless I've missed it, the binary logical operations are missing from this proposal. I suppose they should follow the pattern we already established with `Sort` and `Order`.

My initial thought was, by default, working only with `AND`but I am fine to allow `OR`as well.
What do you think now?

> 6. I believe you're going to have warnings if you put varargs on an interface, so it's better if those operations on `CriteriaRepository` accept a single conjunction of restrictions, instead of multiple restrictions.

I've created a collection representation, what do you think?

otaviojava:
> I've been going back and forth on how best to handle not/negate. There are other options as well beyond what is in this PR. I need to think about this some more.

@njr-11 could you bring it to this PR?

otaviojava:
@njr-11 @gavinking does it makes sense?

gavinking:
OK, so this is quite a bit different now. :) I have two questions:

1. would it be better to implement these new methods as `default` methods of the `Attribute` interface?
2. would it have been better to leave `Restriction` as a `record`, and the also have `Restrict` as the composite type, like what we do with `Sort` and `Order`, rather than the new design here which makes `Restriction` an interface?

otaviojava:
> * would it be better to implement these new methods as `default` methods of the `Attribute` interface?

1. Greate! I moved it might not break anything
2. I've moved to the interface because the `Pattern` will implement it as well and there is the `CompositeRestriction`

njr-11:
> 1. would it be better to implement these new methods as `default` methods of the `Attribute` interface?

This is an excellent idea. It eliminates the problem that would have occurred if any annotation processors were supplying their own *Attribute implementations rather than using the provided ones. With the default methods in place, even if they were doing that, there will be no breaking change at all.

api/src/main/java/jakarta/data/Operator.java
```diff
@@ -0,0 +1,88 @@
...
+    BETWEEN,
+
+    /**
+     * Matches records where the field value is null.
+     *
+     * <p>This operator is used to filter records where the specified field
+     * does not contain a value. Commonly used to find records with unset or
+     * missing data in a particular field.</p>
+     */
+    IS_NULL
```
njr-11, line 84: You can use EQUALS with value of null instead.
Also, this makes no sense if shared with `Is`, which annotates a supplied parameter
```suggestion
```

njr-11 (changes requested):
I added some review comments to start out with.  It is definitely making progress and moving toward converging with how I wrote it up in the other PR

gavinking:
Here's a minimal example of what `Pattern` and `Range` could look like.

```java
public record Pattern(String pattern, boolean caseSensitive) {

    public Pattern ignoringCase() {
        return new Pattern(pattern, caseSensitive);
    }

    public static Pattern is(String literal) {
        return new Pattern(escape(literal), true);
    }

    public static Pattern matches(String pattern) {
        return new Pattern(pattern, true);
    }

    public static Pattern matches(String pattern,
                                  char characterWildcard, char stringWildcard) {
        final String standardized =
                escape(pattern)
                        .replace(characterWildcard, '_')
                        .replace(stringWildcard, '%');
        return new Pattern(standardized, true);
    }

    public static Pattern startsWith(String prefix) {
        return new Pattern(escape(prefix) + '%', true);
    }

    public static Pattern endsWith(String suffix) {
        return new Pattern('%' + escape(suffix), true);
    }

    public static Pattern contains(String substring) {
        return new Pattern('%' + escape(substring) + '%', true);
    }

    private static String escape(String literal) {
        return literal.replace("_", "\\_").replace("%", "\\%");
    }
}
```

And:

```java
public record Range<T>(T lowerBound, T upperBound, boolean open) {

    public static <T> Range<T> between(T lowerBound, T upperBound) {
        return new Range<>(lowerBound, upperBound, false);
    }

    public static <T> Range<T> from(T lowerBound) {
        return new Range<>(lowerBound, null, false);
    }

    public static <T> Range<T> to(T upperBound) {
        return new Range<>(null, upperBound, false);
    }

    public static <T> Range<T> above(T lowerBound) {
        return new Range<>(lowerBound, null, true);
    }

    public static <T> Range<T> below(T upperBound) {
        return new Range<>(null, upperBound, true);
    }
}
```

These are quite expressive, with a minimal amount of code, and almost nothing at all to specify in English.

We could also have `From` and `To` classes for typed upper and lower bounds if necessary. I would probably just make `Range` a class with two little subclasses in the case.

api/src/main/java/jakarta/data/repository/Pattern.java
```diff
@@ -0,0 +1,153 @@
...
+ * Restriction<Book> prefixIgnoreCase = Pattern.prefixedIgnoreCase("title", "Jak");
+ *
+ * Restriction<Book> suffixMatch = Pattern.suffixed("title", "Data");
+ *
+ * Restriction<Book> substringIgnoreCase = Pattern.substringedIgnoreCase("title", "Java");
+ * </pre>
+ *
+ * @param <T> the type of the entity on which the restriction is applied.
+ */
+public record Pattern<T>(String field, String value, boolean ignoreCase) implements Restriction<T> {
```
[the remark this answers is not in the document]
gavinking: Right, this is not quite what I had in mind.

api/src/main/java/jakarta/data/repository/Pattern.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * Creates a case-sensitive pattern match for an exact match.
+     *
+     * @param field the field to apply the pattern on.
+     * @param value the value to match exactly.
+     * @return a Pattern for an exact match.
+     */
+    public static <T> Pattern<T> like(String field, String value) {
```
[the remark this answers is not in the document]
gavinking: I agree, but I also note that `Sort` also has these untypesafe versions, and so there's precedent.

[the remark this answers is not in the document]
njr-11: Well, we wrote Sort prior to adding the metamodel. But in any case, we don't need to continue a precedent that doesn't make sense. I'd rather start with a more streamlined approach that directs users to do things in the recommended way without all the distractions/clutter tempting the to code it less desirable ways.  We can always add in those other parts later on if we find them to be useful.

api/src/main/java/jakarta/data/metamodel/TextAttribute.java
```diff
@@ -40,4 +43,84 @@ public interface TextAttribute<T> extends SortableAttribute<T> {
      */
     Sort<T> descIgnoreCase();
 
+    /**
+     * Creates a `LIKE` restriction for an exact match on the specified text.
+     *
+     * @param text the text to match exactly.
+     * @return a Restriction representing a case-sensitive `LIKE` condition.
```
njr-11, line 50: Jakarta Data has never guaranteed case sensitive comparisons. Some databases either are or have modes in which they are case insensitive, and the Jakarta Data provider has no control over that.  So we very careful wrote the documentation to say only there are ways to specifically request case insensitive comparisons, not that we can ever ensure that you will get case sensitive.

```suggestion
     * Creates a `LIKE` restriction for a match on the specified text.
     *
     * @param text the text to match.
     * @return a Restriction representing a `LIKE` condition.
```

api/src/main/java/jakarta/data/repository/Pattern.java
```diff
@@ -0,0 +1,153 @@
...
+
+
+/**
+ * Represents a pattern-based restriction for matching operations, encapsulating different
+ * options such as prefix, suffix, and substring matching. This implementation
+ * allows flexibility in creating pattern-based conditions directly as `Restriction` instances.
+ *
+ * <p>Example usage:</p>
+ * <pre>
+ * // Case-sensitive exact match
```
njr-11, line 32:
```suggestion
```

api/src/main/java/jakarta/data/repository/Pattern.java
```diff
@@ -0,0 +1,153 @@
...
+    public String value() {
+        return value;
+    }
+
+    /**
+     * Creates a case-sensitive pattern match for an exact match.
+     *
+     * @param field the field to apply the pattern on.
+     * @param value the value to match exactly.
+     * @return a Pattern for an exact match.
```
njr-11, line 71:
```suggestion
     * Creates a pattern to match.
     *
     * @param field the field to apply the pattern on.
     * @param value the value to match.
     * @return a Pattern for a match.
```

api/src/main/java/jakarta/data/Restriction.java
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * Creates a `LIKE` restriction using a `Pattern`, supporting complex `LIKE` operations.
+     * <pre>
+     * Restriction<Book> titleIgnoreCase = Restriction.like("title", Pattern.prefixedIgnoreCase("Java"));
+     * </pre>
+     */
+    static <T> Restriction<T> like(String field, Pattern<T> pattern) {
+        return pattern;
+    }
```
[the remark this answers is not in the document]
otaviojava: Ok, I will remove those!

api/src/main/java/jakarta/data/BasicRestriction.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data;
```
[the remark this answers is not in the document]
otaviojava: Done.

gavinking:
> These are quite expressive

I mean, actually I'm really *loving* the expressiveness and flexibility of being able to write:

```java
/**
 * Return issues matching the given patterned title and description, 
 * created within the given date range, and with one of the given
 * statuses, ordered by the given sorting criterion.
 */
List<Issue> issues(Pattern title, Pattern description, 
                   Range<LocalDate> created, List<Status> status,
                   Sort<Issue> order)
```

That's extremely compact, completely type-safe, completely readable.

api/src/main/java/jakarta/data/repository/Pattern.java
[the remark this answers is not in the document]
otaviojava: I hide the `Pattern`to be used only on `TextAttribute`.
I like the approach of given the user the right direction.

api/src/main/java/jakarta/data/repository/Pattern.java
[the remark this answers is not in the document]
otaviojava: This one is the default package that we are using only on `TextAttribute`. We can remove this one for sure.

otaviojava:
> I added some review comments to start out with. It is definitely making progress and moving toward converging with how I wrote it up in the other PR

Could you check If I missed some comments?
I get that I covered all of those.

njr-11:
> > These are quite expressive
> 
> I mean, actually I'm really _loving_ the expressiveness and flexibility of being able to write:
> 
> ```java
> /**
>  * Return issues matching the given patterned title and description, 
>  * created within the given date range, and with one of the given
>  * statuses, ordered by the given sorting criterion.
>  */
> List<Issue> issues(Pattern title, Pattern description, 
>                    Range<LocalDate> created, List<Status> status,
>                    Sort<Issue> order)
> ```
> 
> That's extremely compact, completely type-safe, completely readable.

That isn't actually type safe, but could be made so with `Pattern<Issue>` which you probably intended to include.

In any case, it would be more compact, more flexible, and also type-safe, to define,

```
List<Issue> issues(Restriction<Issue> restrictions,
                   Order<Issue> order)
```

Then a repository user who, for example, doesn't care about the date range, or maybe wants to also search by author, can still use the method without the repository provider needing to write a new one.

I'm struggling to see the advantage of being able to write repository methods that only offer flexibility on how to filter, but not on which attributes to filter.  I can see a good use case for the repository user wanting to do both, but only the latter seems less useful.

otaviojava:
> That isn't actually type safe, but could be made so with `Pattern<Issue>` which you probably intended to include.
ÍMHO it is:

```java

// Define pattern-based restrictions on the title and description fields
Pattern<Issue> titlePattern = _Issue.title.startsWith("Bug");
Pattern<Issue> descriptionPattern = _Issue.description.containsIgnoreCase("critical");

// Define a range for the created date using the between method directly on _Issue.created
Range<Issue> createdDateRange = _Issue.created.between(LocalDate.of(2023, 1, 1), LocalDate.now());

// Define sorting order (e.g., by created date in descending order)
Sort<Issue> order = _Issue.created.desc();

// Execute the query with the specified criteria
List<Issue> results = issueRepository.issues(titlePattern, descriptionPattern, createdDateRange, List.of(Status.OPEN, Status.IN_PROGRESS), order);

```

gavinking:
> That isn't actually type safe, but could be made so with `Pattern<Issue>` which you probably intended to include.

Not sure what you mean by this.

I think you've misunderstood what I'm saying here. 

> I'm struggling to see the advantage of being able to write repository methods that only offer flexibility on how to filter, but not on which attributes to filter.

Right, I'm saying that there are two ways to use `Pattern` and `Range`. For example:

```java

@Repository
interface Tracker {

    List<Issue> issues(Restriction<Issue> restriction, Order<Issue> order);

    List<Issue> issues(Pattern title, Pattern description, 
                       Range<LocalDate> created, List<Status> status,
                       Sort<Issue> order)

}
```

And calling them looks like:

```
var issues = repo.issues(Restrict.all(_Issue.title.like(Pattern.matches(title),
                                      _Issue.title.like(Pattern.match(description).ignoringCase(),
                                      _Issue.created.in(Range.from(LocaDate.now().minusYears(1))),
                                      Order.by(_Issue.created.asc()));

var issues = repo.issues(Pattern.matches(title), 
                         Pattern.matches(description).ignoringCase(),
                         Range.from(LocaDate.now().minusYears(1)),
                         _Issue.created.asc());

```

So you have one unified API, with very "small, compact" primitives, that kills both use-cases in a way that means that users can smoothly transition from one way of using the API to the other with no "friction".

gavinking:
@njr-11 I know you're modeling this as there are two completely different things:

- "static" repository methods
- repository methods  with "dynamic" restrictions

I'm saying these things exist on a spectrum and we can design this so that there's just one API that handles both cases elegantly and extensibly,

api/src/main/java/jakarta/data/repository/Pattern.java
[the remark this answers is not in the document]
gavinking: @otaviojava In my model `Pattern` isn't a parameterized type. It's just a sort of `String` with wildcards in it, and possibly case-insensitivity.

To get a `Restriction` based on a `Pattern` you do:

```java
_Book.title.like(Pattern.startsWith("Hibernate").ignoringCase())
```

Or, if, like me, you don't think there's anything wrong with static imports:

```java
_Book.title.like(startsWith("Hibernate").ignoringCase())
```

FTR, I'm completely happy with there also being an abbreviation for this, e.g. 

```java
_Book.name.likeIgnoringCase("Hibernate%")
```

But the ability to write `like(Pattern)` is a bit more flexible in certain ways.

njr-11:
I don't see the semi-static approach being that valuable.  If there were a fully static approach, then there can be optimization and a performance benefit to process repository methods (such as at compile time) and prepare reusable statements in advance. That was really the only useful aspect that I think we got from Query by Method Name. I proposed the `Is` annotation to give the Parameter-based automatic queries approach a way of gaining that benefit, after which there would be little reason for anyone to continue using Query by Method Name.  Technically, it would be possible for an approach like Pattern/Range to achieve that benefit as well, but only if there were additional parameter classes (Min, Max, ExclusiveMin, ExclusiveMax, Pattern, PatternIgnoringCase, NotPatternNegated, NotPatternIgnoringCase, and so forth)  to precisely indicate everything needed to generate a query upfront.  I did concede that we could get rid of startsWith/endsWith/contains because those can be represented with the Like pattern.  However, that is still a lot of classes to achieve that, and I doubt we want to go in that direction.  The `Is` annotation still seems the cleanest to me for the static pattern that actually lets the work be done in advance.

As soon as we introduce the possibility to have variability between where it might be comparing a field in one way vs another way, determined by whatever code the application provides to compute the repository method parameter, then none of that upfront work is possible.  I don't mean to argue against the dynamic approach.  There are many scenarios where it is very useful and we need it in the spec for that.  I want to have both approaches so that repositories can leverage each approach where it is suitable, and to benefit from both where it is possible to do so.

api/src/main/java/jakarta/data/repository/Pattern.java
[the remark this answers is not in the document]
otaviojava: Done, I've updated this one as well.

api/src/main/java/jakarta/data/metamodel/BasicRestriction.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * A basic implementation of the `Restriction` interface, representing various conditions
+ * used in repository queries, including equality, comparison, and range checks.
+ *
+ * @param <T> the type of the entity on which the restriction is applied.
+ */
+public record BasicRestriction<T>(String field, Operator operator, Object value) implements Restriction<T> {
+
+    /**
+     * Constructs a `BasicRestriction` with the specified field, operator, and value.
```
njr-11:
```suggestion
     * Constructs a {@code BasicRestriction} with the specified field, operator, and value.
```

otaviojava:
@njr-11 I've make the multiple restriction implement Restriction as you said in the last commit

api/src/main/java/jakarta/data/metamodel/Pattern.java
```diff
@@ -0,0 +1,176 @@
...
+    static <T> Pattern<T> like(String field, String value) {
+        return new Pattern<>(field, value, false);
+    }
+
+    static <T> Pattern<T> prefixed(String field, String value) {
+        return new Pattern<>(field, value + "%", false);
+    }
+
+    /**
+     * Returns a new `Pattern` instance with case-insensitive matching enabled.
```
njr-11, line 162:
```suggestion
     * Returns a new {@code Pattern} instance with case-insensitive matching.
```

api/src/main/java/jakarta/data/metamodel/BasicRestriction.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package jakarta.data.metamodel;
+
+import jakarta.data.Operator;
+
+/**
+ * A basic restriction applied to a single field, representing conditions such as equality,
+ * comparisons, range checks, and pattern matches.
+ *
+ * <p>The `BasicRestriction` interface provides methods for defining simple, singular restrictions
```
njr-11, line 26:
```suggestion
 * <p>The {@code BasicRestriction} interface provides methods for defining simple, singular restrictions
```

api/src/main/java/jakarta/data/metamodel/BasicRestriction.java
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.metamodel;
+
+import jakarta.data.Operator;
+
+/**
+ * A basic restriction applied to a single field, representing conditions such as equality,
```
njr-11, line 23:
```suggestion
 * A basic restriction applies to a single field, representing conditions such as equality,
```

api/src/main/java/jakarta/data/metamodel/MultipleRestriction.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.metamodel;
+
+import java.util.List;
+
+/**
+ * A composite restriction that combines multiple `Restriction` instances using logical operators.
```
njr-11, line 23:
```suggestion
 * A composite restriction that combines multiple {@link Restriction} instances using logical operators.
```

api/src/main/java/jakarta/data/metamodel/MultipleRestriction.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * <p>The `MultipleRestriction` interface allows for combining multiple restrictions, enabling complex
+ * filtering scenarios where multiple conditions must be satisfied. Each contained `Restriction`
+ * can be evaluated based on the logical operator specified by the `Restrict` type.</p>
```
njr-11, line 27:
```suggestion
 * <p>The {@code MultipleRestriction} interface allows for combining multiple restrictions, enabling complex
 * filtering scenarios where multiple conditions must be satisfied. Each contained {@link Restriction}
 * can be evaluated based on the logical operator specified by the {@link Restrict} type.</p>
```

api/src/main/java/jakarta/data/metamodel/Pattern.java
```diff
@@ -0,0 +1,176 @@
...
+ */
+package jakarta.data.metamodel;
+
+import jakarta.data.Operator;
+
+
+/**
+ * Represents a pattern-based restriction for matching operations, encapsulating different
+ * options such as prefix, suffix, and substring matching. This implementation
+ * allows flexibility in creating pattern-based conditions directly as `Restriction` instances.
```
njr-11, line 26:
```suggestion
 * allows flexibility in creating pattern-based conditions directly as {@link Restriction} instances.
```

api/src/main/java/jakarta/data/metamodel/Pattern.java
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * <p>Example usage with metadata attributes:</p>
+ * <pre>
+ * // Case-sensitive exact match
+ * Restriction<Book> exactMatch = Pattern.like(_Book.title, "Jakarta Data");
+ *
```
njr-11, line 32: The first example seems like a bad practice that we should not recommend.  Don't use pattern for exact matching. Use the default equality matching that the spec already has.
```suggestion
```

api/src/main/java/jakarta/data/metamodel/Pattern.java
```diff
@@ -0,0 +1,176 @@
...
+ * // Case-insensitive prefix match
+ * Restriction<Book> prefixIgnoreCase = Pattern.prefixedIgnoreCase(_Book.title, "Jak");
+ *
+ * // Case-sensitive suffix match
```
njr-11, line 36: Per prior comments, we do not guarantee case sensitive.
```suggestion
 * // Suffix match
```

api/src/main/java/jakarta/data/metamodel/Pattern.java
```diff
@@ -0,0 +1,176 @@
...
+ * Restriction<Book> suffixMatch = Pattern.suffixed(_Book.title, "Guide");
```
njr-11, line 37: I don't see why we would want this usage with passing in the metamodel attribute (or its String name) as a parameter rather than using the metamodel attribute to more directly and cleanly obtain the restriction,
```
_Book.title.endsWith("Guide")
```
or supplying the pattern as the value of the `title` parameter for a parameter based automatic query,
```
Pattern.endsWith("Guide")
```

The same applies to the other examples in this section.  None of the examples here seem like good practices.

api/src/main/java/jakarta/data/metamodel/Pattern.java
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * // Case-insensitive substring match
+ * Restriction<Book> substringIgnoreCase = Pattern.substringedIgnoreCase(_Book.title, "Java");
+ * </pre>
+ *
+ * @param <T> the type of the entity on which the restriction is applied.
+ */
+public record Pattern<T>(String field, String value, boolean ignoreCase) implements BasicRestriction<T> {
```
njr-11, line 45: Don't inherit from BasicRestriction, and don't include `field` inside the Pattern.

api/src/main/java/jakarta/data/metamodel/Pattern.java
```diff
@@ -0,0 +1,176 @@
...
+     *
+     * @return the pattern value.
+     */
+    @Override
+    public String value() {
+        return value;
+    }
+
+    /**
+     * Creates a pattern for an exact `LIKE` match on the specified value.
```
njr-11, line 68:
```suggestion
     * Creates a pattern for a {@link Operator#LIKE LIKE} match on the specified value.
```

api/src/main/java/jakarta/data/metamodel/Pattern.java
```diff
@@ -0,0 +1,176 @@
...
+     * This method sets the field to `null`, allowing it to be applied
+     * later to a specific attribute.
+     *
+     * <p>Example usage:</p>
+     * <pre>
+     * Restriction<Book> titlePattern = _Book.title.like(Pattern.like("Jakarta"));
```
njr-11, line 74: Bad practice to use like for equals. Adding a wildcard will make it a better example,
```suggestion
     * Restriction<Book> titlePattern = _Book.title.like(Pattern.like("Jakarta %"));
```

api/src/main/java/jakarta/data/metamodel/Pattern.java
```diff
@@ -0,0 +1,176 @@
...
+     * </pre>
+     *
+     * @param value the exact text to match.
+     * @return a Pattern instance for an exact match.
```
njr-11, line 78:
```suggestion
     * @param value the pattern to match.
     * @return a Pattern instance for a pattern match.
```

api/src/main/java/jakarta/data/metamodel/Pattern.java
```diff
@@ -0,0 +1,176 @@
...
+     */
+    public static <T> Pattern<T> like(String value) {
+        return like(null, value);
```
njr-11, line 81:
```suggestion
    public static <T> Pattern<T> like(String pattern) {
        return like(null, pattern);
```

njr-11 (commented):
I spent a little more time reviewing. Here are some comments, but I'll stop before going further.  It is unfortunate that this PR is trying to achieve a huge amount of things all at once. I don't ever have enough time to make it all the way through it.  In the future, we need to to a better job of proposing things in smaller sizes and gaining consensus are different parts are added more gradually.

api/src/main/java/jakarta/data/metamodel/Pattern.java
otaviojava: Aggree

api/src/main/java/jakarta/data/metamodel/Pattern.java
njr-11: It seems nonsensical to me to have a restriction that doesn't apply to any entity field in particular.  Having a Pattern be independent of a field makes sense to me. A field isn't part of a pattern.

njr-11: > Using the factory method, the user won't see it at all. Thus, it will also work for our metamodel structure, thus:
> 
> ```java
> Restriction<Book> filter = Pattern.like("...");
> Restriction<Book> filter = _Book.title.like("...");
> Restriction<Book> filter = _Book.title.like(Pattern.like("..."));
> ```
> 
> All of those three will work.

That second and third are good. The first one is an invalid restriction that will fail when used.  A Pattern is text, typically with `%` or `_` wildcards.  A Restriction is the combination of an entity attribute, a comparison operator, information about whether to negate or ignore case, and a value to compare with.  A Pattern is not a Restriction.

otaviojava: >  A Pattern is not a Restriction.

That is a good point, on NoSQL to archive it we would use the LIKE operator against the field that the user wish...

So, on JPA we do that without operator? Do you do it on memory instead of returning on the database?

That sounds no efficient.

njr-11: > So, on JPA we do that without operator? Do you do it on memory instead of returning on the database?

No. I don't know why you would conclude that from what I said.

> use the LIKE operator against the field that the user wish...

This part is important. Pattern has no field.  You need the field to come from either a Restriction that is built using the Pattern as its value,

```
Restriction<Book> filter = _Book.title.like(Pattern.startsWith("Jakarta "))
                                      .ignoreCase();
```

or from a repository method indicating which field it pertains to,

```
@Find
List<Book> named(@Is(ANY_CASE) Pattern name);
```

which has all of the same information you could use to construct a Restriction.

gavinking: @otaviojava I agree with Nathan here.

`Pattern` should be a standalone class, not a kind of `Restriction`. It's not a elegant, nor typesafe, to have one field being nullable just completely change the role of a type and what you're supposed to use it for.

otaviojava: @njr-11 @gavinking I got your point.
I've moved Pattern to the `Jakarta.data` package and removed the Restriction interface.


Please, let me know what you think and please, feel free to open PR as well.

njr-11 (changes requested):
> feel free to open PR as well.

We do have all of the Restriction changes only (minus JavaDoc which I don't think we should be adding yet while there are so many API changes still going on) split out under https://github.com/jakartaee/data/pull/895 which Gavin has reviewed.

Regarding Pattern, we are getting closer.  Here is my review of what you have for the Pattern class under this PR.  A general comment is that it's important to account for the possibility that the Strings supplied to endsWith, contains, and so forth will include wildcard characters within them and will behave as patterns as well.  My code suggestions included covering that possibility.

api/src/main/java/jakarta/data/Pattern.java
```diff
@@ -0,0 +1,96 @@
...
+ * options such as prefix, suffix, and substring matching. This implementation
+ * allows flexibility in creating pattern-based conditions directly as {@link Restriction} instances.
+ *
+ * <p>Example usage with metadata attributes:</p>
+ * <pre>
+ * Restriction<Book> prefixIgnoreCase = _Book.title.endsWith("Guide");
+ * </pre>
+ *
+ */
+public record Pattern(String value, boolean ignoreCase) {
```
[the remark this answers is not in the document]
gavinking: I don't agree with this. One of the main nice things about `Pattern` is that it encapsulates case-sensitivity.

By original proposal was:

```java
public record Pattern(String pattern, boolean caseSensitive) {

    public Pattern ignoringCase() {
        return new Pattern(pattern, false);
    }

    public static Pattern is(String literal) {
        return new Pattern(escape(literal), true);
    }

    public static Pattern matches(String pattern) {
        return new Pattern(pattern, true);
    }

    public static Pattern matches(String pattern,
                                  char characterWildcard, char stringWildcard) {
        final String standardized =
                escape(pattern)
                        .replace(characterWildcard, '_')
                        .replace(stringWildcard, '%');
        return new Pattern(standardized, true);
    }

    public static Pattern startsWith(String prefix) {
        return new Pattern(escape(prefix) + '%', true);
    }

    public static Pattern endsWith(String suffix) {
        return new Pattern('%' + escape(suffix), true);
    }

    public static Pattern contains(String substring) {
        return new Pattern('%' + escape(substring) + '%', true);
    }

    private static String escape(String literal) {
        return literal.replace("_", "\\_").replace("%", "\\%");
    }
}
```

Where, the idea was that users would not usually call the constructor to obtain a `Pattern`.

api/src/main/java/jakarta/data/Pattern.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Creates a pattern for a `LIKE` match where values start with the specified prefix.
+     * The `field` is set to `null` initially, allowing it to be assigned to an attribute later.
+     *
+     * @param value the prefix to match at the beginning of the field's value.
+     * @return a Pattern instance for a prefix match.
+     */
+    public static Pattern startsWith(String value) {
+        return new Pattern( value + "%", false);
```
gavinking, line 55: In my `Pattern` class, I `escape()` the `prefix`, because I figured it shouldn't be a pattern.

```java
    public static Pattern startsWith(String prefix) {
        return new Pattern(escape(prefix) + '%', true);
    }
```

Well, I mean, `startsWith()` is not a pattern in Java. I now notice that it *is* a pattern in the query by method name extension, so now I'm not quite sure what to do.

I suppose we should have `startsWIthLiteral()` / `containsLiteral()`, etc.

api/src/main/java/jakarta/data/Pattern.java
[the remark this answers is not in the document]
njr-11: Irrespective of the discussion on ignoringCase, the automatic escaping that you proposed above seems very useful, and helps users avoid unintentional wildcards in their prefix/suffix/substring, which can be literals now instead of patterns.  I like that better than what is under this PR.

api/src/main/java/jakarta/data/Pattern.java
njr-11: I would leave StartsWith/EndsWith/Contains from Query by Method Name alone, not have them support Pattern, and not add additional methods for StartsWithLiteral..., taking only the minimal step of having Query by Method Name support Pattern for [Not]Like[IgnoreCase], which will sufficiently cover all of the operations.

We should consider if we want to change StartsWith/EndsWith/Contains to use literals rather than patterns in Restrictions and Parameter-based Automatic Methods, which are new in the current version.

otaviojava: @gavinking sorry, I did not find your `Pattern` could you either give me or open a PR against this one?

otaviojava: Never mind, I find it.

otaviojava: I used your class and put the javadoc.

api/src/main/java/jakarta/data/Pattern.java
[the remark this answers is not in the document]
otaviojava: Nice, I agreed.
I moved this class to our PR.

otaviojava:
> #895

About Pattern, I used the same Gavin.

On your PR, I moved without keeping out the NOT condition, at least for now.

otaviojava:
@njr-11 @gavinking 

Hey, I've covered all the mentioned points, and I believe what we have here is good enough to merge and become our first draft version.

Once merged, we can go more scalable, I mean, create an Epic task where we can work on:

Spec documentation
Test scenarios
Naturally, experiment and refine the API
Have a Track to discuss the "not" operation.
Finally, create the TCK (once we agree with what we have)

Note: I believe that is a long journey. I would like to propose the first step here.

njr-11:
> @njr-11 @gavinking
> 
> Hey, I've covered all the mentioned points, and I believe what we have here is good enough to merge and become our first draft version.
> 
> Once merged, we can go more scalable, I mean, create an Epic task where we can work on:
> 
> Spec documentation Test scenarios Naturally, experiment and refine the API Have a Track to discuss the "not" operation. Finally, create the TCK (once we agree with what we have)
> 
> Note: I believe that is a long journey. I would like to propose the first step here.

This isn't making progress because you are trying to do too much at once in a single PR.  I stopped reviewing the rest of the code under this PR because I don't agree with what is being done in Pattern and it is unclear to me at this point whether or not we should even have Pattern and Range.  Decisions such as whether to have those, or whether or not to have `@Is` should not be tied to adding the API for `Restriction`.  I don't think we are going to come to consensus on all of these different API proposals all at once.  I ended up splitting out the Restriction piece only under #895 which is what I think we can gain agreement on first.  Gavin reviewed it, and I addressed all of his comments.  If you want to review that PR, we could try to get Restriction in and then come back to some of these other proposals.

otaviojava:
> > @njr-11 @gavinking

I got your point. In my opinion, those should work as a single one; they should either go together or not go, mainly, because those new types, `Pattern` and `Range` should be integrated with this one as well.

* About `Is` it is removed.
* About the `Range` it is the Gavin code with documentation included.

njr-11:
> I got your point. In my opinion, those should work as a single one; they should either go together or not go, mainly, because those new types, `Pattern` and `Range` should be integrated with this one as well.

If `Pattern` and `Range` end up being introduced, it might make sense to be able to create Restrictions from them.  However, the concept of Restriction has no dependency whatsoever on these classes.

#895 shows that to be the case because it has provided `Restriction` without them.  It enables a user to write only,

```
products.search(_Product.name.endsWith(suffix));

products.search(_Product.name.endsWith(suffix).ignoreCase());
```

which is much cleaner than what you have them doing in your PR,

```
products.search(_Product.name.like(new Pattern('%' + suffix, false));

products.search(_Product.name.like(Pattern.endsWith(suffix));
```

Note that you have ignoreCase as default for Pattern with the only way to avoid it being to construct a Pattern instance.  Even if you correct that, you end up with something like,

```
products.search(_Product.name.like(Pattern.endsWith(suffix));

products.search(_Product.name.like(Pattern.endsWith(suffix).ignoreCase());
```

Not only is it unnecessary for Restriction and Pattern to go together, in almost all cases where users want Restrictions, adding in Pattern will only be unwanted extra typing that doesn't provide any additional value to the user.

otaviojava:
> > I got your point. In my opinion, those should work as a single one; they should either go together or not go, mainly, because those new types, `Pattern` and `Range` should be integrated with this one as well.
> 
> If `Pattern` and `Range` end up being introduced, it might make sense to be able to create Restrictions from them. However, the concept of Restriction has no dependency whatsoever on these classes.

About `Pattern` I am using exactly the Gavin one with documentation: https://github.com/jakartaee/data/pull/894#discussion_r1838391611

About you said, we have access to exactly same methods, thus:

You can do those as well:

```
products.search(_Product.name.like(Pattern.endsWith(suffix));

products.search(_Product.name.like(Pattern.endsWith(suffix).ignoreCase());
```

njr-11:
> > > I got your point. In my opinion, those should work as a single one; they should either go together or not go, mainly, because those new types, `Pattern` and `Range` should be integrated with this one as well.

I don't think you are understanding the point.  You seem to be arguing that it would be invalid to add Restriction without Pattern because then you could only do,

```
products.search(_Product.name.endsWith(suffix));

products.search(_Product.name.endsWith(suffix).ignoreCase());
```

but not also,

```
products.search(_Product.name.like(Pattern.endsWith(suffix));

products.search(_Product.name.like(Pattern.endsWith(suffix).ignoreCase());
```

Those latter two examples are redundant, more verbose, and provide no additional value.  This is not a valid reason to claim that Pattern must be introduced along with Restriction.

otaviojava:
> I don't think you are understanding the point. You seem to be arguing that it would be invalid to add Restriction without Pattern because then you could only do,

Yeap, I am lost here because at `TextAttribute` we do have:

```java
    default Restriction<T> like(Pattern pattern)...
```

Thus, we can do this since `name` might be a `TextAttribute`:

```
products.search(_Product.name.like(Pattern.endsWith(suffix)));
products.search(_Product.name.like(Pattern.endsWith(suffix).ignoreCase()));
```

We also have `default Restriction<T> like(String text)` that allows users to set their own terms to like.

Plus, we also have support for `Pattern` as you said.

Thus, we have those as supported. Right now, `TextAttribute` can work either `Pattern` or without `Pattern`, the same as `Pattern´, which can work alone as well.

PR #895, opened 2024-11-05
dynamic query restrictions

njr-11:
This PR contains some ideas that I was experimenting with locally for dynamic query restrictions #829 and hadn't been planning to propose until after a lot more discussion occurred.  However, because code changes are already being proposed and I wanted to illustrate how this should tie into the static metamodel, I decided to share this prematurely so everyone can see what I was originally aiming for with the pattern.

```
@Repository
public interface Products extends CrudRepository<Product, Long> {
    List<Product> search(String namePattern,
                         Restriction<Product> restriction,
                         Order<Product> order);
}
```

Usage with single restriction,
```
    List<Product> found = products.search(pattern,
                                          _Product.price.lessThan(50.0f),
                                          Order.by(_Product.price.desc(),
                                                   _Product.name.asc()));
```

Usage with multiple restrictions,
```
    List<Product> found = products.search(pattern,
                                          Restrict.all(_Product.price.greaterThan(25.0f),
                                                       _Product.price.lessThan(50.0f),
                                                       _Product.name.startsWith(pattern).ignoreCase(),
                                                       _Product.image.notNull()),
                                          Order.by(_Product.price.desc(),
                                                   _Product.name.asc()));
```

This was just the basics, and doesn't yet include not/negation, or Pattern/Range, which might make parts of it unnecessary.  It also just occurred to me that if we had a TextRestriction subclass, which I did not include in this, we could omit the `*IgnoreCase` methods and have `.ignoreCase()` instead.


njr-11:
After a lot of difficulty reviewing another PR that combines several different new parts of Jakarta Data, I decided to just finish off updating this PR with how I expect Restrictions to work.

I was originally thinking that Restrictions would only come from the static metamodel, but others have expressed a desire that users be able to create them like they can for Sort.  I'm okay with that, but not by giving users access to record/implementation class constructors where they can populate them with any combinations of things they like.  With this PR, all implementation classes are made non-public and kept out of the API.  There are 2 places where users can obtain restrictions:

- static metamodel attributes (basic restrictions only)
- the Restrict class (both composite and basic restrictions)

This also aims to de-emphasize interfaces that the end user shouldn't care about (and only Data Providers need, such as Operator and the Composite restriction) out of the main view by placing them in inner classes.

There is no JavaDoc in this on purpose. The idea is to iterate over the design first, and later spend time on JavaDoc and other things.

Other spec function like Pattern, Range, Is, ... are intentionally kept out so we can focus only on getting Restriction correct.

`Restriction` could justifiably be placed in the jakarta.data package alongside some other special classes that are used as repository parameters (Sort, Order, Limit) although we might want to consider if it is worth it to have a separate `jakarta.data.restrict` package to separate it out like we do with `jakarta.data.page` for PageRequest, which is the other special parameter type. I think the latter is a good place for it, so that's where I put it in this PR.

greater/lessThan methods require a Comparable parameter (lots of things can be sortable, but I think they are all Comparable).  TextAttribute methods, and their equivalents on Restrict, require a String parameter.  I added a commented out method showing what it might be appropriate to do with Pattern parameter as well.  But ignore that if you don't want to think about Pattern yet.

I would like to see more error checking for values supplied to Restriction factory methods in general, and identify if we can use any more precise types that would help keep invalid objects out, but we can save that for when we are further along.

api/src/main/java/jakarta/data/restrict/Restrict.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.List;
+import java.util.Set;
+
+import jakarta.data.restrict.Restriction.Basic;
+import jakarta.data.restrict.Restriction.Operator;
+
+// TODO document
+// This is one of two places from which to obtain restrictions.
+// The other place is from static metamodel attributes.
+public enum Restrict {
```
gavinking, line 29: Not sure I love how this breaks the pattern established by `Sort`/`Order`.

gavinking:
> `Restriction` could justifiably be placed in the `jakarta.data` package

This is where I would have put it.

api/src/main/java/jakarta/data/restrict/Restrict.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @SafeVarargs
+    public static <T> Restriction<T> all(Restriction<T>... restrictions) {
+        return new CompositeRestriction<>(ALL, List.of(restrictions));
+    }
+
+    @SafeVarargs
+    public static <T> Restriction<T> any(Restriction<T>... restrictions) {
+        return new CompositeRestriction<>(ANY, List.of(restrictions));
+    }
```
gavinking, line 41: Still no way to do `not()`, is that correct?

api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -17,19 +17,46 @@
...
+import jakarta.data.restrict.Restriction;
 
 /**
  * Represents an entity attribute in the {@link StaticMetamodel}.
  *
  * @param <T> entity class of the static metamodel.
  */
 public interface Attribute<T> {
+
+    default Restriction.Basic<T> equal(Object value) {
```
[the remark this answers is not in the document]
njr-11: updated

api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -17,19 +17,46 @@
...
+        return Restrict.equal(value, name());
+    }
+
+    default Restriction.Basic<T> in(Object... values) {
```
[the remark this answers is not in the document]
njr-11: Yes, we definitely want notIn and some others. I hadn't been thinking to include not* under this PR, but it's trivial to add, so I added it

api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -17,19 +17,46 @@
...
+        if (values == null || values.length == 0)
+            throw new IllegalArgumentException("values are required");
+
+        return Restrict.in(Set.of(values), name());
+    }
+
+    default Restriction.Basic<T> isNull() {
```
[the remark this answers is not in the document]
njr-11: added

api/src/main/java/jakarta/data/metamodel/SortableAttribute.java
```diff
@@ -49,4 +51,19 @@ public interface SortableAttribute<T> extends Attribute<T> {
...
+        return Restrict.greaterThanEqual(value, name());
+    }
+
+    default Restriction.Basic<T> lessThan(Comparable<Object> value) {
+        return Restrict.lessThan(value, name());
+    }
+
+    default Restriction.Basic<T> lessThanEqual(Comparable<Object> value) {
+        return Restrict.lessThanEqual(value, name());
+    }
```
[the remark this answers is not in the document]
njr-11: added

api/src/main/java/jakarta/data/restrict/Restrict.java
njr-11: Is the enum class the only problem here? If so, I can move the enum to another class and make Restrict into a normal class like Order is.  I didn't include this in the commit with the first round of fixes because I want to be sure what you are asking for.

api/src/main/java/jakarta/data/restrict/Restrict.java
njr-11: Because it requires further discussion, I chose to defer it and leave a TODO comment (right after this) for it.  A follow-on PR can add it after we decide how best to do it.

njr-11 (commented):
Added a commit to cover most of Gavin's review comments.
Also, I made an update to ensure ignoreCase is only available to String values, and not to numbers, dates, and so forth.

njr-11:
> > `Restriction` could justifiably be placed in the `jakarta.data` package
> 
> This is where I would have put it.

That's fine. I have moved it to `jakarta.data` in the review fixes commit.

api/src/main/java/jakarta/data/restrict/Restrict.java
gavinking: Yes, I think that's my preference. I don't think it's going to be normal for the user to want to access the enum values directly.

api/src/main/java/jakarta/data/restrict/Restrict.java
gavinking: Yeah no worries, was just checking I had not missed it.

gavinking:
> Added a commit to cover most of Gavin's review comments.

Thanks!

api/src/main/java/jakarta/data/restrict/Restrict.java
njr-11: I agree. The user will typically never look the composite type. The enum values are mainly for the Jakarta Data provider to check which type of composite it is. I made Restrict into a normal class and moved the enum elsewhere. Thanks for pointing this out.

njr-11:
Switched methods such as startsWith(prefixPattern) to startsWith(prefix) and automatically add escaping to spare the user from that level of detail/complexity.

api/src/main/java/jakarta/data/Restriction.java
```diff
@@ -0,0 +1,61 @@
...
+        enum Type {
+            ALL,
+            ANY
+        }
+    }
+
+    interface Text<T> extends Basic<T> {
+        Restriction<T> ignoreCase();
+
+        boolean isAnyCase();
```
otaviojava, line 47: Change this one to: ignoreCase.

gavinking, line 47:
```suggestion
        boolean isCaseSensitive();
```

njr-11: Lacking a better name, we can use Gavin's suggestion of isCaseSensitive.  This method is here for the Jakarta Data provider rather than the user, and we can easily add a disclaimer in the JavaDoc that if a database is incapable of case sensitive comparison, it will ignore a value of true.

api/src/main/java/jakarta/data/TextRestrictionRecord.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data;
+
+// Internal implementation class.
+// The proper way for users to obtain instances is via
+// the static metamodel or Restrict.* methods 
+
+record TextRestrictionRecord<T>(
```
otaviojava, line 24: You can use `DefaultTextRestriction`

njr-11: I suppose another idea would be `TextRestrictionImpl`. That would make it very clear that it is a hidden internal class used for implementation.

otaviojava: I am OK with either `...Record`or `Default`, but I would avoid the `TextRestrictionImpl`

njr-11: Alright. `Default` seemed awkward to me, so I'll leave it as `TextRestrictionRecord` for now. We can rename it later if we think of something better.

api/src/main/java/jakarta/data/Restrict.java
```diff
@@ -0,0 +1,232 @@
...
+
+    public static <T> TextRestriction<T> like(String pattern,
+                                               char charWildcard,
+                                               char stringWildcard,
+                                               String field) {
+        String p = toLikeEscaped(charWildcard, stringWildcard, false, pattern, false);
+        return new TextRestrictionRecord<>(field, Operator.LIKE, ESCAPED, p);
+    }
+
+    public static <T> Restriction<T> not(Object value, String field) {
```
otaviojava, line 137: It is not clear that `not` means, `notEquals´. 
 Please either, use `ne` or `NotEquals`

api/src/main/java/jakarta/data/Restrict.java
```diff
@@ -0,0 +1,232 @@
...
+        return new BasicRestrictionRecord<>(field, NOT, Operator.EQUAL, value);
+    }
+
+    public static <T> TextRestriction<T> not(String value, String field) {
+        return new TextRestrictionRecord<>(field, NOT, Operator.EQUAL, value);
+    }
+
+    public static <T> TextRestriction<T> notContains(String substring, String field) {
```
otaviojava, line 145: Can we remove all of those `no` method factory sequences?

It would be great if we have only two options:  by the method instead `negate`and a single method factory `not`.

It is pretty similar to the Java `Predicate`: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/function/Predicate.html

api/src/main/java/jakarta/data/Restrict.java
```diff
@@ -0,0 +1,232 @@
...
+     * @param charWildcard    single character wildcard, typically _.
+     * @param stringWildcard  0 or more character wildcard, typically %.
+     * @param allowPrevious   whether to allow characters prior to the text.
+     * @param literal text    that is not escaped that must be matched.
+     * @param allowSubsequent whether to allow more characters after the text.
+     * @return escaped pattern.
+     * @throws IllegalArgumentException if the same character is supplied for
+     *                                  both wildcard types.
+     */
+    private static String toLikeEscaped(char charWildcard,
```
otaviojava, line 197: Once everybody agreed with Gavin's `Pattern` class, why not centralize this behavior in that class?

api/src/main/java/jakarta/data/Restriction.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data;
+
+public interface Restriction<T> {
+    boolean isNegated();
+
+    enum Operator {
```
otaviojava, line 23: Please remove this one from the inner class; it is good, mainly for the Javadoc.

api/src/main/java/jakarta/data/metamodel/SortableAttribute.java
```diff
@@ -42,11 +44,30 @@ public interface SortableAttribute<T> extends Attribute<T> {
...
+
     /**
      * Obtain a request for a descending {@link Sort} based on the entity attribute.
      *
      * @return a request for a descending sort on the entity attribute.
      */
     Sort<T> desc();
 
+    default Restriction<T> greaterThan(Comparable<Object> value) {
```
otaviojava, line 58: Should this comparable be `Comparable<? extends Object> value`, I am not sure here.

api/src/main/java/jakarta/data/metamodel/TextAttribute.java
```diff
@@ -33,11 +35,84 @@ public interface TextAttribute<T> extends SortableAttribute<T> {
...
+        return Restrict.like(pattern, name());
+    }
+
+    default TextRestriction<T> like(String pattern,
+                                    char charWildcard,
+                                    char stringWildcard) {
+        return Restrict.like(pattern, charWildcard, stringWildcard, name());
+    }
+
+    default TextRestriction<T> not(String value) {
```
otaviojava, line 88: The same point of using the Java core as reference to us:

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/function/Predicate.html

api/src/main/java/jakarta/data/metamodel/SortableAttribute.java
gavinking: I think it should probably just be `T`.

api/src/main/java/jakarta/data/Restrict.java
gavinking: @otaviojava JDQL has operators like `not between`, `not like`, etc, so it makes sense to me for such things to be reflected here. And it's significantly less verbose.

api/src/main/java/jakarta/data/Restrict.java
gavinking: Yes, I think it's much better to have a nice class which models this basic concept. It's a much semantically stronger construct.

api/src/main/java/jakarta/data/Restrict.java
```diff
@@ -0,0 +1,232 @@
...
+
+    public static <T> Restriction<T> greaterThan(Comparable<Object> value, String field) {
+        return new BasicRestrictionRecord<>(field, Operator.GREATER_THAN, value);
+    }
+
+    public static <T> TextRestriction<T> greaterThan(String value, String field) {
+        return new TextRestrictionRecord<>(field, Operator.GREATER_THAN, value);
+    }
+
+    public static <T> Restriction<T> greaterThanEqual(Comparable<Object> value, String field) {
```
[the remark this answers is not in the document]
otaviojava: Better!

api/src/main/java/jakarta/data/Restrict.java
njr-11: I agree with Gavin on this.

Being able to write code such as,

```
products.search(_Product.description.notContains("refurbished"));
```

is much more straightforward and readable than

```
products.search(_Product.description.contains("refurbished").negated());
```

or 

```
products.search(Restrict.not(_Product.description.contains("refurbished")));
```

api/src/main/java/jakarta/data/Restrict.java
njr-11: I agree, if Pattern gets added, this utility method can move to there.  Also note that the method is private, so the documentation on it is only for our own benefit and will not show in the API, so users won't even notice once it is moved.

api/src/main/java/jakarta/data/Restrict.java
njr-11: Sure, I can update that. Given that we have a `equalTo` method, the equivalent here will be `notEqualTo`

njr-11:
Review comments have been addressed.

otaviojava:
> Review comments have been addressed.

I saw, I am moving the documentation and the unit test that make sense against this PR.

I hope finish it today, thus, you will receive a PR soon.

otaviojava:
@njr-11 done the first step: https://github.com/njr-11/data/pull/1

njr-11:
Thanks for contributing the unit tests. That helps to confirm the correctness of the changes.
I was thinking maybe we should wait on adding JavaDoc until after this PR is merged, so that we don't need to rewrite it if anyone asks for further changes to the initial PR.

otaviojava:
> Thanks for contributing the unit tests. That helps to confirm the correctness of the changes. I was thinking maybe we should wait on adding JavaDoc until after this PR is merged, so that we don't need to rewrite it if anyone asks for further changes to the initial PR.

It does make sense. 
I will focus on the Unit test to find some issues on it.

otaviojava:
@njr-11 one more PR against yours.

njr-11:
> @njr-11 one more PR against yours.

@otaviojava I don't see another PR against this anywhere.

njr-11:
> > > @njr-11 one more PR against yours.
> > 
> > 
> > @otaviojava I don't see another PR against this anywhere.
> 
> This one here: [njr-11#2](https://github.com/njr-11/data/pull/2)

I see it now and have accepted your PR/commits for more unit tests.  Thanks for adding the test coverage.

otaviojava:
> > > > @njr-11 one more PR against yours.
> > > 
> > > 
> > > @otaviojava I don't see another PR against this anywhere.
> > 
> > 
> > This one here: [njr-11#2](https://github.com/njr-11/data/pull/2)
> 
> I see it now and have accepted your PR/commits for more unit tests. Thanks for adding the test coverage.


@njr-11 

This is the last one:

https://github.com/njr-11/data/pull/3

I put the validation on field making the field mandatory

otaviojava:
As a first step, IMHO, it is good to go.

@njr-11 @gavinking, what do you think?

otaviojava (approved):

PR #904, opened 2024-12-05
Resolve build failure from LifecycleEvent:51

njr-11:
I brought in the latest code and tried to build, but it fails here,

```
[WARNING] PMD Failure: jakarta.data.event.LifecycleEvent:51 Rule:AbstractClassWithoutAbstractMethod Priority:3 This abstract class does not have any abstract methods.
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Jakarta Data 1.0.1-SNAPSHOT:
[INFO] 
[INFO] Jakarta Data ....................................... SUCCESS [  5.347 s]
[INFO] Jakarta Data API ................................... FAILURE [ 13.975 s]
[INFO] Jakarta Data Specification ......................... SKIPPED
[INFO] Jakarta Data Technology Compatibility Kit .......... SKIPPED
[INFO] Jakarta Data Technology Compatibility Kit Distribution SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  19.425 s
[INFO] Finished at: 2024-12-05T14:53:33-06:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.26.0:check (default) on project jakarta.data-api: PMD 7.7.0 has found 1 violation. For more details see: ...
```

This PR is one way to resolve the failure.

### DIFF
--- a/api/src/main/java/jakarta/data/event/LifecycleEvent.java
+++ b/api/src/main/java/jakarta/data/event/LifecycleEvent.java
@@ -48,10 +48,10 @@ package jakarta.data.event;
  *
  * @param <E> the entity type
  */
-public abstract class LifecycleEvent<E> {
+public class LifecycleEvent<E> {
     private final E entity;
 
-    public LifecycleEvent(E entity) {
+    protected LifecycleEvent(E entity) {
         this.entity = entity;
     }
 


gavinking:
I think that if the stupid nag bot is giving us its stupid arbitrary nags over *perfectly legit and sensible code* then we should deactivate the stupid bot.

We don't need it.

njr-11:
> I think that if the stupid nag bot is giving us its stupid arbitrary nags over _perfectly legit and sensible code_ then we should deactivate the stupid bot.
> 
> We don't need it.

Yeah, we should consider that as the resolution instead. This certainly wouldn't be the first time we've had to change valid code to comply with tools enforcing their preferences.

gavinking (changes requested):
> "abstract class without abstract method"

What an absurd, arbitrary rule!

Here, `LifecycleEvent` is `abstract` because it should not have direct instances, not because it has abstract operations.

I mean, an abstract type (potentially without operations) is the normal, bog-standard mapping of a disjoint union into an object-oriented type system. This is why I hate linters: they're written by people who have no actual foundation in PL design and just make up arbitrary crap based on feelings.

otaviojava:
We will close because we won't change the code because PMD.

PR #905, opened 2024-12-05
negated restrictions

njr-11:
If I remember the discussion correctly, the 3 ideas for how to do negation of restrictions were:

- restriction.negate()
- Restrict.not(restriction)
- Restrict.notAll(restrictions) / Restrict.notAny(restrictions)

notAll/notAny seemed the most awkward to me, and I realized that the first two are basically the same thing, and it would be trivial to have both ways, where Restrict.not(restriction) is a convenience method that invokes restriction.negate().

### DIFF
--- a/api/src/main/java/jakarta/data/BasicRestrictionRecord.java
+++ b/api/src/main/java/jakarta/data/BasicRestrictionRecord.java
@@ -36,4 +36,33 @@ record BasicRestrictionRecord<T>(
     BasicRestrictionRecord(String field, Operator comparison, Object value) {
         this(field, false, comparison, value);
     }
+
+    @Override
+    public Restriction<T> negate() {
+        boolean newNegation = isNegated;
+        Operator newComparison;
+        switch (comparison) {
+            case GREATER_THAN:
+                newComparison = Operator.LESS_THAN_EQUAL;
+                break;
+            case GREATER_THAN_EQUAL:
+                newComparison = Operator.LESS_THAN;
+                break;
+            case LESS_THAN:
+                newComparison = Operator.GREATER_THAN_EQUAL;
+                break;
+            case LESS_THAN_EQUAL:
+                newComparison = Operator.GREATER_THAN;
+                break;
+            default:
+                newComparison = comparison;
+                newNegation = !isNegated;
+        }
+
+        return new BasicRestrictionRecord<>(
+                field,
+                newNegation,
+                newComparison,
+                value);
+    }
 }
--- a/api/src/main/java/jakarta/data/BasicRestrictionRecord.java
+++ b/api/src/main/java/jakarta/data/BasicRestrictionRecord.java
@@ -25,7 +25,6 @@ import java.util.Objects;
 
 record BasicRestrictionRecord<T>(
         String field,
-        boolean isNegated,
         Operator comparison,
         Object value) implements BasicRestriction<T> {
 
@@ -33,36 +32,11 @@ record BasicRestrictionRecord<T>(
         Objects.requireNonNull(field, "Field must not be null");
     }
 
-    BasicRestrictionRecord(String field, Operator comparison, Object value) {
-        this(field, false, comparison, value);
-    }
-
     @Override
-    public Restriction<T> negate() {
-        boolean newNegation = isNegated;
-        Operator newComparison;
-        switch (comparison) {
-            case GREATER_THAN:
-                newComparison = Operator.LESS_THAN_EQUAL;
-                break;
-            case GREATER_THAN_EQUAL:
-                newComparison = Operator.LESS_THAN;
-                break;
-            case LESS_THAN:
-                newComparison = Operator.GREATER_THAN_EQUAL;
-                break;
-            case LESS_THAN_EQUAL:
-                newComparison = Operator.GREATER_THAN;
-                break;
-            default:
-                newComparison = comparison;
-                newNegation = !isNegated;
-        }
-
+    public BasicRestriction<T> negate() {
         return new BasicRestrictionRecord<>(
                 field,
-                newNegation,
-                newComparison,
+                comparison.negate(),
                 value);
     }
 }
--- a/api/src/main/java/jakarta/data/Operator.java
+++ b/api/src/main/java/jakarta/data/Operator.java
@@ -24,5 +24,40 @@ public enum Operator {
     IN,
     LESS_THAN,
     LESS_THAN_EQUAL,
-    LIKE
+    LIKE,
+    NOT_EQUAL,
+    NOT_IN,
+    NOT_LIKE;
+
+    /**
+     * Returns the operator that is the negation of this operator.
+     *
+     * @return the operator that is the negation of this operator.
+     */
+    Operator negate() {
+        switch (this) {
+            case EQUAL:
+                return NOT_EQUAL;
+            case GREATER_THAN:
+                return LESS_THAN_EQUAL;
+            case GREATER_THAN_EQUAL:
+                return LESS_THAN;
+            case IN:
+                return NOT_IN;
+            case LESS_THAN:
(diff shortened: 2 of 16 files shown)


api/src/test/java/jakarta/data/BasicRestrictionRecordTest.java
```diff
@@ -62,6 +65,23 @@ void shouldCreateBasicRestrictionWithNullValue() {
         });
     }
 
+    @Test
+    void shouldNegateLTERestriction() {
```
otaviojava, line 69: I should have a test scenario case when we negate twice, I mean negate a Restriction that is already `not`

otaviojava: I mean, not remove this one, but include a second one

api/src/test/java/jakarta/data/CompositeRestrictionRecordTest.java
```diff
@@ -71,6 +73,35 @@ void shouldFailIfEmptyRestrictions() {
                 .hasMessage("Cannot create a composite restriction without any restrictions to combine.");
     }
 
+    @Test
+    void shouldNegateCompositeRestriction() {
```
otaviojava, line 77: The same here, the negate, negate scenario.

api/src/test/java/jakarta/data/BasicRestrictionRecordTest.java
njr-11: > I should have a test scenario case when we negate twice, I mean negate a Restriction that is already `not`

Good idea. I added tests for negating a negated restriction here:
https://github.com/jakartaee/data/pull/905/commits/f5d3bb91143dd534ab0cda7089cf80b906a9dbc7

api/src/main/java/jakarta/data/BasicRestrictionRecord.java
```diff
@@ -36,4 +36,33 @@ record BasicRestrictionRecord<T>(
     BasicRestrictionRecord(String field, Operator comparison, Object value) {
         this(field, false, comparison, value);
     }
+
+    @Override
+    public Restriction<T> negate() {
+        boolean newNegation = isNegated;
+        Operator newComparison;
```
gavinking, line 43: This is incredibly anal, I know, but can we avoid the use of non-`final` instance vars here, please?

Also: remind we what version of Java we're targeting? Can we still not use `switch` expressions? I thought we could now.

api/src/main/java/jakarta/data/BasicRestrictionRecord.java
```diff
@@ -36,4 +36,33 @@ record BasicRestrictionRecord<T>(
...
+                break;
+            case LESS_THAN:
+                newComparison = Operator.GREATER_THAN_EQUAL;
+                break;
+            case LESS_THAN_EQUAL:
+                newComparison = Operator.GREATER_THAN;
+                break;
+            default:
+                newComparison = comparison;
+                newNegation = !isNegated;
```
gavinking, line 59: Actually I think it would make more sense to just add `NOT_IN` and `NOT_LIKE` operators (they exist in JDQL), and move all this logic to a `negated()` method on `Operator`.

njr-11: Good idea. That makes it possible for all operations to be negated and then makes sense for negation to move to Operator as you suggested. I added this in commit https://github.com/jakartaee/data/pull/905/commits/c480490190e05b90954166d48eb66d2d970280aa

api/src/main/java/jakarta/data/BasicRestrictionRecord.java
njr-11: > This is incredibly anal, I know, but can we avoid the use of non-`final` instance vars here, please?

The non-final variable went away after I moved the negation detail to Operator, so this is covered now.

> Also: remind we what version of Java we're targeting? Can we still not use `switch` expressions? I thought we could now.

Jakarta EE 11 targets Java SE 21 and Java SE 17.
Jakarta EE 12 will add Java SE 25 (and I assume drop 17 but not really sure).

It would make sense for Data 1.1 to aim for compatibility with all 3 of these Java levels.  I don't know of any concerns with `switch` statements for these levels.

api/src/main/java/jakarta/data/Operator.java
```diff
@@ -24,5 +24,40 @@ public enum Operator {
...
+                return GREATER_THAN;
+            case LIKE:
+                return NOT_LIKE;
+            case NOT_EQUAL:
+                return EQUAL;
+            case NOT_IN:
+                return IN;
+            case NOT_LIKE:
+                return LIKE;
+            default: // should be unreachable
```
otaviojava, line 59: Should it be the opposite?
I mean, NOT_EQUAL returning EQUAL

otaviojava (changes requested):

api/src/main/java/jakarta/data/Operator.java
njr-11: > Should it be the opposite? I mean, NOT_EQUAL returning EQUAL

Otavio, I don't understand your comment in relation to the code.  The negate method code switches NOT_EQUAL to EQUAL, and it switches EQUAL to NOT_EQUAL,

```
            case EQUAL:
                return NOT_EQUAL;
            case GREATER_THAN:
                return LESS_THAN_EQUAL;
            case GREATER_THAN_EQUAL:
                return LESS_THAN;
            case IN:
                return NOT_IN;
            case LESS_THAN:
                return GREATER_THAN_EQUAL;
            case LESS_THAN_EQUAL:
                return GREATER_THAN;
            case LIKE:
                return NOT_LIKE;
            case NOT_EQUAL:
                return EQUAL;
            case NOT_IN:
                return IN;
            case NOT_LIKE:
                return LIKE;
            default: // should be unreachable
```

Maybe I am overlooking something. Could you add a suggestion with the specific correction you want to see made?

api/src/main/java/jakarta/data/BasicRestrictionRecord.java
gavinking: OK, so we can use `switch` expressions then, right?

api/src/main/java/jakarta/data/Operator.java
```diff
@@ -24,5 +24,40 @@ public enum Operator {
...
+                throw new IllegalStateException(name());
+        }
```
gavinking, line 61:
```suggestion
        return switch (this) {
            case EQUAL -> NOT_EQUAL;
            case GREATER_THAN -> LESS_THAN_EQUAL;
            case GREATER_THAN_EQUAL -> LESS_THAN;
            case IN -> NOT_IN;
            case LESS_THAN -> GREATER_THAN_EQUAL;
            case LESS_THAN_EQUAL -> GREATER_THAN;
            case LIKE -> NOT_LIKE;
            case NOT_EQUAL -> EQUAL;
            case NOT_IN -> IN;
            case NOT_LIKE -> LIKE;
        }
```

api/src/main/java/jakarta/data/Operator.java
otaviojava: That was my point, so, please ignore it.

api/src/main/java/jakarta/data/BasicRestrictionRecord.java
njr-11: Yes, I accepted your suggestion to use the more concise `->` instead of `:` and followed that up with a commit to add `;` at the end which was also needed.  This is also nice because allowed us to get past the code checking police without needing to add the useless default throw exception.  It is actually better without the default because if someone adds a new enumerated value to Operator at some point, this method will not compile until they also add the implementation for negating it.

otaviojava (approved):

Issue #539, opened 2024-03-12
[Use Case]: multi-item 'select' lists in JDQL


gavinking:
### As a ...

- [ ] Application user/user of the configuration itself
- [X] API user (application developer)
- [ ] SPI user (container or runtime developer)
- [ ] Specification implementer

### I need to be able to ...

select multiple fields of my entity in a JDQL `select` list, and have them packaged into some nice typesafe `record` object.

### Which enables me to ...

handle "subset" or "simplified" views of the entity data, especially useful for externalization.

### Additional information

For reference, please see a similar proposal for JPQL: https://github.com/jakartaee/persistence/issues/420

My preferred way to handle this is the following:

```java
// declare a record type
record BookSummary(String title, String author) {}

// use it as the return type of a @Query method
@Query("select title, author.name from Book")
List<BookSummary> bookSummaries();
```

The repository method would automatically package the the elements of the `select` list into new instances of the record type. Of course, the "type" of the `select` list must match the signature of the constructor.

People sometimes find this useful for creating DTOs.

This would not be for 1.0, but for a future release.

gavinking:
A related idea, that I'm shamelessly stealing from Blaze-Persistence, is the notion of an "entity view".

From our perspective, it's a nice way to handle projections with `@Find` methods. Consider:

```java
// declare a record type
@Projection(Book.class)
record BookSummary(String title, String isbn, 
                   @Project("length(text)") int length, 
                   @Project("publisher.name") String publisherName) {}

// use it as the return type of a @Find method
@Find
List<BookSummary> bookSummaries();
```

Note that:

- In Blaze-Persistence, an entity view is an interface, but in modern Java, a `record` is a better fit.
- The `@Projection` annotation determines the queried entity class.
- By default, the fields of the record are matched to the field of the entity by name and type, but the `@Project` annotation lets you specify a fragment of JDQL.

This idea fits in _very_ nicely and consistently with #546.

beikov:
For reference, Blaze-Persistence Entity-Views also supports classes and records, but interfaces support multi-inheritance and updatable projections which is why that is preferred.

I wanted to propose a new Jakarta EE project at some point to standardize projections, as I believe it has more applicability than just for Jakarta Persistence and Data.

The idea is that this new Jakarta project provides annotations and means to introspect the model, as well as create and take apart object instances.
It's up to the respective other Jakarta implementations to then support these projection types. Wdyt about that? We should get together and talk about this some time.

[...]

PR #921, opened 2025-01-02
subset of entity attributes - Find methods

njr-11:
This PR contains some of the work for #539
It covers returning one or more entity attributes rather than the entire entity when using the `Find` annotation.
(A separate PR will be needed to cover selecting multiple entity attributes in JDQL and corresponding update to `@Query` methods to allow a record return type there).

This PR will enable users to write code like:

```
@Repository
public interface Cars extends BasicRepository<Car, String> {
    @Find
    @Select(_Car.PRICE)
    Optional<Float> getPrice(String vin);
} 
```

and

```
@Repository
public interface Cars extends BasicRepository<Car, String> {
    record ModelInfo(String make,
                     String model,
                     int year) {}

    @Find
    Optional<ModelInfo> getModelInfo(String vin);
} 
```

or for the case where the record component names don't match the entity attribute names,

```
@Repository
public interface Cars extends BasicRepository<Car, String> {
    record ModelInfo(String model,
                     String manufacturer,
                     int designYear) {}

    @Find
    @Select({_Car.MODEL, _Car.MAKE, _Car.YEAR})
    Optional<ModelInfo> getModelInfo(String vin);
} 
```

or an alternative way of doing the above by placing the annotation onto the record component(s) instead (which is convenient if you have control over the definition of the record and it is used by multiple repository methods),

```
@Repository
public interface Cars extends BasicRepository<Car, String> {
    record ModelInfo(String model,
                     @Select(_Car.MAKE) String manufacturer,
                     @Select(_Car.YEAR) int designYear) {}

    @Find
    Optional<ModelInfo> getModelInfo(String vin);
} 
```


gavinking:
This approach looks basically OK to me, though I have a couple of comments.

The first is that we need a way to specify the entity type when not using `BasicRepository`. This is what I would put in the `@Find` annotation. And I would work really hard to leave other stuff out of `@Find`. I would use a dedicated annotation for projections. So we could write:

```java
@Repository
public interface Cars  {
    record ModelInfo(String model,
                     String manufacturer,
                     int designYear) {}

    @Find(Car.class) 
    @Project({_Car.MODEL, _Car.MAKE, _Car.YEAR})
    Optional<ModelInfo> getModelInfo(String vin);
} 
```

The second comment is that I would like the _option_ of putting `@Project` on the fields of the record:

```java
@Repository
public interface Cars  {
    record ModelInfo(@Project(_Car.MODEL) String model,
                     @Project(_Car.MAKE) String manufacturer,
                     @Project(_Car.YEAR) int designYear) {}

    @Find(Car.class)
    Optional<ModelInfo> getModelInfo(String vin);
} 
```

gavinking:
i.e. I want to cut a very wide berth around horrible stuff like:

```java
    @Find(from = Car.class, value = {_Car.MODEL, _Car.MAKE, _Car.YEAR}) 
    Optional<ModelInfo> getModelInfo(String vin);
```

Let's not box ourselves into something like that.

njr-11:
> The first is that we need a way to specify the entity type when not using `BasicRepository`. This is what I would put in the `@Find` annotation. And I would work really hard to leave other stuff out of `@Find`. I would use a dedicated annotation for projections. So we could write:
> 
> ```java
> @Repository
> public interface Cars {
>     record ModelInfo(String model,
>                      String manufacturer,
>                      int designYear) {}
> 
>     @Find(Car.class) 
>     @Project({_Car.MODEL, _Car.MAKE, _Car.YEAR})
>     Optional<ModelInfo> getModelInfo(String vin);
> } 
> ```

That's a good idea. Having the value of `@Find` be the entity class would eliminate the need to define a primary entity type for the repository.  I'm fine with having a separate annotation to optionally specify the projection, and `@Project` works as a name for it. Another name that I would consider for it is `@Select` which would match the JDQL syntax for achieving the same thing.  And I suppose another advantage of separating out the projection to its own annotation is that it would open up the possibility of using it in combination with a partial JDQL query that omits the `SELECT` clause.

> The second comment is that I would like the option of putting @Project on the fields of the record

That's a nice improvement as well.  This allows for more concise code because it isolates specifying the entity attribute names to a single place (the record) rather than each repository method where it is returned.  Additionally, for records where all of the component names match except one, it could allow overriding only that one.

> i.e. I want to cut a very wide berth around horrible stuff like:
> 
> ```java
>     @Find(from = Car.class, value = {_Car.MODEL, _Car.MAKE, _Car.YEAR}) 
>     Optional<ModelInfo> getModelInfo(String vin);
> ```
> 
> Let's not box ourselves into something like that.

Totally agree. We should never add multiple fields to `@Find`.

I'll get the PR updated.

gavinking:
> Another name that I would consider for it is `@Select` which would match the JDQL syntax

That would be fine, but I usually try to avoid "select" just because selection is really the job of the `where` clause. But I agree that confusion is very unlikely here.

otaviojava:
Once we can archive it using JPQL, do we need a duplicate version to find the annotation?

Using JPQL is:

- Easier than finding annotation (we don't need an extra annotation to link this projection)
- Requires less cognitive load
- Most users are already familiar with the one from JPA.


Once it is a subset, we could use a similar structure:


```java
@Repository
public interface Cars {
    record ModelInfo(String model,
                     String manufacturer,
                     int designYear) {}

    @Query("select new ModelInfo(model, manufacturer, year) from Car")
    Optional<ModelInfo> getModelInfo(String vin);
} 

```

gavinking:
@otaviojava I think it's nice to be able to have both.

```java
    @Find(Car.class) 
    @Project({_Car.MODEL, _Car.MAKE, _Car.YEAR})
    Optional<ModelInfo> getModelInfo(String vin);
```

vs

```java
    @Query("select model, make, year from Car where vin = ?1") 
    Optional<ModelInfo> getModelInfo(String vin);
```

And:

```java
    @Find(Book.class) 
    @Project(_Book.Title)
    List<String> getModelInfo(Restriction<Book> restriction, Order<Book> order);
```

vs

```java
    @Query("select title from Book") 
    List<String> getModelInfo(Restriction<Book> restriction, Order<Book> order);
```

I agree that in these cases the JDQL versions look slightly cleaner, but the `@Project` versions don't look terrible to my eyes.

njr-11:
PR and description are updated to split out the entity attribute names and use the Find value for the entity type instead.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -40,7 +40,7 @@ A superinterface of a repository interface must either:
...
+Repositories perform operations on entities. For repository methods that are annotated with `@Insert`, `@Update`, `@Save`, or `@Delete`, the entity type is determined from the method parameter type. For repository methods that are annotated with a `@Find`, the entity type is determined from the annotation value, if a valid entity type. Otherwise, for `find` and `delete` methods where the return type is an entity, array of entity, or parameterized type such as `List<MyEntity>` or `Page<MyEntity>`, the entity type is determined from the method return type.  For `count`, `exists`, and other `find` and `delete` methods that do not return the entity or accept the entity as a parameter, the entity type cannot be determined from the method signature and a _primary entity type_ must be defined for the repository.
```
gavinking, line 43:
```suggestion
Repositories perform operations on entities. For repository methods that are annotated with `@Insert`, `@Update`, `@Save`, or `@Delete`, the entity type is determined from the method parameter type. For repository methods that are annotated with a `@Find`, the entity type is determined by the annotation `value` member, if an entity type is explicitly specified. Otherwise, for `find` and `delete` methods where the return type is an entity, array of entity, or parameterized type such as `List<MyEntity>` or `Page<MyEntity>`, the entity type is determined from the method return type.  For `count`, `exists`, and other `find` and `delete` methods that do not return the entity or accept the entity as a parameter, the entity type cannot be determined from the method signature and a _primary entity type_ must be defined for the repository.
```

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -26,11 +26,25 @@
 
 
 /**
- * <p>Annotates a repository method returning entities as a parameter-based automatic query method.</p>
+ * <p>Annotates a repository method to be a parameter-based automatic query method
```
gavinking, line 29: Nitpick, but I find "Annotates X to be Y" really clumsy English.

It's fine to write "Declares X to be Y", but I think "Annotates X as Y" was also perfectly OK.

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -55,23 +69,45 @@
  * instances with the given value of the {@code model} field.
  * </p>
  *
- * <p>A method annotated with {@code @Find} must return one of the following types:</p>
+ * <p>A method annotated with {@code @Find} must return one of the following types,
+ * where {@code E} is an entity type, entity attribute type, or record type:</p>
```
gavinking, line 73:
```suggestion
 * where {@code E} is the entity type, entity attribute type, or record type inferred or explicitly specified by the `@Find` annotation:</p>
```

gavinking (commented):
All looks very reasonable. Left minor comments.

spec/src/main/asciidoc/repository.asciidoc
njr-11: Thanks for improving the language/clarity here.

api/src/main/java/jakarta/data/repository/Find.java
njr-11: It looks like I rewrote that sentence when switching around two parts of it. I'll get that corrected back to "as" instead of "to be"

njr-11: Corrected in https://github.com/jakartaee/data/pull/921/commits/d93b48e148052c18563811fee1c931e0a6eff451 along with removal of an extra "a" that I spotted in a different section that I had updated.

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -85,4 +121,30 @@
...
+     * public interface Vehicles {
+     *     &#64;Find(Car.class)
+     *     &#64;Select("price")
+     *     Optional&lt;Float&gt; getPrice(@By("vin") String vehicleIdNum);
+     *
+     *     ...
+     * }
+     * </pre>
+     */
+    Class<?> value = Object.class;
```
gavinking, line 149: I would use `void.class` to indicate "unspecified".

```suggestion
    Class<?> value = void.class;
```

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -85,4 +121,30 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.METHOD)
 public @interface Find {
+    /**
+     * <p>Optionally specifies the entity type to look for in the database.</p>
+     *
+     * <p>The default value, {@code Object.class}, has the special meaning of
```
gavinking, line 127:
```suggestion
     * <p>The default value, {@code void.class}, has the special meaning of
```

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -85,4 +121,30 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.METHOD)
 public @interface Find {
+    /**
+     * <p>Optionally specifies the entity type to look for in the database.</p>
```
gavinking, line 125:
```suggestion
     * <p>Optionally specifies the queried entity type.</p>
```

gavinking (commented):
Additional comments.

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -55,23 +69,45 @@
...
+ * to request that only a subset of entity attributes be retrieved,</p>
+ * <pre>
+ * &#64;Repository
+ * public interface Cars extends BasicRepository&lt;Car, String&gt; {
+ *     record ModelInfo(String make,
+ *                      String model,
+ *                      int year) {}
+ *
+ *     &#64;Find
+ *     Optional&lt;ModelInfo&gt; getModelInfo(@By("vin") String vehicleIdNum);
```
otaviojava, line 95: I enjoy the @gavinking approach on this find one:

https://github.com/jakartaee/data/pull/921#issuecomment-2569090819


```java
   @Find(Car.class) 
   @Project({_Car.MODEL, _Car.MAKE, _Car.YEAR})
   Optional<ModelInfo> getModelInfo(String vin);
```

Thus, I centralized the "builder" into a single place.

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -85,4 +125,30 @@
...
+     * valid entity class as the value if the method does not return entities
+     * and the repository does not otherwise define a primary entity type.</p>
+     *
+     * <p>For example,</p>
+     *
+     * <pre>
+     * &#64;Repository
+     * public interface Vehicles {
+     *     &#64;Find(Car.class)
+     *     &#64;Select("price")
```
otaviojava, line 146: This goes against our spec, once "price" should not be a valid JPQL:

https://github.com/jakartaee/data/blob/main/api/src/main/java/jakarta/data/repository/Query.java#L130-L142

api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -0,0 +1,135 @@
...
+     * <p>A repository method can return a subset of entity attributes per result
+     * by having the result type be a Java record. The {@code Select} annotation
+     * can be used in the following ways to accommodate this.</p>
+     *
+     * <h3>Annotating a Repository Method</h3>
+     *
+     * <p>Place the {@code Select} annotation on a repository find method and
+     * assign the annotation value to be the names of multiple entity attributes,
+     * corresponding to the order and types of the components of the Java record
+     * that is used for the result type.</p>
```
KyleAure, line 88: If a user writes a `@Select` annotation with attributes in the incorrect order from the Record attributes what type of exception should we expect the repository implementation to throw? `MappingException` or something else?

api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -0,0 +1,135 @@
...
+     *                      &#64;Select(_Car.YEAR) int designYear) {}
+     *
+     *     &#64;Find
+     *     Optional&lt;ModelInfo&gt; getModelInfo(@By(_Car.VIN) String vehicleIdNum);
+     * }
+     * </pre>
+     *
+     * <p>For more concise code, the {@code Select} annotation can be omitted from
+     * record components that have the same name as the entity attribute name,
+     * such as {@code model} in the above example.</p>
```
KyleAure, line 128: Will this also require the `--parameters` compiler option? If so should we point that out here?

api/src/main/java/module-info.java
```diff
@@ -843,6 +844,48 @@
  *                                 Sort.asc("name"));
  * </pre>
  *
+ * <h2>Returning subsets of entity attributes</h2>
+ *
+ * <p>In addition to retrieving results that are entities, repository find methds
```
KyleAure, line 849:
```suggestion
 * <p>In addition to retrieving results that are entities, repository find methods
```

api/src/main/java/jakarta/data/repository/Select.java
njr-11: No, unlike method parameter names, which require `--parameters`, record component names are always available.

api/src/main/java/jakarta/data/repository/Select.java
njr-11: Given that it defines the mapping to be whatever it says it is, I don't think there is any way for a provider to consider the ordering to be incorrect.  A person can look at it and say that it doesn't make sense to map `_Car.MAKE` to a record component called `model` and map `_Car.MODEL` to a record component called `make` because a person will understand that such a thing is likely to be undesirable, but a Jakarta Data provider should not be expected to programmatically make that judgement.  Technically, it is okay to do such a thing, although in practice is it extremely unlikely to be desirable.

Some mismatches that could occur would be if there are different numbers of entity attribute mappings vs record components or if the data type of an entity attribute does not match the record component.  These are both detectable, and I think some Jakarta Data providers will detect them at compile time, others at run time, or others might defer the detection to an underlying Jakarta Persistence provider.  So I'm hesitant to require a specific exception type for this.

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -26,11 +26,29 @@
...
+ *
+ * <p>The method return type identifies one of the following:</p>
+ * <ul>
+ * <li>the entity type returned by the query,</li>
+ * <li>a single entity attribute type returned by the query (requires the
+ *     {@link Select} annotation), or</li>
+ * <li>a Java record type representing a subset of entity attributes returned
+ *     by the query. The names of the record components and entity attributes
+ *     must match, or the entity attribute names must be specified by the
+ *     {@link Select} annotation.</li>
```
otaviojava, line 44: Can we also use a regular class here?

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -55,23 +73,45 @@
...
- *     <li>{@code Stream<E>}, or</li>
- *     <li>{@code Page<E>} or {@code CursoredPage<E>}.</li>
+ *     <li>{@code Stream<E>},</li>
+ *     <li>{@code Page<E>}, or</li>
+ *     <li>{@code CursoredPage<E>} (only allowed when {@code E} is the entity type).</li>
  * </ul>
  *
- * <p>An automatic query method annotated {@code Find} returns an entity instance for every record which satisfies the
- * parameter-based conditions.</p>
+ * <p>For example, if a {@code Car} entity has attribute names including {@code make},
```
otaviojava, line 88: What happens if I have the same field?

```java

@Entity 
public class Person {

private CreditCard card;
private String name;
}

@Embedded
public class CreditCard {
private String name;
private String number;
}
```

```java
record ModelInfo(String name) {}
```

In this case, we have both with `name.` Which one should I take?

gavinking: We already have a convention for this which is to use `card_name` for fields of the embeddable.

api/src/main/java/jakarta/data/repository/Find.java
njr-11: > Can we also use a regular class here?

With a little bit of added complexity, this approach could certainly be extended to include non-record classes as well.  However, I would propose that in this PR we stick with record only to start out with, get that working, and then separately consider if we want to add in non-record classes.  Would that be okay with you?

gavinking: I agree with Nathan.

otaviojava: Baby-steps.
I like this approach.

otaviojava (approved):

PR #924, opened 2025-01-06
TCK challenge against boolean result of Query

### LINKED ISSUE #923: [TCK Challenge]: The TCK contains a repository method with an UPDATE query that returns boolean but this is not allowed in the spec and Javadoc
### Specification

https://github.com/jakartaee/data/blob/8adf740df3b0a55bf7255cf6f7e6a53fae468acd/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/MultipleEntityRepo.java#L54

### Assertion

https://github.com/jakartaee/data/blob/8adf740df3b0a55bf7255cf6f7e6a53fae468acd/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java#L2465

### TCK Version

1.0.1

JNoSQL Jakarta Persistence Extension

### Challenge Scenario

Claims that a test asserts requirements over and above that of the specification.

### Full Description

The test `testUpdateQueryWithWhereClause` in https://github.com/jakartaee/data/blob/8adf740df3b0a55bf7255cf6f7e6a53fae468acd/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java#L2465 asserts that the return value from calling the `move` method is `true`. However, it's not clear what the return value should represent.

The `move` method in the [MultipleEntityRepo](https://github.com/jakartaee/data/blob/8adf740df3b0a55bf7255cf6f7e6a53fae468acd/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/MultipleEntityRepo.java#L54) class is annotated with [@Query](https://github.com/jakartaee/data/blob/8adf740df3b0a55bf7255cf6f7e6a53fae468acd/api/src/main/java/jakarta/data/repository/Query.java#L28),which contains an `UPDATE` statement. I didn't find anywhere in the spec of javadoc any mention of what should be returned in this case. 

On the contrary, the spec suggests that returning a `boolean` type is forbidden in this case:

* The [@Query](https://github.com/jakartaee/data/blob/8adf740df3b0a55bf7255cf6f7e6a53fae468acd/api/src/main/java/jakarta/data/repository/Query.java#L28) Javadoc says: "For select statements, the return type of the query method must be consistent with the type returned by the query. For update or delete statements, it must be void, int or long."
* [In the specification document](https://github.com/jakartaee/data/blob/8adf740df3b0a55bf7255cf6f7e6a53fae468acd/spec/src/main/asciidoc/repository.asciidoc?plain=1#L139): "implementations are encouraged to support void as the return type for a query which never returns a result"

### Additional Context

I suggest that the method `move` in `MultipleEntityRepo` is modified to return `void` and remove the assertion. 

A future version of the `Data` spec could introduce the `boolean` return type and clarify what it means. it could e.g. mean that it's true when a `long` type would return > 0, false otherwise. 

It would also be good to clarify what the `int` and `long` return types mean, whether methods should return the number of modified entities or > 0 if some entities modified, or it a value of > 0 means an error. It's not no clearly stated anywhere in the spec of javadocs, I only assume it's the number of modified entities based on the assertions in the TCK tests.

njr-11:
Fixes #923 by switching the TCK test from the invalid `boolean` return value for `@Query("UPDATE ...` to instead have a return type of `int`.  Also, improves documentation of `Query` to clarify what the `int` and `long` return values mean.

A separate PR/issue should be used if we decide that we also want to support a boolean return type.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Query.java
+++ b/api/src/main/java/jakarta/data/repository/Query.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022,2024 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022,2025 Contributors to the Eclipse Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -34,8 +34,14 @@ import java.lang.annotation.Target;
  *
  * <p>The required {@link #value} member specifies the JDQL or JPQL query as a string.</p>
  *
- * <p>For {@code select} statements, the return type of the query method must be consistent with the type returned by
- * the query. For {@code update} or {@code delete} statements, it must be {@code void}, {@code int} or {@code long}.</p>
+ * <p>For {@code select} statements, the return type of the query method must be
+ * consistent with the type returned by the query.</p>
+ *
+ * <p>For {@code update} or {@code delete} statements, the return value must be one of:</p>
+ * <ul>
+ * <li>{@code void}</li>
+ * <li>{@code int} or {@code long}, where the value is the count of matching entities</li>
+ * </ul>
  *
  * <p>Compared to SQL, JDQL allows an abbreviated syntax for {@code select} statements:</p>
  * <ul>
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023,2024 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023,2025 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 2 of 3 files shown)


api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -34,8 +34,14 @@
...
  *
- * <p>For {@code select} statements, the return type of the query method must be consistent with the type returned by
- * the query. For {@code update} or {@code delete} statements, it must be {@code void}, {@code int} or {@code long}.</p>
+ * <p>For {@code select} statements, the return type of the query method must be
+ * consistent with the type returned by the query.</p>
+ *
+ * <p>For {@code update} or {@code delete} statements, the return value must be one of:</p>
+ * <ul>
+ * <li>{@code void}</li>
+ * <li>{@code int} or {@code long}, where the value is the count of matching entities</li>
```
gavinking, line 43:
```suggestion
 * <li>{@code int} or {@code long}, where the value is the number of matching entities</li>
```

api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -34,8 +34,14 @@
...
  *
- * <p>For {@code select} statements, the return type of the query method must be consistent with the type returned by
- * the query. For {@code update} or {@code delete} statements, it must be {@code void}, {@code int} or {@code long}.</p>
+ * <p>For {@code select} statements, the return type of the query method must be
+ * consistent with the type returned by the query.</p>
+ *
+ * <p>For {@code update} or {@code delete} statements, the return value must be one of:</p>
+ * <ul>
+ * <li>{@code void}</li>
+ * <li>{@code int} or {@code long}, where the value is the number of matching entities</li>
```
otaviojava, line 43: Please only include that with a NoSQL database that works with eventual persistence; this value might not be as precise as an ACID database. Such as, a relational database.

njr-11: Sure, I added that in https://github.com/jakartaee/data/pull/924/commits/111205c938788370b5b33cdc944cd4e3ee808ae9

otaviojava (approved):

PR #932, opened 2025-01-07
Create documentation to operator enum, plus create a new test scenario

otaviojava:


gavinking:
@otaviojava Note that [over here](https://github.com/jakartaee/data/issues/829#issuecomment-2574007664) I'm arguing that this enumeration should simply go away. Now, you guys might not go for that. But perhaps the time is not yet ripe for investing effort in documenting all this stuff.

otaviojava:
Ok, I will check the discussion and see it.

PR #933, opened 2025-01-07
Use separate package for Restrictions

### LINKED ISSUE #931: separate package for restrictions
On the topic of further API updates for restrictions, I noticed that 9 of the 13 classes in the `jakarta.data` package, (or 8 if we consider `Operator` to be general purpose) are all for restrictions.  This tells me restrictions ought to have its own package, maybe `jakarta.data.restrict`?  I'm not opening a PR for this yet because it would collide with the rename in #915 and it would be better to do these serially to avoid merge conflicts.  In the mean time, this is a good opportunity to collect feedback from others, if anyone wants a different package name, has thoughts on whether `Operator` should go with it or stay, and so forth.

njr-11:
Fixes #931
This moves the 9 classes that we have for restrictions into their own package so that they don't overwhelm the main `jakarta.data` package.

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/Attribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/Attribute.java
@@ -19,8 +19,8 @@ package jakarta.data.metamodel;
 
 import java.util.Set;
 
-import jakarta.data.restrict.Restrict;
-import jakarta.data.restrict.Restriction;
+import jakarta.data.metamodel.restrict.Restrict;
+import jakarta.data.metamodel.restrict.Restriction;
 
 /**
  * Represents an entity attribute in the {@link StaticMetamodel}.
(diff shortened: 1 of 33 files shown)


otaviojava:
@njr-11 I agree with you that you need to discuss this one more.
IMHO: I only see `Restrictions` to be used with the static model conflicts, thus, it should be on the metadata domain/package

otaviojava:
@njr-11 please, rebase this code.

njr-11:
> @njr-11 please, rebase this code.

Done. Thanks for pointing this out. It was needed to resolve a merge conflict.

> @njr-11 I agree with you that you need to discuss this one more.
> IMHO: I only see `Restrictions` to be used with the static model conflicts, thus, it should be on the metadata domain/package

It could make sense to put it under `metamodel`. I'll wait for the meeting on Wednesday to get feedback on what everyone thinks is the ideal place for it.

otaviojava (approved):

PR #935, opened 2025-01-09
standardize on the term entity attribute

### LINKED ISSUE #906: Can we use more consistent terminology for entity attribute/field/property?
Jakarta Data is currently very inconsistent in how it refers to entity attributes/fields/properties. Or sometimes persistent fields. Or sometimes (persistent) fields and properties.  By using different terms, there is some concern that users, especially newer users who are often the ones who most need to rely on the documentation, will believe they are referring to different things.

Whether or not we go back and try to be more consistent with what we have already done is one matter, but I would like to at least figure out what is preferred going forward before we starting adding documentation for v1.1 and beyond.

I did some quick searches across the Javadoc and spec documents to approximately figure out what the current usage is:

117 occurrences of `attribute`: 108 in Javadoc. 9 in spec doc.
94 occurrences of `property`: 59 in Javadoc. 35 is spec doc.
220 occurrences of `field`: 98 in Javadoc. 122 in spec doc.
11 occurrences of `field or property` and similar: 8 in Javadoc. 3 in spec doc.
52 occurrences of `persistent field`: 19 in Javadoc. 43 in spec doc.

A subset of this usage is encoded into the API and cannot change even if we wanted to:

- `Attribute` is used in class names `Attribute`, `SortableAttribute`, `TextAttribute` within the static metamodel (and also in the *Record implementation classes, which are lesser concern because users should not look at those)

- `property` is a record component name of the `Sort` API.

And we just recently added for v1.1 (so it could still be changed if we want),

- `field` is a record component name of the `BasicRestriction` API.

njr-11:
Fixes #906

Updates existing Javadoc and spec doc to refer to use consistent terminology when referring to entity attributes/fields/properties.

Hopefully this will be easier for our users to understand because we will no longer be switching back and forth between entity attribute / field / property / fields and properties, and so forth everywhere, which gave the appearance of them being different things.


api/src/main/java/jakarta/data/Sort.java
```diff
@@ -152,67 +158,67 @@ public boolean isDescending() {
...
     }
 
     /**
      * Create a {@link Sort} instance with {@link Direction#ASC ascending direction}
      * that does not request case insensitive ordering.
      *
-     * @param <T>      entity class of the sortable property.
-     * @param property the property name to order by
+     * @param <T>           entity class of the sortable entity attribute.
+     * @param attributeName name of the entity attribute to order by
```
otaviojava, line 178: `attributeName`sounds odd, why not just `attribute`?

njr-11: Thanks for the feedback. I can switch it to that. Here is the commit for it, https://github.com/jakartaee/data/pull/935/commits/d82c0449ce9d12950aa13046847fec01943b5744

api/src/main/java/jakarta/data/Sort.java
```diff
@@ -83,24 +83,28 @@
...
+     *     for case insensitive ascending sort on an entity attribute.</li>
+     * <li>{@link #desc(String) Sort.desc(attributeName)}
+     *     for descending sort on an entity attribute.</li>
+     * <li>{@link #descIgnoreCase(String) Sort.descIgnoreCase(attributeName)}
+     *      for case insensitive descending sort on an entity attribute.</li>
      * </ul>
      *
-     * @param property    name of the property to order by.
-     * @param isAscending whether ordering for this property is ascending (true) or descending (false).
+     * @param property    name of the entity attribute to order by.
```
otaviojava, line 106:
```suggestion
     * @param property name of the entity attribute to order by.
```
I would remove this space:

otaviojava: Nope, ignore this.

otaviojava (approved):

PR #962, opened 2025-02-05
implement toString to make Restrictions more readable

njr-11:
Adds implementation of toString on the built-in Restrictions to display in a concise way that is meaningful.

For example,
`name LIKE "Jakarta EE %"`
Or,
`ALL (price GREATER_THAN_EQUAL 50, price LESS_THAN 100)`

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/restrict/BasicRestrictionRecord.java
+++ b/api/src/main/java/jakarta/data/metamodel/restrict/BasicRestrictionRecord.java
@@ -49,14 +49,19 @@ record BasicRestrictionRecord<T>(
      */
     @Override
     public String toString() {
-        StringBuilder s = new StringBuilder();
-        s.append(attribute).append(' ')
+        String valueString = value == null ? "null" : value.toString();
+        StringBuilder builder = new StringBuilder(
+                attribute.length() +
+                comparison.name().length() +
+                valueString.length() +
+                4); // number of additional characters that might be appended
+        builder.append(attribute).append(' ')
          .append(comparison.name()).append(' ');
         if (value instanceof CharSequence) {
-            s.append('"').append(value).append('"');
+            builder.append('"').append(valueString).append('"');
         } else {
-            s.append(value);
+            builder.append(valueString);
         }
-        return s.toString();
+        return builder.toString();
     }
 }
--- a/api/src/main/java/jakarta/data/metamodel/restrict/BasicRestrictionRecord.java
+++ b/api/src/main/java/jakarta/data/metamodel/restrict/BasicRestrictionRecord.java
@@ -39,4 +39,24 @@ record BasicRestrictionRecord<T>(
                 comparison.negate(),
                 value);
     }
+
+    /**
+     * Textual representation of a basic restriction.
+     * For example,
+     * <pre>price LESS_THAN 50.0</pre>
+     *
+     * @return textual representation of a basic restriction.
+     */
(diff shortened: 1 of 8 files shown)


api/src/main/java/jakarta/data/metamodel/restrict/BasicRestrictionRecord.java
```diff
@@ -39,4 +39,24 @@ public BasicRestriction<T> negate() {
...
+    /**
+     * Textual representation of a basic restriction.
+     * For example,
+     * <pre>price LESS_THAN 50.0</pre>
+     *
+     * @return textual representation of a basic restriction.
+     */
+    @Override
+    public String toString() {
+        StringBuilder s = new StringBuilder();
```
otaviojava, line 52: If possible, avoid variables like `s`, please, use something like: 

`textExpression`

or any variation like this.

api/src/main/java/jakarta/data/metamodel/restrict/CompositeRestrictionRecord.java
```diff
@@ -43,4 +43,31 @@ record CompositeRestrictionRecord<T>(
...
+    /**
+     * Textual representation of a composite restriction.
+     * For example,
+     * <pre>ALL (price LESS_THAN 50.0, name LIKE "%Jakarta EE%")</pre>
+     *
+     * @return textual representation of a composite restriction.
+     */
+    @Override
+    public String toString() {
+        StringBuilder s = new StringBuilder(200);
```
otaviojava, line 56: Why this constructor have size and it does not happen with `BasicRestrictRecord`

I also would recommend magic number, generate a constant and explain about this `200`

api/src/main/java/jakarta/data/metamodel/restrict/BasicRestrictionRecord.java
njr-11: I renamed it in the latest commit.

api/src/main/java/jakarta/data/metamodel/restrict/CompositeRestrictionRecord.java
njr-11: Initially, I didn't supply a size, but the static analysis code checking tools that are built in to the build told me to add it. Apparently they didn't notice BasicRestrictRecord.  I updated all 3 methods with more precise estimates and made this one use a constant as you suggested in the latest commit.

otaviojava (approved):

gavinking:
> `ALL (price GREATER_THAN_EQUAL 50, price LESS_THAN 100)`

Why not align the `toString()`s with the equivalent JDQL, so:

```hql
(price >= 50 AND price < 100)
```

gavinking:
Oh, wait, it was already merged? A bit quick, no? I think we should always leave time for reviews/feedback.

njr-11:
> > `ALL (price GREATER_THAN_EQUAL 50, price LESS_THAN 100)`
> 
> Why not align the `toString()`s with the equivalent JDQL, so:
> 
> ```hql
> (price >= 50 AND price < 100)
> ```

That's a nice improvement, and I like how it is more concise.

> Oh, wait, it was already merged? A bit quick, no? I think we should always leave time for reviews/feedback.

I suppose he didn't realize I only created the PR yesterday?  I guess we'll use a separate PR for the improvement.

gavinking:
OK, perfect.

PR #963, opened 2025-02-06
Fix documentation about NoSQL database behavior.

otaviojava:
It fixes: [NoSQL databases need to be allowed to raise errors for function that the database type is not capable o](https://github.com/jakartaee/data/issues/782)

## Changes

Fixes:

- [ ] [TCK cannot require totalElements and totalPages of some NoSQL databases](https://github.com/jakartaee/data/pull/804) - Otavio needs to clarify which types of NoSQL databases can support counts and which cannot.
- [ ] Per #828 document that UnsupportedOperationException can be raised on Delete methods that return an update count when a non-relational database relies on eventual consistency and is incapable of knowing the count of deleted entities.

### DIFF
--- a/api/src/main/java/jakarta/data/page/Page.java
+++ b/api/src/main/java/jakarta/data/page/Page.java
@@ -163,8 +163,16 @@ public interface Page<T> extends Iterable<T> {
      * Returns the total number of pages of query results, if the {@link #pageRequest()}
      * specified that {@linkplain PageRequest#requestTotal the total should be retrieved
      * from the database}.
+     *
+     * <p><strong>Note for NoSQL databases:</strong> The ability to determine the total number of pages
+     * is dependent on support for total element queries in the database. Key-Value and Wide-Column
+     * databases generally do not support this feature. For Graph and Document databases, support may vary
+     * by provider. If the total page count cannot be determined, calling this method will result in an
+     * {@link UnsupportedOperationException}.</p>
+     *
      * @return the total number of pages.
      * @throws IllegalStateException if the total was not retrieved from the database.
+     * @throws UnsupportedOperationException if the database does not support total page count queries.
      */
     long totalPages();
 }
--- a/api/src/main/java/jakarta/data/page/Page.java
+++ b/api/src/main/java/jakarta/data/page/Page.java
@@ -146,8 +146,16 @@ public interface Page<T> extends Iterable<T> {
      * Returns the total number of elements across all pages of query results, if the
      * {@link #pageRequest()} specified that {@linkplain PageRequest#requestTotal the
      * total should be retrieved from the database}.
+     *
+     * <p><strong>Note for NoSQL databases:</strong> Not all NoSQL databases support counting the total
+     * number of elements. This operation is <strong>not supported</strong> for Key-Value and Wide-Column databases.
+     * For Graph and Document databases, support for this operation may vary depending on the provider.
+     * If the database does not support retrieving the total number of elements, calling this method will
+     * result in an {@link UnsupportedOperationException}.</p>
+     *
      * @return the total number of elements across all pages.
      * @throws IllegalStateException if the total was not retrieved from the database.
+     * @throws UnsupportedOperationException if the database does not support total element queries.
      */
     long totalElements();
 
--- a/api/src/main/java/jakarta/data/repository/Query.java
+++ b/api/src/main/java/jakarta/data/repository/Query.java
@@ -44,6 +44,17 @@ import java.lang.annotation.Target;
  *     The value might not be precise on databases that provide eventual consistency.</li>
  * </ul>
  *
+ * <p><strong>Note on NoSQL Databases:</strong></p>
+ * <p>Update and delete operations in NoSQL databases may not support returning {@code int} or {@code long}
+ * due to their eventual consistency model. Unlike traditional ACID-compliant relational databases,
+ * many NoSQL databases do not guarantee immediate consistency across distributed nodes. As a result,
+ * the number of affected entities might not be accurately reported.</p>
+ *
+ * <p>This limitation is not tied to a specific type of NoSQL database but applies across Key-Value,
+ * Wide-Column, Document, and Graph databases. When a NoSQL provider does not support returning
+ * {@code int} or {@code long} for these operations, invoking such a method will result in an
+ * {@link UnsupportedOperationException}.</p>
+ *
  * <p>Compared to SQL, JDQL allows an abbreviated syntax for {@code select} statements:</p>
  * <ul>
  * <li>The {@code from} clause is optional in JDQL. When it is missing, the queried entity is determined by the return


api/src/main/java/jakarta/data/page/Page.java
```diff
@@ -146,17 +146,33 @@ default Stream<T> stream() {
      * Returns the total number of elements across all pages of query results, if the
      * {@link #pageRequest()} specified that {@linkplain PageRequest#requestTotal the
      * total should be retrieved from the database}.
+     *
+     * <p><strong>Note for NoSQL databases:</strong> Not all NoSQL databases support counting the total
```
gavinking, line 150: Can we perhaps use `@apiNote` for this?

api/src/main/java/jakarta/data/page/Page.java
```diff
@@ -146,17 +146,33 @@ default Stream<T> stream() {
...
+     * number of elements. This operation is <strong>not supported</strong> for Key-Value and Wide-Column databases.
+     * For Graph and Document databases, support for this operation may vary depending on the provider.
+     * If the database does not support retrieving the total number of elements, calling this method will
+     * result in an {@link UnsupportedOperationException}.</p>
+     *
      * @return the total number of elements across all pages.
      * @throws IllegalStateException if the total was not retrieved from the database.
+     * @throws UnsupportedOperationException if the database does not support total element queries.
```
njr-11, line 158: It seems confusing to me to add UnsupportedOperationException alongside the existing IllegalStateException which the method already throws "if the total was not retrieved from the database."  I propose that we don't add the new exception or any of the preceding language and just add a sentence to the end of the exception we already have instead:

```suggestion
     * @throws IllegalStateException if the total was not retrieved from the database.
     *                               Some NoSQL databases are not capable of retrieving
     *                               a total number of elements.
```

api/src/main/java/jakarta/data/page/Page.java
```diff
@@ -146,17 +146,33 @@ default Stream<T> stream() {
...
      * Returns the total number of pages of query results, if the {@link #pageRequest()}
      * specified that {@linkplain PageRequest#requestTotal the total should be retrieved
      * from the database}.
+     *
+     * <p><strong>Note for NoSQL databases:</strong> The ability to determine the total number of pages
+     * is dependent on support for total element queries in the database. Key-Value and Wide-Column
+     * databases generally do not support this feature. For Graph and Document databases, support may vary
+     * by provider. If the total page count cannot be determined, calling this method will result in an
+     * {@link UnsupportedOperationException}.</p>
+     *
```
njr-11, line 172: Can reuse the existing exception here, too.
```suggestion
```

api/src/main/java/jakarta/data/page/Page.java
```diff
@@ -146,17 +146,33 @@ default Stream<T> stream() {
...
      * @return the total number of pages.
      * @throws IllegalStateException if the total was not retrieved from the database.
+     * @throws UnsupportedOperationException if the database does not support total page count queries.
```
njr-11, line 175:
```suggestion
     * @throws IllegalStateException if the total was not retrieved from the database.
     *                               Some NoSQL databases are not capable of retrieving
     *                               a total number of elements or pages.
```

api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -44,6 +44,17 @@
...
+ * <p>Update and delete operations in NoSQL databases may not support returning {@code int} or {@code long}
+ * due to their eventual consistency model. Unlike traditional ACID-compliant relational databases,
+ * many NoSQL databases do not guarantee immediate consistency across distributed nodes. As a result,
+ * the number of affected entities might not be accurately reported.</p>
+ *
+ * <p>This limitation is not tied to a specific type of NoSQL database but applies across Key-Value,
+ * Wide-Column, Document, and Graph databases. When a NoSQL provider does not support returning
+ * {@code int} or {@code long} for these operations, invoking such a method will result in an
+ * {@link UnsupportedOperationException}.</p>
+ *
```
njr-11, line 57: Can we more concisely state the allowance to raise UnsupportedOperationException on the related line that is right above this?

```suggestion
 *     The value might not be precise on databases that provide eventual consistency,
 *     in which case some Jakarta Data providers might choose to raise
 *     {@link UnsupportedOperationException} instead of returning an imprecise value.
 *     </li>
 * </ul>
 *
```

api/src/main/java/jakarta/data/page/Page.java
otaviojava: But @njr-11 IMHO: it is a different scenario, I mean: 

- IllegalStateException: if the total was not retrieved from the database.

The point is there are NoSQL databases that cannot do it; IMHO, in this scenario, UnsupportedOperationException fits semantically better.

api/src/main/java/jakarta/data/page/Page.java
otaviojava: I don't like this approach:

IllegalStateException:

> Signals that a method has been invoked at an illegal or inappropriate time.

UnsupportedOperationException:

> Thrown to indicate that the requested operation is not supported.

In our case, it is not about a state, once there are NosQL databases that will never return this value.


IMHO: The semantics matter, and it does not make sense to IllegalStateException or other exceptions, such as NullPointerException, just for reuse exception.

otaviojava:
@njr-11 Thanks for the notes; I have a different way to see the reuse of exceptions. Could you check it and let me know your thoughts?

api/src/main/java/jakarta/data/page/Page.java
njr-11: > I don't like this approach:
> 
> [IllegalStateException](https://docs.oracle.com/en/java/javase//21/docs/api/java.base/java/lang/IllegalStateException.html):
> 
> > Signals that a method has been invoked at an illegal or inappropriate time.
> 
> [UnsupportedOperationException](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/UnsupportedOperationException.html):
> 
> > Thrown to indicate that the requested operation is not supported.
> 
> In our case, it is not about a state, once there are NosQL databases that will never return this value.
> 
> IMHO: The semantics matter, and it does not make sense to IllegalStateException or other exceptions, such as NullPointerException, just for reuse exception.

I don't remember the specifics of why IllegalStateException was chosen during Data 1.0. It might have to do with the fact that the user has gotten the Page into an illegal state by making a PageRequest that requires the Page to have totals despite the database being incapable of retrieving them.  If the Jakarta Data provider did not reject the PageRequest with totals upfront and returned a Page instance, the returned Page instance is in an illegal state because no total was retrieved from the database, and thus no total can be supplied when the user invokes the method to ask for it.  You can argue that it would have been preferable to use UnsupportedOperationException instead.  Either exception can be justified.  But IllegalStateException was already chosen, and the historical precedent matters.  A provider that implemented Data 1.0 would be raising IllegalStateException for this and application code written to Data 1.0 would be trapping for IllegalStateException if it wanted to handle this case, because that is what the API currently says.  I don't see a sufficiently good justification to switch this because the existing behavior does also make sense.  It would be especially confusing to see two different exceptions with causes of `the total was not retrieved from the database` and `the database does not support total page count queries` when the problem that needs to be reported is that the "total was not retrieved from the database because the database does not support total page count queries".  That is confusing to both the Jakarta Data provider who doesn't know which exception to throw and the application developer who doesn't know which exception to catch.

otaviojava: Nathan, my goal is not to overwrite history. I agree with you that we should not break compatibility.  

However, my point is that in **Java**, we have **different exceptions for different contexts**, which helps **differentiate failure scenarios**.

## Java Persistence API Examples**

Using **Java Persistence-related examples**, we can see how the API differentiates errors based on context:

#### 1. [EntityManager.find()](https://jakarta.ee/specifications/persistence/2.2/apidocs/javax/persistence/entitymanager#find(java.lang.Class,java.lang.Object,javax.persistence.LockModeType))

#### 2. [EntityManager.getReference()](https://jakarta.ee/specifications/persistence/2.2/apidocs/javax/persistence/entitymanager#getReference(java.lang.Class,java.lang.Object))

- **Throws:**
  - `IllegalArgumentException` – If the **first argument is not an entity type**, or the **primary key type is invalid or null**.
  - `EntityNotFoundException` – If the **entity state cannot be accessed**.

#### 3. [EntityManager.flush()](https://jakarta.ee/specifications/persistence/2.2/apidocs/javax/persistence/entitymanager#flush())

- **Throws:**
  - `TransactionRequiredException` – If **there is no active transaction** or the entity manager is not **joined to the current transaction**.
  - `PersistenceException` – If the **flush operation fails**.


My point here is **not** to ignore **past decisions** or break compatibility.  

However, we **clearly** have **two different error scenarios**:

1. **Illegal State** → When the **total was not retrieved** from the database or the database **does not load**.
2. **Unsupported Operation** → When the **total will *never* be retrieved** because the **database does not support it**.  

For **case (2)**, it should be `UnsupportedOperationException`, as the operation is inherently **not supported** by the system.

This distinction exists in **Java/Jakarta APIs** for a reason, and we should **apply the same principle** in Jakarta Data.

gavinking: So I think really the mismatch here is that we've specified that `PageRequest.requestTotal()` is enabled by default, when on some NoSQL databases, totals are simply never available.

- _Ideally_, on such platforms, `requestTotal()` would be _disabled_ by default, and passing a `PageRequest` with `requestTotal()` enabled would result in an `IllegalArgumentException` _when the repository method is called_.

- But instead, since it's enabled by default, you've have been thinking that the error occurs later, when `totalElements()` or `totalPages()` is called. This isn't actually semantically clean. :-(

How can we clean this up without imposing on users the requirement to explicitly call `withoutTotal()` every time they create a `PageRequest` on these platforms?

- One option is to simply specify that this setting is completely ignored on platforms which don't support totals. This option is equivalent to what Nathan has been arguing for. It's the simplest solution, but, as I say, it's not very clean.

- A second option would be to say that `requestTotal()` is disabled by default on platforms which don't support totals. We would have to change the implementation of `Pagination` to keep track of whether `requestTotal` was set explicitly by the user, or was just left to be defaulted, but that looks OK. We would have a problem with the signature of `boolean requestTotal()` in `PageRequest`, which we might have to deprecate. Not perfect.

- A third option would be to give users a way to control the default globally `PageRequest.requestTotalByDefault(false)`.  Also not perfect.

njr-11 (changes requested):
See review comments.  There is no good solution for NoSQL, so I'll mostly agree to the behavior that Otavio wants, but I still see this as a more undesirable way for a provider to handle things, so I don't want to see significant portions of the JavaDoc for these methods being cluttered up with it and distracting from more core usage.

api/src/main/java/jakarta/data/page/Page.java
otaviojava: I got it, yeap.

Thank you for both for explaining.

api/src/main/java/jakarta/data/page/Page.java
```diff
@@ -146,8 +146,18 @@ default Stream<T> stream() {
      * Returns the total number of elements across all pages of query results, if the
      * {@link #pageRequest()} specified that {@linkplain PageRequest#requestTotal the
      * total should be retrieved from the database}.
+     * {@apiNote Not all NoSQL databases support counting the total
+     * number of elements. This operation is not supported for Key-Value and Wide-Column databases.
+     * For Graph and Document databases, support for this operation may vary depending on the provider.
+     * If the database does not support retrieving the total number of elements, calling this method will
+     * result in an {@link UnsupportedOperationException}.</p>
+     *}
```
njr-11, line 154: `{@apiNote ...}` in curly braces didn't seem to work when I tried it. Should it instead be formatted like `@param` or `@return` ?

```suggestion
     * @apiNote Not all NoSQL databases support counting the total number of elements.
     *          This operation is not supported for Key-Value and Wide-Column databases.
     *          For Graph and Document databases, support for this operation may vary
     *          depending on the provider. If the database does not support retrieving the
     *          total number of elements, calling this method will result in an
     *          {@link UnsupportedOperationException}.
```

njr-11 (approved):

PR #966, opened 2025-02-07
address review comment to format Restriction toString more like JDQL

njr-11:
Address this review comment from #962

              > `ALL (price GREATER_THAN_EQUAL 50, price LESS_THAN 100)`

Why not align the `toString()`s with the equivalent JDQL, so:

```hql
(price >= 50 AND price < 100)
```

_Originally posted by @gavinking in https://github.com/jakartaee/data/issues/962#issuecomment-2640902362_

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/restrict/Operator.java
+++ b/api/src/main/java/jakarta/data/metamodel/restrict/Operator.java
@@ -18,16 +18,16 @@
 package jakarta.data.metamodel.restrict;
 
 public enum Operator {
-    EQUAL("="),
-    GREATER_THAN(">"),
-    GREATER_THAN_EQUAL(">="),
-    IN("IN"),
-    LESS_THAN("<"),
-    LESS_THAN_EQUAL("<="),
-    LIKE("LIKE"),
-    NOT_EQUAL("<>"),
-    NOT_IN("NOT IN"),
-    NOT_LIKE("NOT LIKE");
+    EQUAL,
+    GREATER_THAN,
+    GREATER_THAN_EQUAL,
+    IN,
+    LESS_THAN,
+    LESS_THAN_EQUAL,
+    LIKE,
+    NOT_EQUAL,
+    NOT_IN,
+    NOT_LIKE;
 
     /**
      * Representation of the operator in query language.
--- a/api/src/main/java/jakarta/data/metamodel/restrict/Operator.java
+++ b/api/src/main/java/jakarta/data/metamodel/restrict/Operator.java
@@ -18,16 +18,36 @@
 package jakarta.data.metamodel.restrict;
 
 public enum Operator {
-    EQUAL,
-    GREATER_THAN,
-    GREATER_THAN_EQUAL,
-    IN,
-    LESS_THAN,
-    LESS_THAN_EQUAL,
-    LIKE,
-    NOT_EQUAL,
-    NOT_IN,
-    NOT_LIKE;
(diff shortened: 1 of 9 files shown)


api/src/main/java/jakarta/data/metamodel/restrict/Operator.java
```diff
@@ -18,16 +18,36 @@
...
+    EQUAL("="),
+    GREATER_THAN(">"),
+    GREATER_THAN_EQUAL(">="),
+    IN("IN"),
+    LESS_THAN("<"),
+    LESS_THAN_EQUAL("<="),
+    LIKE("LIKE"),
+    NOT_EQUAL("<>"),
+    NOT_IN("NOT IN"),
+    NOT_LIKE("NOT LIKE");
```
gavinking, line 30:
```suggestion
    EQUAL,
    GREATER_THAN,
    GREATER_THAN_EQUAL,
    IN,
    LESS_THAN,
    LESS_THAN_EQUAL,
    LIKE,
    NOT_EQUAL,
    NOT_IN,
    NOT_LIKE;
```

api/src/main/java/jakarta/data/metamodel/restrict/Operator.java
```diff
@@ -18,16 +18,36 @@
...
+    /**
+     * Representation of the operator as it appears in query language.
+     * For example, {@link #GREATER_THAN} is represented as {@code >}
+     * in query langugae.
+     *
+     * @return the representation of the operator in query language.
+     */
+    String asQueryLanguage() {
+        return qlRepresentation;
+    }
```
gavinking, line 50:
```suggestion
    /**
     * Representation of the operator as it appears in query language.
     * For example, {@link #GREATER_THAN} is represented as {@code >}
     * in query langugae.
     *
     * @return the representation of the operator in query language.
     */
    String asQueryLanguage() {
        return switch (this) {
            case EQUAL -> "=";
            case GREATER_THAN -> ">";
            case GREATER_THAN_EQUAL -> ">=";
            case IN -> "IN";
            case LESS_THAN -> "<";
            case LESS_THAN_EQUAL -> "<=";
            case LIKE -> "LIKE";
            case NOT_EQUAL -> "<>";
            case NOT_IN -> "NOT IN";
            case NOT_LIKE -> "NOT LIKE";
        };
    }
```

gavinking (commented):
Just a suggestion: in modern Java I would use a `switch` expression instead of adding a constructor to the enum. I find that to be cleaner.

njr-11:
> Just a suggestion: in modern Java I would use a `switch` expression instead of adding a constructor to the enum. I find that to be cleaner.

I had it that way at first but switched to the constructor to be more performant - to return a fixed known value instead of re-computing it every time.  But I suppose this is only for a toString operation that is most likely to be used when dumping information for debug and performance doesn't matter at all then.  I can change it if you like.

gavinking:
Switching on an enum should be very very fast, I believe.

njr-11:
> Switching on an enum should be very very fast, I believe.

It's updated now with your suggestions, plus I added a commit to apply the same pattern to the other enum where I had added constructor parameters.

gavinking:
Awesome, thanks 🙏

gavinking (approved):
thanks Nathan, LGTM

otaviojava:
@njr-11 could you review this PR?

gavinking:
@otaviojava I fixed the conflict.

PR #967, opened 2025-02-11
Allow NoSQL to defer request for totals and raise UnsupportedOperationException later

njr-11:
Accommodate NoSQL databases that are incapable of page totals by clarifying that the request for totals can be deferred until they are asked for, and prepare the user for the possibility of an exception to be raised at that time rather than immediately upon supplying the PageRequest with requestTotals to a repository method.

This goes along with Otavio's #963 where we are trying to figure out a compatible way for NoSQL databases that aren't capable of retrieving totals to be usable for PageRequest instances that by default do not set withoutTotals.

### DIFF
--- a/api/src/main/java/jakarta/data/page/PageRequest.java
+++ b/api/src/main/java/jakarta/data/page/PageRequest.java
@@ -219,11 +219,18 @@ public interface PageRequest {
     int size();
 
     /**
-     * Indicates that a query method which returns a {@link Page}
+     * <p>Indicates that a query method which returns a {@link Page}
      * should retrieve the {@linkplain Page#totalElements() total
      * number of elements} available across all pages. This behavior
      * is enabled by default. To obtain a page request with total
-     * retrieval disabled, call {@link #withoutTotal()}.
+     * retrieval disabled, call {@link #withoutTotal()}.</p>
+     *
+     * <p>The total can be requested from the database before returning
+     * the page, or the request for the total can be deferred until the
+     * {@link Page#totalElements()} or {@link Page#totalPages()} method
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/data/page/PageRequest.java
```diff
@@ -219,11 +219,18 @@ static PageRequest beforeCursor(Cursor cursor, long pageNumber, int maxPageSize,
...
      * number of elements} available across all pages. This behavior
      * is enabled by default. To obtain a page request with total
-     * retrieval disabled, call {@link #withoutTotal()}.
+     * retrieval disabled, call {@link #withoutTotal()}.</p>
+     *
+     * <p>The total can be requested from the database before returning
+     * the page, or the request for the total can be deferred until the
+     * {@link Page#totalElements()} or {@link Page#totalPages()} method
+     * is invoked, which can result in an exception being raised if the
+     * request fails or the database is incapable of it.</p>
```
gavinking, line 232:
```suggestion
     * <p>A repository implementation might obtain a total from the 
     * database before returning the page of results, or might defer 
     * fetching the total until {@link Page#totalElements()} or 
     * {@link Page#totalPages()} method is invoked. In the case of
     * deferred fetching, the call to {@code totalElements()} or
     * {@code totalPages()} raises an exception if the database
     * request fails or if the database is incapable of computing
     * totals.</p>
```

gavinking:
> clarifying that the request for totals can be deferred until they are asked for, and prepare the user for the possibility of an exception to be raised at that time 

This is a good solution to the problem. 👍

gavinking (approved):

PR #969, opened 2025-02-12
TCK test of record result types where the repository method uses Select to identify the entity attributes because record component names do not match

njr-11:
Add TCK tests returning Page/List/Optional/array of record where the record component names do no match the entity attribute names, and the annotation is used to identify the corresponding entity attributes.


KyleAure (approved):
LGTM

otaviojava (approved):

PR #970, opened 2025-02-12
introduce BasicAttribute, ComparableAttribute to make restrictions typesafe

gavinking:
See explanation here: https://github.com/jakartaee/data/issues/829#issuecomment-2654769536

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/impl/BasicAttributeRecord.java
+++ b/api/src/main/java/jakarta/data/metamodel/impl/BasicAttributeRecord.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023,2025 Contributors to the Eclipse Foundation
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


api/src/main/java/jakarta/data/metamodel/impl/BasicAttributeRecord.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
```
njr-11, line 2:
```suggestion
 * Copyright (c) 2025 Contributors to the Eclipse Foundation
```

api/src/main/java/jakarta/data/metamodel/impl/ComparableAttributeRecord.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
```
njr-11, line 2:
```suggestion
 * Copyright (c) 2025 Contributors to the Eclipse Foundation
```

njr-11 (approved):
Very nicely done! I like the choice of the names BasicAttribute and ComparableAttribute, which together with TextAttribute (which did successfully retrofit) become the 3 attribute classes to use in a static metamodel.  Attribute then becomes a superclass that ought to not be used directly in the metamodel.  And SortableAttribute can be portrayed as a type that is only needed for a corner case where an attribute is sortable but not Comparable (I don't currently know of any useful example).

gavinking:
> And SortableAttribute can be portrayed as a type that is only needed for a corner case where an attribute is sortable but not java.lang.Comparable (I don't currently know of any useful example).

One example is `byte[]`, which we have as a basic type :-)

PR #971, opened 2025-02-13
fix up treatment of null restrictions

gavinking:
This makes the following changes:

1. Introduces `UnaryRestriction` and `UnaryOperator` to represent `is null` and `is not null`.
2. Disallows `null` as a value in `BasicRestriction`, since the meaning of `= null` and `<> null` varies between datastores.

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
@@ -36,8 +36,7 @@ public interface BasicAttribute<T,V> extends Attribute<T> {
     }
 
     default Restriction<T> isNull() {
-        //TODO: fix this "= NULL" is not the same thing as "IS NULL" in JDQL
-        return Restrict.equalTo(null, name());
+        return Restrict.isNull(name());
     }
 
     default Restriction<T> notEqualTo(V value) {
@@ -52,8 +51,7 @@ public interface BasicAttribute<T,V> extends Attribute<T> {
     }
 
     default Restriction<T> notNull() {
-        //TODO: fix this "<> NULL" is not the same thing as "IS NOT NULL" in JDQL
-        return Restrict.notEqualTo(null, name());
+        return Restrict.isNotNull(name());
     }
 
 }
--- a/api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
@@ -51,7 +51,7 @@ public interface BasicAttribute<T,V> extends Attribute<T> {
     }
 
     default Restriction<T> notNull() {
-        return Restrict.isNotNull(name());
+        return Restrict.notNull(name());
     }
 
 }
--- a/api/src/main/java/jakarta/data/metamodel/restrict/Restrict.java
+++ b/api/src/main/java/jakarta/data/metamodel/restrict/Restrict.java
@@ -79,6 +79,10 @@ public class Restrict {
         return new TextRestrictionRecord<>(attribute, Operator.EQUAL, value);
     }
 
+    public static <T> UnaryRestriction<T> isNull(String attribute) {
+        return new UnaryRestrictionRecord<>(attribute, UnaryOperator.IS_NULL);
+    }
+
     public static <T, V extends Comparable<V>> Restriction<T> greaterThan(V value, String attribute) {
         return new BasicRestrictionRecord<>(attribute, Operator.GREATER_THAN, value);
     }
@@ -144,6 +148,10 @@ public class Restrict {
 
     public static <T> TextRestriction<T> notEqualTo(String value, String attribute) {
         return new TextRestrictionRecord<>(attribute, Operator.NOT_EQUAL, value);
+    }
+
+    public static <T> UnaryRestriction<T> isNotNull(String attribute) {
+        return new UnaryRestrictionRecord<>(attribute, UnaryOperator.IS_NOT_NULL);
     }
 
     public static <T> TextRestriction<T> notContains(String substring, String attribute) {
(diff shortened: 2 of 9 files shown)


api/src/main/java/jakarta/data/metamodel/restrict/UnaryOperator.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.metamodel.restrict;
+
+public enum UnaryOperator {
```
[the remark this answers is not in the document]
gavinking: A `BasicRestriction` can have a (binary) `Operator`. A `UnaryExpression` can have a `UnaryOperator`. Type safety calls for them to be different types.

otaviojava (approved):

api/src/main/java/jakarta/data/metamodel/restrict/UnaryOperator.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2023,2025 Contributors to the Eclipse Foundation
```
njr-11, line 2:
```suggestion
 * Copyright (c) 2025 Contributors to the Eclipse Foundation
```

api/src/main/java/jakarta/data/metamodel/restrict/UnaryRestriction.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2023,2025 Contributors to the Eclipse Foundation
```
njr-11, line 2:
```suggestion
 * Copyright (c) 2025 Contributors to the Eclipse Foundation
```

api/src/main/java/jakarta/data/metamodel/restrict/UnaryRestrictionRecord.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) 2023,2025 Contributors to the Eclipse Foundation
```
njr-11, line 2:
```suggestion
 * Copyright (c) 2025 Contributors to the Eclipse Foundation
```

api/src/main/java/jakarta/data/metamodel/restrict/UnaryOperator.java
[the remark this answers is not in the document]
njr-11: I agree, as long as `BasicRestriction` and `UnaryRestriction` are two separate classes, it makes sense to have `Operator` and `UnaryOperator`.  If we didn't have `UnaryRestriction` and instead `BasicRestriction` represented both, then it would make sense to have only `Operator`.  The case for keeping them separate would be that the `value()` of `BasicRestriction` is unused in the case of unary.  The case for combining them would be having less interfaces/classes.

api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
```diff
@@ -52,8 +51,7 @@ default Restriction<T> notIn(Set<V> values) {
     }
 
     default Restriction<T> notNull() {
-        //TODO: fix this "<> NULL" is not the same thing as "IS NOT NULL" in JDQL
-        return Restrict.notEqualTo(null, name());
+        return Restrict.isNotNull(name());
```
njr-11, line 54: Recommend switching `isNotNull` to `notNull` in order to align with our other names: `notEqualTo`, `notIn`, `notNull` (from BasicAttribute), `notContains`, `notEndsWith`, `notLike`, `notStartsWith`.  

```suggestion
        return Restrict.notNull(name());
```

api/src/main/java/jakarta/data/metamodel/restrict/Restrict.java
```diff
@@ -146,6 +150,10 @@ public static <T> TextRestriction<T> notEqualTo(String value, String attribute)
         return new TextRestrictionRecord<>(attribute, Operator.NOT_EQUAL, value);
     }
 
+    public static <T> UnaryRestriction<T> isNotNull(String attribute) {
```
njr-11, line 153:
```suggestion
    public static <T> UnaryRestriction<T> notNull(String attribute) {
```

njr-11 (commented):
Looks good - thanks for correcting this! I added a few minor comments/suggestions

api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
gavinking: So I'm not sure what to do here. The issue is that since `null` is a keyword and can't be a method name, it has to be `isNull`. But then `isNull`/`notNull` doesn't conform with the pattern.

api/src/main/java/jakarta/data/metamodel/restrict/UnaryOperator.java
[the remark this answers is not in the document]
gavinking: Right.

api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
njr-11: > So I'm not sure what to do here. The issue is that since `null` is a keyword and can't be a method name, it has to be `isNull`. But then `isNull`/`notNull` doesn't conform with the pattern.

Yeah, I'm not sure either.  At least for now, I'd say name them `Restrict.isNull` and `Restrict.notNull` to be consistent with the `BasicAttribute.isNull` and `BasicAttribute.notNull` methods that we already have.  Then our spec will at least match in case we forget to revisit the naming. And if we do decide to change them later, we can change them all together.

gavinking: OK, fine. I can't think of anything better.

gavinking: Done.

api/src/main/java/jakarta/data/metamodel/restrict/Restrict.java
gavinking: Done.

njr-11 (approved):

PR #975, opened 2025-02-14
introduce Constraint and friends

gavinking:
I'm sharing this work in progress so that you guys can monitor the progress and offer feedback.

Note that this is a minimal-impact first phase of what I have in my head.

There are two problems with this design:

1. We still have `UnaryRestriction` - nullability isn't handled as a Range. Ideally, I would like to see `UnaryRestriction` go away.
2. There's a redundancy between `Operator` and `Range` - really `Operator` is unnecessary in this design, but we could perhaps move it into `Range`.

So in a second phase I would try to address these issues, though I would like to hear some reactions first.

Note that this change removes `TextRestriction` (case insensitivity becomes an aspect of the `Pattern`). If case-insensitive equality is needed:

- we could introduce a `TextValue` range type, or
- `TextRestriction` could be rescued in a simpler form.


gavinking:
I've now done the second phase.

> 1. We still have `UnaryRestriction` - nullability isn't handled as a Range. Ideally, I would like to see `UnaryRestriction` go away.

I've removed `UnaryRestriction` and `UnaryOperator` and replaced them with the `Null` range.

> 2. There's a redundancy between `Operator` and `Range`

I've made `Operator` a thing which is inferred from the `Range` subtype, to eliminate the redundancy. `Operator` could now be removed completely, but I intuit that you guys might like to keep it around.

> - we could introduce a `TextValue` range type, or
> - `TextRestriction` could be rescued in a simpler form.

I've gone with the second option, to demonstrate how this can have a minimal impact on the existing API.

gavinking:
In the third phase, we need to nail down exactly the API for this stuff:

1. Where do application developers go to get instance of `Range`? The `Range` subtypes themselves aren't a perfect place for that. We could perhaps add static methods to the `Range` interface.
2. Should `Range` be a sealed type as I have it here? The reason for doing it is that users aren't allowed to define their own implementations. On the other hand, a Data *provider* might like to define extensions.
3. Should we attempt to hide the `Range` subtypes from clients?

Etc.

But we need to discuss this stuff.

njr-11:
> In the third phase, we need to nail down exactly the API for this stuff:
> 
>     1. Where do application developers go to get instance of `Range`? The `Range` subtypes themselves aren't a perfect place for that. We could perhaps add static methods to the `Range` interface.
> 
>     2. Should `Range` be a sealed type as I have it here? The reason for doing it is that users aren't allowed to define their own implementations. On the other hand, a Data _provider_ might like to define extensions.
> 
>     3. Should we attempt to hide the `Range` subtypes from clients?

I agree with hiding the Range subtypes and having them be obtained from Range. It would otherwise be a lot of clutter for the user to sift through trying to figure what are all the available classes that they can use and where to find them.  Having them be obtained from Range makes them easily discoverable from a single place - the place that users will first go looking because the repository method signature told them to supply something of type `Range`.

A better way to hide them will be to use package protected or private where they will not be part of the API at all rather than in an `impl` package that users can see but need to ignore. There will be the question of how does the Data provider figure out which type of range it is, but the Data provider will need to be invoking methods on the Range anyways to obtain the values against which to compare and so forth, so it is natural to also obtain the Operator from the Range.

Also, while looking over this, I was wondering why we have different method names on Range, for example,
```
Range.upperBound(value)
Range.lowerBound(value)
Range.enumeration(value1, value2, value3)
```
rather than
```
Range.lessThanEqual(value)
Range.greaterThanEqual(value)
Range.in(value1, value2, value3)
```
where the latter exactly match the corresponding method names from Restrict and the static metamodel attribute interfaces (and also generally match up with Query by Method Name).

gavinking:
> I agree with hiding the `Range` subtypes and having them be obtained from `Range`. 

OK, so this is what I actually implemented, the factory methods are currently on `Range` and `TextRange`.

> A better way to hide them will be to use package protected or private where they will not be part of the API at all rather than in an impl package that users can see but need to ignore.

Unfortunately Jakarta Data _implementors_ are going to need to access to them, so what I did was move them to a `.impl` package, which is the strategy we've used for other SPI stuff.

gavinking:
> Also, while looking over this, I was wondering why we have different method names on `Range`

I'm OK with the idea of changing `Range.upperBound(value)` to `Range.lessThanEqual(value)`.

Slightly less keen on `Range.in()`, but if you have a strong opinion on that I guess I don't mind.

njr-11:
> Unfortunately Jakarta Data _implementors_ are going to need to access to them

It isn't yet clear to me why that is.  Range is basically a Restriction that doesn't specify the entity attribute name.  For Restriction, the provider implementation only needs the attribute name/value/operator (plus case insensitivity if TextAttribute). As best I can see, excluding the attribute name which we instead obtain/infer it from the method signature, all of those same things should be similarly obtainable from a Range.  I see methods already there to get the   `range.operator()` and `textrange.caseSensitive()`.  I think only the value is missing, but I'm sure we can figure out how to add that.

gavinking:
> It isn't yet clear to me why that is. 

Because they need the state, which is not always a single value, for example:

- `Interval` has an `upperBound` and a `lowerBound`
- `Enumeration` has a list of values,
- etc.

gavinking:
I mean, sure, you could have `getValue()` that might return a single value or might return a list, or might return an ordered pair of two values, but that's just nasty.

njr-11:
> I mean, sure, you could have `getValue()` that might return a single value or might return a list, or might return an ordered pair of two values, but that's just nasty.

Hmmm. It seems like we do already have that approach for Restriction (using value for single or set of values). I assumed we had something better.

```
        Restriction<Employee> r = _Person.yearOfBirth.in(Set.of(2020, 2022, 2024));
        Object value = ((BasicRestriction<Employee>) r).value(); // this is the Set<Integer>
```

This was an oversight that we didn't think about it, and it needs to be considered and done properly for `Restriction`, too.

gavinking:
>  I assumed we had something better.

Same. I missed that.

gavinking:
Actually I believe `Range` already solves it. You do `basicRestriction.range()` and you switch on its concrete type.

gavinking:
Casting to `BasicRestriction` is perfectly fine, FTR. It's an API.

njr-11:
Yes, casting to BasicRestriction is fine.  Switching the concrete type is not fine if we have the concrete types in an impl package.  Technically it works, but the idea of an impl package is that it's the internal implementation of spec, and a part of the API that anyone should ever use.  And we really shouldn't have externalized it in the first place.  I would like to get away from the impl packages as well given that it's a bad practice that we basically allowed as a compromise to hurriedly get Jakarta Data 1.0 done in time.  The implementation stuff should have been private or package protected.  We should definitely not add more of it.

gavinking:
> the idea of an impl package is that it's the internal implementation of spec

But that's not true ... every one of the existing `impl` packages is an SPI that's intended for use by Jakarta Data implementors. They're all exported by the module descriptor. None of them are module-private.

And there's nothing at all wrong with the pattern of switching over implementations of a sealed interface. That's Java's way of expressing a sum type, and it's a basic programming technique in languages with non-stupid type systems (e.g ML, Haskell, etc).

njr-11:
> But that's not true ... every one of the existing `impl` packages is an SPI that's intended for use by Jakarta Data implementors. They're all exported by the module descriptor. None of them are module-private.

I'm saying they shouldn't have been.  It was a mistake.  Anything that was truly intended to be spi would have been more appropriately place in a package named `spi` rather than `impl`.

gavinking:
(There is, to be clear, no internal implementation detail of Jakarta Data exposed by any exported package.)

njr-11:
> If this is just a question of _naming_ then I have no objection to renaming the package `jakarta.data.metammodel.range.spi`.

If we decide SPI is the right way to go, then yes, that would be appropriate.  Some discussion needs to be had by everyone about what the proper place is for these Range subtypes.  Initially, I wanted to see them all hidden completely, and that is still an option, although we would have some additional work to do in determining a good way to externalize the value/values. Alternatively, they could be SPI as you point out.  Or, they could be API.  There is at least one use case I can see for the latter, although I don't know if it is a good idea or not.  If they became API, then repository methods could be written to enforce/encode more requirements upfront, such as I was trying to achieve with `@Is`, but not quite as nicely because they force more work (wrapping values with ranges) and complexity (figuring out how to create a range instance that is consistent with the intent of the method) onto the repository user.  Still, they could potentially used for that purpose if we were willing to put up with the explosion of repository special parameter types that this would introduce.  Lots of different options here.  I'm not sure which directly is the best to go in at this point.

njr-11:
I have been giving some thought to the terminology that is used in the API for Range.  The ideas behind the pattern are sound, including the concept of splitting up a Restriction (BasicRestriction) into two parts: an entity attribute and what that entity attribute value needs to satisfy in order to be considered matching.  In some cases, this can be easily conceptualized as a range, but for others the term range seems awkward and not very intuitive, especially those related to patterns of text, but some others as well.  I decided to try to think of some different names than `Range` that might make sense.

If I think of an entity attribute whose value "must end with the letters 'EE'", or "must be larger than 10", or "must not be null", or "must contain the text `Jakarta`", or "must match the pattern `Ja%a EE` independent of case", I could describe all of these things as conditions that the entity attribute value must meet.  I think we should consider renaming `Range` to `Condition`.  It makes a lot of sense to me to think of a `Restriction` on an entity attribute as a pairing of the entity attribute with a condition it must meet.  I was already wanting to have consistency with Restrictions on method names, and Condition fits very nicely with this:  `Condition.lessThanEqual(max)`, `Condition.startsWith(prefix)`, `Condition.like(pattern)` ...

I understand that renaming `Range` would also impact the meaningfulness of the names of the subtypes, which we are currently trying to figure out a place for: internal, SPI, or API.  This might help here as well.  If going with the name `Condition`, it would make a lot of sense for the subtypes to simply match the names of the core Restrict/Condition methods (and which also match Query by Method Name) -- `LessThan`, `Like`, `Between`, and so forth.  This added consistency would make it more agreeable to have them on the API or SPI because they would now be less likely to confuse users, and where they do have a potential benefit of enabling a repository author to encode more requirements into methods upfront and name them more meaningfully.  I don't want to push for that too heavily though.  I'd rather see us cover this with `@Is`, but if `@Is` is unwanted by others, this could at least be a decent alternative.

gavinking:
If you want to go down that path, `Constraint` might be better than `Condition`.

gavinking:
what time is the meeting? 9AM EST?

njr-11:
> what time is the meeting? 9AM EST?

Yes, Here's the Jakarta calendar for US Eastern Time, which shows 9AM on Wednesday, February 19:
https://calendar.google.com/calendar/u/0/embed?src=eclipse-foundation.org_e9ki8t2gc75sh07qdh95c8ofvc@group.calendar.google.com&ctz=America/New_York

gavinking:
Thanks, just checking.

gavinking:
I've implemented the things we discussed on the call.

Questions:

1. Do we need to kinds of `TextConstraint`, or it it enough to just have `Like` with two internal implementations, which is what I have just pushed.
2. Do we need `LessThan` and `LessThanOrEqual` instead of the `boolean strict()`. I guess that's better.

gavinking:
> I guess that's better.

I have done it.

api/src/main/java/jakarta/data/metamodel/constraint/Constraint.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
```
njr-11, line 2:
```suggestion
 * Copyright (c) 2025 Contributors to the Eclipse Foundation
```

api/src/main/java/jakarta/data/metamodel/constraint/Constraint.java
```diff
@@ -0,0 +1,64 @@
...
+package jakarta.data.metamodel.constraint;
+
+import jakarta.data.metamodel.restrict.Operator;
+
+import java.util.Set;
+
+public interface Constraint<T> {
+    Operator operator();
+
+    static <T> Constraint<T> value(T value) {
```
njr-11, line 27: Align with Restrict.equalTo
```suggestion
    static <T> Constraint<T> equalTo(T value) {
```

api/src/main/java/jakarta/data/metamodel/constraint/GreaterThanOrEqualRecord.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public Operator operator() {
+        return Operator.GREATER_THAN_EQUAL;
+    }
+
+    @Override
+    public String toString() {
+        return bound.toString();
```
njr-11, line 37: Distinguish GreaterThanEqual.toString output from EqualTo.toString output
```suggestion
        return ">= " + bound.toString();
```

api/src/main/java/jakarta/data/metamodel/constraint/GreaterThanRecord.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public Operator operator() {
+        return Operator.GREATER_THAN;
+    }
+
+    @Override
+    public String toString() {
+        return bound.toString();
```
njr-11, line 37:
```suggestion
        return "> " + bound.toString();
```

api/src/main/java/jakarta/data/metamodel/constraint/InRecord.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof InRecord<?> that
+            && values.equals(that.values);
+    }
+
+    @Override
+    public int hashCode() {
+        return values.hashCode();
+    }
```
njr-11, line 45: The other records have toString, so I assume this one was meant to as well,
```suggestion
    }

    @Override
    public String toString() {
        return "IN " + values;
    }
```

api/src/main/java/jakarta/data/metamodel/constraint/LessThanOrEqualRecord.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public Operator operator() {
+        return Operator.LESS_THAN_EQUAL;
+    }
+
+    @Override
+    public String toString() {
+        return bound.toString();
```
njr-11, line 37:
```suggestion
        return "<= " + bound.toString();
```

api/src/main/java/jakarta/data/metamodel/constraint/LessThanRecord.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public Operator operator() {
+        return Operator.LESS_THAN;
+    }
+
+    @Override
+    public String toString() {
+        return bound.toString();
```
njr-11, line 37:
```suggestion
        return "< " + bound.toString();
```

api/src/main/java/jakarta/data/metamodel/constraint/Like.java
```diff
@@ -0,0 +1,48 @@
...
+        return LikePatternRecord.substring(substring);
+    }
+
+    static Like prefix(String substring) {
+        return LikePatternRecord.prefix(substring);
+    }
+
+    static Like suffix(String substring) {
+        return LikePatternRecord.suffix(substring);
+    }
```
njr-11, line 47: Improves clarity, but feel free to ignore this comment if you like.
```suggestion
    static Like prefix(String prefix) {
        return LikePatternRecord.prefix(prefix);
    }

    static Like suffix(String suffix) {
        return LikePatternRecord.suffix(suffix);
    }
```

api/src/main/java/jakarta/data/metamodel/constraint/LikeLiteralRecord.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public String pattern() {
+        return literal;
+    }
+
+    @Override
+    public Like ignoreCase() {
+        return new LikeLiteralRecord(literal, true);
+    }
```
njr-11, line 43: caseSensitive = false to ignore case
```suggestion
    public Like ignoreCase() {
        return new LikeLiteralRecord(literal, false);
    }
```

api/src/main/java/jakarta/data/metamodel/constraint/LikeLiteralRecord.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public Operator operator() {
+        return Operator.EQUAL;
+    }
+
+    @Override
+    public String toString() {
+        return "'" + literal + "'";
```
njr-11, line 52: Distinguishes this from EqualTo.toString
```suggestion
        return "LIKE '" + literal + "'" + (caseSensitive ? "" : " IGNORE CASE");
```

api/src/main/java/jakarta/data/metamodel/constraint/LikePatternRecord.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof LikePatternRecord that
+            && pattern.equals(that.pattern)
+            && caseSensitive == that.caseSensitive;
+    }
+
+    @Override
+    public int hashCode() {
+        return pattern.hashCode();
```
njr-11, line 112:
```suggestion
        return Objects.hash(pattern, caseSensitive);
```

api/src/main/java/jakarta/data/metamodel/restrict/BasicRestrictionRecord.java
```diff
@@ -21,24 +21,30 @@
...
-        String attribute,
-        Operator comparison,
-        Object value) implements BasicRestriction<T> {
+record BasicRestrictionRecord<T>(String attribute, Constraint<?> constraint, boolean negated)
+        implements BasicRestriction<T> {
 
     BasicRestrictionRecord {
         Objects.requireNonNull(attribute, "Attribute must not be null");
-        Objects.requireNonNull(value, "Value must not be null");
+        Objects.requireNonNull(constraint, "Range must not be null");
```
njr-11, line 33:
```suggestion
        Objects.requireNonNull(constraint, "Constraint must not be null");
```

api/src/main/java/jakarta/data/metamodel/restrict/TextRestrictionRecord.java
```diff
@@ -17,76 +17,46 @@
...
-        Operator comparison,
-        boolean isCaseSensitive,
-        boolean isEscaped,
-        String value) implements TextRestriction<T> {
+record TextRestrictionRecord<T>(String attribute, Like constraint, boolean negated)
+        implements TextRestriction<T> {
 
     TextRestrictionRecord {
         Objects.requireNonNull(attribute, "Attribute must not be null");
+        Objects.requireNonNull(constraint, "Pattern must not be null");
```
njr-11, line 29: Made more generic in case it is a Literal rather than a Pattern
```suggestion
        Objects.requireNonNull(constraint, "Constraint must not be null");
```

api/src/main/java/jakarta/data/metamodel/constraint/EqualToRecord.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public Operator operator() {
+        return Operator.EQUAL;
+    }
+
+    @Override
+    public String toString() {
+        return value instanceof String ? "'" + value + "'" : value.toString();
```
njr-11, line 36: Make distinguishable from other constraints
```suggestion
        return value instanceof String ? "= '" + value + "'" : "= " + value.toString();
```

api/src/main/java/jakarta/data/metamodel/restrict/TextRestrictionRecord.java
```diff
@@ -17,76 +17,46 @@
...
-            builder.append('"').append(valueString).append('"');
-        }
-        if (isEscaped) {
-            builder.append(" ESCAPED");
-        }
-        if (!isCaseSensitive) {
-            builder.append(" IGNORE_CASE");
-        }
-        return builder.toString();
+        return attribute + ' ' + comparison().asQueryLanguage() + ' ' + constraint;
```
njr-11, line 60: Now I see why some constraint records were only printing the value and not the operation.  I proposed change it so that a user will see a more accurate and meaningful output when invoking toString on a Constraint.
```suggestion
        return attribute + ' ' + constraint;
```

api/src/main/java/jakarta/data/metamodel/constraint/LikePatternRecord.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public Operator operator() {
+        return Operator.LIKE;
+    }
+
+    @Override
+    public String toString() {
+        return "'" + pattern + "'";
```
njr-11, line 55: Even after this update, it will look the same as the output of LikeLiteralRecord, so we need to do something further to distinguish. Possibly adding in ESCAPED=true/false?  I can't add it here because that information might have disappeared, so we need to look into that as well.
```suggestion
        return "LIKE '" + pattern + "'" + (caseSensitive ? "" : " IGNORE CASE");
```

api/src/main/java/jakarta/data/metamodel/constraint/NullRecord.java
```diff
@@ -15,13 +15,13 @@
...
-package jakarta.data.metamodel.restrict;
+package jakarta.data.metamodel.constraint;
 
-public interface UnaryRestriction<T> extends Restriction<T> {
-    String attribute();
-
-    UnaryOperator operator();
+import jakarta.data.metamodel.restrict.Operator;
 
+record NullRecord<T>() implements Null<T> {
```
njr-11, line 22: equals and hashCode are missing
```suggestion
record NullRecord<T>() implements Null<T> {
    @Override
    public boolean equals(Object obj) {
        return obj instanceof NullRecord;
    }

    @Override
    public int hashCode() {
        return 0;
    }
```

api/src/main/java/jakarta/data/metamodel/constraint/NullRecord.java
```diff
@@ -15,13 +15,13 @@
...
     @Override
-    UnaryRestriction<T> negate();
+    public Operator operator() {
+        return Operator.NULL;
+    }
```
njr-11, line 26:
```suggestion
    }


    @Override
    public String toString() {
        return "IS NULL";
    }
```

api/src/main/java/jakarta/data/metamodel/restrict/BasicRestrictionRecord.java
```diff
@@ -50,20 +56,12 @@ public BasicRestriction<T> negate() {
...
-            builder.append('"').append(valueString).append('"');
-        } else {
-            builder.append(valueString);
-        }
-        return builder.toString();
+        final Operator comparison = comparison();
+        final String op = comparison.asQueryLanguage();
+        return switch (comparison.arity()) {
+            case 1 -> attribute + ' ' + op;
+            case 2,3 -> attribute + ' ' + op + ' ' + constraint;
```
njr-11, line 63: Goes along with other comments
```suggestion
            case 2,3 -> attribute + ' ' + constraint;
```

njr-11 (commented):
This looks great.  I reviewed some of the files and added a number of minor comments/suggestions.

I didn't review Restrict or TextAttribute because the methods were moved around and that made it difficult to see what was actually changing. Would it be possible to preserve the existing order of methods for the PR, and later use a separate PR if you want to sort them another way. (I think they are currently sorted alphabetically).

I also didn't review the test classes yet.

njr-11:
> 1. Do we need to kinds of `TextConstraint`, or it it enough to just have `Like` with two internal implementations, which is what I have just pushed.

I would say let's go with the approach you chose of not having `TextConstraint` and see if everything works without it (I think it will).

api/src/main/java/module-info.java
```diff
@@ -443,7 +423,7 @@
  *
  * <tr style="vertical-align: top"><td>{@code Between}</td>
  * <td>sortable basic types</td>
- * <td>Requires that the entity's attribute value be within the range specified by two parameters,
+ * <td>Requires that the entity's attribute value be within the constraint specified by two parameters,
```
njr-11, line 426:
```suggestion
 * <td>Requires that the entity's attribute value be within the range specified by two parameters,
```

api/src/main/java/module-info.java
```diff
@@ -766,7 +746,7 @@
  *
  * <p>The number of results returned by a single invocation of a repository
  * find method may be limited by adding a parameter of type {@link Limit}.
- * The results may even be limited to a positioned range. For example,</p>
+ * The results may even be limited to a positioned constraint. For example,</p>
```
njr-11, line 749:
```suggestion
 * The results may even be limited to a positioned range. For example,</p>
```

api/src/main/java/module-info.java
```diff
@@ -775,7 +755,7 @@
  * ...
  * first50 = products.highlyDiscounted(0.30, Limit.of(50));
  * ...
- * second50 = products.highlyDiscounted(0.30, Limit.range(51, 100));
+ * second50 = products.highlyDiscounted(0.30, Limit.constraint(51, 100));
```
njr-11, line 758:
```suggestion
 * second50 = products.highlyDiscounted(0.30, Limit.range(51, 100));
```

njr-11 (commented):
These 3 changes to JavaDoc were probably accidental due to global replace of range with constraint.

api/src/main/java/module-info.java
```diff
@@ -16,26 +16,6 @@
...
-import jakarta.data.repository.Insert;
-import jakarta.data.repository.OrderBy;
-import jakarta.data.repository.Param;
-import jakarta.data.repository.Query;
-import jakarta.data.repository.Repository;
-import jakarta.data.repository.Save;
-import jakarta.data.repository.Select;
-import jakarta.data.repository.Update;
-
 import java.util.Set;
```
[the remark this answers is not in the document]
gavinking: Weird, I didn't mean to remove them, IntelliJ did. Strange that IntelliJ would kill necessary imports.

api/src/main/java/jakarta/data/metamodel/constraint/Between.java
```diff
@@ -15,13 +15,9 @@
...
+package jakarta.data.metamodel.constraint;
 
-public interface UnaryRestriction<T> extends Restriction<T> {
-    String attribute();
-
-    UnaryOperator operator();
-
-    @Override
-    UnaryRestriction<T> negate();
+public interface Between<T extends Comparable<T>> extends Constraint<T> {
```
dstepanov, line 20: Why between is represented as with `GreaterThanOrEqual` and `LessThanOrEqual`? Some data sources might have a native between operation

gavinking: Well, sure, every SQL database has a native `between`, as does JDQL. But it's a very easy job for such databases to do this transformation.

njr-11: Having to do a transformation should only be a concern if an implementation were being asked to observe a pairing of GTE and LTE and identify this as an intended BETWEEN.  Prior to Gavin's changes we had that, but he actually got rid of that in this PR and switched it to `Between` as a first class `Constraint` type.  An implementation will identify BETWEEN by observing the `Between` subinterface the same way it would identify less than from a `LessThan` subinterface.

dstepanov: @njr-11 I'm OK with between, it looks strange to me to include GTE and LTE in the between definition.

api/src/main/java/jakarta/data/metamodel/constraint/LikeLiteralRecord.java
```diff
@@ -0,0 +1,66 @@
...
+record LikeLiteralRecord(String literal, boolean caseSensitive)
+        implements Like {
+
+    public LikeLiteralRecord {
+        Objects.requireNonNull(literal, "Literal must not be null");
+    }
+
+    public LikeLiteralRecord(String literal) {
+        this(literal, true);
+    }
```
njr-11, line 33:
```suggestion
record LikeLiteralRecord(String literal, boolean caseSensitive, boolean escaped)
        implements Like {

    public LikeLiteralRecord {
        Objects.requireNonNull(literal, "Literal must not be null");
        if (escaped) {
            throw new IllegalArgumentException("Literal values must not be escaped");
        }
    }

    public LikeLiteralRecord(String literal) {
        this(literal, true, false);
    }

    LikeLiteralRecord(String literal, boolean caseSensitive) {
        this(literal, caseSensitive, false);
    }
```

api/src/main/java/jakarta/data/metamodel/constraint/LikeLiteralRecord.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof LikeLiteralRecord that
+            && literal.equals(that.literal)
+            && caseSensitive == that.caseSensitive;
+    }
+
+    @Override
+    public int hashCode() {
+        return literal.hashCode();
```
njr-11, line 64:
```suggestion
        return Objects.hash(literal, caseSensitive);
```

api/src/main/java/jakarta/data/metamodel/constraint/LikePatternRecord.java
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.metamodel.constraint;
+
+import jakarta.data.metamodel.restrict.Operator;
+
+import java.util.Objects;
+
+record LikePatternRecord(String pattern, boolean caseSensitive)
```
njr-11, line 24:
```suggestion
record LikePatternRecord(String pattern, boolean caseSensitive, boolean escaped)
```

api/src/main/java/jakarta/data/metamodel/constraint/LikePatternRecord.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof LikePatternRecord that
+            && pattern.equals(that.pattern)
+            && caseSensitive == that.caseSensitive;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(pattern, caseSensitive);
```
njr-11, line 112:
```suggestion
        return Objects.hash(pattern, caseSensitive, escaped);
```

api/src/main/java/jakarta/data/metamodel/constraint/LikePatternRecord.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+        return result.toString();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof LikePatternRecord that
+            && pattern.equals(that.pattern)
+            && caseSensitive == that.caseSensitive;
```
njr-11, line 107:
```suggestion
        return obj instanceof LikePatternRecord that
            && pattern.equals(that.pattern)
            && caseSensitive == that.caseSensitive
            && escaped == that.escaped;
```

api/src/main/java/jakarta/data/metamodel/constraint/LikePatternRecord.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public Operator operator() {
+        return Operator.LIKE;
+    }
+
+    @Override
+    public String toString() {
+        return "LIKE '" + pattern + "'" + (caseSensitive ? "" : " IGNORE CASE");
```
njr-11, line 55:
```suggestion
        return "LIKE '" + pattern + "'" + (caseSensitive ? "" : " IGNORE CASE") + (escaped ? " ESCAPE '\\'" : "");
```

api/src/main/java/jakarta/data/metamodel/constraint/LikePatternRecord.java
```diff
@@ -0,0 +1,114 @@
...
+    public LikePatternRecord {
+        Objects.requireNonNull(pattern, "Pattern must not be null");
+    }
+
+    public LikePatternRecord(String pattern) {
+        this(pattern, true);
+    }
+
+    public LikePatternRecord(String pattern, char charWildcard, char stringWildcard) {
+        this(translate(pattern, charWildcard, stringWildcard));
```
njr-11, line 40:
```suggestion
        this(translate(pattern, charWildcard, stringWildcard), true, true);
```

api/src/main/java/jakarta/data/metamodel/constraint/LikePatternRecord.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public LikePatternRecord ignoreCase() {
+        return new LikePatternRecord(pattern, false);
```
njr-11, line 45:
```suggestion
        return new LikePatternRecord(pattern, false, escaped);
```

api/src/main/java/jakarta/data/metamodel/constraint/LikePatternRecord.java
```diff
@@ -0,0 +1,114 @@
...
+        return new LikePatternRecord(escape(prefix) + STRING_WILDCARD);
+    }
+
+    public static LikePatternRecord suffix(String suffix) {
+        return new LikePatternRecord(STRING_WILDCARD + escape(suffix));
+    }
+
+    public static LikePatternRecord substring(String substring) {
+        return new LikePatternRecord(STRING_WILDCARD + escape(substring) + STRING_WILDCARD);
+    }
```
njr-11, line 68:
```suggestion
    public static LikePatternRecord prefix(String prefix) {
        return new LikePatternRecord(escape(prefix) + STRING_WILDCARD, true, true);
    }

    public static LikePatternRecord suffix(String suffix) {
        return new LikePatternRecord(STRING_WILDCARD + escape(suffix), true, true);
    }

    public static LikePatternRecord substring(String substring) {
        return new LikePatternRecord(STRING_WILDCARD + escape(substring) + STRING_WILDCARD, true, true);
    }
```

api/src/main/java/jakarta/data/metamodel/constraint/Like.java
```diff
@@ -0,0 +1,48 @@
...
+    String pattern();
+    boolean caseSensitive();
+    Like ignoreCase();
+
+    static Like literal(String literal) {
+        return new LikeLiteralRecord(literal);
+    }
+
+    static Like pattern(String pattern) {
+        return new LikePatternRecord(pattern);
```
njr-11, line 30:
```suggestion
        return new LikePatternRecord(pattern, true, false);
```

api/src/test/java/jakarta/data/metamodel/restrict/TextRestrictionRecordTest.java
```diff
@@ -31,72 +32,64 @@ class TextRestrictionRecordTest {
...
 
         SoftAssertions.assertSoftly(soft -> {
             soft.assertThat(restriction.attribute()).isEqualTo("title");
             soft.assertThat(restriction.comparison()).isEqualTo(Operator.LIKE);
-            soft.assertThat(restriction.value()).isEqualTo("%Java%");
-            soft.assertThat(restriction.isCaseSensitive()).isTrue();
-            soft.assertThat(restriction.isEscaped()).isFalse();
+            soft.assertThat(restriction.constraint().pattern()).isEqualTo("%Java%");
+            soft.assertThat(restriction.constraint().caseSensitive()).isTrue();
+//            soft.assertThat(restriction.isEscaped()).isFalse();
```
njr-11, line 43:
```suggestion
            soft.assertThat(((Like) restriction.constraint()).escaped()).isFalse();
```

api/src/test/java/jakarta/data/metamodel/restrict/TextRestrictionRecordTest.java
```diff
@@ -31,72 +32,64 @@ class TextRestrictionRecordTest {
...
 
         SoftAssertions.assertSoftly(soft -> {
             soft.assertThat(restriction.attribute()).isEqualTo("title");
             soft.assertThat(restriction.comparison()).isEqualTo(Operator.NOT_LIKE);
-            soft.assertThat(restriction.value()).isEqualTo("%Java%");
-            soft.assertThat(restriction.isCaseSensitive()).isTrue();
-            soft.assertThat(restriction.isEscaped()).isFalse();
+            soft.assertThat(restriction.constraint().pattern()).isEqualTo("%Java%");
+            soft.assertThat(restriction.constraint().caseSensitive()).isTrue();
+//            soft.assertThat(restriction.isEscaped()).isFalse();
```
njr-11, line 60:
```suggestion
            soft.assertThat(((Like) restriction.constraint()).escaped()).isFalse();
```

api/src/test/java/jakarta/data/metamodel/restrict/TextRestrictionRecordTest.java
```diff
@@ -31,72 +32,64 @@ class TextRestrictionRecordTest {
...
-            soft.assertThat(textRestriction.attribute()).isEqualTo("title");
-            soft.assertThat(textRestriction.comparison()).isEqualTo(Operator.LIKE);
-            soft.assertThat(textRestriction.value()).isEqualTo("%Java%");
-            soft.assertThat(textRestriction.isCaseSensitive()).isFalse();
-            soft.assertThat(textRestriction.isEscaped()).isFalse();
+            soft.assertThat(caseInsensitiveRestriction.attribute()).isEqualTo("title");
+            soft.assertThat(caseInsensitiveRestriction.comparison()).isEqualTo(Operator.LIKE);
+            soft.assertThat(caseInsensitiveRestriction.constraint().pattern()).isEqualTo("%Java%");
+            soft.assertThat(caseInsensitiveRestriction.constraint().caseSensitive()).isFalse();
+//            soft.assertThat(textRestriction.isEscaped()).isFalse();
```
njr-11, line 76:
```suggestion
            soft.assertThat(((Like) caseInsensitiveRestriction.constraint()).escaped()).isFalse();
```

api/src/test/java/jakarta/data/metamodel/restrict/TextRestrictionRecordTest.java
```diff
@@ -31,72 +32,64 @@ class TextRestrictionRecordTest {
...
 
         SoftAssertions.assertSoftly(soft -> {
             soft.assertThat(restriction.attribute()).isEqualTo("title");
             soft.assertThat(restriction.comparison()).isEqualTo(Operator.LIKE);
-            soft.assertThat(restriction.value()).isEqualTo("%Java%");
-            soft.assertThat(restriction.isCaseSensitive()).isTrue();
-            soft.assertThat(restriction.isEscaped()).isTrue();
+            soft.assertThat(restriction.constraint().pattern()).isEqualTo("%Java%");
+            soft.assertThat(restriction.constraint().caseSensitive()).isTrue();
+//            soft.assertThat(restriction.isEscaped()).isTrue();
```
njr-11, line 92:
```suggestion
            soft.assertThat(((Like) restriction.constraint()).escaped()).isTrue();
```

api/src/test/java/jakarta/data/metamodel/restrict/TextRestrictionRecordTest.java
```diff
@@ -130,53 +123,53 @@ void shouldNegateNegatedRestriction() {
...
 
         SoftAssertions.assertSoftly(soft -> {
             soft.assertThat(restriction.attribute()).isEqualTo("author");
             soft.assertThat(restriction.comparison()).isEqualTo(Operator.NOT_EQUAL);
-            soft.assertThat(restriction.value()).isEqualTo("John Doe");
-            soft.assertThat(restriction.isCaseSensitive()).isTrue();
-            soft.assertThat(restriction.isEscaped()).isFalse();
+            soft.assertThat(restriction.constraint().pattern()).isEqualTo("John Doe");
+            soft.assertThat(restriction.constraint().caseSensitive()).isTrue();
+//            soft.assertThat(restriction.isEscaped()).isFalse();
```
njr-11, line 166:
```suggestion
            soft.assertThat(((Like) restriction.constraint()).escaped()).isFalse();
```

api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
```diff
@@ -30,69 +31,65 @@ static class Book {
...
+        BasicRestriction<Book> numChaptersGT10Basic = numChaptersLTE10Basic.negate();
 
         SoftAssertions.assertSoftly(soft -> {
             soft.assertThat(numChaptersLTE10Basic.comparison()).isEqualTo(Operator.LESS_THAN_EQUAL);
-            soft.assertThat(numChaptersLTE10Basic.value()).isEqualTo(10);
+            soft.assertThat(numChaptersLTE10Basic.constraint()).isEqualTo(Constraint.lessThanOrEqual(10));
 
             soft.assertThat(numChaptersGT10Basic.comparison()).isEqualTo(Operator.GREATER_THAN);
-            soft.assertThat(numChaptersGT10Basic.value()).isEqualTo(10);
+            soft.assertThat(numChaptersGT10Basic.constraint()).isEqualTo(Constraint.lessThanOrEqual(10));
```
njr-11, line 66: This is a bit odd, and it has to do with how we left negation in Restriction and don't have it as part of the Constraint.  After this PR goes in, I would like to consider moving negation to Constraint because that will open up more possibilities for using repository methods that have a `Constraint` parameter.

njr-11 (commented):
This review mostly concerns preserving a way for the implementation to know whether or not we have added escaping to a pattern. This information is needed to correctly interpret the pattern value.

api/src/main/java/jakarta/data/metamodel/constraint/Between.java
njr-11: > @njr-11 I'm OK with between, it looks strange to me to include GTE and LTE in the between definition.

Right, you have a good point there. This would be more straightforward as
```
    T lowerBound();
    T upperBound();
```
I was caught up on "Some data sources might have a native between operation" because I did not see how that mattered.

gavinking: I'm OK with that change. Originally the idea was that we would accommodating both open and closed intervals (an open interval being one with non-strict) bounds. But now that `Interval` was renamed to `Between`, and `GreaterThanOrEqual` was split from `GreaterThan`, this doesn't work out.

api/src/main/java/jakarta/data/metamodel/constraint/Like.java
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.metamodel.constraint;
+
+public interface Like extends Constraint<String> {
+    String pattern();
```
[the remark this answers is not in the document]
gavinking: `string()`?

api/src/main/java/jakarta/data/metamodel/constraint/Like.java
```diff
@@ -0,0 +1,48 @@
...
+    boolean caseSensitive();
```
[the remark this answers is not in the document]
gavinking: Yeah I don't understand what you mean by that, nor what the semantics of `escaped()` was supposed to be. It didn't seem to make any sense. We don't have such a thing in JDQL.

- If you were thinking of something like the syntax `LIKE pattern_value [ESCAPE escape_character]` in SQL or JPQL, then that would require an explicit `escapeCharacter()`. But we don't have that today.
- Or, if you were thinking it was a way to specify a literal pattern, that's covered here by the fact that we have both `Like` and `LikeLiteral`.

api/src/main/java/jakarta/data/metamodel/constraint/LikeLiteralRecord.java
gavinking: It definitely doesn't make sense to say that a literal string is escaped or not. What could that possibly mean?

njr-11: > It definitely doesn't make sense to say that a literal string is escaped or not. What could that possibly mean?

Agree. It makes no sense for a literal to be escaped. But `LikeLiteralRecord` is an implementation of a `Like` interface that in some cases will be an escaped thing (although never when it is a literal).  The point is that the implementor who receives a `Like` constraint needs to known whether or not the string is escaped.  Having said that, using a record component that is forced to always be false was a bad choice for how to get an `escaped()` method that always returns false for  `LikeLiteralRecord`.  I'll remove the suggestions for that and replace them with a method that just returns false.

api/src/main/java/jakarta/data/metamodel/constraint/Like.java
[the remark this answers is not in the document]
gavinking: Oh wait, I think I understood. I'm assuming that `\%` or `\_` are always interpretable as escaped wildcards, but in fact that's not what we've specified. And so your `escaped()` is saying "`\` is the escape character". OK, but that's very confusing and it's not consistent with JPQL.

So what we should do instead is add `escapeCharacter()` to be consistent with how JPQL and ANSI SQL work. (And we should add `ESCAPE` to JDQL.)

api/src/main/java/jakarta/data/metamodel/constraint/LikeLiteralRecord.java
njr-11, line 33:
```suggestion
    }

    @Override
    public boolean escaped() {
        return false;
    }
```

api/src/main/java/jakarta/data/metamodel/constraint/Like.java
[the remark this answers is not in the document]
gavinking: I added `boolean pattern()` to make it easy to distinguish true patterns for literal strings, and  `Character escape()` to allow for a selectable escape character.

api/src/main/java/jakarta/data/metamodel/constraint/Between.java
gavinking: Done.

api/src/main/java/jakarta/data/metamodel/restrict/Restrict.java
```diff
@@ -142,109 +122,48 @@ public static <T> Restriction<T> not(Restriction<T> restriction) {
...
-        return new TextRestrictionRecord<>(attribute, Operator.NOT_LIKE, ESCAPED, p);
+    public static <T> TextRestriction<T> notLike(String pattern, String attribute) {
+        return new TextRestrictionRecord<>(attribute, Like.pattern(pattern), true);
     }
 
-    public static <T> TextRestriction<T> notStartsWith(String prefix, String attribute) {
-        String pattern = toLikeEscaped(CHAR_WILDCARD, STRING_WILDCARD, false, prefix, true);
-        return new TextRestrictionRecord<>(attribute, Operator.NOT_LIKE, ESCAPED, pattern);
+    public static <T> TextRestriction<T> notLike(String pattern, char charWildcard, char stringWildcard, String attribute) {
+        return new TextRestrictionRecord<>(attribute, Like.pattern(pattern, charWildcard, stringWildcard), true);
```
njr-11, line 158: `notLike` should match `like` with respect to escape character
```suggestion
    public static <T> TextRestriction<T> notLike(String pattern, char charWildcard, char stringWildcard, char escape, String attribute) {
        return new TextRestrictionRecord<>(attribute, Like.pattern(pattern, charWildcard, stringWildcard, escape), true);
```

api/src/main/java/jakarta/data/metamodel/constraint/LikePatternRecord.java
```diff
@@ -0,0 +1,128 @@
...
+        this(pattern, true, escape);
+    }
+
+    public LikePatternRecord(String pattern, char charWildcard, char stringWildcard) {
+        this(translate(pattern, charWildcard, stringWildcard));
+    }
+
+    public LikePatternRecord(String pattern, char charWildcard, char stringWildcard, char escape) {
+        this(translate(pattern, charWildcard, stringWildcard), escape);
+    }
```
njr-11, line 49: The `translate` method will need to know which escape character is being used so that it can insert the correct one.
```suggestion
    public LikePatternRecord(String pattern, char charWildcard, char stringWildcard, char escape) {
        this(translate(pattern, charWildcard, stringWildcard, escape), escape);
    }
```

njr-11 (commented):
Proposed some fixes related to the added escape character.

api/src/test/java/jakarta/data/metamodel/restrict/TextRestrictionRecordTest.java
```diff
@@ -31,72 +32,64 @@ class TextRestrictionRecordTest {
...
     }
 
     @Test
     void shouldCreateTextRestrictionWithEscapedValue() {
         TextRestrictionRecord<String> restriction = new TextRestrictionRecord<>(
                 "title",
-                Operator.LIKE,
-                true,
-                "%Java%"
+                Like.pattern("%Java%")
```
njr-11, line 84: adding escape character that can be asserted
```suggestion
                Like.pattern("%Java%", '_', '%', '^')
```

njr-11 (commented):
Updated suggestions to the unit tests based on the outcome of using an escape character rather than a boolean.

gavinking:
@njr-11 I'm happy to keep adding all these little boolean fields to the hash code computation, but:

1. I'm sure it's not necessary, and 
2. I'm skeptical it's helpful.

Every call to `Objects.hash()` is an array instantiation to compute an `int` that's not cached. (OTOH, `String.hashCode()` is cached by the `String` instance.) Whatever benefit we get from the tiny increase in entropy that results from including the `boolean` in the hash code computation is probably wiped out by the cost of instantiating and cleaning up an array. No?

Well, I dunno, this sort of micro-optimization stuff is definitively not my jam, so perhaps what I'm saying is wrong.

api/src/main/java/jakarta/data/metamodel/constraint/LikePatternRecord.java
```diff
@@ -0,0 +1,128 @@
...
+        for (int i = 0; i<pattern.length(); i++) {
+            final char ch = pattern.charAt(i);
+            if (ch == charWildcard) {
+                result.append(CHAR_WILDCARD);
+            } else if (ch == stringWildcard) {
+                result.append(STRING_WILDCARD);
+            } else {
+                if (ch == STRING_WILDCARD || ch == CHAR_WILDCARD || ch == ESCAPE) {
+                    result.append(ESCAPE);
+                }
```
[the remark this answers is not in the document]
gavinking: yes of course, my bad, done.

api/src/main/java/jakarta/data/metamodel/constraint/LikePatternRecord.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Override
+    public Operator operator() {
+        return Operator.LIKE;
+    }
+
+    @Override
+    public String toString() {
+        return "LIKE '" + string + "'" + (caseSensitive ? "" : " IGNORE CASE");
```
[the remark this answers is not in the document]
gavinking: done.

api/src/main/java/jakarta/data/metamodel/restrict/Restrict.java
gavinking: actually it's an overload.

api/src/test/java/jakarta/data/metamodel/restrict/TextRestrictionRecordTest.java
gavinking: I added a new test

njr-11 (commented):
Up to this point, I have waited on reviewing `TextAttribute` and `Restrict` in this PR because many of the methods were reordered.  As I am now starting to look at it, I found that several methods have been lost -- the ones I noticed so far are methods to obtain TextRestrictions for GreaterThan/LessThan comparisons that are independent of case.  We shouldn't be removing those, but I also understand this PR has a lot in it and I don't want to make it any bigger or more complex, so I think I can accept omitting them for now to make progress, assuming they weren't removed on purpose and we agree we can bring them back afterward.  If we are going to do that, I would like to see the original method order restored here so that when doing the later work to restore them it will be immediately clear from a diff of this PR what we are trying to bring back.  After that, I would say we just approve this PR, merge it, and follow up with a short list of remaining actions later.

gavinking:
>  I have waited on reviewing `TextAttribute` and `Restrict` in this PR because many of the methods were reordered.

I reordered the methods of `TextAttribute` because they were previously in a completely chaotic order that made it incredibly difficult to see their symmetries:

- what is `contains()` doing up there getting in between `ascIgnoreCase()` and `descIgnoreCase()`?
- what on earth gave `endsWith()` the right to go to the head of the line?
- what did `startsWith()` do wrong that got him sent to the _back_ of the line?
- etc.

On the other hand, I don't think I made any major change to the ordering of things in `Restrict` and I think perhaps you're just misreading the diff.

> t, I would like to see the original method order restored here

-1. The original order is rubbish.

>  I found that several methods have been lost -- the ones I noticed so far are methods to obtain TextRestrictions for GreaterThan/LessThan comparisons that are independent of case.

Ah yes, apologies, my bad, I had absolutely intended to call this out but with so many moving parts and so much going on here, I simply forgot.

So the thing is that having case-insensitive `<` and `>` adds a surprising amount of complexity to this, and I'm not clear that it's worth it. Case-insensitive _sorting_ is certainly very common (and we have that), but I'm not clear that `lower(x)<lower(:value)` is something you really see very often.

Nor am I clear on the best way to do it. We could, for example:

1. restore `TextRestriction` in its full glory with an explosion of subtypes, or
2. introduce a way to apply functions to attribute references, something like `_Book.title.lowercase()`. 

Between these two possibilities, I strongly suspect that 2 is a much, much better way to go, but such a change _definitely_ doesn't belong in _this_ PR.

So, yes, it was intentional that I removed these operations for now, with the intention of discussing the best way to recover them, assuming ya'll decide that they're definitely worth having. That is to say, we can certainly add these operations back, but I would like to see it done as a separate unit, so that people can properly appreciate their complexity cost and weigh the options.

gavinking:
> I don't think I made any major change to the ordering of things in Restrict and I think perhaps you're just misreading the diff.

Oh, looks like I moved `startsWith` and `notStartsWith` because they broke the symmetry by not following the pattern established by all other `x`/`notX` pairs.

njr-11:
> That is to say, we can certainly add these operations back, but I would like to see it done as a separate unit, so that people can properly appreciate their complexity cost and weigh the options.

I'm in full agreement with addressing that in a separate PR.  I was just hoping to get confirmation that we are going to pursue bringing the lost methods back.  I have a couple of ideas on how we might be able to do it that should be minimally invasive, and I'll plan to look into that after this is merged and I can try things out more easily.  I'll add an approval to this PR to move ahead.
@otaviojava are you okay with this approach?  I haven't seen any review comments from you.

With regard to moving methods around, I'm not saying don't do it, but for the future I would recommend isolating that sort of thing to a separate PR to avoid obscuring other changes to the same code.  I almost missed it this time.

njr-11 (approved):

njr-11:
All review comments have been addressed or agreed to be deferred to follow-on work, so I'll merge this and begin looking into some of the follow-on work.

PR #976, opened 2025-02-14
ComparableAttribute and BasicAttribute are preferred

njr-11:
This PR contains updates to make `ComparableAttribute` and `BasicAttribute` be preferred for the static metamodel over `SortableAttribute` and `Attribute`.

`Attribute` is now documented as the common supertype that all metamodel attribute interfaces inherit from.
`SortableAttribute` is discouraged for most usage.
Spec and JavaDoc examples are updated to show ComparableAttribute and BasicAttribute instead.
Elsewhere in spec and JavaDoc, we now list the provided attribute subtypes, along with what they are for, further promoting the use of TextAttribute/ComparableAttribute/BasicAttribute.

As a second commit in this PR, I moved the new internal implementation classes out of the public package where they would be visible as part of the API (it was a mistake that we ever allowed others there, but it was too late in the release cycle to correct it) and put them as package protected under the interface so that they will not be externalized to the user.  Instead, the proper way to obtain instances is with the *Attribute.of(name) method. See examples that were also updated. This is cleaner and avoids externalizing implementation, which not only avoids cluttering the API, but also preserves our flexibility in the future to make changes to those implementations that would otherwise be considered a breaking change.  I would like to eventually deprecate the .impl package altogether, but I did not do that in this PR.


api/src/main/java/jakarta/data/metamodel/SortableAttribute.java
```diff
@@ -18,18 +18,17 @@
...
- * <li>enum attributes</li>
- * <li>time attributes</li>
- * <li>boolean attributes</li>
- * <li>{@linkplain TextAttribute textual attributes}</li>
- * </ul>
+ * <p>Wherever possible, it is preferable to use {@link ComparableAttribute}
+ * instead because it provides more function. Use {@code SortableAttribute}
+ * only where you have an entity attribute upon which query results can be sorted,
+ * but cannot otherwise be compared within query restrictions. For some databases,
+ * entity attributes of type {@code byte[]} fall under this category.</p>
```
gavinking, line 31:
```suggestion
 *<p>A {@code SortableAttribute} may be used to sort query results.
 * When an attribute type is a subtype of {@link java.lang.Comparable>, 
 * use of {@link ComparableAttribute} is usually preferred, since a
 * {@code SortableAttribute} may not not used in order-based query 
 * restrictions. Direct use of {@code SortableAttribute} is appropriate 
 * for attributes of type {@code byte[]}.</p>
```

api/src/main/java/jakarta/data/metamodel/SortableAttribute.java
```diff
@@ -49,4 +48,18 @@ public interface SortableAttribute<T> extends Attribute<T> {
...
+     * <p>Creates a static metamodel {@code SortableAttribute} representing the
+     * entity attribute with the specified name.</p>
+     *
+     * @param <T> entity class of the static metamodel.
+     * @param name the name of the entity attribute.
+     * @return instance of {@code SortableAttribute}.
+     */
+    static <T> SortableAttribute<T> of(String name) {
+        if (name == null)
+            throw new IllegalArgumentException("entity attribute name is required");
```
gavinking, line 61: Ive started using `Objects.requireNonNull()` for this.

```suggestion
        Objects.requireNonNull(name, "entity attribute name is required");
```

api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
```diff
@@ -54,4 +61,29 @@ default Restriction<T> notNull() {
...
+     * entity attribute with the specified name.</p>
+     *
+     * @param <T> entity class of the static metamodel.
+     * @param <V> type of entity attribute (or wrapper type if primitive).
+     * @param name the name of the entity attribute.
+     * @return instance of {@code BasicAttribute}.
+     */
+    static <T,V> BasicAttribute<T,V> of(String name) {
+        if (name == null)
+            throw new IllegalArgumentException("entity attribute name is required");
```
gavinking, line 75:
```suggestion
        Objects.requireNonNull(name, "entity attribute name is required");
```

api/src/main/java/jakarta/data/metamodel/ComparableAttribute.java
```diff
@@ -43,4 +75,39 @@ default Restriction<T> lessThanEqual(V value) {
...
+     * entity attribute with the specified name.</p>
+     *
+     * @param <T> entity class of the static metamodel.
+     * @param <V> type of entity attribute (or wrapper type if primitive).
+     * @param name the name of the entity attribute.
+     * @return instance of {@code ComparableAttribute}.
+     */
+    static <T, V extends Comparable<V>> ComparableAttribute<T, V> of(String name) {
+        if (name == null)
+            throw new IllegalArgumentException("entity attribute name is required");
```
gavinking, line 89:
```suggestion
        Objects.requireNonNull(name, "entity attribute name is required");
```

api/src/main/java/jakarta/data/metamodel/TextAttribute.java
```diff
@@ -117,6 +118,21 @@ default TextRestriction<T> notStartsWith(String prefix) {
...
+     * <p>Creates a static metamodel {@code TextAttribute} representing the
+     * entity attribute with the specified name.</p>
+     *
+     * @param <T> entity class of the static metamodel.
+     * @param name the name of the entity attribute.
+     * @return instance of {@code TextAttribute}.
+     */
+    static <T> TextAttribute<T> of(String name) {
+        if (name == null)
+            throw new IllegalArgumentException("entity attribute name is required");
```
gavinking, line 131:
```suggestion
        Objects.requireNonNull(name, "entity attribute name is required");
```

gavinking (commented):
Some suggestions.

api/src/main/java/jakarta/data/metamodel/impl/ComparableAttributeRecord.java
```diff
@@ -1,34 +0,0 @@
...
-    @Override
-    public Sort<T> asc() {
-        return Sort.asc(name);
-    }
-
-    @Override
-    public Sort<T> desc() {
-        return Sort.desc(name);
-    }
-}
```
gavinking, line 34: So, given how trivial these are (they're not a burden in terms of maintenance), I would not delete these unless we're also deprecating the existing implementations in Jakarta Data 1.1. I think it's completely fine to have that conversation, but it seems out of scope for this pull request.

api/src/main/java/jakarta/data/metamodel/SortableAttribute.java
njr-11: This has one minor type, which I'll correct later, but otherwise looks good

api/src/main/java/jakarta/data/metamodel/impl/ComparableAttributeRecord.java
njr-11: Sure, I'll add them back in and we can split that discussion out separately.

njr-11: Commit https://github.com/jakartaee/data/pull/976/commits/0427cdd04f753336683a8e5e7de40984a87aa65c is ready for review. It adds the implementation classes back in, fixes two typos, adds the import for usage of java.util.Ojbects, and adds a clarification to account for primitives like `int` corresponding to ComparableAttribute on the basis of the wrapper class, even though the primitive itself isn't Comparable

api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -18,7 +18,21 @@
...
+ * <p>The following subtypes are provided:</p>
+ * <ul>
+ * <li>{@link TextAttribute} for entity attributes that represent text,
+ *     typically of type {@link String}.</li>
+ * <li>{@link ComparableAttribute} for entity attributes that represent other
+ *     sortable and comparable values, such as {@code int}, {@link Long},
+ *     {@code boolean}, {@link java.time.LocalDateTime}, and enumerations.</li>
+ * <li>{@link SortableAttribute} for entity types that are sortable, but not
+ *     comparable. Generally this subtype is unused, but in some databases
+ *     might apply to {@code byte[]} attributes. </li>
```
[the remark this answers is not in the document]
njr-11: I didn't write this very well.  "Some databases" was intended to refer to that some databases might allow sorting on byte array, not to imply that there is ever a case where byte[] is Comparable.  I improved the wording in commit https://github.com/jakartaee/data/pull/976/commits/7073877333d0987180969ae170d382f2af42aaa0

njr-11:
Merge conflicts have been resolved so this PR is ready for its review to be completed.  The prior review comments should be resolved now as well.

api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -18,7 +18,21 @@
...
+ * <p>The following subtypes are provided:</p>
+ * <ul>
+ * <li>{@link TextAttribute} for entity attributes that represent text,
+ *     typically of type {@link String}.</li>
+ * <li>{@link ComparableAttribute} for entity attributes that represent other
+ *     sortable and comparable values, such as {@code int}, {@link Long},
+ *     {@code boolean}, {@link java.time.LocalDateTime}, and enumerations.</li>
+ * <li>{@link SortableAttribute} for entity types that are sortable, but not
+ *     comparable. Generally this subtype is unused but is applicable for
+ *     databases that allow sorting on {@code byte[]} attributes. </li>
```
gavinking, line 32:
```suggestion
 *     comparable, for example, the type {@code byte[]} when the
 *     database allows sorting by binary fields. </li>
```

njr-11:
@gavinking this PR is now updated to add the entity class to the `.of` method as discussed during the Jakarta Data meeting, which is done in commit https://github.com/jakartaee/data/pull/976/commits/2585d22849049cffd2eacc0c162700dabf41c1ca
Please review again.

gavinking (approved):
LGTM!

PR #985, opened 2025-02-24
Find value should be annotation member, not a constant

njr-11:
A typo caused the value member of the Find annotation to be a constant rather than a member of the annotation. This PR corrects it.


gavinking (approved):
Oops.

otaviojava (approved):

Issue #19, opened 2022-09-19
Asynchronous repository methods and integration with Jakarta Concurrency


njr-11:
**Description**

As a:

- [ ] Application user/user of the configuration itself
- [x] API user (application developer)
- [ ] SPI user (container or runtime developer)
- [ ] Specification implementer

...I need to be able to:

write repository methods that run asynchronously

...which enables me to:

avoid tying up the requesting thread waiting for queries to finish so that I can complete other work in the mean time.

Note: This issue does not cover reactive programming models like Flow. That is a separate discussion.

The preference would be to rely on the data access provider's own async support, but Jakarta Persistence/JPA (which are JDBC based) doesn't have async support.  Lacking that, Jakarta Concurrency [@Asynchronous](https://jakarta.ee/specifications/concurrency/3.0/apidocs/jakarta/enterprise/concurrent/asynchronous) could provide what we need in this area and fit quite well with just some minor language added to the Jakarta Data spec to make it possible.

If we specify that repository methods can have return types of [CompletionStage](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/CompletionStage.html) and [CompletableFuture](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/CompletableFuture.html) (as some implementations already do), then based on the data access provider's level of support for async:
- If the data access provider has its own async support, use that and the stage/future just gets completed with the result upon completion.
- Otherwise, Jakarta Data could state that if a repository interface method is annotated with Concurrency `@Asynchronous` then that same `@Asynchronous` annotation is also applied to the bean instance.  Doing that will automatically trigger the Concurrency 3.0 interceptor.  Then all the implementation needs to do is return an already-completed future, relying on the Concurrency `@Asynchronous` interceptor to cover making it async.

```
@Asynchronous
CompletionStage<List<Customer>> findByNumOrdersGreaterThan(int threshold);

...
findByNumOrdersGreaterThan(2).thenAccept(this::sendLastestSalesPromotion);
```

Other benefits of this will be that if the data access provider, such as based on a Jakarta Persistence/JPA implementation, requires the async thread to have services of the Jakarta EE platform available, such as performing a java:comp/env lookup of a data source, or to be able to run under a transaction, or to have the same security credentials on the thread for performing the database operation, the Jakarta Concurrency spec will have taken care of all of that for us.

keilw:
[...]

Issue #17, opened 2022-09-19
Be clearer about "reactive" repositories


keilw:
**Description**

The current [ReactiveRepository](https://github.com/jakartaee/data/blob/main/api/src/main/java/jakarta/data/repository/ReactiveRepository.java) is meaningless and confusing, because it creates a fake impression of being a "Generic Reactive" repository type which does not work.

[RxJavaReactiveStreams](https://github.com/ReactiveX/RxJavaReactiveStreams) highlights, the two different incompatible concepts of these two (RxJava and Reactive Streams), add the JDK Async [Flow](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.html) and you have a possible combination between 3 incompatible types. 

[java.util.concurrent](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/package-summary.html) also says nothing about "reactive", hence calling it along the lines of Micronaut [AsyncCrudRepository](https://github.com/micronaut-projects/micronaut-data/blob/master/data-model/src/main/java/io/micronaut/data/repository/async/AsyncCrudRepository.java
) seems more sensitive. 
@graemerocher said in https://github.com/jakartaee/data/issues/8#issuecomment-1247785287, something like `FlowRepository` would be OK. If Micronaut's `AsyncCrudRepository` (working with [CompletableFuture](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/CompletableFuture.html)) was also worth supporting, then both
`AsyncCrudRepository` or say `FutureCrudRepository` might exist side-by-side with a `FlowRepository` or `FlowCrudRepository` (if others like pageable variants also apply) while the generic "reactive" term sounds meaningless like a source for buzzword bingo where no common denominator could be found between all the mentioned **concurrent/async** APIs the JDK provides as `Flow` and `Future` are not interpoerable or build on top of each other either.

otaviojava:
Great point @keilw when we talk about this topic, we can work with ```FlowRepository```

hantsy:
Like Spring Data JPA, Java 8 @Async CompletableFuture/CompletionStage can used in generic blocking API.

Aligned with Java/Jakarta, I would like used Java 9 Flow by default for the ReactiveRepository. SmallRye Mutiny is switched to Java 9 Flow by default, which can be used as backend implementations.

hantsy:
In Spring framework, it used Reactor by default, and used a RactiveAdaterRegistry to register other Reactive Streams lib, and convert between them. 

[...]

PR #987, opened 2025-02-24
initial proposal for reactive/async repositories

gavinking:
See #19 and #17.

Looking for feedback here!

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -833,7 +833,7 @@ where:
 - `R m(P1 p1, P2 p2, ...)` is a legal repository method signature according to the previous sections of this chapter, or, in the special case that `R` is the type `java.lang.Void`, `void m(P1 p1, P2 p2, ...)` is a legal repository method signature according to the previous sections of this chapter, and
 - `F` is a parameterized type representing a value which might not yet have been computed and which is supported by the Jakarta Data provider.
 
-NOTE: Every Jakarta Data provider is encouraged, but not required, to support asynchronous repository methods returning `java.util.CompletionStage`.
+NOTE: Every Jakarta Data provider is encouraged, but not required, to support asynchronous repository methods returning `java.util.concurrent.CompletionStage`.
 
 For example, the following is an asynchronous parameter-based query method:
 
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -833,7 +833,7 @@ where:
 - `R m(P1 p1, P2 p2, ...)` is a legal repository method signature according to the previous sections of this chapter, or, in the special case that `R` is the type `Void`, `void m(P1 p1, P2 p2, ...)` is a legal repository method signature according to the previous sections of this chapter, and
 - `F<T>` is a parameterized type representing a value which might not yet have been computed and which is supported by the Jakarta Data provider.
 
-NOTE: Every Jakarta Data is encouraged, but not required, to support asynchronous repository methods returning `java.util.CompletionStage`.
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -818,3 +818,42 @@ The following order, with the lower number having higher precedence, is used whe
...
+
+An asynchronous repository method has a signature of form:
+
+[source,java]
+F<R> m(P1 p1, P2 p2, ...)
+
+where:
+
+- `R m(P1 p1, P2 p2, ...)` is a legal repository method signature according to the previous sections of this chapter, or, in the special case that `R` is the type `Void`, `void m(P1 p1, P2 p2, ...)` is a legal repository method signature according to the previous sections of this chapter, and
+- `F<T>` is a parameterized type representing a value which might not yet have been computed and which is supported by the Jakarta Data provider.
```
njr-11, line 834: I assume you meant `R` here because there is no `T` in the above.
```suggestion
- `F<R>` is a parameterized type representing a value which might not yet have been computed and which is supported by the Jakarta Data provider.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -818,3 +818,42 @@ The following order, with the lower number having higher precedence, is used whe
...
+
+NOTE: Every Jakarta Data is encouraged, but not required, to support asynchronous repository methods returning `java.util.CompletionStage`.
```
njr-11, line 836: Minor typo:
```suggestion
NOTE: Every Jakarta Data provider is encouraged, but not required, to support asynchronous repository methods returning `java.util.CompletionStage`.
```

njr-11 (commented):
Nicely written up and good initial set of changes.

spec/src/main/asciidoc/repository.asciidoc
gavinking, line 834:
```suggestion
- `F` is a parameterized type representing a value which might not yet have been computed and which is supported by the Jakarta Data provider.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -818,3 +818,42 @@ The following order, with the lower number having higher precedence, is used whe
 
 A repository method that does not fit any of the above patterns and is not handled as a vendor-specific extension to the specification must either result in an error at build time or raise `UnsupportedOperationException` at runtime.
 
+=== Asynchronous repositories
+
+An _asynchronous repository method_ is a repository method which returns an object representing a value which will be eventually obtained from the database, but which might not yet be available.
```
gavinking, line 823:
```suggestion
An _asynchronous repository method_ is a repository method which returns an object representing a value which will eventually be obtained from the database, but which might not yet be available.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -818,3 +818,42 @@ The following order, with the lower number having higher precedence, is used whe
...
+An asynchronous repository method is permitted to return such an object immediately, and access the database asynchronously.
+
+An asynchronous repository method has a signature of form:
+
+[source,java]
+F<R> m(P1 p1, P2 p2, ...)
+
+where:
+
+- `R m(P1 p1, P2 p2, ...)` is a legal repository method signature according to the previous sections of this chapter, or, in the special case that `R` is the type `Void`, `void m(P1 p1, P2 p2, ...)` is a legal repository method signature according to the previous sections of this chapter, and
```
gavinking, line 833:
```suggestion
- `R m(P1 p1, P2 p2, ...)` is a legal repository method signature according to the previous sections of this chapter, or, in the special case that `R` is the type `java.lang.Void`, `void m(P1 p1, P2 p2, ...)` is a legal repository method signature according to the previous sections of this chapter, and
```

PR #988, opened 2025-02-24
add TextAttribute.ignoreCase

gavinking:
This is just an idea. WDYT?


gavinking:
Note that an alternative, but very similar-in-spirit approach would be to add `uppercased()` and `lowercased()` to `TextAttribute`. I'm not sure which option I prefer.

gavinking:
Yeah, so I should make it clear that this was just a hand-wave at what I would actually do. The full impl of this would involve giving `BasicRestrictionRecord` and `TextRestrictionRecord` a reference to the actual `BasicAttribute` or `TextAttribute`, instead of just to its `name`.

And at the same time, I would probably so something like get rid of a bunch of the operations on `Restrict`.

And that would mean that users who don't want to use the _static_ metamodel could write something like:

```java
TextRestriction<Book> titleEq = 
        TextAttribute.for(Book.class, "title").ignoreCase().startsWith("Jakarta Data");
```
I don't see how you're thinking of making your approach work, since:

```java
TextRestriction<Book> titleEq = 
        Restrict.equalTo("title", "Jakarta Data").ignoreCase(); // wrong, ERROR
```

doesn't pass the typechecker. (There's no way to correctly infer the type parameter of `TextRestriction`.)

Well, actually I just realized we already have that bug with `negate()`. I guess it was never tested with the non-static-metamodel usage pattern.

njr-11:
> I don't see how you're thinking of making your approach work, since:
> 
> ```java
> TextRestriction<Book> titleEq = 
>         Restrict.equalTo("title", "Jakarta Data").ignoreCase(); // wrong, ERROR
> ```
> 
> doesn't pass the typechecker. (There's no way to correctly infer the type parameter of `TextRestriction`.)

There is a way to do it, albeit somewhat ugly,

```
        TextRestriction<Book> titleEq = 
                Restrict.<Book>equalTo("title", "Jakarta Data").ignoreCase();
```

I'm not saying `Restrict` is the way I would prefer doing things. I was mainly aiming not to break it or do anything too invasive to it because the aim of my PR is different - to restore lost methods of TextAttribute, which I'm hoping we can do sooner rather than later before things drift further and we lose track of them.  We should have this other discussion as well, and maybe we should even consider getting rid of the Restrict.* way of obtaining BasicRestriction/TextRestriction altogether now that we have Constraint.  But hopefully we can discuss these things after the other PR.

gavinking:
OK, Nathan, I realized that what I pushed was too bare-bones, because all I was looking for was to start a discussion and get you guys thinking. I had no idea you had actually started implementing stuff.

And you couldn't look at what I pushed and see that whole picture.

So I've *very quickly*, and a bit messily, implemented the whole idea.

njr-11:
> OK, Nathan, I realized that what I pushed was too bare-bones, because all I was looking for was to start a discussion and get you guys thinking. I had no idea you had actually started implementing stuff.
> 
> And you couldn't look at what I pushed and see that whole picture.
> 
> So I've _very quickly_, and a bit messily, implemented the whole idea.

I would have rather this remained a discussion because now we have a large merge conflict between the two PRs that will be a mess for one of us to resolve.  I tried to point out when merging #975 that I would immediately begin working on bringing back the lost methods and some other deferred issues that I spotted, but didn't make that clear enough.

njr-11:
I reviewed the changes in this PR, and I see the following covered:

- New `.of` methods added to the *Attribute classes.
- Switching the Restrict.* methods to accept a *Attribute rather than a String attribute name.
- Moving `ignoreCase()` to `TextAttribute`.

### New `.of` methods added to the *Attribute classes
I like this idea. We have another PR #976 that already adds it, but without the class parameter, having inferred it from the assignment that is typical of the static metamodel. I can see why you might want to have a signature with the class parameter if manually constructing instances inline for single use to invoke a method.
```
books.search( TextAttribute.for(Book.class, "title").startsWith("Jakarta Data") );
```
But I would also say that even here it would be a better practice to isolate usage of hard coded strings to constants, where it can infer the class,
```
static final TextAttribute<Book> BOOK_TITLE = TextAttribute.of("title");
...
books.search( BOOK_TITLE.startsWith("Jakarta Data") );
```
I think the `of` method without the Class is good enough and we don't really need a signature that also has Class, but I'm fine with including it if you think it is important.

### Switching the Restrict.* methods to accept a *Attribute rather than a String attribute name
In order to use these methods, the user would need to first obtain a *Attribute instance,
```
books.search( Restrict.startsWith("Jakarta Data", TextAttribute.for(Book.class, "title")) );
```
But if a user has first obtained the *Attribute instance, it would be easier to avoid `Restrict` altogether and do,
```
books.search( TextAttribute.for(Book.class, "title").startsWith("Jakarta Data") );
```
Switching the parameter to *Attribute doesn't appear to be useful. I can understand that you don't like the signatures with `String`, and I would be in favor of just removing these Restrict.* methods that create Basic/TextRestriction altogether. Between the ability to obtain Restrictions from  *Attribute and the ability to use `Constraint`, I don't think we need this variant of the pattern anymore.  I would be willing to remove them in my PR.  Let me know if you would be in favor of that resolution.

### Moving `.ignoreCase()` to `TextAttribute`
I definitely agree with moving it away from `Like` and getting rid of the `LikeLiteralRecord`.  The question is whether to move it to `TextRestriction` as done under #990 or move it to `TextAttribute` as done here.  In considering this, we need to also consider `.negate()`, which is currently on `Restriction`/`BasicRestriction`/`TextRestriction`.  It makes sense for a customer to be able to obtain a negated Restriction and a Restriction that ignores case.  On the other hand, I don't think that a negated Attribute makes sense (I know you didn't propose this). An Attribute that ignores case seems a little less awkward, but still somewhat awkward. They are most meaningful on Restriction rather than Attribute.

gavinking:
> I think the of method without the `Class` is good enough and we don't really need a signature that also has Class, but I'm fine with including it if you think it is important.

I *do* think it's important yes, because otherwise use of these APIs without the static metamodel is just a minefield of weird type inference errors. Yes, sure, I agree there's a disciplined way of doing it using `TextAttribute<Book> BOOK_TITLE` as you suggest, but:

- that's just a handwritten static metamodel, and 
- there's no good way for the user to easily discover that it's the intended way to use the API.

Now, if we were telling everyone to always use the static metamodel there would be no problem here. But I think we decided that wasn't what we were going to do.


> I would be in favor of just removing these Restrict.* methods that create Basic/TextRestriction altogether.

Agree. I would make them internal, or just inline them into the `XxxxxAttribute` interfaces.

gavinking:
>  An Attribute that ignores case seems a little less awkward, but still somewhat awkward.

Actually I think you're right, hold on, let me try something....

gavinking:
OK, what I just pushed illustrates the approach much better.

But to summarize, the difference between the two approaches is:

Here, in this PR:

- a `Like` pattern defines its own case-sensitivity, which I believe is useful and very natural
- the only kind of `TextRestriction` is one with a `Like` constraint
- but you can use `TextAttribute.lowercased()`  or `TextAttribute.uppercased()` to do case-insensitive `equalTo()`, `between()`, etc, in a perfectly-natural way

In #990, on the other hand:

- a `Like` pattern does not define its own case-sensitivity
- a `TextRestriction` can now have any kind of constraint, and may decorate the constraint with case-insensitivitiy via `ignoreCase()`

Either are legit approaches. But FWIW, if I'm not mistaken, this approach is simpler with a smaller impact.

njr-11 (commented):
Also, if I am understanding this correctly, I can obtain a `Constraint` from an `uppercased()` attribute as follows:
```
Like constraint = _Book.title.uppercased().startsWith("Jakarta").constraint();
```
and if I supply that constraint to the following method which accepts a `Constraint` on `title`, I will get Books that start with "Jakarta" and not "JAKARTA",
```
books.search(_Book.title.uppercased().startsWith("Jakarta").constraint());
```
which is because information on casing is stored in 2 different places.

It's because casing is its own property of TextAttribute and does not impact the independent Constraint piece of the TextAttribute, and then only the Constraint part is obtained and used.  It makes sense when you know the implementation and logic behind it.  But it won't be intuitive to users, who will be tripped up by it.

gavinking:
Right, you can have a case-insensitive pattern in `Like`, and you can have an uppercased/lowercased attribute value.

But in this design there's no such thing as a case-insensitive equality or between constraint. (Nor is there in JDQL, JPQL, SQL, etc.)

njr-11:
> But in this design there's no such thing as a case-insensitive equality or between constraint. (Nor is there in JDQL, JPQL, SQL, etc.)

It's turning out that this approach and the one from #990 are more similar than I realized, with the main difference being where it most makes sense to put the case information and how to present that to the user.  A lot of the other aspects are common.  Either way, `Restrict.*` methods that have a problem with typing need to be omitted from the API.  Either way we want to be able to construct the Attribute subtypes with an `.of` method, including one that accepts a `Class` parameter to enforce the type. Either way, case-insensitive EqualTo, NotEqualTo, GTE, LTE, GT, LT, Between Constraints are not possible from the approach itself, with another mechanism needed to provide that instead, like what I did in #990.  It should also be noted that #990 fixes a variety of other issues that came in with Constraint (such as lack of negation for constraints, unnecessary Operator, missed type parameters, and so forth). The changes to case representation are maybe even the smaller part of it, so my recommendation is to proceed with reviewing #990, agree on something to merge independent of the casing aspect, with the understanding that immediately afterward, we will finish the discussion of where to best place the case information within the API: on Restriction or Attribute.  With all of the other common pieces out of the way, I think it will be easier for other specification participants besides just you and me to have a clear understanding of the options for the API for casing.

njr-11 (commented):
A couple of changes here are good ideas that I think we should include regardless of the decision of whether to put ignoreCase() on Restriction or uppercased()/lowercased() on TextAttribute.

- Adding the `.of` methods. This overlapped another PR that also did the same where it was already merged.
- Switching from `String attribute;` to `BasicAttribute<T,V> attribute();` on `BasicRestriction` and likewise for `TestRestriction`.

I think we should wait on the discussion of Restriction.ignoreCase() vs TextAttribute.upper/lowercased() until after we have chosen what we will do for case-insensitive constraints because that might end up making a solution to the former unnecessary, depending on what option is chosen for the latter.

otaviojava:
I guess we can close this one, right?

gavinking:
Yes, `TextExpression` now defines `lower()` and `upper()`.

PR #989, opened 2025-02-24
minimal possible change to allow multiple-element select lists

gavinking:
Starting point for discussion. This is probably not quite right as written.

### DIFF
--- a/spec/src/antlr/JDQL.g4
+++ b/spec/src/antlr/JDQL.g4
@@ -20,7 +20,7 @@ select_item
     | aggregate_expression
     ;
 select_items
-    : state_field_path_expression state_field_path_expression+
+    : state_field_path_expression {, state_field_path_expression}+
     ;
 
 orderby_clause : 'ORDER' 'BY' orderby_item (',' orderby_item)*;
--- a/spec/src/antlr/JDQL.g4
+++ b/spec/src/antlr/JDQL.g4
@@ -13,11 +13,14 @@ where_clause : 'WHERE' conditional_expression;
 set_clause : 'SET' update_item (',' update_item)*;
 update_item : state_field_path_expression '=' (scalar_expression | 'NULL');
 
-select_clause : 'SELECT' select_item;
+select_clause : 'SELECT' (select_item | select_items);
 select_item
     : state_field_path_expression
     | id_expression
     | aggregate_expression
+    ;
+select_items
+    : state_field_path_expression state_field_path_expression+
     ;
 
 orderby_clause : 'ORDER' 'BY' orderby_item (',' orderby_item)*;
--- a/spec/src/main/asciidoc/query-language.asciidoc
+++ b/spec/src/main/asciidoc/query-language.asciidoc
@@ -413,7 +413,10 @@ The `where` clause is always optional. When it is missing, there is no restricti
 
 ==== Select clause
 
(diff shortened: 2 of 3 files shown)


spec/src/antlr/JDQL.g4
```diff
@@ -13,12 +13,15 @@ where_clause : 'WHERE' conditional_expression;
...
 
-select_clause : 'SELECT' select_item;
+select_clause : 'SELECT' (select_item | select_items);
 select_item
     : state_field_path_expression
     | id_expression
     | aggregate_expression
     ;
+select_items
+    : state_field_path_expression state_field_path_expression+
```
njr-11, line 23: Need to include a comma to preserve consistency with JDQL
```suggestion
    : state_field_path_expression {, state_field_path_expression}+
```

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -413,7 +413,10 @@ The `where` clause is always optional. When it is missing, there is no restricti
 
 ==== Select clause
 
-The `select` clause, with syntax given by `select_clause`, specifies a path expression which determines the values returned by the query. The path expression is evaluated for each record which satisfies the restriction imposed by the `where` clause, as specified in <<Path expressions>>, and the value of the last element of the path expression is added to the query results.
+The `select` clause, with syntax given by `select_clause`, specifies one or more path expressions which determine the values returned by the query. Each path expression is evaluated for each record which satisfies the restriction imposed by the `where` clause, as specified in <<Path expressions>>, and a tuple containing the resulting values is added to the query results.
+
+NOTE: When a `select` clause contains more than one path expression, the `@Select` annotation is used to map the elements of the tuple to constructor parameters of the query return type.
+When the `select` clause contains only one path expression, the `@Select` annotation is not used, and the query directly returns the values of the path expression.
```
njr-11, line 419: I'm trying to decide if we should allow this usage of `@Select` on record components in combination with `@Query`.  The scenario that I see this directed at is:

```java
public record ModelInfo (
    @Select("make") String manufacturer,
    String model,
    @Select("modelYear") int year) {}

@Query("SELECT model, modelYear, make FROM Car WHERE vin = ?1")
Optional<ModelInfo> getModelInfo(String vin);
```

The above allows the JDQL query to select the entity attributes in any order, with the `@Select` annotation (and matching record component name?) correcting it.

The alternative to this would be to require the JDQL query to use the same order as the record components,

```java
public record ModelInfo (
    String manufacturer,
    String model,
    int year) {}

@Query("SELECT make, model, modelYear FROM Car WHERE vin = ?1")
Optional<ModelInfo> getModelInfo(String vin);
```

In that case, we would state that `Select` either must not be used (or is ignored?) on records that are used in the results of a `@Query` method.

njr-11 (commented):
I'm sorry I didn't notice this PR when it was opened.  I added a few comments because I need to give more thought to what is best in one case. We can discuss it on today's Jakarta Data call.

spec/src/antlr/JDQL.g4
gavinking: Indeed.

gavinking:
@njr-11 updated with the changes we talked about on today's call. WDYT?

njr-11 (approved):
Looks good. Approving.

PR #1000, opened 2025-03-03
Javadoc should indicate how boolean is sorted

njr-11:
While writing a test, I checked the Javadoc table of basic types to see whether `true` or `false` is ordered first.  Although this table has information on how some other types are sorted, it does not mention anything for boolean.  I did find the answer elsewhere in the specification document.  This PR adds the information to the table in the JavaDoc.

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -237,7 +237,7 @@ import java.util.Set;
  * <br>{@code int} and {@link Integer}
  * <br>{@code long} and {@link Long}
  * <br>{@code short} and {@link Short}</td>
- * <td></td></tr>
+ * <td>Boolean values are sorted such that {@code false} &lt; {@code true}.</td></tr>
  *
  * <tr style="vertical-align: top"><td>Binary data</td>
  * <td>{@code byte[]}</td>


api/src/main/java/module-info.java
```diff
@@ -237,7 +237,7 @@
  * <br>{@code int} and {@link Integer}
  * <br>{@code long} and {@link Long}
  * <br>{@code short} and {@link Short}</td>
- * <td></td></tr>
+ * <td>Boolean values are sorted such that {@code false} &lt; {@code true}.</td></tr>
```
otaviojava, line 240: Do we have this control?

E.g: I do believe that SQLite, DynamoDB and CouchDB & JSON-based Databases.

But I did go deep on this topic.

gavinking (approved):

api/src/main/java/module-info.java
njr-11: @otaviojava I am not adding any new requirement here.  This same requirement already exists in the specification, section [5.4.6 Ordering](https://jakarta.ee/specifications/data/1.0/jakarta-data-1.0#_ordering) where it states,
> Boolean values must be ordered so that `false < true` is satisfied.

This PR only copies the existing requirement to the Javadoc so that users can more easily find the information.

otaviojava: Ok, but at NoSQL, there is no this guarantee. Unless it queries by memory, it is not the most efficient way to do it.

Anyway, once, those are exceptions. I guess it it fine.

PR #1002, opened 2025-03-04
TCK tests of Select annotation on record component to identify the entity attribute

njr-11:
TCK tests that of repository methods that return a subset of entity attributes as a Java record, where the record component uses the Select annotation to identify a corresponding entity attribute.  Some of the tests also include specifying the Find annotation value to identify the entity.  Others assume the primary entity type.


mswatosh (approved):

njr-11:
resolved merge conflicts

PR #1003, opened 2025-03-05
interoperability with the Jakarta Concurrency specification

njr-11:
Clarifies how the support for asynchronous repository methods works in relation to the Jakarta Concurrency specification's `@Asynchronous`.


api/src/main/java/module-info.java
```diff
@@ -1109,6 +1111,39 @@
  * {@code @Transactional} is called in an environment where both Jakarta Transactions and
  * Jakarta CDI are available, the semantics of this annotation are observed during execution
  * of the repository operation.</p>
+ *
+ * <h2>Jakarta Concurrency</h2>
+ *
+ * <p>When Jakarta Concurrency is available, a repository method that returns
+ * {@code void}, {@link CompletionStage}, or {@link CompletableFuture} can be
```
gavinking, line 1118:
```suggestion
 * <p>When Jakarta Concurrency is available, an asynchronous repository method may be
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -853,7 +854,7 @@ CompletionStage<Void> insertBook(Book book);
 
 
 An _asynchronous repository_ is a repository which declares asynchronous repository methods.
-A Jakarta Data provider is not required to support mixing synchronous and asynchronous repository methods within the same repository interface.
+A mixture of synchronous and asynchronous repository methods can be used within the same repository if all of the asynchronous methods are annotated with the `@Asynchronous` annotation to have the Jakarta Concurrency provider handle the asynchronous behavior. Otherwise, the Jakarta Data provider is not required to support mixing synchronous and asynchronous repository methods within the same repository interface.
```
gavinking, line 857: I would like to say something here which allows _reactive_ repositories to _not_ have to support `@Asynchronous`, because it would be some sort of weird noop for them.

njr-11: Yes, that makes sense.  How about this wording?  https://github.com/jakartaee/data/pull/1003/commits/818dfd8c98414fc7029a9e01b609e1889ce8fa90  It disallows `@Asynchronous` in that case.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -853,7 +854,7 @@ CompletionStage<Void> insertBook(Book book);
 
 
 An _asynchronous repository_ is a repository which declares asynchronous repository methods.
-A Jakarta Data provider is not required to support mixing synchronous and asynchronous repository methods within the same repository interface.
+A mixture of synchronous and asynchronous repository methods can be used within the same repository if all of the asynchronous methods are annotated with the `@Asynchronous` annotation to have the Jakarta Concurrency provider handle the asynchronous behavior. Otherwise, the Jakarta Data provider is not required to support mixing synchronous and asynchronous repository methods within the same repository interface. The `@Asynchronous` annotation must not be used on reactive repositories.
```
gavinking, line 857:
```suggestion
A mixture of synchronous and asynchronous repository methods can be used within the same repository if all of the asynchronous methods are annotated with the `@Asynchronous` annotation to have the Jakarta Concurrency provider handle the asynchronous behavior. Otherwise, the Jakarta Data provider is not required to support mixing synchronous and asynchronous repository methods within the same repository interface. The `@Asynchronous` annotation must not be used on repositories implemented using reactive streams.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -833,17 +833,18 @@ where:
...
 - `F` is a parameterized type representing a value which might not yet have been computed and which is supported by the Jakarta Data provider.
 
-NOTE: Every Jakarta Data provider is encouraged, but not required, to support asynchronous repository methods returning `java.util.concurrent.CompletionStage`.
+NOTE: Repository methods that are annotated with the `@Asynchronous` annotation from the Jakarta Concurrency specification must be able to return the types `java.util.concurrent.CompletionStage` and `java.util.concurrent.CompletableFuture`, in which case the Jakarta Data provider returns an already completed `CompletionStage` or `CompletableFuture` such that the Jakarta Concurrency provider is able to control the asynchronous behavior. Every Jakarta Data provider is otherwise encouraged, but not required, to support asynchronous repository methods returning `java.util.concurrent.CompletionStage`.
```
gavinking, line 836:
```suggestion
NOTE: Repository methods annotated with the `@Asynchronous` annotation from the Jakarta Concurrency specification must be able to return the types `java.util.concurrent.CompletionStage` and `java.util.concurrent.CompletableFuture`, in which case the Jakarta Data provider returns an already completed `CompletionStage` or `CompletableFuture` such that the Jakarta Concurrency provider is able to control the asynchronous behavior. Every Jakarta Data provider is otherwise encouraged, but not required, to support asynchronous repository methods returning `java.util.concurrent.CompletionStage`.
```

gavinking: I'm not sure we should allow `CompletableFuture` here. I think `CompletionStage` is enough.

njr-11: The Jakarta Concurrency spec explicitly states `CompletableFuture` as one of the 3 return types to use with [@Asynchronous](https://jakarta.ee/specifications/concurrency/3.1/apidocs/jakarta.concurrency/jakarta/enterprise/concurrent/asynchronous), so it seems correct to mention it:

> Methods with the following return types can be annotated to be asynchronous methods:
>
>  * [CompletableFuture](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/CompletableFuture.html)
>  * [CompletionStage](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/CompletionStage.html)
>  * void

gavinking: > The Jakarta Concurrency spec explicitly states CompletableFuture as one of the 3 return types

Yes, sure, I saw that, but that doesn't mean that Jakarta Data should require that providers support this.

And, I'm not an expert on this stuff, but it seems to me that it's better to _not_ expose the `complete()` method to repository clients. It's the responsibility of the repository implementation to call this operation, and clients should not be messing with it. 

AFAIK, `CompletionStage` is the right level of abstraction for clients.

njr-11: > it seems to me that it's better to _not_ expose the `complete()` method to repository clients

Yeah, I agree. I was only thinking about it relation to Find operations. But if a user has an asynchronous update/delete/insert, we wouldn't want them to be able to forcibly complete or cancel its CompletableFuture because it would be uncertain to the application whether the change was made or not.  I added commit https://github.com/jakartaee/data/pull/1003/commits/79c9fa5be19bb4717ed28471c949e9436fe7049c to remove CompletableFuture.

njr-11:
The remaining review comment was addressed under https://github.com/jakartaee/data/pull/1003/commits/79c9fa5be19bb4717ed28471c949e9436fe7049c and this is now ready for review again.

api/src/main/java/module-info.java
```diff
@@ -1109,6 +1111,37 @@
  * {@code @Transactional} is called in an environment where both Jakarta Transactions and
  * Jakarta CDI are available, the semantics of this annotation are observed during execution
  * of the repository operation.</p>
+ *
+ * <h2>Jakarta Concurrency</h2>
+ *
+ * <p>When Jakarta Concurrency is available, an asynchronous repository method may be
+ * annotated with {@code jakarta.enterprise.concurrent.Asynchronous} to cause the
+ * method to run asynchronously to the method invoker, as outlined by the section
+ * titled Asynchronous Methods in the Jakarta Concurrency specification.
```
gavinking, line 1120:
```suggestion
 * titled <em>Asynchronous Methods</em> in the Jakarta Concurrency specification.
```

api/src/main/java/module-info.java
```diff
@@ -1109,6 +1111,37 @@
...
+ * A {@code CompletionStage} return type must be parameterized with a type that
+ * would be a valid return type for a non-asynchronous repository method.</p>
```
gavinking, line 1122:
```suggestion
 * The return type must be {@code CompletionStage<R>} where {@code R} is a type 
 * that would be a valid return type for a non-asynchronous repository method.</p>
```

api/src/main/java/module-info.java
```diff
@@ -1109,6 +1111,37 @@
...
+ *
+ * <p>For example, in the following example, the {@code setPriceAsync} method
```
gavinking, line 1124:
```suggestion
 * <p>In the following example, the method {@code setPriceAsync()}
```

api/src/main/java/module-info.java
```diff
@@ -1109,6 +1111,37 @@
...
+ * immediately returns a {@code CompletionStage<Integer>} to the caller.
+ * When the {@code UPDATE} finishes, the {@code CompletionStage} completes
```
gavinking, line 1126:
```suggestion
 * After the operation is performed by the database, the {@code CompletionStage} completes
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -833,17 +833,18 @@ where:
 - `R m(P1 p1, P2 p2, ...)` is a legal repository method signature according to the previous sections of this chapter, or, in the special case that `R` is the type `java.lang.Void`, `void m(P1 p1, P2 p2, ...)` is a legal repository method signature according to the previous sections of this chapter, and
 - `F` is a parameterized type representing a value which might not yet have been computed and which is supported by the Jakarta Data provider.
 
-NOTE: Every Jakarta Data provider is encouraged, but not required, to support asynchronous repository methods returning `java.util.concurrent.CompletionStage`.
+NOTE: Repository methods annotated with the `@Asynchronous` annotation from the Jakarta Concurrency specification must be able to return the type `java.util.concurrent.CompletionStage`, in which case the Jakarta Data provider returns an already completed `CompletionStage` such that the Jakarta Concurrency provider is able to control the asynchronous behavior. Every Jakarta Data provider is otherwise encouraged, but not required, to support asynchronous repository methods returning `java.util.concurrent.CompletionStage`.
```
gavinking, line 836:
```suggestion
NOTE: Every Jakarta Data provider is otherwise encouraged, but not required, to support asynchronous repository methods returning `java.util.concurrent.CompletionStage`.

NOTE: A repository method annotated with the `@Asynchronous` annotation from the Jakarta Concurrency specification is permitted to declare the return type `java.util.concurrent.CompletionStage`. 
In this case, the Jakarta Data provider must return an already-completed `CompletionStage` so that the Jakarta Concurrency provider is able to control the asynchronous behavior.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -853,7 +854,7 @@ CompletionStage<Void> insertBook(Book book);
 
 
 An _asynchronous repository_ is a repository which declares asynchronous repository methods.
-A Jakarta Data provider is not required to support mixing synchronous and asynchronous repository methods within the same repository interface.
+A mixture of synchronous and asynchronous repository methods can be used within the same repository if all of the asynchronous methods are annotated with the `@Asynchronous` annotation to have the Jakarta Concurrency provider handle the asynchronous behavior. Otherwise, the Jakarta Data provider is not required to support mixing synchronous and asynchronous repository methods within the same repository interface. The `@Asynchronous` annotation must not be used on repositories implemented using reactive streams.
```
gavinking, line 857:
```suggestion
A repository may declare a mixture of synchronous and asynchronous repository methods if every asynchronous method is annotated with the `@Asynchronous` annotation, so that Jakarta Concurrency provides the asynchronous behavior. 
Otherwise, the Jakarta Data provider is not required to support mixing synchronous and asynchronous repository methods within the same repository interface. 
The `@Asynchronous` annotation must not be used on repositories implemented using reactive streams.
```

gavinking (commented):
Some minor suggestions on language.

api/src/main/java/module-info.java
gavinking: Not entirely sure of the best way to phrase this one, but "{@code UPDATE}" sounds too much like we're assuming this is SQL.

spec/src/main/asciidoc/repository.asciidoc
gavinking: There's a bit of a hole here; `@Asynchronous` methods don't (I assume) propagate the transaction context. So we should say something about what transaction context they run in. Does the async interceptor get applied before or after the tx management interceptor?

gavinking: Ah, there, no hole, found this:

> The Jakarta EE Product Provider must assign the interceptor for asynchronous methods to have priority of `Interceptor.Priority.PLATFORM_BEFORE + 5`. Interceptors with a lower priority, such as `Transactional`, must run on the thread where the asynchronous method executes, rather than on the submitting thread. When an asynchronous method is annotated as `Transactional`, the transactional types which can be used are: `TxType.REQUIRES_NEW`, which causes the method to run in a new transaction, and `TxType.NOT_SUPPORTED`, which causes the method to run with no transaction. All other transaction attributes must result in [UnsupportedOperationException](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/UnsupportedOperationException.html?is-external=true) upon invocation of the asynchronous method.

That's fine.

api/src/main/java/module-info.java
njr-11: `void` is also an acceptable return type (and currently works), so I'll plan to add that in after committing this suggestion.

njr-11:
Thanks for the review. I added in all of the suggestions and followed up with one commit that also mentions `void` as a valid return type, and I saw your other comment about `Transactional` and realized it would be a good idea to include that annotation on the example where it is making an update to the database to help remind users.

gavinking (approved):
🙏

PR #1005, opened 2025-03-06
Case insensitive Constraints via IgnoreCase annotation

njr-11:
Option 1 for case insensitive Constraints:
Give the repository method author an `@IgnoreCase` annotation with which to indicate when a Constraint should be applied independent of case.  Typically it will be known upfront that the database contains a mixture of case, and it will be more convenient to specify this information in a single place on the repository method signature rather than everywhere it is used (where there is also the risk of forgetting to do it in one place).

Example usage:

```
List<Book> booksTitled(@IgnoreCase Like title);

found = library.booksTitled(Like.pattern("Developing applications with Jakarta Data"));
```

### DIFF
--- a/api/src/main/java/jakarta/data/repository/IgnoreCase.java
+++ b/api/src/main/java/jakarta/data/repository/IgnoreCase.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.repository;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import jakarta.data.constraint.AtLeast;
+import jakarta.data.constraint.AtMost;
+import jakarta.data.constraint.Between;
+import jakarta.data.constraint.EqualTo;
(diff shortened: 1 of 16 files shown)


otaviojava:
Comparing those two options, I would stay with this one.

@gavinking, what do you think?

gavinking:
> what do you think?

I think I would not introduce a new top-level construct in our API for something marginally-useful with perfectly reasonable alternative solutions (JDQL, `Restriction`s).

njr-11:
Rebased this onto latest and then resolved various soft merge conflicts:
- Javadoc needed updates to match up with renamed packages
- Honor the recent line length limitation
- Refer to entity attributes rather than fields

I do see this PR as still relevant because there is no answer for ignoring case in what I would consider to be the recommended pattern:

- Encode everything upfront into the repository method that is fixed and known in advance, sparing the repository user from needing to re-specify it on every use.
- Have a Restriction parameter to allow for everything else that a repository user might want to customize at run time.

Without `@IgnoreCase`, the only way to encode ignoring case upfront is to use Query by Method Name or JDQL, neither of which the spec states will work with Restriction, and the former is also discouraged in general.

njr-11:
Rebased this again onto the latest.  I also went through code examples in the spec and identified where `@IgnoreCase` will be useful.  Pretty much any time an application takes input from a user (for example a product name or name pattern entered into a search form) or where the database itself could have varying case entries, this is going to be useful.  Restrictions provide a way to perform case insensitive comparisons for when the search criteria is determined at run time.   Query by Method Name pattern handles it for search criteria determined in advance, but Query by Method Name is discouraged.  This annotation fills the remaining gap to handle it for constraints on parameter-based Find/Delete.

otaviojava:
@GavinKing, we need to discuss this one. How about our next Jakarta Data meeting?

gavinking:
Yes, sorry I couldn't be there today. I was at an event.

api/src/main/java/jakarta/data/repository/IgnoreCase.java
```diff
@@ -0,0 +1,98 @@
...
+ * <li>{@link Like}</li>
+ * <li>{@link NotBetween} (not available for {@code @Is})</li>
+ * <li>{@link NotEqualTo}</li>
+ * <li>{@link NotLike}</li>
+ * </ul>
+ *
+ * <p>For example,</p>
+ *
+ * <pre>
+ * &#64;Repository
```
otaviojava, line 61: Please, just update to use Javadoc code tag :)

gavinking:
It's my opinion that it's quite likely that if we let this stew some more, we will think of a cleverer solution in future.

PR #1009, opened 2025-03-10
Fix typo, include example Query with SELECT returning record, and idea for more concise usage

njr-11:
This PR includes the following:
Fixes a mistake where an entity class name in an example doesn't match the example code where it is used.
The javadoc overview on methods returning Java records is specific to `@Find`. It should also allow for `@Query`.
An idea for a more concise for of `@Query` returning Java records, where the `SELECT` clause can be omitted entirely if the record component names match.

The latter is done as its own commit so that it could be removed from the PR in case it turns out to be unwanted for some reason.  Let me know.


gavinking (approved):
I think this is fine for now, though at some point I think we're going to have to nail the semantics down a bit more precisely, probably a in new section of the prose spec.

PR #1014, opened 2025-03-12
TCK tests for Query methods returning Java record

njr-11:
TCK tests for scenarios where:

- JDQL query returns Java record type that has the Select annotation on record components
- JDQL query returns Java record type where the record components determine the entity attributes retrieved
- JDQL query with a SELECT of multiple entity attributes returns Java record type


KyleAure (approved):
These tests look good to me.

PR #1016, opened 2025-03-13
add PageRequest.page(long)

gavinking:
Does anyone remember if there was a good reason that `PageRequest` doesn't offer the ability to mutate the page number?

Because to me it feels much more natural to write `Page.ofSize(PAGE_SIZE).page(pageNumber)` than `Page.ofPage(pageNumber).size(PAGE_SIZE)`.

### DIFF
--- a/api/src/main/java/jakarta/data/page/PageRequest.java
+++ b/api/src/main/java/jakarta/data/page/PageRequest.java
@@ -241,6 +241,17 @@ public interface PageRequest {
 
     /**
      * <p>Creates a new page request with the same pagination information,
+     * but with the specified page number.</p>
+     *
+     * @param pageNumber the page number.
+     * @return a new instance of {@code PageRequest}.
+     *         This method never returns {@code null}.
+     * @since 1.1
+     */
+    PageRequest page(long pageNumber);
+
+    /**
+     * <p>Creates a new page request with the same pagination information,
      * but with the specified maximum page size. When a page is retrieved
      * from the database, the number of elements in the page must be equal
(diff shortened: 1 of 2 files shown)


njr-11 (approved):
> Does anyone remember if there was a good reason that `PageRequest` doesn't offer the ability to mutate the page number?

Not really sure, but I think there might have been some concern about users trying to scroll through pages with this method rather than checking `Page.hasNext()` and then `Page.nextPageRequest()`?
 
> Because to me it feels much more natural to write `Page.ofSize(PAGE_SIZE).page(pageNumber)` than `Page.ofPage(pageNumber).size(PAGE_SIZE)`.

I agree.  A user will typically first decide how big they want the pages to be and after. that which page number they want.  It is backwards for a user to say that they interested in a page 3 regardless of how big the pages are and then after that decide what the size of the page is.

gavinking:
> Not really sure, but I think there might have been some concern about users trying to scroll through pages with this method rather than checking `Page.hasNext()` and then `Page.nextPageRequest()`?

Hrm, yeah, I guess that sorta rings a bell. But now I doubt it's something we need to be too concerned about.

otaviojava:
> Does anyone remember if there was a good reason that `PageRequest` doesn't offer the ability to mutate the page number?
> 
> Because to me it feels much more natural to write `Page.ofSize(PAGE_SIZE).page(pageNumber)` than `Page.ofPage(pageNumber).size(PAGE_SIZE)`.

Because we forget it :).
Glad to put it on it

PR #1017, opened 2025-03-14
Disallow multiple Select values on a record component

njr-11:
This is an idea for a possible improvement to `@Select` that will switch its `value` from an array to a single item to prevent the following which do not make sense:

- `@Select({ "attributeName1", "attributeName2", ... })` on a single record component
- `@Select({})`

The only place where we need multiple entity attribute names is on a repository method, and so this PR proposes to make the annotation repeatable (as we have for `OrderBy`) on a method,

```
@Repeatable(Select.List.class)
```

```
    @Target(ElementType.METHOD) // RECORD_COMPONENT is intentionally omitted
    @interface List {
```

See the updated TCK tests for examples of usage.

It's a little bit more to type when specified on a method that returns a record. However, it does format nicely and has consistency with existing usage of `OrderBy`. Also, `@Select` on a repository method is useful when you want to map results to a third-party provided record class that you can't modify to add the `@Select` to or you are sharing the same record class for different purposes and cannot have the `@Select` there or need to override it. These are both less often encountered scenarios outside of core usage, so it could make sense to trade off making it at little bit more verbose here to better enforce valid usage and eliminate the possibility of doing some nonsensical things with it.

What do others think about this?  I'm fine with it either way and this is pretty minor, but I thought it was a potential improvement that deserved to be considered.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Select.java
+++ b/api/src/main/java/jakarta/data/repository/Select.java
@@ -19,6 +19,7 @@ package jakarta.data.repository;
 
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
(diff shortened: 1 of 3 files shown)


gavinking (approved):
Good idea.

Issue #459, opened 2024-02-20
Proposal for Enhanced Data Persistence Pattern[Vote]


otaviojava:
### As a ...

- [X] Application user/user of the configuration itself
- [X] API user (application developer)
- [ ] SPI user (container or runtime developer)
- [ ] Specification implementer

### I need to be able to ...

Building upon the Query Language Proposal ([#458](https://github.com/jakartaee/data/issues/458)), this proposal suggests introducing a new Data Persistence pattern to augment the Jakarta Data Template interface. This enhancement aims to provide a more data-agnostic approach to Data Access Object (DAO) implementation, offering greater flexibility and consistency across different data storage technologies.

### Which enables me to ...

The primary goal of this proposal is to initiate a discussion and a vote regarding introducing the Enhanced Data Persistence pattern into the Jakarta Data project.

**Key Features:**

1. **Standardized Operations:** The existing Jakarta Data Template Interface offers a standardized set of methods for common database operations, ensuring consistency across the application.

2. **Template Pattern:** Following the template pattern, the interface simplifies the implementation of data access logic, making it more accessible and modular.

3. **Flexibility for Data-Agnostic DAOs:** The proposed pattern enriches the Template interface with additional capabilities, empowering developers to implement Data Access Objects (DAOs) in a data-agnostic manner. This flexibility allows for the seamless switching between different data storage technologies.

4. **Support for Fluent API Querying:** The addition of a fluent query API enables developers to construct expressive and database-agnostic queries directly through the DAO interface, enhancing readability and maintainability.

5. **Integration with Criteria Queries:** The proposal includes support for Criteria queries, enhancing the flexibility and expressiveness of data retrieval operations. Further discussion on this feature will be conducted in Issue [#SOON](https://github.com/jakartaee/data/issues/XYZ).

**Example Usage:**

```java
List<Car> cars = template
    .select(Car.class)
    .where(_Car.city.eq("Rome").and(_Car.active.isTrue()))
    .orderBy(_Car.name.asc())
    .result();
```

**Benefits:**

1. **Enhanced Productivity:** The Enhanced Data Persistence pattern significantly increases productivity by reducing boilerplate code and simplifying common database operations.

2. **Consistent Data Access:** Developers can rely on a consistent API for data access, promoting maintainability and ease of understanding across the application.

3. **Data-Agnostic Implementation:** By decoupling DAOs from specific data storage technologies, the pattern promotes adaptability and future-proofing of the application architecture.

[...]

PR #1021, opened 2025-03-19
Expressions

gavinking:
This is a rough straw man for a possible path to take with expressions.

I'm posting this here in an incomplete and uncompilable state, so that you guys can look at the interfaces, before we go and spend any time on adding implementing `record` types.

FTR this turned out a lot different to&mdash;and somewhat more complicated than&mdash;what I perhaps anticipated, but I actually quite like the result. You can write stuff like:

```java
Restriction<Entity> like = _Entity.message.lower().prepend("hello").like("hello %");
Restriction<Entity> notBetween = _Entity.count.abs().times(12L).between(-1L, 1L).negate();
Restriction<Entity> greaterThan = _Entity.decimal.minus(BigDecimal.TEN).satisfies(greaterThan(BigDecimal.ZERO));
```

Note that this is missing the ability to obtain restrictions that compare expressions or different attributes.

Related issue: #459


api/src/main/java/jakarta/data/metamodel/TextExpression.java
```diff
@@ -0,0 +1,66 @@
...
+    TextExpression<T> prepend(Expression<T,String> string);
+    TextExpression<T> append(Expression<T,String> string);
+
+    TextExpression<T> upper();
+    TextExpression<T> lower();
+
+    TextExpression<T> left(int length);
+    TextExpression<T> right(int length);
+
+    Expression<T,Integer> length();
```
njr-11, line 36:
```suggestion
    NumericExpression<T,Integer> length();
```

So that you can write

        Restriction<Book> titleLT50 = _Book.title.length().lessThan(50);

instead of:

        Restriction<Book> titleLT50 = _Book.title.length().satisfies(Constraint.lessThan(50));

njr-11 (commented):
Good idea starting with the interfaces. I pulled them in locally and this will allow us to experiment with different usage, and see what kinds of Restrictions it will become possible for a user to write with it. At first glance this looks quite nice. I added one initial comment that I spotted on a first attempt at experimenting.

api/src/main/java/jakarta/data/metamodel/Expression.java
```diff
@@ -0,0 +1,36 @@
...
+
+import jakarta.data.metamodel.constraint.Constraint;
+import jakarta.data.metamodel.restrict.Restriction;
+
+import java.util.Set;
+
+public interface Expression<T,V> {
+
+    Restriction<T> equalTo(V value);
+    Restriction<T> notEqualTo(V value);
```
njr-11, line 28:
```suggestion
    Restriction<T> equalTo(Expression<T, V> expression);
    Restriction<T> equalTo(V value);
    Restriction<T> notEqualTo(Expression<T, V> expression);
    Restriction<T> notEqualTo(V value);
```

so that you can compare two attributes to each other:

```
        Restriction<Book> autobiography = _Book.title.equalTo(_Book.author);
```

api/src/main/java/jakarta/data/metamodel/NumericExpression.java
```diff
@@ -0,0 +1,34 @@
...
+
+    NumericExpression<T,N> plus(N other);
+    NumericExpression<T,N> minus(N other);
+    NumericExpression<T,N> times(N other);
+    NumericExpression<T,N> divide(N other);
+
+    NumericExpression<T,N> plus(Expression<T,N> other);
+    NumericExpression<T,N> minus(Expression<T,N> other);
+    NumericExpression<T,N> times(Expression<T,N> other);
+    NumericExpression<T,N> divide(Expression<T,N> other);
```
njr-11, line 33:
```suggestion
    NumericExpression<T,N> divideBy(N other);

    NumericExpression<T,N> plus(Expression<T,N> other);
    NumericExpression<T,N> minus(Expression<T,N> other);
    NumericExpression<T,N> times(Expression<T,N> other);
    NumericExpression<T,N> divideBy(Expression<T,N> other);
```
I like that `divide` is shorter, but it might read better to have `divideBy`.  I'm not really sure which is better, but I thought I'd bring it up.

An example,

```
        Restriction<Book> shortChaptered =
                _Book.numPages.divide(_Book.numChapters.plus(1))
                        .lessThanEqual(5);
```

or

```
        Restriction<Book> shortChaptered =
                _Book.numPages.divideBy(_Book.numChapters.plus(1))
                        .lessThanEqual(5);
```

gavinking: I'm not at all attached to the naming of these methods.

api/src/main/java/jakarta/data/metamodel/Expression.java
gavinking: Was just thinking/wondering if it's more complicated than that, but maybe not, perhaps that's all that's needed.

gavinking:
Mainly because I was curious to see what it would look like, I just added a way to do "navigability". We would grow two new interfaces, but it seems to me that they're self-contained, with (surprisingly) no impact elsewhere.

So you could write:

```java
TextRestriction<Book> authorNameStartsWithGavin = 
        _Book.authors.navigate(_Author.name).startsWith("Gavin");
```

api/src/main/java/jakarta/data/metamodel/Expression.java
gavinking: > thinking/wondering if it's more complicated than that

So the complications arise when you ask if these sort of cross-field restrictions are made from `Constraint`s. After some reflection, I don't think there are. They would be a new sort of `Restriction`. That's actually fine, but it's probably not something we want to bite off at first (it can easily be added).

gavinking:
As a first step on implementation, I've generalized `BasicRestriction` so that it applied to `Expression`s, and not just `BasicAttribute`s. This is pretty clean.

api/src/main/java/jakarta/data/metamodel/NavigableExpression.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package jakarta.data.metamodel;
+
+public interface NavigableExpression<T,U> {
+
+    default <V> NavigableExpression<T, V> navigate(NavigableAttribute<V, V> attribute) {
+        throw new UnsupportedOperationException("not yet implemented");
+    }
+
+    default TextExpression<T> navigate(TextAttribute<U> attribute) {
```
njr-11, line 26: It is unfortunate that this will require the user to input information that should already be known,

```java
_Book.author.navigate(_Author.name).equalTo(authorName);
```

But I haven't been able to think of any way around it.  The closest I could come up with is to have the metamodel class supply the attribute (`_Author`) when constructing the NavigableAttribute, and have,
```
_Book.author.navigate().equalTo(authorName);
```
although this doesn't work due to warnings about invoking instance methods on a static class and we would also lose the information about which entity the association came from.

njr-11 (commented):
I added a few minor comments, but I would say it all looks good to continue on with the next steps.

api/src/main/java/jakarta/data/metamodel/Expression.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    default BasicRestriction<T,V> notNull() {
+        return new BasicRestrictionRecord<>(this, NotNull.instance());
+    }
+
+    // TODO: should this be called restrict() ?
+    default BasicRestriction<T,V> satisfies(Constraint<V> constraint) {
+        return new BasicRestrictionRecord<>(this, constraint);
+    }
```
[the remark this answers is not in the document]
gavinking: I think it's the sort of thing that can be *very* useful for generic/framework code built on top of Jakarta Data.

api/src/main/java/jakarta/data/metamodel/NavigableExpression.java
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.metamodel;
+
+public interface NavigableExpression<T,U> {
```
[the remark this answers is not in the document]
gavinking: Yeah, exactly, that's why I went with "navigable", bit I'm not at all attached to it if you can think of something better.

api/src/main/java/jakarta/data/metamodel/NavigableExpression.java
gavinking: the only way I can see to do better would be to generate additional types, something approximately like this:

```java
@StaticMetamodel(Book.class)
@Generated("org.hibernate.processor.HibernateProcessor")
public interface _Book {
    TextAttribute<Book,String> isbn = ... ;
    TextAttribute<Book,String> title = ... ;

    _Author.Navigable<Book>  authors = .... ; 

    interface Navigable<T> extends NavigableAttribute<T,Book> {
        Expression<Book,String> isbn = ... ;
        Expression<Book,String> title = ... ;
        ...
    }
}
```

So you could write `_Book.authors.name.equalTo(authorName)` or `_Author.books.title.startsWith("Jakarta")`.

I suppose that could be made to work but it does mean generating that extra type.

gavinking: (Perhaps `NavigableAttribute` is not needed in this approach. Need to prototype something.)

gavinking:
I just pushed a strawman implementation of the new operations, based on the idea of a `Function` type which extends `Expression`. I think it's ... actually okay.

gavinking:
OK, folks, I've done a bunch of work on filling out the implementation of this stuff, and cleaning things up to give everything a consistent approach. I'm pretty happy with it, though I do still have a couple of unanswered questions.

gavinking:
(I recommend actually checking out the branch, instead of trying to review the diff!)

api/src/main/java/jakarta/data/metamodel/ComparableAttribute.java
```diff
@@ -109,22 +78,3 @@ static <T, V extends Comparable<V>> ComparableAttribute<T, V> of(
     }
 }
 
-/**
- * Hidden internal implementation of ComparableAttribute.
- *
- * @param <T> entity class of the static metamodel.
- * @param <V> type of entity attribute (or wrapper type if primitive).
- */
-record ComparableAttributeRecord<T,V extends Comparable<V>>(String name)
```
njr-11, line 118: This PR mostly does well with keeping the introduced implementation classes/records as package protected and hidden, except not the *AttributeRecord classes.
I'm very much opposed to introducing more implementation classes into the API.  This class here was demonstrating one way you can still have implementation for Attribute classes and avoid including it in the API.  I suppose it's fine if we're going to temporarily do this and put it back before releasing, but I want to be sure that we actually do that.

api/src/main/java/jakarta/data/metamodel/ComparableExpression.java
```diff
@@ -0,0 +1,74 @@
...
+import jakarta.data.metamodel.constraint.GreaterThanOrEqual;
+import jakarta.data.metamodel.constraint.LessThan;
+import jakarta.data.metamodel.constraint.LessThanOrEqual;
+import jakarta.data.metamodel.constraint.NotBetween;
+import jakarta.data.metamodel.restrict.BasicRestriction;
+
+public interface ComparableExpression<T,V extends Comparable<?>>
+        extends Expression<T, V> {
+
+    // QUESTION: do we really need to expose BasicRestriction here
```
njr-11, line 31: No, I can't see a reason to return BasicRestriction rather than Restriction. The only operation for the end user on BasicRestriction is .negate() and that is also on Restriction, so we should return Restriction given that we don't have a need for anything more. That will make autocomplete simpler and more straightforward because it won't be cluttered up with the extra SPI operations that only Jakarta Data providers will be using from BasicRestriction.

api/src/main/java/jakarta/data/metamodel/Expression.java
```diff
@@ -0,0 +1,78 @@
...
+import jakarta.data.metamodel.constraint.NotIn;
+import jakarta.data.metamodel.constraint.NotNull;
+import jakarta.data.metamodel.constraint.Null;
+import jakarta.data.metamodel.restrict.BasicRestriction;
+
+import java.util.Set;
+
+public interface Expression<T,V> {
+
+    // QUESTION: do we really need to expose BasicRestriction here
```
njr-11, line 32: Same as in the other case. Let's use Restriction unless/until a need is shown to be more specific.

api/src/main/java/jakarta/data/metamodel/NavigableExpression.java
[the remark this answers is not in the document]
njr-11: Resolving this conversation. We can use the separate issue with the table of attributes types/metamodel classes to consider if we want to keep Navigable or split into types that are more specific.

api/src/main/java/jakarta/data/metamodel/expression/FunctionExpression.java
```diff
@@ -15,13 +15,13 @@
...
 
-public interface TextRestriction<T> extends BasicRestriction<T, String> {
-    @Override
-    TextRestriction<T> negate();
-    @Override
-    Constraint<String> constraint();
+import java.util.List;
+
+public interface FunctionExpression<T,V> extends Expression<T,V> {
+    String name();
```
njr-11, line 25: We can fix this later, but it seems like this should be an enumeration of possible functions. Or if not that, at least there should be String constants for each function name defined by the spec.

njr-11 (commented):
Overall this looks great.  We will need to merge it for the rest of us to really experiment with it, and I'm fine with that.  My biggest concern is with seeing more record impl classes go onto the API, which as I pointed out in one of the comments you mostly avoided, except in the case of *AttributeRecord.  I do want some assurance that we will not be releasing those and can hide them instead, but with that plan in place, I won't make you clean them up here before merging.  This PR is already enormous.  We also need to decide between using an enumeration or string constants for functions.  Hard coding string values that the Data provider needs to somehow know is not the right way to do things, but is fine as an intermediate step.

gavinking:
>  I do want some assurance that we will not be releasing those and can hide them instead

Yes, I know, as I said, I think we need to discuss the strategy for that. For now I just want everything consistent with what we had in 1.0, until we decide what the correct solution is.

Let's talk about it on the next call.

api/src/main/java/jakarta/data/metamodel/expression/FunctionExpression.java
gavinking: I think string constants would be a good idea. I'm working on the assumption that functions are eventually going to be an open world.

api/src/main/java/jakarta/data/metamodel/Expression.java
gavinking: OK, fine by me.

api/src/main/java/jakarta/data/metamodel/expression/FunctionExpression.java
gavinking: Done!

api/src/main/java/jakarta/data/metamodel/Expression.java
gavinking: Done.

api/src/main/java/jakarta/data/metamodel/ComparableExpression.java
gavinking: All done.

gavinking:
@njr-11 @otaviojava Would you guys like me to  add `BooleanExpression` and `BooleanAttribute` to this PR, with `isTrue()` and `isFalse()` methods?

If so, I will add them. Otherwise I will leave that for when we decide it's really needed.

njr-11:
> @njr-11 @otaviojava Would you guys like me to add `BooleanExpression` and `BooleanAttribute` to this PR, with `isTrue()` and `isFalse()` methods?
> 
> If so, I will add them. Otherwise I will leave that for when we decide it's really needed.

Let's leave it for later if we decide to do it at all.  At this point, I would say that there isn't a good justification for needing to add a BooleanExpression because you can already do `.equalTo(true/false)`

njr-11 (approved):

otaviojava:
> @njr-11 @otaviojava Would you guys like me to add `BooleanExpression` and `BooleanAttribute` to this PR, with `isTrue()` and `isFalse()` methods?
> 
> If so, I will add them. Otherwise I will leave that for when we decide it's really needed.

I said that at the meeting, I would like more having this boolean, even as next PR.

api/src/main/java/jakarta/data/metamodel/Expression.java
```diff
@@ -0,0 +1,77 @@
...
+
+        return BasicRestriction.of(this, NotIn.values(values));
+    }
+
+    default Restriction<T> notNull() {
+        return BasicRestriction.of(this, NotNull.instance());
+    }
+
+    // TODO: should this be called restrict() ?
+    default Restriction<T> satisfies(Constraint<V> constraint) {
```
otaviojava, line 64: as next PR, I would go to: **restrict** method.

api/src/main/java/jakarta/data/metamodel/expression/TextFunctionExpression.java
```diff
@@ -0,0 +1,48 @@
...
+package jakarta.data.metamodel.expression;
+
+import jakarta.data.metamodel.TextExpression;
+
+import java.util.List;
+
+public interface TextFunctionExpression<T>
+        extends FunctionExpression<T,String>, TextExpression<T> {
+
+    String CONCAT = "concat";
```
otaviojava, line 27: at `NumericOperatorExpression` you put it at an enum.

I did not understand why I was not following the same way.

api/src/main/java/jakarta/data/metamodel/expression/TextFunctionExpression.java
```diff
@@ -0,0 +1,48 @@
...
+    String UPPER = "upper";
+    String LOWER = "lower";
+    String LEFT = "left";
+    String RIGHT = "right";
+
+    static <T> TextFunctionExpression<T> of(String name, TextExpression<T> argument) {
```
otaviojava, line 33: I would go with an enum with the function that we support.

Otherwise, I could put:

```java
   return TextFunctionExpression.of("gavin", string, this);
  
 ``` 
 
 Unless, I am missing somepoint

gavinking: @otaviojava Remember that:

- JPQL defines extra functions that don't exist in Jakarta Data, and
- actual datastore have their own functions, even user-written functions.

so as I said above, eventually it's *extremely* likely that we'll want to let people call functions which aren't required by the spec.

api/src/main/java/jakarta/data/metamodel/expression/TextFunctionExpression.java
gavinking: Short answer: because user-defined functions are a thing (it's an open world) whereas user-defined operators are not a thing (at least not in this context).

api/src/main/java/jakarta/data/metamodel/Expression.java
gavinking: That's fine.

otaviojava (approved):

PR #1028, opened 2025-03-24
proposed improvements to javadoc

gavinking:


api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -25,26 +25,30 @@
...
+ * of a record type. When applied to a record component, the {@code Select}
+ * annotation affects every repository method which returns the record type
+ * and which does not have a {@code @Select} annotation.</li>
+ * </ul>
  *
  * <p>This annotation must not be used in other locations.</p>
  *
+ * <p>Any repository method affected by a {@code @Select} annotation must
+ * {@linkplain Find#value explicitly specify} the queried entity type.</p>
+ *
```
njr-11, line 51: I think the point of this is to ensure users don't lose the information about which entity the query pertains to, and it is definitely good to say something about that.  However, the wording here with "explicitly specify" linking to `Find.value()` might be interpreted as ruling out two other ways of explicitly specifying: relying on the primary entity type or a JDQL `FROM` clause.  Maybe try this to avoid that confusion:

```suggestion
 * <p>Any repository method affected by a {@code @Select} annotation must
 * explicitly specify the queried entity type. This is accomplished by
 * including the {@code @Find} annotation {@link Find#value value},
 * a {@code FROM} clause (if using {@link Query @Query}), or the
 * primary entity type of the repository.</p>
 *
```

njr-11 (commented):
Good updates. I only spotted one place that needs clarification to avoid possible misinterpretation.

api/src/main/java/jakarta/data/repository/Select.java
gavinking: @njr-11 I though we decided that `@Select` can't be used with `@Query`. If it can be, we need more changes than this.

njr-11: > @njr-11 I though we decided that `@Select` can't be used with `@Query`. If it can be, we need more changes than this.

There is a subtle aspect to the wording of the proposed addition here.  It says `affected by a {@code @Select} annotation` but maybe it was intended to say or mean `annotated by a {@code @Select} annotation`. If that were the case, it would not be odds with `@Query` at all. `@Select` cannot annotate a repository method that is annotated `@Query`, but `@Select` can be used on the components of a record type that is returned by a `@Query` method, which is also a way of affecting the repository method.

From `@Query`:
https://github.com/jakartaee/data/blob/527155b0ffd02cc9249200fa1be9ebe0bbbab91e/api/src/main/java/jakarta/data/repository/Query.java#L38-L42

Example in Javadoc:
https://github.com/jakartaee/data/blob/527155b0ffd02cc9249200fa1be9ebe0bbbab91e/api/src/main/java/module-info.java#L917-L928

gavinking: OK, fine, but then we need to mention in the javadoc for `@Select` that it affects `@Query` methods too, because currently it's written like `@Select` is only for use with `@Find`.

njr-11: > OK, fine, but then we need to mention in the javadoc for `@Select` that it affects `@Query` methods too, because currently it's written like `@Select` is only for use with `@Find`.

I agree.  `@Select` is currently written to avoid contradicting that usage, but it is only implicit in that `@Select` has language about "When used on a record component ..." and records can be returned by `@Query` and `@Query` says that record components can be annotated with `@Select`.  It's very indirect for a user to be able to piece that all together.  I'll see if I can propose an update to wording to be more explicit here.

api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -25,26 +25,30 @@
 import java.lang.annotation.Target;
 
 /**
- * <p>Determines which entity attribute(s) are returned by a {@link Find}
- * operation.</p>
+ * <p>Determines the attribute or attributes of an entity which are returned
+ * by a {@linkplain Find parameter-based} query method.</p>
```
njr-11, line 29:
```suggestion
 * <p>Determines the attribute or attributes of an entity which are returned
 * by a query.</p>
```

api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -25,26 +25,30 @@
...
+ * <li>At most one {@code @Select} annotation may be applied to a repository
+ * method that returns an entity attribute type.</li>
+ * <li>Multiple {@code @Select} annotations may be applied to a repository
+ * method which returns a record type, one annotation for each component of
+ * the record type. The number and order of {@code @Select} annotations must
+ * match the number and order of record components.</li>
+ * <li>At most one {@code @Select} annotation may be applied to a component
+ * of a record type. When applied to a record component, the {@code Select}
+ * annotation affects every repository method which returns the record type
+ * and which does not have a {@code @Select} annotation.</li>
```
njr-11, line 44:
```suggestion
 * annotation affects every repository {@link Find @Find} method which returns
 * the record type and which does not have a {@code @Select} annotation, and
 * it applies to every repository {@link Query @Query} method which returns
 * the record type.</li>
```

njr-11 (commented):
Added some possible ways of mentioning the usage with `@Query`

api/src/main/java/jakarta/data/repository/Select.java
otaviojava: > OK, fine, but then we need to mention in the javadoc for @Select that it affects @Query methods too, because currently it's written like @Select is only for use with @Find.

I am super afraid of creating a Frankenstein at the API.
I would vote to keep the `Select`annotation only combined with `Find`.

Otherwise, we need to describe any aspect, of the combinations, such as:

```java

@Select({"id", "title", "edition"})
@Select("select title, from Book")
List<Book> find(String title);
```

njr-11: > I am super afraid of creating a Frankenstein at the API. I would vote to keep the `Select`annotation only combined with `Find`.
> 
> Otherwise, we need to describe any aspect, of the combinations, such as:
> 
> ```java
> @Select({"id", "title", "edition"})
> @Select("select title, from Book")
> List<Book> find(String title);
> ```

@otaviojava No one is proposing that we allow `@Select` and `@Query` to annotate same method (I think that is what you intended to show above rather than a `@Select` with JDQL in it).

This PR isn't changing what is allowed or adding anything new to Jakarta Data. It's only aiming to make the documentation of `@Select` more clear.  One of the patterns that it wasn't clear on is the ability to put `@Select` onto record components.  Record components can be returned by `@Query` methods as well as `@Find` methods.  Our documentation for `@Query` already states this is valid and we have other Javadoc giving [examples of this usage](https://github.com/jakartaee/data/blob/527155b0ffd02cc9249200fa1be9ebe0bbbab91e/api/src/main/java/module-info.java#L917-L928) and TCK scenarios for it.  The `@Select` documentation was already consistent with it as well, although didn't explicitly call it out.  All we are proposing to do here is have the `@Select` javadoc explicity call out what elsewhere in the documentation we already say is possible.

gavinking:
I just pushed. WDYT of the changes?

api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -25,23 +25,27 @@
...
+ * <li>Multiple {@code @Select} annotations may be applied to a repository
+ * method which returns a record type, one annotation for each component of
+ * the record type. The number and order of {@code @Select} annotations must
+ * match the number and order of record components.</li>
+ * <li>At most one {@code @Select} annotation may be applied to a component
+ * of a record type. When applied to a record component, the {@code Select}
+ * annotation affects every {@link Find @Find} method which returns the
+ * record type and which does not have a {@code @Select} annotation, and
+ * every {@link Query @Query} method which returns the record type and does
+ * not have an explicit {@code SELECT} clause.</li>
```
njr-11, line 47: minor grammar correction:
```suggestion
 * it affects every {@link Query @Query} method which returns the record type
 * and does not have an explicit {@code SELECT} clause.</li>
```

api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -25,23 +25,27 @@
...
- * <p>When used on a record component, the {@code Select} annotation
- * value is used for all repository methods that lack this
- * annotation and for which the record is the result type.</p>
+ * <ul>
+ * <li>At most one {@code @Select} annotation may be applied to a repository
+ * method that returns an entity attribute type.</li>
+ * <li>Multiple {@code @Select} annotations may be applied to a repository
+ * method which returns a record type, one annotation for each component of
+ * the record type. The number and order of {@code @Select} annotations must
+ * match the number and order of record components.</li>
```
njr-11, line 41: Clarifies that this is allowed only with `@Find` and not `@Query`, 
```suggestion
 * {@code @Find} method which returns a record type, one annotation for each
 * component of the record type. The number and order of {@code @Select}
 * annotations must match the number and order of record components.</li>
```

api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -25,23 +25,27 @@
...
- * multiple {@code Select} annotations may annotate the method, one for each
- * record component. The number and order of {@code Select} annotations must
- * match the number and order of record components.</p>
- *
- * <p>When used on a record component, the {@code Select} annotation
- * value is used for all repository methods that lack this
- * annotation and for which the record is the result type.</p>
+ * <ul>
+ * <li>At most one {@code @Select} annotation may be applied to a repository
+ * method that returns an entity attribute type.</li>
```
njr-11, line 37: Clarification that this is allowed only with `@Find` and not `@Query`
```suggestion
 * <li>At most one {@code @Select} annotation may be applied to a repository
 * {@code @Find} method that returns an entity attribute type.</li>
```

api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -25,23 +25,27 @@
...
- * <p>Determines which entity attribute(s) are returned by a {@link Find}
- * operation.</p>
+ * <p>Determines the attribute or attributes of an entity which are returned
+ * by a {@linkplain Find parameter-based} or {@linkplain Query annotated}
+ * query method.</p>
  *
- * <p>This annotation can be specified on a repository find method or
- * on a record component.</p>
+ * <p>This annotation may be applied to a repository method annotated
+ * {@code @Find}, or to a component of a record type.</p>
```
njr-11, line 33: Minor grammar correction:
```suggestion
 * {@code @Find} or to a component of a record type.</p>
```

njr-11 (approved):
This looks great.
I spotted two very minor grammar corrections and two other places where we should clarify that were are speaking only of `@Find` and not `@Query` to avoid misinterpretation.

gavinking:
OK, I have pushed again, with your suggestions, and some rewriting of the jdoc of `@Find`.

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -27,48 +27,49 @@
...
+ * <li>be of type {@link jakarta.data.Limit}, {@link jakarta.data.Sort},
+ *     {@link jakarta.data.Order}, or {@link jakarta.data.page.PageRequest}.</li>
  * </ul>
- * <p>The query is inferred from the method parameters which match attributes of the entity.
- * </p>
- * <p>There is no specific naming convention for methods annotated with {@code @Find}; they may be named arbitrarily,
- * and their names do not carry any semantic meaning defined by the Jakarta Data specification.
- * </p>
+ * <p>The query is inferred from the method parameters which match attributes of
+ *  entity.</p>
```
njr-11, line 67: A word was lost when reformatting:
```suggestion
 * <p>The query is inferred from the method parameters which match attributes of
 * the entity.</p>
```

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -27,48 +27,49 @@
...
- *     must match, or the entity attribute names must be specified by the
- *     {@link Select} annotation.</li>
+ * <li>If the method return involves an entity type, according to the rules
+ *     specified below, the return type determines the queried entity type.</li>
+ * <li>Otherwise, if the method return type does not involve an entity type,
+ *     the queried entity type is the {@linkplain #value entity class explicitly
+ *     specified by the annotation}, if any, or the primary entity type of the
+ *     repository.</li>
  * </ul>
  *
```
njr-11, line 44: This paragraph was lost during the rewrite.  This was added to ensure it is implementable based on Jakarta Persistence/JPQL. (At the time it was written the JDQL for this pattern didn't exist yet, but JDQL needs to remain a subset of JPQL and cannot go beyond it or become incompatible with it)
```suggestion
 * <p>When returning a single entity attribute or a subset of entity attributes,
 * the attribute types must be single-valued basic types, not multiple-valued
 * types such as collections, arrays, or associations.</p>
 *
```

njr-11 (changes requested):
The rewritten text looks fine, but I spotted 2 places where things were lost. One is a paragraph and the other was a single word that was lost when adding a line break.

api/src/main/java/jakarta/data/repository/Find.java
gavinking: It wasn't lost. It was changed to:

```java
 * <p>When the return type is not the queried entity type, every returned
 * attribute must be a basic attribute.</p>
 ```

which uses a term with a precise definition in the spec.

Comment (njr-11) [reply api/src/main/java/jakarta/data/repository/Find.java:44]: > It wasn't lost. It was changed to:
> 
> ```java
>  * <p>When the return type is not the queried entity type, every returned
>  * attribute must be a basic attribute.</p>
> ```
> 
> which uses a term with a precise definition in the spec.

I didn't notice that because it also moved to a different a place. This is fine. No issue here.

File: api/src/main/java/jakarta/data/repository/Select.java [hunk 10]
```diff
@@ -25,23 +25,27 @@
...
+ * method that returns an entity attribute type.</li>
+ * <li>Multiple {@code @Select} annotations may be applied to a {@code @Find}
+ * method which returns a record type, one annotation for each component of
+ * the record type. The number and order of {@code @Select} annotations must
+ * match the number and order of record components.</li>
+ * <li>At most one {@code @Select} annotation may be applied to a component
+ * of a record type. When applied to a record component, the {@code Select}
+ * annotation affects every {@link Find @Find} method which returns the
+ * record type and which does not have a {@code @Select} annotation, and
+ * affects every {@link Query @Query} method which returns the record type
```
njr-11, line 46: Sorry I wasn't very clear about the grammar correction on this one. The comma can be used before "and" when the "and" joins two clauses that could be sentences of their own.  I had proposed adding "it" here so there would be a subject in the second clause.  Another option would be removing the comma, but that seemed more confusing because the sentence is already long and has another occurrences of "and" that is used differently.  In any case, this is so trivial that I'm adding the approval regardless of whether you decide to change it.

```suggestion
 * record type and which does not have a {@code @Select} annotation, and it
 * affects every {@link Query @Query} method which returns the record type
```

njr-11 (approved):

api/src/main/java/jakarta/data/repository/Select.java
gavinking: > The comma can be used before "and" when the "and" joins two clauses that could be sentences of their own. 

I've been going along with this for a while now, but don't think it's actually correct. Do you have a source for it? ChatGPT thinks I'm allowed to use ", and" in this circumstance:

Screenshot 2025-03-26 at 8 42 03 PM
Screenshot 2025-03-26 at 8 42 32 PM

That is to say, this is a "complex list of clauses", not just "two phrases", because each of the items has a noun and a verb, IIUC.

njr-11: I'm not expert on grammar.  When I saw it, it looked wrong, so I googled "when to use comma before and".
Google's AI told me: `Use a comma before "and" when it connects two independent clauses. An independent clause can stand alone as a sentence because it has a subject and a verb.`

Not entirely trusting AI to get things right, I also clicked on the search result,
https://www.grammarly.com/blog/punctuation-capitalization/comma-before-and/
which says,
"Use a comma before _and_ when connecting two independent clauses."
and
"Don’t use a comma before _and_ when connecting an independent clause to a dependent clause ..."

After that, I didn't look any further because it didn't seem worth it.

gavinking: I mean, I don't know English grammar either because it's not taught in Australian schools. But usually my intuition on these things is correct.

With respect to trusting language models on this sort of stuff, well, actually I trust them more than any other source. They're *really* good at this shit. I don't use them to actually _write_ stuff, because from what I've seen they tend to produce sorta bland and boring writing, but for analyzing text they're magical.

DeepSeek agrees with ChatGPT, by the way, and has an even better explanation for why this sentence was correct as originally written.

Screenshot 2025-03-26 at 9 03 13 PM

Notice that DeepSeek has parsed and understood the sentence perfectly, even though this is a sufficiently grammatically complicated sentence that two _native speakers are arguing over how to punctate it_!

njr-11: It's almost like DeepSeek is claiming it's okay because the second clause becomes an independent clause due to its length and complexity rather than it actually being valid as a sentence on its own. Grammar is complex and enough time has been spent on this, so I'd say just merge it.

I'm curious what the right answer is for when this comes up again in something I'm reviewing, so maybe I'll ask our technical writers who are experts on grammar.  They review all the other documentation we write for users, but for some reason, never the specifications.

gavinking: I don't think it's an independent clause; merely a "complex clause" that occurs in a list of complex clauses. 

Remember that English is not a CFG!

njr-11: FYI - The answer I got back from our tech writer is that neither of our approaches were any good here.  The sentence is too long and should have been split in two.  If it had been a shorter sentence then the answer would have been to remove the comma before "and" because the sentence has just one subject.

gavinking: OK, fine, I put back in the colon which is actually how I had originally written this sentence.

otaviojava (approved):

PR #1031, opened 2025-03-25
javadoc should point to the First annotation as a solution to NonUniqueResultException

njr-11:
Javadoc should point to the `@First` annotation as a solution to `NonUniqueResultException` rather than `Limit.of(1)`, which relies on the user of the method to always supply a `1` value.

Also, our main Javadoc starting page that highlights available spec function should mention `@First` and include an example of it.  That page also has a side-by-side comparison of a parameter-based find method vs a Query by Method Name find method and it would be nice to include `@First` there as well now that we have an alternative to the Query by Method Name `findFirst#By` name pattern.

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -21,6 +21,7 @@ import java.util.Set;
 import jakarta.data.Limit;
 import jakarta.data.Order;
 import jakarta.data.Sort;
+import jakarta.data.exceptions.NonUniqueResultException;
 import jakarta.data.metamodel.StaticMetamodel;
 import jakarta.data.page.PageRequest;
 import jakarta.data.repository.BasicRepository;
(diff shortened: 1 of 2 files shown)


api/src/main/java/module-info.java
```diff
@@ -745,10 +747,11 @@
  *
  * <pre>
  * // Query by Method Name
- * Vehicle[] findByMakeAndModelAndYear(String makerName, String model, int year, Sort&lt;?&gt;... sorts);
+ * Vehicle[] findFirst50ByMakeAndModelAndYear(String makerName, String model, int year, Sort&lt;?&gt;... sorts);
  *
  * // parameter-based conditions
  * &#64;Find
+ * &#64;First(50)
  * Vehicle[] searchFor(String make, String model, int year, Sort&lt;?&gt;... sorts);
```
njr-11, line 755: This is the section that compares Query by Method Name with Parameter-based find

gavinking (approved):

Issue #1036, opened 2025-03-29
Reify the entity type in Attribute


gavinking:
Currently an `Attribute` only knows its name, and there is nothing which reifies the type variable `T` at runtime.

Especially since `Attributes` can now be used in compound expressions, I think we should add a `getDeclaringType()` method which returns either:

- the Java `Class<T>`, or perhaps even
- our own metamodel object, some analog of JPA's `ManagedType`.

At this stage I can't see a strong reason for needing a full-blown runtime metamodel like JPA has, so I would lean toward the first option for now.

otaviojava:
Shall we close this one here?

gavinking:
Yes, it's done, I believe.

PR #1037, opened 2025-03-29
Add Attribute.declaringType()

gavinking:
Simplest possible solution to #1036.

### DIFF
--- a/api/src/test/java/jakarta/data/metamodel/TextAttributeTest.java
+++ b/api/src/test/java/jakarta/data/metamodel/TextAttributeTest.java
@@ -26,7 +26,7 @@ import org.junit.jupiter.api.Test;
 class TextAttributeTest {
 
 
-    private final TextAttribute<String> testAttribute = () -> "testAttribute";
+    private final TextAttribute<String> testAttribute = TextAttribute.of(String.class, "testAttribute");
 
     @Test
     void shouldCreateContainsRestriction() {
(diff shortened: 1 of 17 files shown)


api/src/test/java/jakarta/data/metamodel/TextAttributeTest.java
```diff
@@ -26,7 +26,7 @@
 class TextAttributeTest {
 
 
-    private final TextAttribute<String> testAttribute = () -> "testAttribute";
+    private final TextAttribute<String> testAttribute = TextAttribute.of(String.class, "testAttribute");
```
njr-11, line 29: It is too confusing to use String.class (which is also the entity attribute type) as the entity class.  Here's a simple mock entity,

```suggestion
    static class Author {
        long id;
        String name;
        String testAttribute;
    }

    private final TextAttribute<Author> testAttribute = TextAttribute.of(Author.class, "testAttribute");
```

njr-11 (approved):
Looks good. I suggested one minor clarification in the test.

api/src/test/java/jakarta/data/metamodel/TextAttributeTest.java
gavinking: OK, but that meant fixing the whole test.

njr-11: > OK, but that meant fixing the whole test.

Sorry about that. I'm not sure how we allowed that test in in the first place with it confusing the entity attribute type with the entity type.

PR #1038, opened 2025-03-29
Note warning about statement on NoSQL databases

otaviojava:


spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -485,10 +485,42 @@ A `select` statement, with syntax given by `select_statement`, returns data to t
 
 An `update` statement, with syntax given by `update_statement`, updates each record which satisfies the restriction imposed by the `where` clause, and returns the number of updated records to the client.
 
+[WARNING]
+====
+Support for update statements may vary across NoSQL databases.
```
gavinking, line 490:
```suggestion
Support for `update` statements varies across NoSQL databases.
```

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -485,10 +485,42 @@ A `select` statement, with syntax given by `select_statement`, returns data to t
...
+In such cases, the Jakarta Data provider must throw an `UnsupportedOperationException` when a repository method declares an `int` or `long` return type for an update query that the data store cannot fulfill accurately.
+====
+
 ==== Delete statements
 
 A `delete` statement, with syntax given by `delete_statement`, deletes each record which satisfies the restriction imposed by the `where` clause, and returns the number of deleted records to the client.
 
+[WARNING]
+====
+Support for delete statements may vary significantly across NoSQL databases.
```
gavinking, line 511:
```suggestion
Support for `delete` statements varies significantly across NoSQL databases.
```

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -485,10 +485,42 @@ A `select` statement, with syntax given by `select_statement`, returns data to t
...
+
+In general:
+
+- **Key-Value databases** often only support delete operations by key. Queries with complex `where` clauses may not be supported, and returning the count of deleted records is typically not possible.
+- **Wide-Column databases** may have limited support for deletions with conditions beyond the primary key or partition key. Additionally, support for returning a count of deleted records may require the use of secondary indexes or may be entirely unsupported.
+- **Document and Graph databases** are more likely to support conditional deletes. However, the behavior and performance characteristics (especially around consistency and the returned delete count) can differ between providers.
+
+Due to the **eventual persistence** behavior that is common in many NoSQL databases, even when a delete operation appears to succeed, the underlying system may not provide an accurate or immediate count of affected records. As a result, the returned count may be **imprecise** or delayed.
```
gavinking, line 519:
```suggestion
Due to the _eventual persistence_ behavior common in many NoSQL databases, even when a delete operation appears to succeed, the underlying system may not provide an accurate or immediate count of affected records. As a result, the returned count may be imprecise or delayed.
```

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -485,10 +485,42 @@ A `select` statement, with syntax given by `select_statement`, returns data to t
...
+====
+Support for update statements may vary across NoSQL databases.
+
+In particular:
+
+- **Key-Value databases** typically only support updates by key. They do not allow conditional updates using a `where` clause, and returning the number of updated records is generally unsupported.
+- **Wide-Column databases** may restrict updates to records identified by primary or partition keys. Conditional updates or returning update counts may not be supported unless secondary indexes are used.
+- **Document and Graph databases** are more likely to support conditional updates and return the number of updated records. However, the semantics and consistency guarantees may vary depending on the provider.
+
+Due to the **eventual persistence** behavior common in many NoSQL databases, returning the number of updated records as an `int` or `long` might not be supported. Even when supported, the value may be **imprecise**, reflecting an estimate or partial result rather than an exact count.
```
gavinking, line 498:
```suggestion
Due to the _eventual persistence_ behavior common in many NoSQL databases, returning the number of updated records as an `int` or `long` might not be supported. Even when supported, the value may be imprecise, reflecting an estimate or partial result rather than an exact count.
```

gavinking (approved):

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -485,10 +485,42 @@ A `select` statement, with syntax given by `select_statement`, returns data to t
...
+- **Wide-Column databases** may have limited support for deletions with conditions beyond the primary key or partition key. Additionally, support for returning a count of deleted records may require the use of secondary indexes or may be entirely unsupported.
+- **Document and Graph databases** are more likely to support conditional deletes. However, the behavior and performance characteristics (especially around consistency and the returned delete count) can differ between providers.
+
+Due to the _eventual persistence_ behavior common in many NoSQL databases, even when a delete operation appears to succeed, the underlying system may not provide an accurate or immediate count of affected records. As a result, the returned count may be imprecise or delayed.
+
+If a NoSQL provider does not support the execution of conditional deletes or cannot determine the number of deleted records reliably, a `delete` operation that returns an `int` or `long` must throw an `UnsupportedOperationException`.
```
njr-11, line 521: The last two paragraphs seem to contradict each other. One says that "the returned count may be imprecise or delayed" and the other says that if it "cannot determine the number of deleted records reliably, a `delete` operation that returns an `int` or `long` must throw an `UnsupportedOperationException`."

I'm also not sure how much background information/generalization/prediction of what some databases might or might not do we should be writing into the Jakarta Data spec.  Would it be enough to more concisely state something like,

```suggestion
If a NoSQL database is not capable of the execution of conditional deletes or cannot determine the number of deleted records reliably for a `delete` operation that returns an `int` or `long`, the `delete` operation must throw an `UnsupportedOperationException`.
```

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -485,10 +485,42 @@ A `select` statement, with syntax given by `select_statement`, returns data to t
...
+- **Wide-Column databases** may restrict updates to records identified by primary or partition keys. Conditional updates or returning update counts may not be supported unless secondary indexes are used.
+- **Document and Graph databases** are more likely to support conditional updates and return the number of updated records. However, the semantics and consistency guarantees may vary depending on the provider.
+
+Due to the _eventual persistence_ behavior common in many NoSQL databases, returning the number of updated records as an `int` or `long` might not be supported. Even when supported, the value may be imprecise, reflecting an estimate or partial result rather than an exact count.
+
+Additionally, in databases with **append-only semantics**—such as many time-series and wide-column databases—the `update` operation may behave more like an `insert`, and repeated updates to the same record may not overwrite previous values.
+
+In such cases, the Jakarta Data provider must throw an `UnsupportedOperationException` when a repository method declares an `int` or `long` return type for an update query that the data store cannot fulfill accurately.
```
njr-11, line 502: More concise wording to align with the other suggestion:

```suggestion
If a NoSQL database is not capable of conditional updates or cannot determine the number of matching records reliably for an `update` operation that returns an `int` or `long`, the `update` operation must throw an `UnsupportedOperationException`.

Additionally, in databases with **append-only semantics**—such as many time-series and wide-column databases—the `update` operation may behave more like an `insert`, and repeated updates to the same record might not overwrite previous values.
```

njr-11 (commented):
I made a couple of suggestions to focus on when the provider raises the error and to omit the background and generalizations that don't provide any guarantees to the user or requirements on providers.  I wonder if we should have a separate section of the specification to focus on this sort of guidance to NoSQL users.

njr-11 (approved):

Issue #1040, opened 2025-03-29
Numeric promotions for NumericExpressions


gavinking:
Currently, `NumericExpression` is much less useful than it should be, because arithmetic operations can only handle identical numeric types. So, for example, you can't multiply a price by a quantity. There's two ways we could attack this:

1. we could add binary static methods to `NumericExpression` which faithfully implement the numeric promotions defined by the specification, or
2. we could add methods which do type conversions to `NumericExpression`, at minimum, `asDouble()`, `asLong()`, `asBigDecimal()`, and `asBigInteger()`.

I prefer option 2, but it does pose a requirement to support typecasting, which JDQL does not contain. Therefore, providers would need to be able to deal with things like casting `Double` to `Long`. The burden here could be reduced by saying that providers are only required to support widening conversions.

PR #1041, opened 2025-03-29
add numeric type promotions in simplest possible way

gavinking:
see #1040

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/NumericExpression.java
+++ b/api/src/main/java/jakarta/data/metamodel/NumericExpression.java
@@ -17,8 +17,12 @@
  */
 package jakarta.data.metamodel;
 
+import jakarta.data.metamodel.expression.NumericCast;
 import jakarta.data.metamodel.expression.NumericFunctionExpression;
 import jakarta.data.metamodel.expression.NumericOperatorExpression;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
 
 import static jakarta.data.metamodel.expression.NumericFunctionExpression.ABS;
 import static jakarta.data.metamodel.expression.NumericFunctionExpression.NEG;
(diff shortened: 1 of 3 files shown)


api/src/main/java/jakarta/data/metamodel/expression/NumericCastRecord.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.metamodel.expression;
+
+import jakarta.data.metamodel.NumericExpression;
+
+public record NumericCastRecord<T, N extends Number & Comparable<N>>
```
njr-11, line 22:
```suggestion
record NumericCastRecord<T, N extends Number & Comparable<N>>
```

api/src/main/java/jakarta/data/metamodel/NumericExpression.java
```diff
@@ -62,4 +66,17 @@ default NumericExpression<T,N> times(NumericExpression<T,N> other) {
     default NumericExpression<T,N> divide(NumericExpression<T,N> other) {
         return NumericOperatorExpression.of(DIVIDE, this, other);
     }
+
+    default NumericExpression<T,Long> asLong() {
```
otaviojava, line 70: Does it make sense to include asInt as well?

gavinking: It might. I'm just not sure what we think about narrowing conversions...

gavinking: I think it would be difficult to commit to that, given the heterogeneity of the datastore we're targeting.

gavinking:
Great, thanks!

PR #1045, opened 2025-03-31
Fix compile issue at _AsciiCharacter

otaviojava:


njr-11 (changes requested):
We should not change the TCK for this.  The compile failure suggests that a breaking change has been merged.

gavinking:
I guess we could add a `default` implementation to `Attribute`.

gavinking:
```java
default Class<T> declaringType() {
    throw new UnsupportedOperationException("not implemented");
}
```

njr-11:
> I guess we could add a `default` implementation to `Attribute.declaringType()`.

Yes, that's the same thing I was thinking. I created #1048 to do that.

njr-11:
Closing at not valid.  Instead we fixed the breaking change and compile failure under #1048

PR #1048, opened 2025-03-31
Fix breaking change with declaringType()

njr-11:
Fixes the compile failure in the TCK by fixing the breaking change from Attribute.declaringType

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/Attribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/Attribute.java
@@ -48,8 +48,15 @@ public interface Attribute<T> {
 
     /**
      * Obtain the Java class which declares this entity attribute.
+     * This is only guaranteed to be known if a static <code>of</code> method,
+     * such as {@link BasicAttribute#of(Class, String, Class)},
+     * was used to obtain the instance.
+     *
      * @return the declaring class
+     * @throws UnsupportedOperationException if the declaring type is not known.
      * @since 1.1
      */
-    Class<T> declaringType();
+    default Class<T> declaringType() {
+        throw new UnsupportedOperationException();
+    }
 }
--- a/api/src/main/java/jakarta/data/metamodel/Attribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/Attribute.java
@@ -48,9 +48,9 @@ public interface Attribute<T> {
 
     /**
      * Obtain the Java class which declares this entity attribute.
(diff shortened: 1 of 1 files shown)


gavinking (approved):

api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -48,8 +48,15 @@ public interface Attribute<T> {
 
     /**
      * Obtain the Java class which declares this entity attribute.
+     * This is only guaranteed to be known if a static <code>of</code> method,
+     * such as {@link BasicAttribute#of(Class, String, Class)},
+     * was used to obtain the instance.
```
gavinking, line 53:
```suggestion
     * @apiNote This is only guaranteed to be known if a static <code>of</code> method,
     *.         such as {@link BasicAttribute#of(Class, String, Class)}, was used to obtain 
     *          the instance.
```

njr-11:
Merging this right away (thanks @gavinking for reviewing so quickly!) to resolve the build failure.

njr-11:
> Merging this right away (thanks @gavinking for reviewing so quickly!) to resolve the build failure.

Oops, just saw a suggestion come up. I'll get that in first.

api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -48,8 +48,15 @@ public interface Attribute<T> {
...
+     *          such as {@link BasicAttribute#of(Class, String, Class)}, was used to obtain 
+     *          the instance.
+     *
      * @return the declaring class
+     * @throws UnsupportedOperationException if the declaring type is not known.
      * @since 1.1
      */
-    Class<T> declaringType();
+    default Class<T> declaringType() {
+        throw new UnsupportedOperationException();
```
otaviojava, line 60: Please, at least put a message on this exception.
I am not a massive fan of this approach, mainly because the interface starts breaking the Liskov Substitution Principle

gavinking: Given that the 1.1 spec can perfectly reasonably *require* that 1.1 providers *do* implement that method, and that there is no expectation that the 1.1 API jar can be used with a 1.0 provider, I don't see much of a problem here.

Is it unfortunate that code generated with 1.0 might fail when you call a method defined by 1.1? I mean, yeah, I guess. But since the solution is just "recompile your program with a 1.1 implementation", I just don't see it as a big deal at all.

PR #1054, opened 2025-04-02
add TemporalAttribute and related classes

### LINKED ISSUE #1022: LocalDate and LocalDateTime don't work with ComparableAttribute
Entity attribute types java.time.LocalDate and java.time.LocalDateTime are Comparable, but they implement `Comparable<ChronoLocalDate>` and `Comparable<ChronoLocalDateTime>` respectively rather than `Comparable<LocalDate>` and `Comparable<LocalDateTime>`.

I noticed this when attempting to use the static metamodel for `LocalDate`, where the following doesn't compile,

```
        ComparableAttribute<Author, LocalDate> birthday =
                ComparableAttribute.of(Author.class, BIRTHDAY, LocalDate.class);
```

njr-11:
Add TemporalAttribute and related static metamodel classes

Fixes #1022

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/TemporalExpression.java
+++ b/api/src/main/java/jakarta/data/metamodel/TemporalExpression.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.metamodel;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.temporal.Temporal;
+
+import jakarta.data.metamodel.expression.CurrentDate;
+import jakarta.data.metamodel.expression.CurrentDateTime;
+import jakarta.data.metamodel.expression.CurrentTime;
+
+public interface TemporalExpression<T, V extends Temporal & Comparable<? extends Temporal>>
+        extends ComparableExpression<T, V> {
+
+    // TODO How much does it matter to be able to supply the entity class?
+    // For now, I included it in one place (here) and not in the other (.of)
+    static <T> TemporalExpression<T, LocalDate> currentDate(Class<T> entityClass) {
+        return CurrentDate.of();
+    }
+
+    static <T> TemporalExpression<T, LocalDateTime> currentDateTime(Class<T> entityClass) {
+        return CurrentDateTime.of();
+    }
+
+    static <T> TemporalExpression<T, LocalTime> currentTime(Class<T> entityClass) {
+        return CurrentTime.of();
+    }
+
+    // TODO add methods to obtain the year, month, day, hour, and so forth
+    // after EXTRACT is added to the query language in some future release
+
+    // Methods for year/month/day would apply to Instant, LocalDateTime, and LocalDate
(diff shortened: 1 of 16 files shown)


api/src/main/java/jakarta/data/metamodel/TemporalExpression.java
```diff
@@ -0,0 +1,75 @@
...
+
+import jakarta.data.metamodel.expression.CurrentDate;
+import jakarta.data.metamodel.expression.CurrentDateTime;
+import jakarta.data.metamodel.expression.CurrentTime;
+
+public interface TemporalExpression<T, V extends Temporal & Comparable<? extends Temporal>>
+        extends ComparableExpression<T, V> {
+
+    // TODO How much does it matter to be able to supply the entity class?
+    // For now, I included it in one place (here) and not in the other (.of)
```
gavinking, line 33: I think in the below cases the unbound type variable is completely fine, and we don't need to make people pass in a class as a witness.

If only Java had a bottom type!

api/src/main/java/jakarta/data/metamodel/expression/CurrentDate.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package jakarta.data.metamodel.expression;
+
+import jakarta.data.metamodel.TemporalExpression;
+
+import java.time.LocalDate;
+
+public interface CurrentDate<T> extends TemporalExpression<T,LocalDate> {
+
+    static <T> CurrentDate<T> of() {
```
gavinking, line 26: I would call it `now()` since that what it's called by `LocalDate` and friends.

api/src/main/java/jakarta/data/metamodel/TemporalExpression.java
gavinking: P.S. It might seem I'm being inconsistent here because I insisted on passing the class in other places. But in the other places the type variable represents a real thing (that we now actually reify). Here it doesn't. If we could declare this operation as:

```java
static <T> TemporalExpression<Nothing, LocalDate> currentDate();
```

with `Nothing` being the bottom type, then we would do that. Unfortunately Java's type system is very incomplete.

Also, in the actual usage patterns for these operations, the type argument can always be easily and correctly inferred. We're never going to be assigning `currentDate()` to a `var`.

gavinking: Ohh, wait wait wait, this is contravariant not covariant. 

So a bit item on my todo list was to consider how all this shit is impacted by inheritance. That is, an `Attribute<Book>` is also an `Attribute<AcademicBook>`, and so some operations are currently declared with signatures that are too restrictive.

I have to think a bit hard about that, but one direction this could go in is that, e.g. `greaterThan` might get declared like this:

```java
default Restriction<T> greaterThan(Expression<? super T,V> value)
```

That is, we could allow an attribute value to be compared against a the value of an attribute declared by a supertype. [Unfortunately, there's an asymmetry here which is inelegant.]

And then we could declare `currentDate()` like this:

```java
static TemporalExpression<Object, LocalDate> currentDate();
```

Letting you write `_Book.publicationDate.greaterThan( currentDate());`.

gavinking: >  and so some operations are currently declared with signatures that are too restrictive.

To clarify: I'm talking here about all the "commented-out" operations.

njr-11:
Rebased onto latest and resolved a soft merge conflict with the removal of `.attribute()`

api/src/main/java/jakarta/data/metamodel/TemporalExpression.java
```diff
@@ -0,0 +1,52 @@
...
+        return CurrentDate.now();
+    }
+
+    static <T> TemporalExpression<T, LocalDateTime> currentDateTime() {
+        return CurrentDateTime.now();
+    }
+
+    static <T> TemporalExpression<T, LocalTime> currentTime() {
+        return CurrentTime.now();
+    }
```
gavinking, line 42:
```suggestion
    static TemporalExpression<Object, LocalDate> localDate() {
        return CurrentDate.now();
    }

    static TemporalExpression<Object, LocalDateTime> localDateTime() {
        return CurrentDateTime.now();
    }

    static TemporalExpression<Object, LocalTime> localTime() {
        return CurrentTime.now();
    }
```

njr-11:
I merged Gavin's suggestion/review comment to rename the currentDate methods to localDate, after which I rebased onto latest and addressed merge conflicts. This PR is ready for review again.

gavinking (approved):
LGTM

PR #1059, opened 2025-04-04
move *Expression classes into the expression package with the other *Expression classes

njr-11:
We have a `jakarta.data.metamodel.expression` package with some of the *Expression classes and interfaces there, but others in the base `jakarta.data.metamodel` package.  This PR moves them all over to `jakarta.data.metamodel.expression` so they will be in the same place.  Also, there is a `jakarta.data.metamodel.path` package that has most things for Path, except for the `Path` class, so I have moved that as well under this PR.

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
@@ -17,6 +17,7 @@
  */
 package jakarta.data.metamodel;
 
+import jakarta.data.metamodel.expression.Expression;
 import jakarta.data.metamodel.impl.BasicAttributeRecord;
 
 import java.util.Objects;
--- a/api/src/main/java/jakarta/data/metamodel/ComparableAttribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/ComparableAttribute.java
@@ -17,6 +17,7 @@
  */
 package jakarta.data.metamodel;
 
+import jakarta.data.metamodel.expression.ComparableExpression;
 import jakarta.data.metamodel.impl.ComparableAttributeRecord;
 
 import java.util.Objects;
--- a/api/src/main/java/jakarta/data/metamodel/NavigableAttribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/NavigableAttribute.java
@@ -17,6 +17,7 @@
  */
 package jakarta.data.metamodel;
 
+import jakarta.data.metamodel.expression.NavigableExpression;
 import jakarta.data.metamodel.impl.NavigableAttributeRecord;
 
 import java.util.Objects;
--- a/api/src/main/java/jakarta/data/metamodel/NumericAttribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/NumericAttribute.java
@@ -17,6 +17,7 @@
  */
 package jakarta.data.metamodel;
 
+import jakarta.data.metamodel.expression.NumericExpression;
 import jakarta.data.metamodel.impl.NumericAttributeRecord;
 
 import java.util.Objects;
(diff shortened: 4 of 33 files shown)


gavinking:
The previous organization was done so that APIs declared in the outer package would not have dependencies to APIs in the nested package. With this change, you have `TextAttribute` in the outer package extending `TextEpression` in the nested package.

njr-11:
This PR has merge conflicts and will need to be set aside for now because it collides with too many other places where we have changes proposed.  Thanks for pointing out the reasoning why the packages were arranged this way:

> The previous organization was done so that APIs declared in the outer package would not have dependencies to APIs in the nested package.

However, it doesn't appear this is actually being followed currently.  For example,

 - `NumericExpression` in `jakarta.data.metamodel` currently depends on `jakarta.data.metamodel.expression` for `NumericCast`, `NumericOperatorExpression`, and `NumericFunctionExpression`
 - `TextExpression` in `jakarta.data.metamodel` currently depends on `jakarta.data.metamodel.expression` for `TextFunctionExpression` and `NumericFunctionExpression`
 - `NavigableExpression` in jakarta.data.metamodel` currently depends on `jakarta.data.metamodel.path` for `ComparablePath`, `NavigablePath`, `NumericPath`, and `TextPath`.

gavinking:
>> The previous organization was done so that APIs declared in the outer package would not have dependencies to APIs in the nested package. 

> However, it doesn't appear this is actually being followed currently

That's not what I'm saying. I said that the **APIs** of those classes don't expose the nested packages. They do use the nested packages _in their internal unexposed implementation_, but the client doesn't see that.

njr-11:
> **APIs** of those classes don't expose the nested packages. They do use the nested packages _in their internal unexposed implementation_, but the client doesn't see that.

In addition to the above, this PR has become very out of sync with main with all of the other stuff going in, so I'll start over on it.  One part of it that did meet the above was moving Path to the the path package, so I'll start with that.  We do need to figure out something better for the classes under the expression package because we currently have some Expression classes there and others under metamodel which is confusing.  At worst, all of it could move to the same package, but I'll have to think more about whether there could be other ways to logically split it up that would make sense.

gavinking:
> we currently have some Expression classes there and others under metamodel

Yes, I've been meaning to say that all this stuff (`expression`, `path`, `restrict`) should be moved up out of the `metamodel` namespace. There should be a new top-level package for `Expression`, `ComparableExpression`, etc, and the current packages `metamodel.expression`, `metamodel.path`, `metamodel.restrict` should be moved to be children of the new package.

gavinking:
> One part of it that did meet the above was moving `Path` to the the path package

Yeah, looks like `Path` and `NavigablePath` are not exposed by expressions, and so they can be moved to the sub-package.

njr-11:
> Yes, I've been meaning to say that all this stuff (`expression`, `path`, `restrict`) should be moved up out of the `metamodel` namespace. There should be a new top-level package for `Expression`, `ComparableExpression`, etc, and the current packages `metamodel.expression`, `metamodel.path`, `metamodel.restrict` should be moved to be children of the new package.

That sounds like a good approach. I added a commit to mostly do that. Where I ran into problems was that API in restrict relies on expression, so I made them peers rather than restrict a subpackage of expression.  I was also left with what to call the subpackage of expression in which to place some more specific subtypes of expressions.  I noticed more than half of them were for literals, so I created a `expression.literal` package for those, and then was left with the temporal functions and cast. I wasn't sure what to call a package for those and have at least temporarily given it the name `expression.function`.  Otherwise everything seemed to split up nicely.  Let me know if this works.

gavinking (approved):
Looks good, only one comment: previously `NumericOperatorExpression` and `FunctionExpression` were not in the "main package" next to `Expression` and friends. I think I liked it like that because they are not really "API".

WDYT?

njr-11:
> Looks good, only one comment: previously `NumericOperatorExpression` and `FunctionExpression` were not in the "main package" next to `Expression` and friends. I think I liked it like that because they are not really "API".
> 
> WDYT?

I suppose it would make sense to put them under the `expression.function` package that I just created. The package name actually works well for that. I added a commit to do that.

PR #1061, opened 2025-04-05
Restrictions that compare against Expressions

njr-11:
This PR tries out splitting BasicRestriction into one that compares against values (as it currently does) named ValueRestriction and another that compares against Expressions, named ExpressionRestriction.  To get this to work, I had to get around the requirement for the constraint values of LessThan, GreaterThan, and so on to be Comparable.  ComparableExpression is not currently Comparable, but to get this to work, I made it be, and that mostly worked. This might not be where we want to end up long term, but it could a temporary step on the way there. With this, I'm now able to write several unit tests that compare against Expressions and successfully preserve all of the information.


njr-11:
I finished off the updates that are needed to make `ComparableExpression` be `Comparable` so that these expressions can be supplied as the constraint value to LessThan/GreaterThan, and so forth.  This didn't turn out as awkward as I thought and could be a valid solution for being able to have Restrictions that compare against other Attributes and Expressions, for example,
```
_Book.copyrightYear.greaterThan(_Book.firstPublishedYear);

_Book.currentPrice.lessThanEqual(_Book.discountPrice.times(1.1));
```

The alternatives that I thought of so far are:

- Remove the `extends Comparable<?>` from LessThan, GreaterThan, and other constraint subtypes
- Duplicate all of the LessThan/GreaterThan, ... constraint subtypes to allow a variant that builds around Expression instead of a Comparable value.  I really don't like this option.
- Get rid of Constraint and go back to separate Operator and value.

Maybe there are other solutions that I haven't thought of yet.

njr-11:
Rebased and pushed updates to resolve various soft merge conflicts with #1060
This is ready for review again.

njr-11:
This is now updated per the review comment from @gavinking to take the approach of having separate methods on constraint subtypes for values vs expressions rather than reusing a single method that accepts both.

gavinking:
> Is anyone planning on reviewing this PR? 

There's a lot there but I don't see anything wrong. Certainly nothing that needs fixing now.

I'm not, like, a superfan of this simply because it's, well, again, _a lot_.

But on the other hand I started us down the `Expression`s path and this is clearly where that path logically leads and that was obvious right from the start. And this useful functionality, for sure, no question. I'm pretty sure users are going to like it quite a lot more than I do :-)

otaviojava:
@njr-11, by any chance, can you find a strategy to break this PR into smaller ones?

gavinking:
>  However, some changes like this one don't really allow for that 

Yeah it seems like a pretty integral change to me.

gavinking (approved):

PR #1067, opened 2025-04-16
squash a bunch of warnings in IntelliJ

gavinking:
1. unnecessary @SuppressWarnings
2. assertEquals instead of more specialized asserts
3. method references

### DIFF
--- a/api/src/test/java/jakarta/data/metamodel/AttributeTest.java
+++ b/api/src/test/java/jakarta/data/metamodel/AttributeTest.java
@@ -47,7 +47,6 @@ class AttributeTest {
             BasicAttribute.of(Person.class, "testAttribute", String.class);
     @Test
     void shouldCreateEqualToRestriction() {
-        @SuppressWarnings("unchecked")
         BasicRestriction<Person, String> restriction =
                 (BasicRestriction<Person, String>) testAttribute.equalTo("testValue");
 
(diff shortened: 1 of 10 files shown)


api/src/test/java/jakarta/data/metamodel/AttributeTest.java
```diff
@@ -47,7 +47,6 @@ static class Person {
             BasicAttribute.of(Person.class, "testAttribute", String.class);
     @Test
     void shouldCreateEqualToRestriction() {
-        @SuppressWarnings("unchecked")
```
njr-11, line 50: I tried removing `SuppressWarnings` here and it caused Ecilpse to show the warning,
```
Type safety: Unchecked cast from Restriction<AttributeTest.Person> to BasicRestriction<AttributeTest.Person,String>
```
It is unfortunate that default settings in code editors don't agree with each other here.

I also noticed that after removing `SuppressWarnings` here, the following shows up in build output:
```
.../data/api/src/test/java/jakarta/data/metamodel/AttributeTest.java: /Users/njr/lgit/data/api/src/test/java/jakarta/data/metamodel/AttributeTest.java uses unchecked or unsafe operations.
.../data/api/src/test/java/jakarta/data/metamodel/AttributeTest.java: Recompile with -Xlint:unchecked for details.
```
which is under,
```
[INFO] --- compiler:3.14.0:testCompile (default-testCompile) @ jakarta.data-api ---
```

gavinking: Oh wow that is so weird, IntelliJ is totally complaining about those `@SuppressWarnings` being unnecessary, but now I look more closely, they're totally necessary. That's an unchecked cast for sure. WTF

gavinking:
Closing in favor of #1090.

PR #1070, opened 2025-04-23
Enhance and include new test scenarions on the BasicRestrictionRecordTest

otaviojava:
A great work has been done on this point.

I included the `DisplayName` annotation to each scenario, including constraint validation, plus I've removed the warning annotation.

### DIFF
--- a/api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
+++ b/api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
@@ -167,4 +167,29 @@ class BasicRestrictionRecordTest {
         });
     }
 
+    @DisplayName("should create NotNull constraint correctly")
+    @Test
+    void shouldCreateNotNullRestriction() {
+        BasicRestriction<Book, String> restriction = (BasicRestriction<Book, String>) _Book.title.isNull().negate();
+
+        SoftAssertions.assertSoftly(soft -> {
+            soft.assertThat(restriction.expression()).isEqualTo(_Book.title);
+            soft.assertThat(restriction.constraint()).isEqualTo(NotNull.instance());
+            soft.assertThat(restriction.constraint()).isInstanceOf(NotNull.class);
+        });
+    }
+
+    @DisplayName("should create Between constraint correctly")
+    @Test
+    void shouldCreateBetweenRestriction() {
+        BasicRestriction<Book, Integer> restriction = (BasicRestriction<Book, Integer>) _Book.numChapters.between(5, 15);
+
+        SoftAssertions.assertSoftly(soft -> {
+            soft.assertThat(restriction.expression()).isEqualTo(_Book.numChapters);
+            soft.assertThat(restriction.constraint()).isEqualTo(Between.bounds(5, 15));
+            soft.assertThat(restriction.constraint()).isInstanceOf(Between.class);
+        });
+    }
+
+
 }
--- a/api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
+++ b/api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
@@ -129,4 +129,19 @@ class BasicRestrictionRecordTest {
                 .isInstanceOf(NullPointerException.class)
                 .hasMessage("Value must not be null");
     }
+
+    @Test
+    void shouldCreateGreaterThanOrEqualRestriction() {
+        @SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
```diff
@@ -23,109 +23,235 @@
...
         });
     }
 
     @Test
+    @DisplayName("should create NotEqualTo from 'notEqualTo' method directly")
     void shouldSupportNegatedRestrictionUsingDefaultConstructor() {
-        @SuppressWarnings("unchecked")
-        BasicRestriction<Book, String> negatedRestriction =
-                (BasicRestriction<Book, String>) _Book.author.notEqualTo("Unknown");
+        BasicRestriction<Book, String> restriction = (BasicRestriction<Book, String>) _Book.author.notEqualTo("Unknown");
```
gavinking, line 111: Maybe we should just use `var` here, so we can stop fighting over whether this line is too long :)

gavinking (commented):
LGTM

api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
```diff
@@ -23,109 +23,235 @@
...
 import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
 
 
 class BasicRestrictionRecordTest {
 
+
     @Test
+    @DisplayName("should create an EqualTo restriction by default when using 'equalTo' without negation")
     void shouldCreateBasicRestrictionWithDefaultNegation() {
-        @SuppressWarnings("unchecked")
```
njr-11, line 35: Getting rid of `SuppressWarnings` causes complaints like the following to show up in the build output. I would say keep it and don't remove it.  Also, new tests that you added ought to include it where needed.
```
[...]/data/api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java: [...]/data/api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java uses unchecked or unsafe operations.
[...]/data/api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java: Recompile with -Xlint:unchecked for details.
```

api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
njr-11: In general, I would say the change here to merge 2 lines into a single 118 character line that developers with less acute eyesight will need to scroll right to read doesn't add any value.

njr-11 (commented):
Thanks for adding more tests here. I left a comment about the removal and non-usage of `SuppressWarnings`. Compare your build output with/without that change.

api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
gavinking: > developers with less acute eyesight

I resemble that remark.

otaviojava:
@njr-11 I did what @gavinking said about using `var`, the line becomes shorter, that is fine to keep in on line, please, let me know what you think about it.

api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
```diff
@@ -23,109 +23,236 @@
...
 import org.assertj.core.api.SoftAssertions;
+import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
+import java.util.Set;
+
 import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
 
 
+@SuppressWarnings("unchecked")
```
otaviojava, line 34: @njr-11 I put it on the whole class, could you check it now?

otaviojava:
> Thanks for adding more tests here. I left a comment about the removal and non-usage of `SuppressWarnings`. Compare your build output with/without that change.

You're welcome. 

I will focus on increasing test coverage. Please check the style, as my goal is to use this style, following our guide, to improve to at least 90%.

api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
njr-11: That works.

njr-11 (approved):

PR #1072, opened 2025-04-24
Can we use vargargs instead of Set?

otaviojava:
Can we use varags here instead of `Set`?

This is a small change, but it simplifies the goal, which is to move the input without creating a Set instance.

```java
   var restriction =  _Book.author.in("Alice", "Bob");
```

instead of:

```java
   var restriction =  _Book.author.in(Set.of("Alice", "Bob"));
```

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/Expression.java
+++ b/api/src/main/java/jakarta/data/metamodel/Expression.java
@@ -26,7 +26,6 @@ import jakarta.data.metamodel.constraint.Null;
 import jakarta.data.metamodel.restrict.BasicRestriction;
 import jakarta.data.metamodel.restrict.Restriction;
 
-import java.util.Set;
 
 public interface Expression<T,V> {
 
--- a/api/src/main/java/jakarta/data/metamodel/Expression.java
+++ b/api/src/main/java/jakarta/data/metamodel/Expression.java
@@ -35,6 +35,14 @@ public interface Expression<T,V> {
     }
 
     default Restriction<T> in(Set<V> values) {
+        if (values == null || values.isEmpty()){
+            throw new IllegalArgumentException("values are required");
+        }
+
+        return BasicRestriction.of(this, In.values(values));
+    }
+
+    default Restriction<T> in(V... values) {
         if (values == null || values.isEmpty())
             throw new IllegalArgumentException("values are required");
 
--- a/api/src/main/java/jakarta/data/metamodel/Expression.java
+++ b/api/src/main/java/jakarta/data/metamodel/Expression.java
@@ -34,18 +34,10 @@ public interface Expression<T,V> {
         return BasicRestriction.of(this, Constraint.equalTo(value));
     }
 
-    default Restriction<T> in(Set<V> values) {
-        if (values == null || values.isEmpty()){
+    default Restriction<T> in(V... values) {
+        if (values == null || values.length == 0){
             throw new IllegalArgumentException("values are required");
         }
-
-        return BasicRestriction.of(this, In.values(values));
-    }
-
-    default Restriction<T> in(V... values) {
-        if (values == null || values.isEmpty())
-            throw new IllegalArgumentException("values are required");
-
         return BasicRestriction.of(this, In.values(values));
     }
 
@@ -57,9 +49,10 @@ public interface Expression<T,V> {
         return BasicRestriction.of(this, NotEqualTo.value(value));
     }
 
-    default Restriction<T> notIn(Set<V> values) {
-        if (values == null || values.isEmpty())
+    default Restriction<T> notIn(V... values) {
+        if (values == null || values.length == 0){
             throw new IllegalArgumentException("values are required");
+        }
 
         return BasicRestriction.of(this, NotIn.values(values));
     }
--- a/api/src/test/java/jakarta/data/metamodel/AttributeTest.java
+++ b/api/src/test/java/jakarta/data/metamodel/AttributeTest.java
@@ -77,7 +77,7 @@ class AttributeTest {
     void shouldCreateInRestriction() {
         @SuppressWarnings("unchecked")
         BasicRestriction<Person, String> restriction =
-                (BasicRestriction<Person, String>) testAttribute.in(Set.of("value1", "value2"));
+                (BasicRestriction<Person, String>) testAttribute.in("value1", "value2");
 
         SoftAssertions.assertSoftly(soft -> {
             soft.assertThat(restriction).isInstanceOf(BasicRestriction.class);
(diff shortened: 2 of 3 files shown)


gavinking:
I usually prefer to accept `List` or `Set` because it's easy to get a list or set from an array, but it's an absolute pain to get an array from a `List` or `Set` due to the fact that Java doesn't reify type arguments. But in this case we can just overload `in()` to accept both `Set` and varargs, just like we do on `Constraint`.

otaviojava:
> I usually prefer to accept `List` or `Set` because it's easy to get a list or set from an array, but it's an absolute pain to get an array from a `List` or `Set` due to the fact that Java doesn't reify type arguments. But in this case we can just overload `in()` to accept both `Set` and varargs, just like we do on `Constraint`.

I've included both options.

gavinking:
> I've included both options.

Oh wait, no, my bad. The reason to not have a non-`static`:

```java
default Restriction<T> in(V... values) {
    return BasicRestriction.of(this, In.values(values));
}
```

is that you get the world's stupidest compiler warning grrrrr:

> Possible heap pollution from parameterized vararg type

There was a reason we left it out.

otaviojava:
At the end, the @njr-11, I've included this option, I rebased it, and I've included more test scenarios.

gavinking:
> There was a reason we left it out.

Well, perhaps it's not a real problem. It's a warning in our code, not in user code. I seem to recall that Java *used* to also report a warning on the caller side, but that appears to no longer be the case.

gavinking:
@otaviojava See https://github.com/jakartaee/data/pull/1075, which sorts out the problems, at least for me.

api/src/main/java/jakarta/data/metamodel/Expression.java
```diff
@@ -46,13 +46,11 @@ default Restriction<T> in(Set<V> values) {
         return BasicRestriction.of(this, In.values(values));
     }
 
-    default Restriction<T> in(
-            @SuppressWarnings("unchecked") V... values) {
+    default Restriction<T> in(@SuppressWarnings("unchecked") V... values) {
```
gavinking, line 49: Do these warning suppressions actually work for you, @otaviojava? They don't work for me, at least not when they're placed on the parameter.

otaviojava: It did not work for me either.

otaviojava: Indeed, my IDE recommends that I remove those annotations.

gavinking: OK, thx.

otaviojava:
@njr-11 I've updated the title, after your code, it became more test, could you review to move it forward, please?

api/src/test/java/jakarta/data/metamodel/AttributeTest.java
```diff
@@ -89,7 +89,7 @@ void shouldCreateInRestriction() {
 
     @Test
     void shouldThrowExceptionForEmptyInRestriction() {
-        assertThatThrownBy(() -> testAttribute.in(Set.of()))
+        assertThatThrownBy(() -> testAttribute.in((Set<String>) null))
```
njr-11, line 92: A test that says it covers empty should actually pass in empty, not null,
```suggestion
        assertThatThrownBy(() -> testAttribute.in(Set.of()))
```

api/src/test/java/jakarta/data/metamodel/AttributeTest.java
```diff
@@ -110,7 +110,7 @@ void shouldCreateNotInRestriction() {
 
     @Test
     void shouldThrowExceptionForEmptyNotInRestriction() {
-        assertThatThrownBy(() -> testAttribute.notIn(Set.of()))
+        assertThatThrownBy(() -> testAttribute.notIn((Set<String>) null))
```
njr-11, line 113: A test that says it covers empty should actually pass in empty, not null,
```suggestion
        assertThatThrownBy(() -> testAttribute.notIn(Set.of()))
```

api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
```diff
@@ -209,6 +210,30 @@ void shouldCreateInRestriction() {
...
+            soft.assertThat(restriction.expression()).isEqualTo(_Book.author);
+            soft.assertThat(restriction.constraint()).isEqualTo(NotIn.values("Alice", "Bob"));
+            soft.assertThat(restriction.constraint()).isInstanceOf(NotIn.class);
+        });
+    }
+
+    @DisplayName("should create In, using Set, constraint correctly")
+    @Test
+    void shouldCreateInAsSetRestriction() {
+        var restriction = (BasicRestriction<Book, String>) _Book.author.in(List.of("Alice", "Bob"));
```
njr-11, line 225: A test that explicitly says it is using `Set` should actually use `Set`,
```suggestion
    @DisplayName("should create In, using Set, constraint correctly")
    @Test
    void shouldCreateInAsSetRestriction() {
        var restriction = (BasicRestriction<Book, String>) _Book.author.in(Set.of("Alice", "Bob"));
```

api/src/test/java/jakarta/data/metamodel/restrict/BasicRestrictionRecordTest.java
```diff
@@ -209,6 +210,30 @@ void shouldCreateInRestriction() {
...
+            soft.assertThat(restriction.expression()).isEqualTo(_Book.author);
+            soft.assertThat(restriction.constraint()).isEqualTo(In.values("Alice", "Bob"));
+            soft.assertThat(restriction.constraint()).isInstanceOf(In.class);
+        });
+    }
+
+    @DisplayName("should create NotIn, using Set, constraint correctly")
+    @Test
+    void shouldCreateNotInAsSetRestriction() {
         var restriction = (BasicRestriction<Book, String>) _Book.author.in(List.of("Alice", "Bob")).negate();
```
njr-11, line 237: A test that explicitly says it is using `Set` should actually use `Set`,
```suggestion
    @DisplayName("should create NotIn, using Set, constraint correctly")
    @Test
    void shouldCreateNotInAsSetRestriction() {
        var restriction = (BasicRestriction<Book, String>) _Book.author.in(Set.of("Alice", "Bob")).negate();
```

njr-11 (commented):
Thanks for adding test coverage.  I added a few suggestions to make the test logic match what the test says it is doing in a few places.

otaviojava:
> Thanks for adding test coverage. I added a few suggestions to make the test logic match what the test says it is doing in a few places.

Thanks @njr-11 done.

PR #1073, opened 2025-04-24
Enhance the tests on CompositeRestrictionRecordTest

otaviojava:


api/src/main/java/jakarta/data/metamodel/restrict/CompositeRestrictionRecord.java
```diff
@@ -38,10 +40,12 @@ record CompositeRestrictionRecord<T>(
             throw new IllegalArgumentException(
                     "Cannot create a composite restriction without any restrictions to combine.");
         }
+        Objects.requireNonNull(type, "type must not be null");
```
otaviojava, line 43: I've included the validation to avoid null on type.

api/src/main/java/jakarta/data/metamodel/restrict/CompositeRestrictionRecord.java
```diff
@@ -38,10 +40,12 @@ record CompositeRestrictionRecord<T>(
...
+        restrictions.forEach(r -> Objects.requireNonNull(r, "Restriction must not be null"));
     }
 
     CompositeRestrictionRecord(Type type, List<Restriction<T>> restrictions) {
-        this(type, restrictions, false);
+        this(type, new ArrayList<>(restrictions), false);
```
otaviojava, line 48: I did a copy of the List to avoid modification later as reference.

api/src/main/java/jakarta/data/metamodel/restrict/CompositeRestrictionRecord.java
```diff
@@ -38,10 +40,12 @@ record CompositeRestrictionRecord<T>(
             throw new IllegalArgumentException(
                     "Cannot create a composite restriction without any restrictions to combine.");
         }
+        Objects.requireNonNull(type, "type must not be null");
+        restrictions.forEach(r -> Objects.requireNonNull(r, "Restriction must not be null"));
```
otaviojava, line 44: I believe that the restrictions `List` should not have a null item.

api/src/main/java/jakarta/data/metamodel/restrict/CompositeRestrictionRecord.java
gavinking: Good catch. But in fact this is not enough. We need to also call `unmodifiableList()` everywhere the record is instantiated..

gavinking: The lack of immutable collections in Java is a massive hole in the whole design of `record` types :-(

gavinking: Actually, @otaviojava, if you check Nathan's latest code, none of this is necessary. All code paths which instantiate `CompositeRestrictionRecord` do in fact call `List.of()`.

gavinking: (Though there were related problems with `in()`, which I have fixed in #1075.)

api/src/main/java/jakarta/data/metamodel/restrict/CompositeRestrictionRecord.java
gavinking: Agreed.

api/src/main/java/jakarta/data/metamodel/restrict/CompositeRestrictionRecord.java
otaviojava: Hum, let me rebase this PR as well.

otaviojava: You were right, removing it.

otaviojava:
@njr-11 WDYT?

api/src/test/java/jakarta/data/metamodel/restrict/CompositeRestrictionRecordTest.java
```diff
@@ -58,144 +59,231 @@ static class Author {
...
+
+        CompositeRestrictionRecord<Author> duplicate = new CompositeRestrictionRecord<>(
+                CompositeRestriction.Type.ALL, List.of(titleRestriction, nameRestriction));
 
         SoftAssertions.assertSoftly(soft -> {
             soft.assertThat(composite.type()).isEqualTo(CompositeRestriction.Type.ALL);
-            soft.assertThat(composite.restrictions()).containsExactly(restriction1, restriction2);
+            soft.assertThat(composite.type()).isNotNull();
             soft.assertThat(composite.isNegated()).isFalse();
+            soft.assertThat(composite.isNegated()).isInstanceOf(Boolean.class);
```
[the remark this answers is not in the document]
otaviojava: you are right, my bad, removing this one.

Thanks

api/src/test/java/jakarta/data/metamodel/restrict/CompositeRestrictionRecordTest.java
```diff
@@ -58,144 +59,231 @@ static class Author {
...
-                (CompositeRestriction<Author>) Restrict.all(ageLessThan50, nameStartsWithDuke);
-        CompositeRestriction<Author> allNegated = all.negate();
-        CompositeRestriction<Author> notAll =
-                (CompositeRestriction<Author>) Restrict.not(all);
+        var ageLessThan50 = _Author.age.lessThan(50);
+        var nameStartsWithDuke = _Author.name.startsWith("Duke ");
 
-        SoftAssertions.assertSoftly(soft -> {
-            soft.assertThat(all.isNegated()).isEqualTo(false);
+        CompositeRestriction<Author> all = (CompositeRestriction<Author>) Restrict.all(ageLessThan50, nameStartsWithDuke);
```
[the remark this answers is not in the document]
otaviojava: I can update the code style to break the build after this size; at the moment, it will break after `180`. It is up to us to define it.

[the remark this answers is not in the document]
otaviojava: anyway, I used var to decrease the line size, could you check it and let me know what you think?

otaviojava:
> Again, thanks for adding more tests. Some parts of this seemed to be just reformatting code and renaming variables, so hopefully I was able to spot all of the substantive changes for reviewing. I'm not a fan of combining multiple lines into really long single lines and if I ever need to edit the code will likely split it back up in order to be able to read it wherever that is done excessively.

we can define the right size and break the compile if it goes over this size.

We need to modify this line here: https://github.com/jakartaee/data/blob/main/api/src/main/resources/checkstyle.xml#L32

gavinking:
> I'm not a fan of combining multiple lines into really long single lines 

I agree. For context I had perfect eyesight when I was young but I don't anymore and so small fonts disagree with me. And I always work on a 14-inch laptop screen.

I understand that some other folks are comfortable with small fonts on their huge desktop monitors, but for me scrolling right is just a sucky experience, and anyway code with "ragged" formatting just looks aesthetically very ugly to my eyes.

Short line length is where it's at, IMO. Break 'em at 80 chars.

njr-11 (approved):

otaviojava:
> > I'm not a fan of combining multiple lines into really long single lines
> 
> I agree. For context I had perfect eyesight when I was young but I don't anymore and so small fonts disagree with me. And I always work on a 14-inch laptop screen.
> 
> I understand that some other folks are comfortable with small fonts on their huge desktop monitors, but for me scrolling right is just a sucky experience, and anyway code with "ragged" formatting just looks aesthetically very ugly to my eyes.
> 
> Short line length is where it's at, IMO. Break 'em at 80 chars.

@njr-11 do you agree with 80 as well?

otaviojava:
I will merge because the line size is being addressed in another PR.

njr-11:
> @njr-11 do you agree with 80 as well?

Yes

otaviojava:
> > @njr-11 do you agree with 80 as well?
> 
> Yes

Ok, I have one more PR today: https://github.com/jakartaee/data/pull/1078

I will create this 80 size validation as next PR.

otaviojava:
I am working on the line with size 80 as well.

Once it breaks on some points, it might take more time than I expected, but I will finish at most by tomorrow morning.

PR #1074, opened 2025-04-24
Create test scenarions to the impl package

otaviojava:
I know this package is deprecated, but I would still love to have some tests to ensure the behavior.

### DIFF
--- a/api/src/test/java/jakarta/data/metamodel/impl/AttributeRecordTest.java
+++ b/api/src/test/java/jakarta/data/metamodel/impl/AttributeRecordTest.java
@@ -65,19 +65,11 @@ class AttributeRecordTest {
     @Test
     @DisplayName("should not be equal if names differ")
     void shouldNotEqualDifferentNames() {
-        AttributeRecord<SimpleEntity> id = new AttributeRecord<>("id");
-        AttributeRecord<SimpleEntity> desc = new AttributeRecord<>("description");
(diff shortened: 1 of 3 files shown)


api/src/test/java/jakarta/data/metamodel/impl/AttributeRecordTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Test
+    @DisplayName("should create attribute record with name and validate interface type")
+    void shouldCreateAttributeRecordWithName() {
+        var attribute = new AttributeRecord<>("id");
+
+        SoftAssertions.assertSoftly(soft -> {
+            soft.assertThat(attribute.name()).isEqualTo("id");
+            soft.assertThat(attribute).isInstanceOf(Attribute.class);
+            soft.assertThat(attribute.name()).isNotBlank();
```
njr-11, line 49: A few of these assertions are somewhat amusing, like checking that a value is not blank after having already checked that it was equal to a particular non-blank String value 2 lines above.  But everything looks fine. Approving.

njr-11 (approved):

api/src/test/java/jakarta/data/metamodel/impl/AttributeRecordTest.java
otaviojava: Ok, I will remove those and merge it.

PR #1075, opened 2025-04-24
sort out issues with in()

gavinking:
I think that using `Set` here was okay when we only had literal values. But now that we also have `Expression`s, I think it should be a `List`. Otherwise we commit to always having well-defined equality for expressions, which I would not want to do. Also. the generated SQL would be (in principle) non-deterministic which is (in principle) a problem for prepared statement caching.

I've also added appropriate calls to `unmodifiableList()`.

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/Expression.java
+++ b/api/src/main/java/jakarta/data/metamodel/Expression.java
@@ -27,7 +27,7 @@ import jakarta.data.metamodel.constraint.Null;
 import jakarta.data.metamodel.restrict.BasicRestriction;
 import jakarta.data.metamodel.restrict.Restriction;
 
-import java.util.Set;
+import java.util.Collection;
 
 public interface Expression<T,V> {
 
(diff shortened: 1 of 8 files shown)


api/src/main/java/jakarta/data/metamodel/Expression.java
```diff
@@ -39,20 +39,20 @@ default Restriction<T> equalTo(Expression<? super T,V> expression) {
...
-            @SuppressWarnings("unchecked") V... values) {
+    @SuppressWarnings("unchecked")
+    default Restriction<T> in(V... values) {
         return BasicRestriction.of(this, In.values(values));
     }
 
-    default Restriction<T> in(
-            @SuppressWarnings("unchecked") Expression<? super T,V>... expressions) {
+    @SuppressWarnings("unchecked")
+    default Restriction<T> in(Expression<? super T,V>... expressions) {
```
gavinking, line 55: This `@SuppressWarnings` annotation was in the wrong place, at least according to _my_ version of `javac` and IntelliJ. Apparently there's been some movement here, because `@SuppressWarnings` didn't used to be enough to get rid of this useless warning. I'm very happy they're _finally_ provided a way to make it go away!

otaviojava:
@gavinking with @njr-11, a big PR. Could you merge/rebase your branch to see the impact?

otaviojava (approved):

api/src/main/java/jakarta/data/metamodel/Expression.java
njr-11: As stated in the other PR, moving it to the method gets rid of complaints in the build output, so this is a good update.

njr-11 (approved):

PR #1079, opened 2025-04-24
Follow consistent style for messages

njr-11:
While reviewing code, I noticed the way in which our error messages is written is all over the place.

I did a search and found that Objects.requireNonNull appears 87 times in our API. Here are some of the different styles that I spotted when browsing through the search results:

- value must not be null
- Value must not be null.
- value is required
- Value is required.
- no message at all in a few cases

For messages that we supply to IllegalArgumentException that we construct for the metamodel, we do a better job of being consistent. In a few cases the message is all lower case with no punctuation, like "values are required", but in most cases it is a sentence beginning with a capital letter and ending with a period, such as "Values must not be empty."

This PR aims to make all of the messages consistent.

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
@@ -42,9 +42,9 @@ public interface BasicAttribute<T,V> extends Attribute<T>, Expression<T,V> {
     static <T,V> BasicAttribute<T,V> of(Class<T> entityClass,
                                         String name,
                                         Class<V> attributeType) {
-        Objects.requireNonNull(entityClass, "Entity class is required.");
-        Objects.requireNonNull(name, "Entity attribute name is required.");
-        Objects.requireNonNull(attributeType, "Entity attribute type is required.");
+        Objects.requireNonNull(entityClass, "Entity class is required");
+        Objects.requireNonNull(name, "Entity attribute name is required");
+        Objects.requireNonNull(attributeType, "Entity attribute type is required");
 
         return new BasicAttributeRecord<>(entityClass, name);
     }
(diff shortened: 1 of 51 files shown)


njr-11:
> Good idea, but one nitpick: I don't think we need to end each error message with a period.

Yeah, I wasn't sure which of the various styles to pick.  It is updated to remove the ending `.` now.

gavinking (approved):

api/src/main/java/jakarta/data/metamodel/BasicAttribute.java
```diff
@@ -42,9 +42,9 @@ public interface BasicAttribute<T,V> extends Attribute<T>, Expression<T,V> {
     static <T,V> BasicAttribute<T,V> of(Class<T> entityClass,
                                         String name,
                                         Class<V> attributeType) {
-        Objects.requireNonNull(entityClass, "entity class is required");
-        Objects.requireNonNull(name, "entity attribute name is required");
-        Objects.requireNonNull(attributeType, "entity attribute type is required");
+        Objects.requireNonNull(entityClass, "Entity class is required");
+        Objects.requireNonNull(name, "Entity attribute name is required");
```
otaviojava, line 46: But @njr-11, once the parameter is "name," it should match with the variable, right?

I mean, I passed the name parameter as `null`, and I received the "Entity attribute name is required" sounds odd.

otaviojava: Maybe we could have a wrapper class to handle this message, such as:

```java
Validation. requireNonNull(name, "name);
```

Where can we centralize the error message?

otaviojava: @njr-11 that you be nice.

One more question, about all the messages be the same:

Like:

```java
 Objects.requireNonNull(entityClass, "Entity class is required");
 Objects.requireNonNull(name, "Entity attribute name is required");
```

should it be this instead:


```java
 Objects.requireNonNull(entityClass, "entityClass class is required");
 Objects.requireNonNull(name, "name attribute name is required");
```

I mean, I believe the message name should match the parameter name, right?

njr-11: It definitely should not be `"name attribute name is required"`. That doesn't make any sense. From your other example, `"entityClass class is required"`, I think you are asking to ensure we have the exact variable name in the message rather than splitting it into separate words.  So `entityClass` rather than `Entity class`.  I'm fine with that.  In order to match the consistent style of beginning with a capital letter, we could have something like, `"The entityClass argument is required."`

PR #1084, opened 2025-04-29
Enhance test text expression

otaviojava:


api/src/test/java/jakarta/data/metamodel/TextAttributeTest.java
```diff
@@ -32,127 +34,106 @@ static class Author {
...
-        Like like = (Like) restriction.constraint();
-        StringLiteral<?> literal = ((StringLiteral<?>) like.pattern());
+        var restriction = (BasicRestriction<?, ?>) _Author.testAttribute.endsWith("testValue");
 
         SoftAssertions.assertSoftly(soft -> {
-            soft.assertThat(restriction.expression()).isEqualTo(testAttribute);
-            soft.assertThat(literal.value()).isEqualTo("%testValue");
+            soft.assertThat(restriction.constraint()).isInstanceOf(Like.class);
+            soft.assertThat(restriction.expression()).isEqualTo(_Author.testAttribute);
+            soft.assertThat(restriction.constraint()).isEqualTo(Like.suffix("testValue"));
```
gavinking, line 76: I'm not sure I would say that:

    assertThat(restriction.constraint()).isEqualTo(Like.suffix("testValue"))

is exactly the same thing as:

    assertThat(literal.value()).isEqualTo("%testValue")

I guess I would maybe leave both assertions there.

njr-11: I agree. They are definitely not the same assertions.  Losing the assertion on `.value()` means the test will pass even if `Like.suffix` is wrongly implemented -- as long as the wrong implementation returns a constraint that is equal whatever `.endsWith` returned, which is probably `Like.suffix` invoked on the same parameter.

api/src/test/java/jakarta/data/metamodel/expression/ComparableLiteralRecordTest.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    // Static metamodel
+    interface _Book {
+        String TITLE = "title";
+        String PAGES = "pages";
+
+        ComparableExpression<Book, String> title = new ComparableLiteralRecord<>(TITLE);
+        ComparableExpression<Book, Integer> pages = new ComparableLiteralRecord<>(100);
+    }
```
njr-11, line 41: This is wrong usage.  Static metamodel classes use *Attribute, not Expression.
```suggestion
```

api/src/test/java/jakarta/data/metamodel/expression/StringLiteralTest.java
```diff
@@ -0,0 +1,139 @@
...
+import org.assertj.core.api.SoftAssertions;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+class StringLiteralTest {
+
+    interface _SimpleEntity {
+        StringLiteral<SimpleEntity> name = StringLiteral.of("name");
+    }
+
```
njr-11, line 32: If this was supposed to be a static metamodel class (the naming pattern looks like it), it is incorrect usage to have something other than an Attribute or String constant in it,
```suggestion
```

api/src/test/java/jakarta/data/metamodel/expression/ComparableLiteralRecordTest.java
otaviojava: Thank you, I am fixing.

otaviojava:
@njr-11 how about now?

gavinking (approved):

Issue #1086, opened 2025-04-30
Keep track of documentation needed for Restrictions, Expressions, and Constraints


njr-11:
We deferred the creation of documentation for Restrictions, Expressions, and Constraints due to a large amount of churn in those classes.  This issue is opened to track the work that is needed, as well as to try to logically split it up in to parts that have been more stable vs parts that might be more likely to continue to see changes. There is also a lot of similarity in many of the API classes (for example the constraint subtypes or core expression subtypes), and when we are ready to work on those, an efficient way to do so would be to work on just a single class first and gain agreement on that before replicating it to all of the similar classes. That will save a lot of work for both the reviewers and author of the PR.

From what seems most stable to most likely to see more changes:

1. - [ ] Restriction (basic and composite) - this piece has remained stable for a while now, and I'm going to look into starting on the Javadoc for it.
2. - [ ] Expression and 1 of the subtypes from the expression package.
3. - [ ] The other Expression subtypes from the expression package.
4. - [ ] Literal and 1 of the subtypes from the expression.literal package.
5. - [ ] The remaining subtypes from the expression.literal package.
6. - [ ] NumericOperatorExpression
7. - [ ] FunctionExpression, TextFunctionExpression, NumericFunctionExpression
8. - [ ] CurrentDate, CurrentDateTime, CurrentTime
9. - [ ] NumericCast (will need to be able to document which types of casting are valid)
10. - [ ] Path and 1 of the subtypes from the expression.path package
11. - [ ] The remaining subtypes from the expression.path package
12. - [ ] Constraint and 1 of the subtypes from the constraint package
13. - [ ] The remaining subtypes from the constraint package
14. - [ ] Double check for other API classes that have been added but not documented

In addition to the above, the specification document. I wasn't sure where to place that in the order.

We need to be careful when writing examples into documentation to avoid using capability that is still in question or is expected to be deferred post 1.1, such as constraints and expressions that compare against other expressions and attributes rather than values.

PR #1087, opened 2025-04-30
Javadoc for Restriction

njr-11:
First item under #1086
Adds Javadoc for Restriction.  I think the Restriction interface and Restrict have remained fairly stable for a while now and should be safe to add Javadoc for.

### DIFF
--- a/api/src/main/java/jakarta/data/restrict/Restriction.java
+++ b/api/src/main/java/jakarta/data/restrict/Restriction.java
@@ -125,8 +125,8 @@ public interface Restriction<T> {
      * <p>For example, a basic restriction that represents an {@linkplain GreaterThan exclusive upper bound}
      * on a value is negated as an {@linkplain LessThanOrEqual inclusive lower bound} on the value.</p>
      *
-     * <p>A basic restriction that represents {@link Like} a pattern is negated
-     * as {@link NotLike} the pattern.</p>
+     * <p>A basic restriction that represents {@linkplain Like matching}  a pattern is negated
+     * as a restriction to {@link NotLike not match} the pattern.</p>
      *
      * <p>A composite restriction that requires satisfying
      * {@linkplain CompositeRestriction.Type#ANY at least one} restriction
--- a/api/src/main/java/jakarta/data/restrict/Restriction.java
+++ b/api/src/main/java/jakarta/data/restrict/Restriction.java
@@ -77,41 +77,8 @@ import jakarta.data.repository.Find;
  *                     Order.by(_Car.price.desc()));
  * </pre>
  *
- * <p>Entity class for the above example:</p>
- * <pre>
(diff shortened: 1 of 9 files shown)


api/src/main/java/jakarta/data/restrict/Restriction.java
```diff
@@ -17,6 +17,131 @@
...
+ *     TextAttribute&lt;Car&gt; model = TextAttribute.of(
+ *             Car.class, MODEL);
+ *     NumericAttribute&lt;Car,Integer&gt; price = NumericAttribute.of(
+ *             Car.class, PRICE. int.class);
+ *     NumericAttribute&lt;Car,Integer&gt; year = NumericAttribute.of(
+ *             Car.class, YEAR. int.class);
+ *     TextAttribute&lt;Car&gt; vin = TextAttribute.of(
+ *             Car.class, VIN);
+ * }
+ * </pre>
```
gavinking, line 114: We already have a code example like this on `@StaticMetamodel`. I would not repeat it here. If you want to repeat it somewhere, `Attribute` would be a better place IMO.

api/src/main/java/jakarta/data/restrict/Restriction.java
```diff
@@ -17,6 +17,131 @@
...
+ *
+ * @param <T> entity type.
+ */
 public interface Restriction<T> {
+
+    /**
+     * <p>Returns the negation of this restriction.</p>
+     *
+     * <p>For example, a basic restriction that represents {@link GreaterThan}
+     * a value is negated as {@link LessThanOrEqual} the value.</p>
```
gavinking, line 126:
```suggestion
     * <p>For example, a basic restriction that represents an {@linkplain GreaterThan exclusive upper bound}
     * on a value is negated as an {@linkplain LessThanOrEqual inclusive lower bound} on the value.</p>
```

api/src/main/java/jakarta/data/restrict/Restriction.java
```diff
@@ -17,6 +17,131 @@
...
+     *
+     * <p>A basic restriction that represents {@link Like} a pattern is negated
+     * as {@link NotLike} the pattern.</p>
```
gavinking, line 129:
```suggestion
     * <p>A basic restriction that represents {@linkplain Like matching}  a pattern is negated
     * as a restriction to {@link NotLike not match} the pattern.</p>
```

gavinking (approved):

api/src/main/java/jakarta/data/restrict/Restriction.java
njr-11: Yeah, that's probably a good idea. I moved it to `Attribute` as you suggested, and we can just refer to that when we have examples like this that make non-trivial use of the metamodel.

PR #1090, opened 2025-05-01
squash warnings in IntelliJ

gavinking:
Redo of #1067


njr-11 (approved):
Approving. All of the changes look fine, but there are a lot of them. Hopefully none of it causes new warnings to appear in other IDEs.

PR #1091, opened 2025-05-02
Javadoc for Expression

njr-11:
Covers item 2 from #1086 - adding JavaDoc for the Expression class before proceeding to do likewise for the subtypes.

### DIFF
--- a/api/src/main/java/jakarta/data/expression/Expression.java
+++ b/api/src/main/java/jakarta/data/expression/Expression.java
@@ -37,7 +37,7 @@ import java.util.Collection;
  * {@linkplain jakarta.data.expression.function function}.</p>
  *
  * <p>Expressions are used to compose {@linkplain Restriction restrictions}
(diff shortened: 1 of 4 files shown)


api/src/main/java/jakarta/data/expression/Expression.java
```diff
@@ -24,71 +24,358 @@
...
 
 import java.util.Collection;
 
+/**
+ * <p>An expression represents an {@linkplain Attribute entity attribute},
+ * {@linkplain jakarta.data.expression.literal literal value}, or
+ * {@linkplain jakarta.data.expression.function function}.</p>
+ *
+ * <p>Expressions are used to compose {@linkplain Restriction restrictions}
+ * that allow applications to supply query criteria at run time. For example,
```
gavinking, line 40:
```suggestion
 * that allow applications to supply query criteria at runtime. For example,
```

gavinking (approved):
LGTM

PR #1098, opened 2025-05-08
Address TODO to use switch after moving up to Java 21

njr-11:
Address 2 TODO comments in the code to replace with a switch statement after moving up to Java 21 compilation.

### DIFF
--- a/api/src/main/java/jakarta/data/expression/literal/NumericLiteralRecord.java
+++ b/api/src/main/java/jakarta/data/expression/literal/NumericLiteralRecord.java
@@ -30,14 +30,13 @@ record NumericLiteralRecord<T, N extends Number & Comparable<N>>
 
     @Override
     public String toString() {
-        String suffix = switch (value) {
-            case Long l       -> "L";
-            case Float f      -> "F";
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/data/expression/literal/NumericLiteralRecord.java
```diff
@@ -30,21 +30,14 @@ record NumericLiteralRecord<T, N extends Number & Comparable<N>>
...
-        }
+        String suffix = switch (value) {
+            case Long l       -> "L";
+            case Float f      -> "F";
+            case Double d     -> "D";
+            case BigInteger i -> "BI";
+            case BigDecimal d -> "BD";
+            default           -> "";
+        };
         return value.toString() + suffix;
```
gavinking, line 41: How about just:

```suggestion
        return value + switch (value) {
            case Long l -> "L";
            case Float v -> "F";
            case Double v -> "D";
            case BigInteger i -> "BI";
            case BigDecimal d -> "BD";
            default -> "";
        };
```

gavinking (approved):

PR #1101, opened 2025-05-09
1086 javadoc for literals

njr-11:
This PR adds JavaDoc for Literal and all of its subtypes.  While adding the Javadoc for the toString methods of these literals, I noticed that the output was inconsistent in some cases with query language, so I added updates to correct that and some unit tests.

### DIFF
--- a/api/src/main/java/jakarta/data/expression/literal/ComparableLiteral.java
+++ b/api/src/main/java/jakarta/data/expression/literal/ComparableLiteral.java
@@ -23,12 +23,40 @@ import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
+import java.util.UUID;
 
 import jakarta.data.expression.ComparableExpression;
 
+/**
+ * <p>A {@linkplain Literal literal} for a sortable value, such as
+ * {@link Boolean}, {@link Character}, enumeration, and {@link UUID} values.
+ * </p>
+ *
+ * @param <T> entity type.
+ * @param <V> entity attribute type.
+ * @since 1.1
+ */
 public interface ComparableLiteral<T, V extends Comparable<?>>
         extends ComparableExpression<T, V>, Literal<T, V> {
 
+    /**
+     * <p>Creates a {@code ComparableLiteral} or subtype of
+     * {@code ComparableLiteral} that represents the given value.</p>
+     *
+     * <p>The most specific subtype of {@code ComparableLiteral}, such as
+     * {@link NumericLiteral#of(Number) NumericLiteral},
+     * {@link StringLiteral#of(String) StringLiteral}, or
+     * {@link TemporalLiteral#of(java.time.temporal.Temporal) TemporalLiteral},
+     * should be used instead wherever possible.</p>
+     *
+     * @param <T>   entity type.
+     * @param <V>   entity attribute type.
+     * @param value an immutable value or a mutable value that must never be
+     *              modified after it is supplied to this method. Must never be
+     *              {@code null}.
+     * @return a {@code ComparableLiteral} representing the value.
+     * @throws NullPointerException if the value is {@code null}.
+     */
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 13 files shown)


api/src/main/java/jakarta/data/expression/literal/ComparableLiteral.java
```diff
@@ -23,12 +23,40 @@
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
+import java.util.UUID;
 
 import jakarta.data.expression.ComparableExpression;
 
+/**
+ * <p>A {@linkplain Literal literal} for a sortable value, such as
+ * {@link Boolean}, {@link Character}, enumeration, and {@link UUID} values.
```
gavinking, line 32:
```suggestion
 * <p>A {@linkplain Literal literal} expression for a sortable value, such as a
 * {@link Boolean}, {@link Character}, enumerated, or {@link UUID} value.
```

api/src/main/java/jakarta/data/expression/literal/ComparableLiteral.java
```diff
@@ -23,12 +23,40 @@
...
 
+    /**
+     * <p>Creates a {@code ComparableLiteral} or subtype of
+     * {@code ComparableLiteral} that represents the given value.</p>
+     *
+     * <p>The most specific subtype of {@code ComparableLiteral}, such as
+     * {@link NumericLiteral#of(Number) NumericLiteral},
+     * {@link StringLiteral#of(String) StringLiteral}, or
+     * {@link TemporalLiteral#of(java.time.temporal.Temporal) TemporalLiteral},
+     * should be used instead wherever possible.</p>
```
gavinking, line 50: Is this better as an `@apiNote`?

It's almost like we need a `@jakartaDataImplementorNote`.

api/src/main/java/jakarta/data/expression/literal/TemporalLiteralRecord.java
```diff
@@ -34,12 +35,21 @@ record TemporalLiteralRecord<T, V extends Temporal & Comparable<? extends Tempor
...
+            case LocalDateTime d ->
+                "{ts '" + d.toLocalDate() + ' ' + d.toLocalTime() + "'}";
+            case LocalDate d ->
+                "{d '" + value + "'}";
+            case LocalTime t ->
+                "{t '" + value + "'}";
+            default ->
+                "{TemporalLiteral '"
+                        + value.getClass().getName() + " '"
+                        + value + "'}";
```
gavinking, line 52: This will probably need to be revisited in light of https://github.com/jakartaee/persistence/issues/453.  (Currently an issue in the Persistence project, but really belongs in Query.)

gavinking (approved):

api/src/main/java/jakarta/data/expression/literal/TemporalLiteralRecord.java
njr-11: > This will probably need to be revisited in light of [jakartaee/persistence#453](https://github.com/jakartaee/persistence/issues/453). (Currently an issue in the Persistence project, but really belongs in Query.)

Yes, definitely. I noticed Jakarta Persistence 3.2 was very sparse on this topic and only deferred to JDBC escape syntax for timestamp/date/time. I'm glad to see the issue open for it, and I'll get a corresponding issue opened against Jakarta Data to align with whatever the outcome is.

api/src/main/java/jakarta/data/expression/literal/ComparableLiteral.java
njr-11: > It's almost like we need a `@jakartaDataImplementorNote`.

I was actually intending this for the end user when they have a situation where they need to create a literal, encouraging them to use the most specific subtype so that they will get the most usefulness out of it.  I suppose that might already be obvious to them though if they had the use case for wanting the literal in the first place then they probably also know which type they want.

gavinking: Ah ok.

PR #1104, opened 2025-05-12
IllegalArgumentException for division by 0 and negative RIGHT/LEFT size arg

njr-11:
IllegalArgumentException for a couple of error paths due to bad input:
*  .divide(0)
*  .right(negative) and .left(negative)


gavinking (approved):
Not sure this is something I would have bothered with, but it's fine.

PR #1107, opened 2025-05-19
Reuse same error messages instead of duplicating

njr-11:
This addresses a comment from @otaviojava in a prior review to reuse our error messages across code.

To do this, I replaced occurrences of requireNonNull because that would have required looking up error messages in the resource bundle and substituting in the values to supply it to that method on the normal codepath when no error occurs. This seemed wasteful. We should only need incur that cost when an error is detected.

I chose to create a single utility class to centralize obtaining the messages and substituting in the values to make the code cleaner, which needed to be public so it could be accessed everywhere, but we will not generate Javadoc for it or export it from the module.  I'm not sure if this approach is acceptable.  If not, other options include having that class as package protected in each package, or omitting it entirely and having slightly more verbose code wherever we need to obtain error messages.

### DIFF
--- a/api/src/main/resources/jakarta/data/messages/Messages.properties
+++ b/api/src/main/resources/jakarta/data/messages/Messages.properties
@@ -1,0 +1,29 @@
+#
+# Copyright (c) 2025 Contributors to the Eclipse Foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+001.arg.required=The {0} argument is required.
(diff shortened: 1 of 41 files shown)


api/src/main/resources/jakarta/data/messages/Messages.properties
```diff
@@ -0,0 +1,29 @@
...
+002.no.elements=The {0} list cannot be empty.
+003.null.element=The {0} list cannot not have a null element.
+004.arg.negative=The {0} argument cannot be negative.
+005.zero.not.allowed=The {0} argument cannot be 0.
+006.zero.size.key=The key must consist of at least one element.
+007.wildcard.conflict=Cannot use the same character ({0}) for both wildcards.
+008.escape.conflict=Cannot use the same character ({0}) for both a wildcard \
+ and escape character.
+009.unknown.number.type=Unexpected subtype of Number: {0}.
+010.unknown.total=A total count of elements is not available.
```
gavinking, line 29: Personally, I would leave off the periods.

gavinking (approved):

api/src/main/resources/jakarta/data/messages/Messages.properties
njr-11: > Personally, I would leave off the periods.

Sure, that's fine. Some of the messages were inconsistent and I had went with the approach that I'm used to from our tech writers asking us to do for our own products. I checked with other Jakarta EE specs do for messages and it looks like Jakarta Servlet leaves off the period, so we can align with that.  It's easy to switch now with all of the messages in one place.

PR #1110, opened 2025-05-20
findAll documentation of error path is misleading

njr-11:
The method,
```
    @Find
    Page<T> findAll(PageRequest pageRequest, Order<T> sortBy);
````
has Javadoc that currently states,
```
     * @throws UnsupportedOperationException for Key-Value and Wide-Column
     *                                       databases when the
     *                                       {@link
     *                                       PageRequest.Mode#CURSOR_NEXT} or
     *                                       {@link
     *                                       PageRequest.Mode#CURSOR_PREVIOUS}
     *                                       pagination mode is selected.
```
It's perfectly fine to disallow the cursored PageRequest when Key-Value and Wide-Column databases are used.  However, given that the method is returning Page and not CursoredPage, it would already be necessary to reject the method whenever a cursored PageRequest is supplied, regardless of whether it is Key-Value or Wide-Column.  The message here is then misleading because a reader can be easily confused that a statement that says it is unsupported for Key-Value and Wide-Column might imply that it _is_ supported for other database types.  This PR updates the Javadoc to make it clear that cursored PageRequests are never supported on this method.


njr-11:
Updated per review comments on the Jakarta Data call that the exception can be omitted entirely given that it is already not allowed to request a cursored page for a method that returns Page.

PR #1113, opened 2025-05-21
toString for remaining expressions

njr-11:
Implement toString for the remaining functions, attributes, and path expressions. Include some tests.


api/src/test/java/jakarta/data/metamodel/NumericExpressionTest.java
```diff
@@ -199,4 +225,21 @@ void shouldCastToBigDecimal() {
...
         });
     }
+
+    @Test
+    @DisplayName("toString output must include parentheses where needed")
+    void shouldIncludeParenthesesInToString() {
+        NumericLiteral<Invoice, Integer> hundred = NumericLiteral.of(100);
+        // TODO this does not compile when the above is included directly
+        // within the following without the cast. Do literals need an of method
+        // that explicitly includes the entity class?
```
gavinking, line 235: We will need to fix that. Like what we did with `localDate()`, etc.

gavinking: Fix looks like it involves three changes:

```java
    static <N extends Number & Comparable<N>> NumericLiteral<Object,N> of(N value) {
        return new NumericLiteralRecord<>(value);
    }
```

```java
    Expression<? super T, N> left();

    Expression<? super T, N> right();
```

```java
record NumericOperatorExpressionRecord<T, N extends Number & Comparable<N>>
        (Operator operator, NumericExpression<? super T, N> left,
         NumericExpression<? super T, N> right)
```

We will have to audit stuff to see if there are any similar things hiding.

njr-11: It looks like you already have a PR open to cover thie above TODO (good idea to split it out from this one) so I'll review that.

api/src/test/java/jakarta/data/metamodel/NumericExpressionTest.java
```diff
@@ -199,4 +226,24 @@ void shouldCastToBigDecimal() {
...
+    @Test
+    @DisplayName("toString output must include parentheses where needed")
+    void shouldIncludeParenthesesInToString() {
+        @SuppressWarnings("unchecked")
+        NumericLiteral<Invoice, Integer> hundred =
+                (NumericLiteral<Invoice, Integer>)
+                (NumericLiteral<?, Integer>) NumericLiteral.of(100);
+
+        // TODO this does not compile when the above is included directly
+        // within the following without the cast.
```
njr-11, line 239: > Fix looks like it involves three changes:
> 
> ```java
>     static <N extends Number & Comparable<N>> NumericLiteral<Object,N> of(N value) {
>         return new NumericLiteralRecord<>(value);
>     }
> ```

After bringing in #1114, it didn't turn out as I was hoping.  Now we need to cast from `NumericLiteral<Object, Integer>` to `NumericLiteral<Invoice, Integer>` rather than what we had before to define the entity type,

```
NumericLiteral<Invoice, Integer> hundred = NumericLiteral.of(100);
```

I think we would need to get rid of the entity type entirely from Literals, but that isn't really possible because they inherit from Expressions.

gavinking: Show me the actually concrete example you're trying to make work.

njr-11: > Show me the actually concrete example you're trying to make work.

It's the following line of code, but with `NumericLiteral.of(100)` within it,

```
        NumericExpression<Invoice, Integer> expression = _Invoice.amount.times(
                NumericLiteral.of(100).minus(_Invoice.percentDiscount))
                .divide(100);
```

gavinking: Yes but that object is not useful on its own. You need to put it in a predicate. I need the context.

gavinking: Oh it's because we also need to fix all these:

```java
    default NumericExpression<T, N> plus(NumericExpression<T, N> expression) {
        return NumericOperatorExpression.of(PLUS, this, expression);
    }
```

To:

```java
    default NumericExpression<T, N> plus(NumericExpression<? super T, N> expression) {
        return NumericOperatorExpression.of(PLUS, this, expression);
    }
```

njr-11: > Oh it's because we also need to fix all these:
> 
> ```java
>     default NumericExpression<T, N> plus(NumericExpression<T, N> expression) {
>         return NumericOperatorExpression.of(PLUS, this, expression);
>     }
> ```
> 
> To:
> 
> ```java
>     default NumericExpression<T, N> plus(NumericExpression<? super T, N> expression) {
>         return NumericOperatorExpression.of(PLUS, this, expression);
>     }
> ```

Ah, I hadn't thought of looking there for some reasons. I'll give that a try. Seems promising.

njr-11: > Oh it's because we also need to fix all these:
> 
> ```java
>     default NumericExpression<T, N> plus(NumericExpression<T, N> expression) {
>         return NumericOperatorExpression.of(PLUS, this, expression);
>     }
> ```
> 
> To:
> 
> ```java
>     default NumericExpression<T, N> plus(NumericExpression<? super T, N> expression) {
>         return NumericOperatorExpression.of(PLUS, this, expression);
>     }
> ```

That didn't quite work, but I was able to do something similar in that area (under the latest commit) after which I was able to remove the TODO.  I think we will definitely need more testing with Literals as the starting point in different places to see if there are more issues like this, but I'll save that for another PR.  This one was only supposed to be focused on the toString output.

api/src/main/java/jakarta/data/expression/NumericExpression.java
```diff
@@ -184,7 +184,8 @@ default NumericExpression<T, N> divide(N divisor) {
      * @return an expression for the function that computes the sum.
      * @throws NullPointerException if the supplied value is null.
      */
-    default NumericExpression<T, N> plus(NumericExpression<T, N> expression) {
+    default <U extends T> NumericExpression<U, N> plus(
+            NumericExpression<U, N> expression) {
```
gavinking, line 188: This doesn't look quite right to me. I want:

    _Book.pages.add(NumericLiteral.of(1))

to be accepted and have the type `NumericExpression<Book,Integer>`.

I don't care so much about:

     NumericLiteral.of(1).plus(_Book.pages)

because that's a less conventional/natural way to write the expression.

njr-11: I added a commit to make the Literal path work as well (at the cost of 4 overloaded methods).  We will likely want to think more about how to solve this problem.  My aim here is just to get something working to unblock the progress of this PR in adding implementation of toString and being able to test it.  Hopefully we can put this in now and use a separate issue to revisit how we want to cover literals.

api/src/main/java/jakarta/data/expression/NumericExpression.java
```diff
@@ -257,6 +258,46 @@ default <U extends T> NumericExpression<U, N> divide(
...
+        @SuppressWarnings("unchecked")
+        NumericExpression<T, N> literal =
+            (NumericExpression<T, N>) addendLiteral;
+
+        return NumericOperatorExpression.of(PLUS, this, literal);
+    }
+
+    // TODO Javadoc
+    default NumericExpression<T, N> minus(
+            NumericLiteral<Object, N> subtrahendLiteral) {
```
gavinking, line 273: This is a layer breaker. We don't want to leak `NumericLiteral` onto the API of `NumericExpression`.

api/src/main/java/jakarta/data/expression/NumericExpression.java
gavinking: > You could certainly write code differently as you suggested (and as I would recommend for a real application), but that wouldn't be useful in the documentation of the `times(Expression)` method because it performs the times operation on a value rather than an `Expression`.

What I'm saying is that this is more of an SPI than something that's meant for writing queries with.

njr-11: > What I'm saying is that this is more of an SPI than something that's meant for writing queries with.

> Remember: we only introduced `NumericLiteral` in order to be able to abstract over literal values and expressions _internally_. We never really intended people to use `NumericLiteral` directly, since the interesting operations are all overloaded to just accept a literal value.

I don't agree with what you are saying that the interesting/important operations with literals have the value on the righthand size.  Many do, but other don't.  When you have entities with attributes that are ratios and percentages, it is going to be common to want to compute `1 - ratio` or `100 - percentage` and things like that, and as you pointed out, you might even see some reciprocals.  Users will naturally try to achieve this with Literals because Literals are a type of Expression in the API, whether you intended them to be used as SPI or otherwise.  In general, we have been okay including what we consider SPI to be part of the API because what we are thinking of it as not being harmful or confusing there. (For example, methods to get values from Constraints). Having Literal in the API as a subtype of Expression but making it very difficult to compile with it that way will be seen as buggy and poor design.

njr-11: I opened a separate issue #1116 to cover what we end up doing with Literal and put a couple of proposals there.  Let me know if you need anything else done with this current PR #1113 .  I already removed the 4 methods and corresponding lines of tests for usage like _Book.numPages.plus(Literal.of(1))

gavinking: (Sorry, the last two days have been completely nuts for me, and I didn't have time to properly explain or even fully think through what I was trying to explain. I was just working from pure intuition.)

> I don't agree with what you are saying that the interesting/important operations with literals having the value on the righthand size. Many do, but other don't. When you have entities with attributes that are ratios and percentages, it is going to be common to want to compute 1 - ratio or 100 - percentage and things like that, and as you pointed out, you might even see some reciprocals.

Right, but we both independently thought of a solution for these non-commuting binary operators, which is to just add `subtractedFrom()` and `dividedInto()`. For completeness, we could also consider adding `prefixOf()` and `suffixOf()` to `StringExpression`, but I guess I don't see them as very common operations. [It would also make sense, but the way, to add a privileged operation for the multiplicative inverse, just like we have one for the additive inverse.]

So anyway, that's what I mean about I think we're going to be able to get away with assuming that literals get introduced "on the right". I wasn't trying to say that this was some kind of iron law or something, just that it's something that is probably going to be "good enough", given the unfortunate limitations of generic types in Java, and given that we're not trying to boil the ocean.

At worst, we can certainly introduce symmetric forms of these operations as static methods, but I don't really see that as needed so far. Seems to me from experience with the JPA criteria API that `_Pricing.discount.subtractedFrom(100)` is actually quite a lot more friendly to the user than `Expression.subtract(100, _Pricing.discount)`, and so I've been concentrating on trying to keep everything as instance methods. But in principle nothing stops us from providing both forms.

PR #1114, opened 2025-05-21
proposed fix to NumericLiteral and friends

gavinking:


njr-11 (approved):
This looks like a good approach. Literals are independent of any particular entity type and so an entity type should not need to be specified for them.

njr-11:
Merging this so that we can remove the TODO from #1113

PR #1124, opened 2025-06-02
Update Jakarta Data model with more references with the version 1.1

otaviojava:
Please, understand that what I did was based on what I am implementing, but I can not understand very well.

Please feel free to review and make any necessary corrections.

Thank you.

PR #1127, opened 2025-06-04
java.time.Year as a temporal attribute

njr-11:
#1124 added java.time.Year (which we discussed on the Jakarta Data call and agreed on) to the list of temporal types.

This PR adds the rest of the support for it to the spec.

### DIFF
--- a/api/src/main/java/jakarta/data/spi/expression/literal/ComparableLiteral.java
+++ b/api/src/main/java/jakarta/data/spi/expression/literal/ComparableLiteral.java
@@ -23,6 +23,7 @@ import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
+import java.time.Year;
 import java.util.UUID;
 
 import jakarta.data.expression.ComparableExpression;
@@ -87,6 +88,8 @@ public interface ComparableLiteral<V extends Comparable<?>>
             return (ComparableLiteral<V>) TemporalLiteral.of(d);
         } else if (value instanceof LocalTime t) {
             return (ComparableLiteral<V>) TemporalLiteral.of(t);
+        } else if (value instanceof Year y) {
+            return (ComparableLiteral<V>) TemporalLiteral.of(y);
         } else {
             return new ComparableLiteralRecord<>(value);
         }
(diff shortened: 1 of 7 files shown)


api/src/main/java/jakarta/data/spi/expression/literal/ComparableLiteral.java
```diff
@@ -87,6 +88,8 @@ static <V extends Comparable<?>> ComparableLiteral<V> of(V value) {
             return (ComparableLiteral<V>) TemporalLiteral.of(d);
         } else if (value instanceof LocalTime t) {
             return (ComparableLiteral<V>) TemporalLiteral.of(t);
+        } else if (value instanceof Year y) {
```
otaviojava, line 91: instead of this huge amount of if else, can we use a Map instead?


```java
private static final Map<Class<?>, Function<Object, ComparableLiteral<?>>> LITERAL_FACTORIES = Map.ofEntries(
    Map.entry(String.class, v -> StringLiteral.of((String) v)),
    Map.entry(Integer.class, v -> NumericLiteral.of((Integer) v)),
    Map.entry(Long.class, v -> NumericLiteral.of((Long) v)),
    Map.entry(Float.class, v -> NumericLiteral.of((Float) v)),
    Map.entry(Double.class, v -> NumericLiteral.of((Double) v)),
    Map.entry(Byte.class, v -> NumericLiteral.of((Byte) v)),
    Map.entry(Short.class, v -> NumericLiteral.of((Short) v)),
    Map.entry(BigInteger.class, v -> NumericLiteral.of((BigInteger) v)),
    Map.entry(BigDecimal.class, v -> NumericLiteral.of((BigDecimal) v)),
    Map.entry(Instant.class, v -> TemporalLiteral.of((Instant) v)),
    Map.entry(LocalDate.class, v -> TemporalLiteral.of((LocalDate) v)),
    Map.entry(LocalDateTime.class, v -> TemporalLiteral.of((LocalDateTime) v)),
    Map.entry(LocalTime.class, v -> TemporalLiteral.of((LocalTime) v)),
    Map.entry(Year.class, v -> TemporalLiteral.of((Year) v))
);

@SuppressWarnings("unchecked")
public static <V> ComparableLiteral<V> toLiteral(Object value) {
 
    Function<Object, ComparableLiteral<?>> factory = LITERAL_FACTORIES.get(value.getClass());
   
}
```

gavinking (approved):

Issue #470, opened 2024-02-21
[Use Case]: Annotations for use with Jakarta Persistence


gavinking:
### As a ...

- [ ] Application user/user of the configuration itself
- [X] API user (application developer)
- [ ] SPI user (container or runtime developer)
- [ ] Specification implementer

### I need to be able to ...

write a portable repository backed by JPA.

### Which enables me to ...

be more independent of the Jakarta Data provider.

### Additional information

So I know we all said we didn't want to do this in this release. But with all the machinery we've defined, it's now frankly trivial.

We would add:

- lifecycle annotations `@Persist`, `@Merge`, `@Refresh`, `@Lock`, `@Remove`, and `@Detach`, and
- query annotations `@JPQL` and I guess also `@SQL`

in new package `jakarta.data.orm`.

The semantics of the lifecycle annotations are completely defined by the JPA spec. Because they're specific to JPA/ORM, we have zero work to do, it's just a `@see` annotation that points to the corresponding operation of `EntityManager`. 

It's similarly clear what `@JPQL` does. On the other hand, perhaps we don't need it and all we need to do is give people explicit permission to put their JPQL in the `@Query` annotation. (I can see arguments both ways.)

Finally, if I recall correctly, at some stage someone (Otavio I guess) had proposed a built-in `JpaRepository`. I don't argue that we need it, but if you all still want it, it's completely trivial to define in terms of the annotations listed above.

So why is this now so easy when previously it was "hard". Well, before, we were trying to shoehorn JPA into the existing `@Save` annotation or whatever. But now that our model of lifecycle annotations is extensible, it's no longer hard.

Dunno. I guess I feel like it's now become a bit hard to justify _not_ doing it.

Reactions?

gavinking:
> Finally, if I recall correctly, at some stage someone (Otavio I guess) had proposed a built-in `JpaRepository`. I don't argue that we need it, but if you all still want it, it's completely trivial to define in terms of the annotations listed above.

See #143.

gavinking:
Please see #477.

gavinking:
Alright, so I guess I need to rewind a bit here and fully explain why I'm now proposing to do something we said we wouldn't do.

CRUD vs JPA
---------------------

So, right from the start I've been pointing out the deep, fundamental differences in semantics between the semantics of JPA, and the semantics which were contemplated for simple repository lifecycle methods.

Let's start by considering a simple `@Insert` operation, vs JPA's `persist()`. Among other things:

[...]

PR #1128, opened 2025-06-05
add annotations for declaring lifecycle methods of stateful repositories

gavinking:
see #470

Draft for these annotations. Open issues:

1. Do we need `@StatefulRepository` ?
2. How should a `@Remove` method treat an unmanaged entity? Choices are:
    -  error, as I've written here, or 
    - ignore it, if the repository can determine that the entity does not exist in the database, which is what JPA does.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/stateful/Detach.java
+++ b/api/src/main/java/jakarta/data/repository/stateful/Detach.java
@@ -17,7 +17,11 @@
  */
 package jakarta.data.repository.stateful;
 
-import java.lang.annotation.*;
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
 /**
  * <p>Lifecycle annotation for repository methods of stateful repositories
--- a/api/src/main/java/jakarta/data/repository/stateful/Detach.java
+++ b/api/src/main/java/jakarta/data/repository/stateful/Detach.java
@@ -1,0 +1,46 @@
+package jakarta.data.repository.stateful;
+
+import java.lang.annotation.*;
+
+/**
+ * <p>Lifecycle annotation for repository methods of stateful repositories
+ * which perform detach operations. The {@code Detach} annotation must
+ * not be applied to a method of a stateless repository.
+ * </p>
+ * <p>The {@code Detach} annotation indicates that the annotated repository
+ * method immediately makes one or more entities unmanaged, removing them
+ * from the current persistence context associated with the repository.
+ * </p>
+ * <p>An {@code Detach} method accepts an instance or instances of an entity
+ * class. The method must have exactly one parameter whose type is either:
+ * </p>
+ * <ul>
+ *     <li>the class of the entity to be inserted, or</li>
+ *     <li>{@code List<E>} or {@code E[]} where {@code E} is the class of the
+ *     entities to be inserted.</li>
+ * </ul>
+ * <p>The annotated method must be declared {@code void}.
+ * </p>
+ * <p>If an unmanaged entity is passed to a {@code Detach} method, the call
+ * has no effect.
+ * </p>
+ * <p>Every Jakarta Data provider which supports stateful repositories is
+ * required to accept a {@code Detach} method which conforms to this signature.
+ * Application of the {@code Detach} annotation to a method with any other
+ * signature is not portable between Jakarta Data providers. Furthermore,
+ * support for stateful repositories is optional. A Jakarta Data provider
+ * is not required to support stateful repositories.
+ * </p>
+ * <p>Annotations such as {@code @Find}, {@code @Query}, {@code @Persist},
+ * {@code @Merge}, {@code @Remove}, {@code @Refresh}, and {@code Detach} are
+ * mutually-exclusive. A given method of a repository interface may have at
+ * most one {@code @Find} annotation, lifecycle annotation, or query annotation.
+ * </p>
+ *
+ * @since 1.1
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+@interface Detach {
+}
--- a/api/src/main/java/jakarta/data/repository/stateful/Detach.java
+++ b/api/src/main/java/jakarta/data/repository/stateful/Detach.java
@@ -36,9 +36,9 @@ import java.lang.annotation.Target;
  * class. The method must have exactly one parameter whose type is either:
  * </p>
  * <ul>
- *     <li>the class of the entity to be inserted, or</li>
+ *     <li>the class of the entity to be detached, or</li>
  *     <li>{@code List<E>} or {@code E[]} where {@code E} is the class of the
- *     entities to be inserted.</li>
+ *     entities to be detached.</li>
  * </ul>
  * <p>The annotated method must be declared {@code void}.
  * </p>
--- a/api/src/main/java/jakarta/data/repository/stateful/Detach.java
+++ b/api/src/main/java/jakarta/data/repository/stateful/Detach.java
@@ -1,3 +1,20 @@
+/*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
 package jakarta.data.repository.stateful;
 
 import java.lang.annotation.*;
--- a/api/src/main/java/jakarta/data/repository/stateful/Detach.java
+++ b/api/src/main/java/jakarta/data/repository/stateful/Detach.java
@@ -32,7 +32,7 @@ import java.lang.annotation.Target;
  * method immediately makes one or more entities unmanaged, removing them
  * from the current persistence context associated with the repository.
  * </p>
- * <p>An {@code Detach} method accepts an instance or instances of an entity
+ * <p>A {@code Detach} method accepts an instance or instances of an entity
  * class. The method must have exactly one parameter whose type is either:
  * </p>
  * <ul>
--- a/api/src/main/java/jakarta/data/repository/stateful/Merge.java
+++ b/api/src/main/java/jakarta/data/repository/stateful/Merge.java
@@ -17,7 +17,11 @@
  */
 package jakarta.data.repository.stateful;
 
-import java.lang.annotation.*;
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
 /**
  * <p>Lifecycle annotation for repository methods of stateful repositories
--- a/api/src/main/java/jakarta/data/repository/stateful/Merge.java
+++ b/api/src/main/java/jakarta/data/repository/stateful/Merge.java
@@ -40,9 +40,9 @@ import java.lang.annotation.Target;
  * class. The method must have exactly one parameter whose type is either:
  * </p>
  * <ul>
- *     <li>the class of the entity to be inserted, or</li>
+ *     <li>the class of the entity to be merged, or</li>
  *     <li>{@code List<E>} or {@code E[]} where {@code E} is the class of the
- *     entities to be inserted.</li>
+ *     entities to be merged.</li>
  * </ul>
  * <p>The annotated method must have a return type that is the same as the
  * type of its parameter.
--- a/api/src/main/java/jakarta/data/repository/stateful/Merge.java
+++ b/api/src/main/java/jakarta/data/repository/stateful/Merge.java
@@ -1,0 +1,62 @@
+package jakarta.data.repository.stateful;
+
+import java.lang.annotation.*;
+
+/**
+ * <p>Lifecycle annotation for repository methods of stateful repositories
+ * which perform merge operations. The {@code Merge} annotation must
+ * not be applied to a method of a stateless repository.
+ * </p>
+ * <p>The {@code Merge} annotation indicates that the annotated repository
+ * method copies the state of one of more unmanaged entities to managed
+ * entities belonging to the current persistence context associated with the
+ * repository. The merge might change the persistent state of the managed
+ * entities, and might cause that state to be inserted or updated in the
+ * database. This typically occurs later, when the persistence context is
+ * flushed.
+ * </p>
+ * <p>A {@code Merge} method accepts an instance or instances of an entity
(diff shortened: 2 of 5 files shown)


gavinking:
One point about this: I'm assuming we're never going to want a `@Refresh`-like operation for stateless repositories.

If we *are* going to want that, it's going to suck a bit that `@Refresh` is declared in the `stateful` sub-package.

But I'm struggling to see how it's useful for stateless repositories. You can just use `Find`.

api/src/main/java/jakarta/data/repository/stateful/Detach.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * <p>Lifecycle annotation for repository methods of stateful repositories
+ * which perform detach operations. The {@code Detach} annotation must
+ * not be applied to a method of a stateless repository.
+ * </p>
+ * <p>The {@code Detach} annotation indicates that the annotated repository
+ * method immediately makes one or more entities unmanaged, removing them
+ * from the current persistence context associated with the repository.
+ * </p>
+ * <p>An {@code Detach} method accepts an instance or instances of an entity
```
njr-11:
```suggestion
 * <p>A {@code Detach} method accepts an instance or instances of an entity
```

api/src/main/java/jakarta/data/repository/stateful/Detach.java
```diff
@@ -0,0 +1,67 @@
...
+ * </p>
+ * <p>If an unmanaged entity is passed to a {@code Detach} method, the call
+ * has no effect.
+ * </p>
+ * <p>Every Jakarta Data provider which supports stateful repositories is
+ * required to accept a {@code Detach} method which conforms to this signature.
+ * Application of the {@code Detach} annotation to a method with any other
+ * signature is not portable between Jakarta Data providers. Furthermore,
+ * support for stateful repositories is optional. A Jakarta Data provider
+ * is not required to support stateful repositories.
```
njr-11, line 53: Adjust language to avoid optional features:
```suggestion
 * <p>A Jakarta Data provider that is capable of returning
 * {@code jakarta.persistence.EntityManager} from a resource accessor method is
 * required to accept a {@code Detach} method which conforms to this signature.
 * A {@code Detach} method that is not accepted must raise
 * {@link UnsupportedOperationException} or be rejected at compile time.
 * Application of the {@code Detach} annotation to a method with any other
 * signature is not portable between Jakarta Data providers.
```

api/src/main/java/jakarta/data/repository/stateful/Merge.java
```diff
@@ -0,0 +1,83 @@
...
+ * instance is unmanaged, the method returns a managed entity instance which
+ * must be distinct from the argument entity, but which holds the same
+ * persistent state.
+ * </p>
+ * <p>Every Jakarta Data provider which supports stateful repositories is
+ * required to accept a {@code Merge} method which conforms to this signature.
+ * Application of the {@code Merge} annotation to a method with any other
+ * signature is not portable between Jakarta Data providers. Furthermore,
+ * support for stateful repositories is optional. A Jakarta Data provider is
+ * not required to support stateful repositories.
```
njr-11, line 61: Adjust language to avoid optional features:
```suggestion
 * <p>A Jakarta Data provider that is capable of returning
 * {@code jakarta.persistence.EntityManager} from a resource accessor method is
 * required to accept a {@code Merge} method which conforms to this signature.
 * A {@code Merge} method that is not accepted must raise
 * {@link UnsupportedOperationException} or be rejected at compile time.
 * Application of the {@code Merge} annotation to a method with any other
 * signature is not portable between Jakarta Data providers.
```

api/src/main/java/jakarta/data/repository/stateful/Merge.java
```diff
@@ -0,0 +1,83 @@
...
+ * </p>
+ * <p>An event of type {@link jakarta.data.event.PreInsertEvent} must be
+ * raised before each record is inserted in the database. An event of type
+ * {@link jakarta.data.event.PostInsertEvent} must be raised after each record
+ * is successfully inserted. Similarly, an event of type
+ * {@link jakarta.data.event.PreUpdateEvent} must be raised before each record
+ * is updated. An event of type {@link jakarta.data.event.PostUpdateEvent}
+ * must be raised after each record is successfully updated.
+ * </p>
```
njr-11, line 70: It seems like implementations that rely on a database upsert operation might be unable to identify whether operation ends up being an insert or an update until after it happens, or maybe even not at all.  I noticed we don't have any corresponding requirement to sent insert/update events on Save.
```suggestion
```

api/src/main/java/jakarta/data/repository/stateful/Persist.java
```diff
@@ -0,0 +1,78 @@
...
+ * deletion via a call to a {@link Remove} method. If the instance was
+ * scheduled for deletion, then the {@code Persist} method undoes the effect
+ * of the {@code Remove} method.
+ * </p>
+ * <p>Every Jakarta Data provider which supports stateful repositories is
+ * required to accept a {@code Persist} method which conforms to this signature.
+ * Application of the {@code Persist} annotation to a method with any other
+ * signature is not portable between Jakarta Data providers. Furthermore,
+ * support for stateful repositories is optional. A Jakarta Data provider
+ * is not required to support stateful repositories.
```
njr-11, line 59: Adjust language to avoid optional features:
```suggestion
 * <p>A Jakarta Data provider that is capable of returning
 * {@code jakarta.persistence.EntityManager} from a resource accessor method is
 * required to accept a {@code Persist} method which conforms to this signature.
 * A {@code Persist} method that is not accepted must raise
 * {@link UnsupportedOperationException} or be rejected at compile time.
 * Application of the {@code Persist} annotation to a method with any other
 * signature is not portable between Jakarta Data providers.
```

api/src/main/java/jakarta/data/repository/stateful/Refresh.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * <p>Lifecycle annotation for repository methods of stateful repositories
+ * which perform refresh operations. The {@code Refresh} annotation must
+ * not be applied to a method of a stateless repository.
+ * </p>
+ * <p>The {@code Refresh} annotation indicates that the annotated repository
+ * method immediately reloads the state of the given entity from the database,
+ * overwriting the persistent state previously held by the entity.
+ * </p>
+ * <p>An {@code Refresh} method accepts a managed instance or instances of an
```
njr-11:
```suggestion
 * <p>A {@code Refresh} method accepts a managed instance or instances of an
```

api/src/main/java/jakarta/data/repository/stateful/Refresh.java
```diff
@@ -0,0 +1,68 @@
...
+ * </p>
+ * <p>If an unmanaged entity is passed to a {@code Refresh} method, the method
+ * must throw {@link IllegalArgumentException}.
+ * </p>
+ * <p>Every Jakarta Data provider which supports stateful repositories is
+ * required to accept a {@code Refresh} method which conforms to this signature.
+ * Application of the {@code Refresh} annotation to a method with any other
+ * signature is not portable between Jakarta Data providers. Furthermore,
+ * support for stateful repositories is optional. A Jakarta Data provider
+ * is not required to support stateful repositories.
```
njr-11, line 54: Adjust language to avoid optional features:
```suggestion
 * <p>A Jakarta Data provider that is capable of returning
 * {@code jakarta.persistence.EntityManager} from a resource accessor method is
 * required to accept a {@code Refresh} method which conforms to this signature.
 * A {@code Refresh} method that is not accepted must raise
 * {@link UnsupportedOperationException} or be rejected at compile time.
 * Application of the {@code Refresh} annotation to a method with any other
 * signature is not portable between Jakarta Data providers.
```

api/src/main/java/jakarta/data/repository/stateful/Remove.java
```diff
@@ -0,0 +1,77 @@
...
+ * <p>Lifecycle annotation for repository methods of stateful repositories
+ * which perform remove operations. The {@code Remove} annotation must
+ * not be applied to a method of a stateless repository.
+ * </p>
+ * <p>The {@code Remove} annotation indicates that the annotated repository
+ * method schedules one or more managed entities for deletion in the database.
+ * Deletion might occur immediately, when the annotated repository method is
+ * invoked, or it might occur later, when the persistence context is flushed.
+ * </p>
+ * <p>An {@code Remove} method accepts a managed instance or instances of an
```
njr-11:
```suggestion
 * <p>A {@code Remove} method accepts a managed instance or instances of an
```

api/src/main/java/jakarta/data/repository/stateful/Remove.java
```diff
@@ -0,0 +1,77 @@
...
+ * insertion in the database via a call to a {@link Persist} method, but was
+ * not yet inserted in the database, then the {@code Refresh} method undoes the
+ * effect of the {@code Persist} method.
+ * </p>
+ * <p>Every Jakarta Data provider which supports stateful repositories is
+ * required to accept a {@code Remove} method which conforms to this signature.
+ * Application of the {@code Remove} annotation to a method with any other
+ * signature is not portable between Jakarta Data providers. Furthermore,
+ * support for stateful repositories is optional. A Jakarta Data provider
+ * is not required to support stateful repositories.
```
njr-11, line 58: Adjust language to avoid optional features:
```suggestion
 * <p>A Jakarta Data provider that is capable of returning
 * {@code jakarta.persistence.EntityManager} from a resource accessor method is
 * required to accept a {@code Remove} method which conforms to this signature.
 * A {@code Remove} method that is not accepted must raise
 * {@link UnsupportedOperationException} or be rejected at compile time.
 * Application of the {@code Remove} annotation to a method with any other
 * signature is not portable between Jakarta Data providers.
```

api/src/main/java/jakarta/data/repository/stateful/Remove.java
```diff
@@ -0,0 +1,77 @@
...
+ * </p>
+ * <p>An event of type {@link jakarta.data.event.PreDeleteEvent} must be
+ * raised before each record is inserted in the database. An event of type
+ * {@link jakarta.data.event.PreDeleteEvent} must be raised after each record
+ * is successfully inserted.
```
njr-11:
```suggestion
 * <p>An event of type {@link jakarta.data.event.PreDeleteEvent} must be
 * raised before each record is deleted from the database. An event of type
 * {@link jakarta.data.event.PreDeleteEvent} must be raised after each record
 * is successfully deleted.
```

api/src/main/java/jakarta/data/repository/stateful/Merge.java
gavinking: Well that's an interesting point. I believe the answer is that JPA-style `merge()` cannot actually be implemented using upserts (at least not if you have owned collections or orphan deletion), but I would have to think about it a bit more carefully.

api/src/main/java/jakarta/data/repository/stateful/Persist.java
gavinking: Not sure what you mean by this. Any database is in principle _capable_ of being used as the backing store for a programming model based on stateful persistence contexts.

The question is whether the _Jakarta Data implementation_ supports/implements that.

njr-11: The Jakarta EE profiles/platform can't have optional features, so I was trying to figure out how we can define the requirement that won't make it optional.  Given that we cannot tie it to database capability as you pointed out, maybe we can tie it to whether the Jakarta Data provider is backed by Jakarta Persistence, although I was hoping not to do that so that NoSQL could also provide stateful.  I'll try to think of something.

gavinking: > maybe we can tie it to whether the Jakarta Data provider is backed by Jakarta Persistence

That was what I intended to do. But I think that sort of requirement belongs in the spec, not in the Javadoc.

gavinking:
> Personally I am not sure this is necessary. 

It's one of the top requests we have from users, and a barrier to easy migration for people coming from Spring Data or from DeltaSpike, so that's why the need for this turns out to be more acute than I had estimated.

> or a JPA specific extension

This is an option. We could indeed put it in a separate `jar`. I have no objection to that.

api/src/main/java/jakarta/data/repository/stateful/Persist.java
njr-11: > > maybe we can tie it to whether the Jakarta Data provider is backed by Jakarta Persistence
> 
> That was what I intended to do. But I think that sort of requirement belongs in the spec, not in the Javadoc.

Right, all the discussion about the Data Provider would be better moved into the spec.  The end user, who typically only reads Javadoc for what they are using, does need a small statement of some sort in the Javadoc to let them know when it is valid for their application to use stateful.  That could end up being a single line in `StatefulRepository` Javadoc if we end up introducing that interface, but we still need to think about that more.

gavinking: I don't ... disagree ... but I was adapting this language which we have in the pre-existing lifecycle annotations:

```
 * <p>All Jakarta Data providers are required to accept an {@code Insert} method which conforms to this signature.
 * Application of the {@code Insert} annotation to a method with any other signature is not portable between Jakarta
 * Data providers.
```

njr-11:
> How should a `@Remove` method treat an unmanaged entity? Choices are:
> 
>     * error, as I've written here, or
> 
>     * ignore it, if the repository can determine that the entity does not exist in the database, which is what JPA does.

I would vote for consistency with JPA.  Ideally it should be a valid implementation to pass these operations directly through to a stateful EntityManager.

gavinking:
> my question is the name.

Mine too.

> Are you sure about the name be: `jakarta-data-stateful-api` ?

Nope :-/

otaviojava:
> > my question is the name.
> 
> Mine too.
> 
> > Are you sure about the name be: `jakarta-data-stateful-api` ?
> 
> Nope :-/

I will create the module anyway; in the worse scenario we can rename it.

otaviojava:
> > in the worse scenario we can rename it.
> 
> Yeah, that's what I was thinking.

https://github.com/jakartaee/data/pull/1133

api/src/main/java/jakarta/data/repository/stateful/Persist.java
njr-11: > I don't ... disagree ... but I was adapting this language which we have in the pre-existing lifecycle annotations:
> 
> ```
>  * <p>All Jakarta Data providers are required to accept an {@code Insert} method which conforms to this signature.
>  * Application of the {@code Insert} annotation to a method with any other signature is not portable between Jakarta
>  * Data providers.
> ```

The above language is fine because it doesn't introduce anything optional.  First, it puts a requirement on all providers that indicates the intended usage of the API.  Nothing is optional there.  The part about "any other signature is not portable" is unnecessary and would be otherwise assumed, but it's fine to call out that if used in ways other than specified it won't be portable. That isn't defining an optional feature because it doesn't define any feature at all.  That second part corresponds to where you wrote,

```
 * Application of the {@code Persist} annotation to a method with any other
 * signature is not portable between Jakarta Data providers.
```

and I didn't propose changes to that wording at all.

The problem with optional features is where we explicitly state:
```
A Jakarta Data provider
 * is not required to support stateful repositories.
```

and also where optionality can be inferred from the wording, `Every Jakarta Data provider which supports stateful repositories`, which implies there can also be Jakarta Data providers that don't support stateful repositories.

gavinking:
We can close this one now, right? 

And start work on the spec side of this....

otaviojava:
> We can close this one now, right?
> 
> And start work on the spec side of this....

Yep, I copied and pasted your code and accepted all the reviews.
Maybe start checking what is there and fix if there is something wrong.

PR #1135, opened 2025-06-07
Update the code structure using Java 21 features when it makes sense

otaviojava:
This pull request includes several changes aimed at improving code readability, simplifying expressions, and enhancing test assertions. Key updates include replacing method calls with more concise alternatives, switching from `if-else` chains to `switch` expressions, and simplifying test assertions for better readability.

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -45,6 +45,7 @@ import jakarta.data.restrict.Restrict;
 import jakarta.data.restrict.Restriction;
 
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import java.util.Set;
 
 /**
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -45,7 +45,6 @@ import jakarta.data.restrict.Restrict;
 import jakarta.data.restrict.Restriction;
 
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
 import java.util.Set;
 
 /**
--- a/api/src/test/java/jakarta/data/OrderTest.java
+++ b/api/src/test/java/jakarta/data/OrderTest.java
@@ -45,9 +45,7 @@ class OrderTest {
         var list = List.of(Sort.desc("priority"));
         var order = Order.by(list);
 
-        SoftAssertions.assertSoftly(soft -> {
-            soft.assertThat(order.sorts()).containsExactlyElementsOf(list);
-        });
+        SoftAssertions.assertSoftly(soft -> soft.assertThat(order.sorts()).containsExactlyElementsOf(list));
     }
 
     @Test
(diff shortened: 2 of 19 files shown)


api/src/main/java/module-info.java
```diff
@@ -45,7 +45,6 @@
 import jakarta.data.restrict.Restriction;
 
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
```
njr-11: I think you were only able to remove this because the Javadoc mistakenly has `{@code CompletionStage}` rather than `{@link CompletionStage}`

api/src/test/java/jakarta/data/OrderTest.java
```diff
@@ -45,9 +45,7 @@ void shouldCreateOrderFromList() {
         var list = List.of(Sort.desc("priority"));
         var order = Order.by(list);
 
-        SoftAssertions.assertSoftly(soft -> {
-            soft.assertThat(order.sorts()).containsExactlyElementsOf(list);
-        });
+        SoftAssertions.assertSoftly(soft -> soft.assertThat(order.sorts()).containsExactlyElementsOf(list));
```
njr-11: This does not appear to do anything except increase the length of a line beyond the recently agreed upon line length limitation.
```suggestion
        SoftAssertions.assertSoftly(soft ->
            soft.assertThat(order.sorts()).containsExactlyElementsOf(list));
```
It looks like this was done in several other places.

api/src/main/java/module-info.java
otaviojava: returned the import.

api/src/test/java/jakarta/data/spi/expression/literal/LiteralRecordTest.java
```diff
@@ -63,16 +61,12 @@ void shouldNotEqualDifferentValues() {
...
+        SoftAssertions.assertSoftly(soft -> soft.assertThat(one).isNotEqualTo(two));
     }
 
     @Test
     @DisplayName("should throw NullPointerException when value is null")
     void shouldThrowWhenValueIsNull() {
-        org.junit.jupiter.api.Assertions.assertThrows(NullPointerException.class, () -> {
-            new LiteralRecord<>(null);
-        }, "The value argument is required");
+        org.junit.jupiter.api.Assertions.assertThrows(NullPointerException.class, () -> new LiteralRecord<>(null), "The value argument is required");
```
njr-11: Another example of where the PR makes formatting changes that condense code onto really long lines (in this case 150 characters in a single line) contrary to the agreed upon line length limit.

```suggestion
        org.junit.jupiter.api.Assertions.assertThrows(
                NullPointerException.class,
                () -> new LiteralRecord<>(null),
                "The value argument is required");
```

njr-11 (approved):
> @njr-11 how about now?

That's fine. I approved it, but my prior comment about formatting-only changes that combine code onto really long single lines that violate the line length limit was a general comment, not just a single place where it was done.  I'll note another one here, but I'd prefer not to be focused on trivial things like this in the code reviews.  Hopefully we can rely on the code authors to enforce this sort of thing themselves.

PR #1136, opened 2025-06-08
Generate documentation to jakarta.data.constraint package

otaviojava:
Continuing my journey with the spec implementation, I've generated brief documentation for the classes in the `jakarta.data.constraint` package.

I took a baby step to document those interfaces, where I've included documentation at the interface level only, avoiding the methods.


njr-11 (changes requested):
The tracking issue #1086 for adding Javadoc has Constraint further down the list, in a place that indicates we have not agreed that it should be started yet, per the discussion on the Jakarta Data calls.

In any case, the documentation in this PR is covering Restrictions rather than Constraints, so it should not be merged regardless.

PR #1140, opened 2025-06-12
Include more information on Restriction

otaviojava:
I've included the documentation spec of `Restriction`, which is based on what I am doing; thus, it does not mean that it is the right direction (I put it here also to discuss).

I wrote mainly to explain some corner cases that I encountered during the process.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -833,6 +833,44 @@ When a repository method is called with a null value as an argument to one of it
 
 NOTE: The behavior of a query method when the method is called with a null argument is not defined by this specification, and is not portable between Jakarta Data providers.
 
+==== Restriction parameters
+
+The `Restriction<T>` type is a special parameter type that can be used to find records and entities from the repositories. A repository method may declare a parameter of this type, which allows the application to specify additional restrictions on the query results.
+Different from the other queries, such as, `Find` annotation, the `Restriction<T>` can generate dynamic queries, furthermore, it includes the capability of doing a safe and fluent query combined with <<metamodel-java-processor>>
+
+To use the `Restriction<T>` it should be declared as the first parameter of the repository method, and it must not be used in combination with any of the following annotations: `@Query`, `@Save`, `@Insert`, `@Find`, `@Update`, or `@Delete`.
+
+The code below shows an example of a repository method that uses a `Restriction<T>` where two different restrictions are applied to the `Product` entity. The first restriction filters products by name, and the second restriction filters products by price.
+
+[source,java]
+----
+@Repository
+public interface ProductRepository {
+   List<Product> restriction(Restriction<Product> restriction);
+}
+...
+
+var pencils = repository.restriction(_Product.name.equalTo("Pencil"));
+var expensive = repository.restriction(_Product.price.greaterThan(BigDecimal.TEN));
+----
+
+You can combine multiple restrictions using the `any` and `all` methods of the `jakarta.data.restrict.Restrict` type. The following example shows how to combine two restrictions to find products that are expensive pencils:
+
+[source,java]
+----
+var expensivePencils = repository.restriction(Restrict.all(_Product.name.equalTo("Pencil"), _Product.price.greaterThan(BigDecimal.TEN)));
+----
+
+The Restriction queries can be used with:
+
+* `OrderBy` annotations to specify the order of the results.
+* `Sort` and `Order` parameters to specify the order of the results.
+* `PageRequest` parameters to specify pagination and sorting of the results.
+* `Limit` parameters to limit the number of results returned.
+* Return types that are for pagination, such as `Page`, `CursoredPage`.
+* Return types that are for collections, such as `List`, `Set`, `Stream`, or and `T[]`.
+- Return types for a query which returns a single result of type `T`, the type `T` itself, or `Optional<T>`,
+
 === Asynchronous repositories
 
 An _asynchronous repository method_ is a repository method which returns an object representing a value which will eventually be obtained from the database, but which might not yet be available.
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -877,13 +877,17 @@ Valid method signatures include:
 ----
 public interface ProductRepository {
 
-    List<Product> restriction(Restriction<Product> restriction);
+    @Find
+    List<Product> search(Restriction<Product> restriction);
 
-    Page<Product> restriction(Restriction<Product> restriction, PageRequest pageRequest);
+    @Find
+    Page<Product> search(Restriction<Product> restriction, PageRequest pageRequest, Order<Product> order);
 
-    List<Product> restriction(Restriction<Product> restriction, Sort<Product> order);
+    @Find
+    List<Product> search(Restriction<Product> restriction, Sort<Product> order);
 
-    List<Product> restriction(Restriction<Product> restriction, Order<Product> order);
+    @Query("WHERE (product.numAvailable > 0)")
+    List<Product> search(Restriction<Product> restriction, Order<Product> order);
 
     @OrderBy(_Product.PRICE)
     List<Product> restrictionOrderByPriceAsc(Restriction<Product> restriction);
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -287,9 +287,10 @@ A repository method must throw `UnsupportedOperationException` if it has:
 - a parameter of type `PageRequest` or `Limit`, in combination with the keyword `First`,
 - a `@First` annotation, in combination with the keyword `First`, or
 - more than one parameter of type `Order`.
-- more than one parameter of type `Restriction`.
-- a `Restriction` is not at the first position in the parameter.
-- a `Restriction` is combinded with `Query`, `Save`, `Insert`, `Find`, `Update`, or `Delete` annotations.
+- It declares more than one parameter of type `Restriction`.
+- It declares a parameter of type `Restriction` that is not the first method parameter.
+- It declares a `Restriction` parameter in combination with any of the following annotations:
+`@Query`, `@Save`, `@Insert`, `@Find`, `@Update`, or `@Delete`.
 
 Alternatively, a Jakarta Data provider is permitted to reject such a repository method declaration at compile time.
 
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -288,7 +288,6 @@ A repository method must throw `UnsupportedOperationException` if it has:
 - a `@First` annotation, in combination with the keyword `First`,
 - more than one parameter of type `Order`,
 - more than one parameter of type `Restriction`, or
-- a `Restriction` parameter and is a lifecycle method.
 
 Alternatively, a Jakarta Data provider is permitted to reject such a repository method declaration at compile time.
 
@@ -857,8 +856,6 @@ The following is *invalid* because the method does not follow the Query by Metho
 ----
 List<Product> findByName(Restriction<Product> restriction);
 ----
-
-In both cases, Restriction<T> must not be used with lifecycle methods.
 
 The example below shows a repository method that accepts a restriction to filter `Product` entities:
 
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -851,7 +851,7 @@ List<Product> search(Restriction<Product> restriction);
 Page<Product> findAvailable(Restriction<Product> restriction, PageRequest pageRequest, Order<Product> order);
 ----
 
-The following is *invalid* and will throw a `MappingException`:
+The following is *invalid* because the method does not follow the Query by Method Name pattern of supplying a parameter per condition defined by the method name, nor does the method have an annotation that allows a `Restriction` parameter to be used.
 
 [source,java]
 ----
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -898,6 +898,7 @@ Valid method signatures include:
 ----
 public interface ProductRepository {
 
+    @Find
     List<Product> search(Restriction<Product> restriction);
 
     @Find
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -844,11 +844,11 @@ For example:
 
 [source,java]
 ----
-@Query("WHERE (product.numAvailable > 0)")
-List<Product> search(Restriction<Product> restriction, Order<Product> order);
-
 @Find
-Page<Product> findAvailable(Restriction<Product> restriction, PageRequest pageRequest);
+List<Product> search(Restriction<Product> restriction);
+
+@Query("WHERE (product.numAvailable > 0)")
+Page<Product> findAvailable(Restriction<Product> restriction, PageRequest pageRequest, Order<Product> order);
 ----
 
 The following is *invalid* and will throw a `MappingException`:
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -836,7 +836,7 @@ The `Restriction<T>` type is a special parameter used to dynamically construct q
 
 When using `Restriction<T>` in repository methods:
 
-- The method *must* be annotated with either `@Find`, '@Delete', or `@Query`. Without one of these annotations, the use of `Restriction<T>` is *not valid* and results in an `UnsupportedException`.
+- The use of `Restriction<T>` is only considered valid in portable applications when used as a parameter to methods annotated with `@Find`, `@Delete`, or `@Query`. Methods using `Restriction<T>` without one of these annotations are not covered by the specification. Jakarta Data providers may reject such methods by throwing an `UnsupportedOperationException`, or they may choose to support them as vendor-specific extensions.
 - The filtering that is defined by the `@Query` annotation or parameters of the method that is annotated `@Find` or `@Delete` is combined with the additional filtering specified by the `Restriction<T>` parameter such that matching entities must satisfy both.
 
 For example:
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -836,7 +836,7 @@ The `Restriction<T>` type is a special parameter used to dynamically construct q
 
 When using `Restriction<T>` in repository methods:
 
-- The use of `Restriction<T>` is only considered valid in portable applications when used as a parameter to methods annotated with `@Find`, `@Delete`, or `@Query`. Methods using `Restriction<T>` without one of these annotations are not covered by the specification. Jakarta Data providers may reject such methods by throwing an `UnsupportedOperationException`, or they may choose to support them as vendor-specific extensions.
+- The use of `Restriction<T>` is only considered valid in portable applications when used as a parameter to methods annotated with `@Find`, `@Delete`, or `@Query`. Methods using `Restriction<T>` without one of these annotations are not covered by the specification. Jakarta Data providers may reject such methods by throwing an `UnsupportedOperationException`.
 - The filtering defined by the `@Query` annotation or the parameters of methods annotated with `@Find` or `@Delete` is combined with any additional filtering specified by a `Restriction<T>` parameter.
 
 This specification defines support for the `Restriction<T>` parameter only on methods annotated with `@Find`, `@Delete`, or `@Query`.
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -287,7 +287,7 @@ A repository method must throw `UnsupportedOperationException` if it has:
 - a parameter of type `PageRequest` or `Limit`, in combination with the keyword `First`,
 - a `@First` annotation, in combination with the keyword `First`, or
 - more than one parameter of type `Order`.
-- It declares more than one parameter of type `Restriction`.
+- more than one parameter of type `Restriction`, or
 - a `Restriction` parameter and is a lifecycle method.
 
 Alternatively, a Jakarta Data provider is permitted to reject such a repository method declaration at compile time.
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -837,7 +837,7 @@ The `Restriction<T>` type is a special parameter used to dynamically construct q
 
 When using `Restriction<T>` in repository methods:
 
-- The method *must* be annotated with either `@Find` or `@Query`. Without one of these annotations, the use of `Restriction<T>` is *not valid* and results in a `MappingException`.
+- The method *must* be annotated with either `@Find`, '@Delete', or `@Query`. Without one of these annotations, the use of `Restriction<T>` is *not valid* and results in an `UnsupportedException`.
 - When `@Find` or `@Query` is present, `Restriction<T>` can appear in *any position* in the parameter list. In these cases, the `Restriction<T>` provides the full query logic programmatically, and any query derivation from the method name is ignored.
 
 For example:
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -289,6 +289,7 @@ A repository method must throw `UnsupportedOperationException` if it has:
 - more than one parameter of type `Order`.
 - more than one parameter of type `Restriction`.
 - a `Restriction` is not at the first position in the parameter.
+- a `Restriction` is combinded with `Query`, `Save`, `Insert`, `Find`, `Update`, or `Delete` annotations.
 
 Alternatively, a Jakarta Data provider is permitted to reject such a repository method declaration at compile time.
 
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -287,6 +287,8 @@ A repository method must throw `UnsupportedOperationException` if it has:
 - a parameter of type `PageRequest` or `Limit`, in combination with the keyword `First`,
 - a `@First` annotation, in combination with the keyword `First`, or
 - more than one parameter of type `Order`.
+- more than one parameter of type `Restriction`.
+- a `Restriction` is not at the first position in the parameter.
 
 Alternatively, a Jakarta Data provider is permitted to reject such a repository method declaration at compile time.
 
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -898,7 +898,9 @@ public interface ProductRepository {
     @OrderBy(_Product.ID)
     CursoredPage<Product> cursoredPage(Restriction<Product> restriction, PageRequest pageRequest);
 
+    @Find
     @OrderBy(_Product.PRICE)
+    @OrderBy(_Product.ID)
     Page<Product> page(Restriction<Product> restriction, PageRequest pageRequest);
 }
 ----
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -288,8 +288,7 @@ A repository method must throw `UnsupportedOperationException` if it has:
 - a `@First` annotation, in combination with the keyword `First`, or
 - more than one parameter of type `Order`.
 - It declares more than one parameter of type `Restriction`.
-- It declares a `Restriction` parameter in combination with any of the following annotations:
-`@Query`, `@Save`, `@Insert`, `@Find`, `@Update`, or `@Delete`.
+- a `Restriction` parameter and is a lifecycle method.
 
 Alternatively, a Jakarta Data provider is permitted to reject such a repository method declaration at compile time.
 
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -837,7 +837,15 @@ The `Restriction<T>` type is a special parameter used to dynamically construct q
 When using `Restriction<T>` in repository methods:
 
 - The use of `Restriction<T>` is only considered valid in portable applications when used as a parameter to methods annotated with `@Find`, `@Delete`, or `@Query`. Methods using `Restriction<T>` without one of these annotations are not covered by the specification. Jakarta Data providers may reject such methods by throwing an `UnsupportedOperationException`, or they may choose to support them as vendor-specific extensions.
-- The filtering that is defined by the `@Query` annotation or parameters of the method that is annotated `@Find` or `@Delete` is combined with the additional filtering specified by the `Restriction<T>` parameter such that matching entities must satisfy both.
+- The filtering defined by the `@Query` annotation or the parameters of methods annotated with `@Find` or `@Delete` is combined with any additional filtering specified by a `Restriction<T>` parameter.
+
+
+Vendors may choose to support additional usage of `Restriction<T>` on other types of methods or annotations. Such usages are considered non-portable and outside the scope of this specification.
+
+[NOTE]
+====
+Applications that use `Restriction<T>` outside the supported method types defined by this specification may not be portable across Jakarta Data implementations.
+====
 
 For example:
 
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -893,8 +893,10 @@ public interface ProductRepository {
     @OrderBy(_Product.PRICE)
     List<Product> byPriceAsc(Restriction<Product> restriction);
 
+    @Find
     @OrderBy(_Product.PRICE)
-    CursoredPage<Product> cursor(Restriction<Product> restriction, PageRequest pageRequest);
+    @OrderBy(_Product.ID)
+    CursoredPage<Product> cursoredPage(Restriction<Product> restriction, PageRequest pageRequest);
 
     @OrderBy(_Product.PRICE)
     Page<Product> page(Restriction<Product> restriction, PageRequest pageRequest);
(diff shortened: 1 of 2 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -287,6 +287,10 @@ A repository method must throw `UnsupportedOperationException` if it has:
 - a parameter of type `PageRequest` or `Limit`, in combination with the keyword `First`,
 - a `@First` annotation, in combination with the keyword `First`, or
 - more than one parameter of type `Order`.
+- It declares more than one parameter of type `Restriction`.
+- It declares a parameter of type `Restriction` that is not the first method parameter.
```
gavinking: I don't see the reason for this second limitation.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -287,6 +287,10 @@ A repository method must throw `UnsupportedOperationException` if it has:
...
+- It declares a `Restriction` parameter in combination with any of the following annotations:
+`@Query`, `@Save`, `@Insert`, `@Find`, `@Update`, or `@Delete`.
```
gavinking: `Restriction` can be used with `@Query` or `@Find`.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +833,81 @@ When a repository method is called with a null value as an argument to one of it
 
 NOTE: The behavior of a query method when the method is called with a null argument is not defined by this specification, and is not portable between Jakarta Data providers.
 
+==== Restriction parameters
+
+The `Restriction<T>` type is a special parameter used to dynamically construct queries in a fluent and type-safe manner. Unlike fixed queries defined via annotations such as `@Find` or `@Query`, restrictions allow the application to define conditions programmatically, often using a generated metamodel class (see <<metamodel-java-processor>>).
+
```
gavinking: The idea is that `Restriction` is used _together_ with `@Find` or `@Query`.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -287,6 +287,10 @@ A repository method must throw `UnsupportedOperationException` if it has:
 - a parameter of type `PageRequest` or `Limit`, in combination with the keyword `First`,
 - a `@First` annotation, in combination with the keyword `First`, or
 - more than one parameter of type `Order`.
+- It declares more than one parameter of type `Restriction`.
```
njr-11: For consistency with the style of the prior bullet points,
```suggestion
- more than one parameter of type `Restriction`, or
```

spec/src/main/asciidoc/repository.asciidoc
njr-11: It is the intended usage of `Restriction` that restrictions determined at run time be added onto what is predefined.  For example,
```
products.search(namePattern, Restrict.all(_Product.price.lessThanEqual(maxPrice),
                                          _Product.color.in(Color.BLUE)));
```
```suggestion
```

spec/src/main/asciidoc/repository.asciidoc
```suggestion
- a `Restriction` parameter and is a lifecycle method.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -287,6 +287,10 @@ A repository method must throw `UnsupportedOperationException` if it has:
 - a parameter of type `PageRequest` or `Limit`, in combination with the keyword `First`,
 - a `@First` annotation, in combination with the keyword `First`, or
 - more than one parameter of type `Order`.
```
njr-11: Now that this isn't the final bullet point anymore, it is no longer appropriate to include the word `or` here. It needs to be moved to the new second-to-last point.
```suggestion
- a `@First` annotation, in combination with the keyword `First`,
- more than one parameter of type `Order`.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +833,81 @@ When a repository method is called with a null value as an argument to one of it
 
 NOTE: The behavior of a query method when the method is called with a null argument is not defined by this specification, and is not portable between Jakarta Data providers.
 
+==== Restriction parameters
+
+The `Restriction<T>` type is a special parameter used to dynamically construct queries in a fluent and type-safe manner. Unlike fixed queries defined via annotations such as `@Find` or `@Query`, restrictions allow the application to define conditions programmatically, often using a generated metamodel class (see <<metamodel-java-processor>>).
```
```suggestion
The `Restriction<T>` type is a special parameter used to dynamically construct queries in a fluent and type-safe manner. Valid as parameters to methods that perform queries, such as methods annotated `@Find` or `@Delete`, restrictions allow the application to define additional conditions programmatically, often using a generated metamodel class (see <<metamodel-java-processor>>).
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +833,81 @@ When a repository method is called with a null value as an argument to one of it
...
+
+To use `Restriction<T>`, it must appear as the first parameter in a repository method and must not be combined with any of the following annotations: `@Query`, `@Save`, `@Insert`, `@Find`, `@Update`, or `@Delete`.
```
```suggestion
To use `Restriction<T>`, it must appear after any parameters for query conditions that are defined upfront and must not be used on a lifecycle method.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +833,81 @@ When a repository method is called with a null value as an argument to one of it
...
+
+[source,java]
+----
+@Repository
+public interface ProductRepository {
+   List<Product> restriction(Restriction<Product> restriction);
+}
+...
+var pencils = repository.restriction(_Product.name.equalTo("Pencil"));
+var expensive = repository.restriction(_Product.price.greaterThan(BigDecimal.TEN));
```
njr-11: This was not a valid repository method because it neither is annotated `@Find`, `@Query`, nor does it follow Query by Method Name.
```suggestion
   @Find
   List<Product> search(@By(_Product.NAME) name, Restriction<Product> restriction);
}
...
List<Product> redPencils = products.search("pencil", _Product.color.equalTo(Color.RED));
List<Product> expensivePencils = products.search("pencil", _Product.price.greaterThan(BigDecimal.TEN));
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +833,81 @@ When a repository method is called with a null value as an argument to one of it
...
+To combine multiple conditions, use the static methods `Restrict.all(...)` or `Restrict.any(...)`. These allow you to express conjunctions (AND) or disjunctions (OR) in a fluent way:
+
+[source,java]
+----
+var expensivePencils = repository.restriction(
+    Restrict.all(
+        _Product.name.equalTo("Pencil"),
+        _Product.price.greaterThan(BigDecimal.TEN)
+    )
+);
```
```suggestion
List<String> expensiveRedPencils = products.search(
    "pencil",
    Restrict.all(
        _Product.color.equalTo(Color.RED),
        _Product.price.greaterThan(BigDecimal.TEN)
    )
);
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +833,81 @@ When a repository method is called with a null value as an argument to one of it
...
+----
+
+Repository methods using `Restriction<T>` support the following:
+
+* `OrderBy` annotation to define static sorting.
+* `Sort<T>` or `Order<T>` parameters to define dynamic sorting.
+* `PageRequest` for pagination and sorting.
+* `Limit` for limiting the number of results.
+* Return types such as `Page<T>`, `CursoredPage<T>`, `List<T>`, `Set<T>`, `Stream<T>`, or arrays `T[]`.
+* Single-result queries may return `T` or `Optional<T>`.
```
njr-11: All of this is implied by the type of repository method (`@Find`, `@Query`, `@Delete`) upon which the Repository parameter appears. Duplicating the information risks becoming out of sync, so we should not do that here. 
```suggestion
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +833,81 @@ When a repository method is called with a null value as an argument to one of it
...
+----
+public interface ProductRepository {
+
+    List<Product> restriction(Restriction<Product> restriction);
+
+    Page<Product> restriction(Restriction<Product> restriction, PageRequest pageRequest);
+
+    List<Product> restriction(Restriction<Product> restriction, Sort<Product> order);
+
+    List<Product> restriction(Restriction<Product> restriction, Order<Product> order);
```
```suggestion
    @Find
    List<Product> search(Restriction<Product> restriction);

    @Find
    Page<Product> search(Restriction<Product> restriction, PageRequest pageRequest, Order<Product> order);

    @Find
    List<Product> search(Restriction<Product> restriction, Sort<Product> order);

    @Query("WHERE (product.numAvailable > 0)")
    List<Product> search(Restriction<Product> restriction, Order<Product> order);
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +833,81 @@ When a repository method is called with a null value as an argument to one of it
...
+
+    @OrderBy(_Product.PRICE)
+    List<Product> restrictionOrderByPriceAsc(Restriction<Product> restriction);
```
```suggestion
    @Find
    @OrderBy(_Product.PRICE)
    List<Product> byPriceAsc(Restriction<Product> restriction);
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +833,81 @@ When a repository method is called with a null value as an argument to one of it
...
+
+    @OrderBy(_Product.PRICE)
+    CursoredPage<Product> cursor(Restriction<Product> restriction, PageRequest pageRequest);
```
njr-11: Pagination requires consistent and predictable order. Multiple products might have the same price.
```suggestion
    @Find
    @OrderBy(_Product.PRICE)
    @OrderBy(_Product.ID)
    CursoredPage<Product> cursoredPage(Restriction<Product> restriction, PageRequest pageRequest);
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +833,81 @@ When a repository method is called with a null value as an argument to one of it
...
+
+    @OrderBy(_Product.PRICE)
+    Page<Product> page(Restriction<Product> restriction, PageRequest pageRequest);
```
```suggestion
    @Find
    @OrderBy(_Product.PRICE)
    @OrderBy(_Product.ID)
    Page<Product> page(Restriction<Product> restriction, PageRequest pageRequest);
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +833,81 @@ When a repository method is called with a null value as an argument to one of it
...
+}
+----
+
+[NOTE]
+====
+- `Restriction<T>` instances are immutable and can be reused across multiple repository calls without side effects.
+
+- Passing `null` as the `Restriction<T>` argument results in a `NullPointerException`.
+
+- Repository methods returning `void` are not valid when using `Restriction<T>`. This constraint type is intended for querying only and must return a value.
```
njr-11: The `void` return type is important when Restriction is used on `@Delete` methods.  Defer to what is already defined for repository methods on enforcing return types.
```suggestion
```

spec/src/main/asciidoc/repository.asciidoc
otaviojava: Ok, I saw gavin comment as well.
Thus, time to return to the implementation.

At least, now I have this clear.

thanks.

njr-11:
> For example, what happens in this scenario?
> 
> ```java
> List<Product> findByName(Restriction<Product>  restriction);
> ```

MappingException because findByName requires a parameter to provide the name and none is provided.


> Even here:
> 
> ```java
> @Find 
> List<Product> findByName(@By("name") String name, Restriction<Product>  restriction);
> ```

The above is the recommended usage.  I think I have stated this elsewhere, but a well-written repository method should aim to enforce all of the fixed requirements upfront (in this case, that it requires a name to be specified) and then allow the user to optionally provide additional restrictions at run time where that makes sense.  In this case, it seems reasonable that a web application would have an interface that requires the user to input a product name and makes all other search criteria optional.  This method works well for that (although it could probably also benefit from some sorting and pagination).

otaviojava:
> I don't know what you mean. DTO?

Data Transfer Object, at the end, it is our Projection record:

https://github.com/jakartaee/data/issues/1141

otaviojava:
> List<Product> findByName(Restriction<Product>  restriction);

@njr-11 can we agree that it would be a valid scenario here?

otaviojava:
> > For example, what happens in this scenario?
> > ```java
> > List<Product> findByName(Restriction<Product>  restriction);
> > ```
> 

@njr-11 can we agree that it would be a valid scenario here?

In general, at least in my case, I see the programmatic program being used alone and then combined with a fixed one.

> 
> > Even here:
> > ```java
> > @Find 
> > List<Product> findByName(@By("name") String name, Restriction<Product>  restriction);
> > ```
> The above is the recommended usage. 

For me, it is just an option. Don't allow the user to have a dynamic query using restriction. It does not make sense to me, and it does not help in a scenario where I have a filter in my resource, for example, in the REST application using JAX-RS.

```java
@BeanParam //JAX-RS
public class Filter {
   @ QueryParam
   private BigDecimal maxPrice;

   @ QueryParam(
      private BigDecimal minPrice;

...
 }
```

```java
List<Resctirion<Product>> restrictions =...
if(filter.maxPrice != null){
include restriction
}
```

njr-11:
> > > For example, what happens in this scenario?
> > > ```java
> > > List<Product> findByName(Restriction<Product>  restriction);
> > > ```
> 
> @njr-11 can we agree that it would be a valid scenario here?

No, that usage is not valid.  It does not satisfy the rules for Query by Method Name because it has no method parameter supplying the name value.  It does not satisfy the rules for any other pattern because it lacks a method annotation (such as `@Find` or `@Query`).  `Restriction` is not a new repository method pattern/type for Jakarta Data with its own set of rules.  `Restriction` is just another special parameter type, like `Limit`, `Sort`, `Order`, ... that adds some additional meaning (filtering provided at runtime) on top of an existing repository method pattern.  So you need to start out with a method that follows a valid method pattern in the first place, then add `Restriction` to it to gain the additional filtering.

> For me, it is just an option. Don't allow the user to have a dynamic query using restriction. It does not make sense to me, and it does not help in a scenario where I have a filter in my resource

It's difficult to understand what you have said here, but I think you are implying that I've said we will not "allow the user to have a dynamic query using restriction".  That isn't true at all.  I have not said that and nothing in my code review suggestions or comments would lead to that outcome (Let me know if there is a specific place where I made an error in that, and we'll fix it).  Hopefully you are just misunderstanding my statement of the recommended usage, "a well-written repository method should aim to enforce all of the fixed requirements upfront and then allow the user to optionally provide additional restrictions at run time where that makes sense." There are going to be cases where the number of fixed requirements that can be made upfront is 0, and the user provides everything as restrictions at run time. Usage like that is completely consistent with that statement, and I've even written example usage like that into the spec Javadoc,
https://github.com/jakartaee/data/blob/5c03b8f60d6a407b8a1981dfeacafc6a01725a90/api/src/main/java/jakarta/data/expression/TextExpression.java#L134

otaviojava:
@njr-11 

> Hopefully you are just misunderstanding my statement of the recommended usage

Yes, and reply before your suggestions. It does make sense, I've updated the documentation with those information.

Please, let me know what you think.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -285,8 +285,10 @@ A repository method must throw `UnsupportedOperationException` if it has:
 - a parameter of type `PageRequest` and a parameter of type `Limit`,
 - a `@First` annotation and a parameter of type `PageRequest` or `Limit`,
 - a parameter of type `PageRequest` or `Limit`, in combination with the keyword `First`,
-- a `@First` annotation, in combination with the keyword `First`, or
+- a `@First` annotation, in combination with the keyword `First`,
 - more than one parameter of type `Order`.
```
njr-11: needs a comma instead of period because there are now more points listed
```suggestion
- more than one parameter of type `Order`,
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +831,108 @@ When a repository method is called with a null value as an argument to one of it
 
 NOTE: The behavior of a query method when the method is called with a null argument is not defined by this specification, and is not portable between Jakarta Data providers.
 
+==== Restriction parameters
+
+The `Restriction<T>` type is a special parameter used to dynamically construct queries in a fluent and type-safe manner. Valid as parameters to methods that perform queries, such as methods annotated `@Find` or `@Delete`, restrictions allow the application to define additional conditions programmatically, often using a generated metamodel class (see <<metamodel-java-processor>>).
+
+When using `Restriction<T>` in repository methods:
+
+- The method *must* be annotated with either `@Find` or `@Query`. Without one of these annotations, the use of `Restriction<T>` is *not valid* and results in a `MappingException`.
```
njr-11: Missed `@Delete` with query conditions.
It's fine if the spec doesn't require the ability to have `Restriction` on Query by Method Name, given that model is being de-emphasized. Although I'm not sure if we forbid a Data Provider from supporting it.  Regardles,`MappingException` would not be the correct exception.
```suggestion
- The method *must* be annotated with either `@Find`, '@Delete', or `@Query`. Without one of these annotations, the use of `Restriction<T>` is *not valid* and results in an `UnsupportedException`.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +831,108 @@ When a repository method is called with a null value as an argument to one of it
...
+- When `@Find` or `@Query` is present, `Restriction<T>` can appear in *any position* in the parameter list. In these cases, the `Restriction<T>` provides the full query logic programmatically, and any query derivation from the method name is ignored.
```
njr-11: `Restriction` should follow the same rules as other special parameter types.  I'm pretty sure that for `@Query`, positional/ordinal parameters are required to come first.  Also, `Restriction` does not provide the full query logic.  The query logic is the combination of the querying the method would normally define and the filtering defined by the `Restriction`.  The definitions of `@Find`, `@Query`, and `@Delete` already require ignoring the method name.
```suggestion
- The filtering that is defined by the `@Query` annotation or parameters of the method that is annotated `@Find` or `@Delete` is combined with the additional filtering specified by the `Restriction<T>` parameter such that matching entities must satisfy both.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +831,108 @@ When a repository method is called with a null value as an argument to one of it
...
+
+For example:
+
+[source,java]
+----
+@Query("WHERE (product.numAvailable > 0)")
+List<Product> search(Restriction<Product> restriction, Order<Product> order);
+
+@Find
+Page<Product> findAvailable(Restriction<Product> restriction, PageRequest pageRequest);
```
njr-11: I think you mixed these up.  The Query value of one matches the name of the other. And the one with pagination lacks an order, which is a requirement of pagination.
```suggestion
@Find
List<Product> search(Restriction<Product> restriction);

@Query("WHERE (product.numAvailable > 0)")
Page<Product> findAvailable(Restriction<Product> restriction, PageRequest pageRequest, Order<Product> order);
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +831,108 @@ When a repository method is called with a null value as an argument to one of it
...
+----
+
+The following is *invalid* and will throw a `MappingException`:
```
njr-11: Now that you've decided Query by Method Name won't support Restriction, there are two possible errors that could be raised by the method.  MappingException for the lack of a parameter for the name.  UnsupportedOperationException for the attempt to supply Restriction to a method where it is not allowed. I don't think we can tell the Data Provider to choose one over the other. Either error would be valid.

```suggestion
The following is *invalid* because the method does not follow the Query by Method Name pattern of supplying a parameter per condition defined by the method name, nor does the method have an annotation that allows a `Restriction` parameter to be used.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +831,108 @@ When a repository method is called with a null value as an argument to one of it
...
+
+[source,java]
+----
+List<Product> findByName(Restriction<Product> restriction);
+----
+
+In both cases, Restriction<T> must not be used with lifecycle annotations such as @Save, @Insert, @Update, or @Delete.
```
njr-11: To avoid confusion over Delete, which is sometimes a lifecycle method and sometimes performs a delete query depending on the parameter.
```suggestion
In both cases, Restriction<T> must not be used with lifecycle methods.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +831,108 @@ When a repository method is called with a null value as an argument to one of it
...
+   @Find
+   List<Product> search(@By(_Product.NAME) name, Restriction<Product> restriction);
+
+   @Find
+   List<Product> search(Restriction<Product> restriction);
+}
+...
+List<Product> redPencils = products.search("pencil", _Product.color.equalTo(Color.RED));
+List<Product> expensivePencils = products.search("pencil", _Product.price.greaterThan(BigDecimal.TEN));
+List<Product> expensivePencils = products.search(_Product.price.greaterThan(BigDecimal.TEN));
```
njr-11: Avoid confusion because this query did not specify the product name of "pencil"
```suggestion
List<Product> expensiveItems = products.search(_Product.price.greaterThan(BigDecimal.TEN));
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +831,108 @@ When a repository method is called with a null value as an argument to one of it
...
+----
+
+
+Valid method signatures include:
+
+[source,java]
+----
+public interface ProductRepository {
+
+    List<Product> search(Restriction<Product> restriction);
```
njr-11, line 906: Not a valid repository method without `@Find`
```suggestion
    @Find
    List<Product> search(Restriction<Product> restriction);
```

njr-11 (commented):
Looking a lot better. I added a few more comments.

spec/src/main/asciidoc/repository.asciidoc
gavinking: This sentence is incorrect. Please remove it.

1. It's perfectly OK for a `default` method to accept a `Restriction`.
2. And it's perfectly legit for a Jakarta Data provider to define other annotations which go beyond what is included in the specification. The spec is written in an a way so that annotations like `@Find` and `@Query` belong to an open-ended set.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -285,8 +285,10 @@ A repository method must throw `UnsupportedOperationException` if it has:
 - a parameter of type `PageRequest` and a parameter of type `Limit`,
 - a `@First` annotation and a parameter of type `PageRequest` or `Limit`,
 - a parameter of type `PageRequest` or `Limit`, in combination with the keyword `First`,
-- a `@First` annotation, in combination with the keyword `First`, or
-- more than one parameter of type `Order`.
+- a `@First` annotation, in combination with the keyword `First`,
+- more than one parameter of type `Order`,
+- more than one parameter of type `Restriction`, or
+- a `Restriction` parameter and is a lifecycle method.
```
gavinking: This last bullet point is unnecessary, since the context of the whole section is query methods.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +831,109 @@ When a repository method is called with a null value as an argument to one of it
 
 NOTE: The behavior of a query method when the method is called with a null argument is not defined by this specification, and is not portable between Jakarta Data providers.
 
+==== Restriction parameters
+
+The `Restriction<T>` type is a special parameter used to dynamically construct queries in a fluent and type-safe manner. Valid as parameters to methods that perform queries, such as methods annotated `@Find` or `@Delete`, restrictions allow the application to define additional conditions programmatically, often using a generated metamodel class (see <<metamodel-java-processor>>).
+
+When using `Restriction<T>` in repository methods:
+
+- The method *must* be annotated with either `@Find`, '@Delete', or `@Query`. Without one of these annotations, the use of `Restriction<T>` is *not valid* and results in an `UnsupportedException`.
```
gavinking: This is not correct.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +831,109 @@ When a repository method is called with a null value as an argument to one of it
...
+----
+
+The following is *invalid* because the method does not follow the Query by Method Name pattern of supplying a parameter per condition defined by the method name, nor does the method have an annotation that allows a `Restriction` parameter to be used.
+
+[source,java]
+----
+List<Product> findByName(Restriction<Product> restriction);
+----
+
+In both cases, Restriction<T> must not be used with lifecycle methods.
```
gavinking: Please remove this passage.

gavinking (changes requested):
This is a lot better but it still contains unnecessary (and borderline wrong) prescriptive statements.

We need to avoid accidentally sneaking in limitations on what vendors can do as extensions to the specification. The nature of Jakarta Data is that vendors are _always_ going to need to support things which go beyond the spec.

spec/src/main/asciidoc/repository.asciidoc
otaviojava: We use this week's meeting to discuss this point.

To confirm, the key point here is the portability in this scenario, right?

```java
List<Product> findByName(Restriction<Product>  restriction);
```

Once the restriction is the first parameter, it should overwrite the method query to use the restriction instead.

gavinking: The point is that the specification does not say what happens in this case, because there might be some other annotation on that method which the specification doesn't know about, but which the Jakarta Data provider _does_. We have defined annotations like `@Find`, `@Delete`, or `@Query` as canonical representatives of open-ended sets of such "query" or "lifecycle" annotations, to allow for extension by implementations.

For example, Hibernate Data Repositories supports an `@SQL` annotation for queries written in native SQL. `@SQL` query methods don't support `Restriction`s, but the specification can't possibly know that.

Look, I've been trying to say lots of times that the spec needs to treat `Restriction` as a variation on `Order`. If the spec doesn't make a certain statement about `Order`, then it shouldn't make it about `Restriction` either. We want to avoid any divergence here.

njr-11: > To confirm, the key point here is the portability in this scenario, right?
> 
> ```java
> List<Product> findByName(Restriction<Product>  restriction);
> ```

@otaviojava you keep bringing up this same invalid example, and I don't think you are grasping why it is not valid. I'll try to go over it again.

To start with, the method has no annotations (neither spec-defined nor vendor-defined), but its name prefix of `find` matches the Query by Method Name pattern, so it is interpreted as Query by Method Name. The method name specifies one condition, `Name`, so there must be a `name` attribute of the `Product` entity, where the type of the `name` entity attribute match the first method parameter type.  You didn't show the `Product` entity, so we will assume a `name` attribute exists on it, and with type of `String` because that is what most applications would use for a product name.  The `findByName` method does have a parameter, but it is of type `Restriction<Product>`.  Here, we have a decision to make in the spec whether `Restriction` is allowed a special parameter on Query by Method Name. It could go either way, but regardless, this `findByName` method is invalid.  Here are the options:

1. If `Restriction` is not a special parameter type for Query by Method Name, then it is the parameter to use as  the `name`, but it has the wrong attribute type, so the Query by Method Name method is invalid.
2. Otherwise, if `Restriction` is a special parameter type for Query by Method Name, then there are 0 parameters supplied to meet the Query by Method Name conditions. But this `findByName` method requires 1 parameter (`Name`), and none is supplied, so it is invalid.

gavinking: Just want to clarify that my response does not contradict Nathan's, even if they seem to be saying different things.

otaviojava: @njr-11, that is not the focus; my point is more about the limitation, as Gavin said.

Given this sentence, this is wrong. 

```java
//hope this one is better, but please, ignore the method name right now!
List<Product> search(Restriction<Product>  restriction);
```

The provider could generate its annotation and behavior, as Gavin mentioned:

```java
@SQL(" any query here")
List<Product> search(Restriction<Product>  restriction);
```

By specification, this is invalid because it does not contain the required annotations: `@Find`, ' @Delete', or `@Query`.


On JNoSQL, we also have specializations that only make sense to NoSQL, such as, `CQL`, `Cypher`

Hope I make myself clear.

njr-11: > By specification, this is invalid because it does not contain the required annotations: `@Find`, ' @delete', or `@Query`.

That's only because you proposed writing the specification that way.  The specification needs to state that `Restriction` is valid as a special parameter to these 3 types of repository methods.  If properly written that way, then it doesn't stop vendors from defining additional vendor-specific support for `Restriction` with their own vendor-specific annotations.  There is no need whatsoever for the specification to define support for `Restriction` on unannotated methods that don't fit any valid pattern in order for vendors to do the latter.

otaviojava: @njr-11 @gavinking what do you think now?

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -285,8 +285,9 @@ A repository method must throw `UnsupportedOperationException` if it has:
 - a parameter of type `PageRequest` and a parameter of type `Limit`,
 - a `@First` annotation and a parameter of type `PageRequest` or `Limit`,
 - a parameter of type `PageRequest` or `Limit`, in combination with the keyword `First`,
-- a `@First` annotation, in combination with the keyword `First`, or
-- more than one parameter of type `Order`.
+- a `@First` annotation, in combination with the keyword `First`,
+- more than one parameter of type `Order`,
+- more than one parameter of type `Restriction`, or
```
```suggestion
- more than one parameter of type `Order`, or
- more than one parameter of type `Restriction`.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +830,107 @@ When a repository method is called with a null value as an argument to one of it
...
 
+==== Restriction parameters
+
+The `Restriction<T>` type is a special parameter used to dynamically construct queries in a fluent and type-safe manner. Valid as parameters to methods that perform queries, such as methods annotated `@Find` or `@Delete`, restrictions allow the application to define additional conditions programmatically, often using a generated metamodel class (see <<metamodel-java-processor>>).
+
+When using `Restriction<T>` in repository methods:
+
+- The use of `Restriction<T>` is only considered valid in portable applications when used as a parameter to methods annotated with `@Find`, `@Delete`, or `@Query`. Methods using `Restriction<T>` without one of these annotations are not covered by the specification. Jakarta Data providers may reject such methods by throwing an `UnsupportedOperationException`, or they may choose to support them as vendor-specific extensions.
+- The filtering that is defined by the `@Query` annotation or parameters of the method that is annotated `@Find` or `@Delete` is combined with the additional filtering specified by the `Restriction<T>` parameter such that matching entities must satisfy both.
```
gavinking: Does this exactly reflect language we have elsewhere for `Sort` and `Order`?

njr-11: > Does this exactly reflect language we have elsewhere for `Sort` and `Order`?

This is a good point.  `@Query` has the following documentation, which means it can only be used with `Sort`, `Order`, and `@OrderBy` when the query does not include an `ORDER BY` clause,

```
     * <p>If the annotated repository method accepts other forms of sorting
     * (such as a parameter of type {@link Sort}), it is the responsibility of
     * the application programmer to compose the query so that an
     * {@code ORDER BY} clause can be validly appended to the text of the
     * query.</p>
```

Equivalent language would mean that `Restriction` can only be used with `@Query` when the query does not contain a `WHERE` clause.  This would not be very useful, so we will need to think more about whether the consistency is important, or if possibly `Restriction` shouldn't be required to work with `@Query` at all.

otaviojava: IMHO: In this case, it does make sense as an exception.

otaviojava: @gavinking

njr-11: > IMHO: In this case, it does make sense as an exception.

I would not say as an exception.  There will be many cases where a more complex query is needed that requires query language and the application also would like to be able to add restrictions at run time.  To make it work, a Data Provider will need to be aware of the beginning and end of the WHERE clause within a query.  But I think providers already needed to have that awareness anyway for the sake of cursor-based pagination.

gavinking: AFAICT it's not very hard to add additional restrictions to a JDQL query, given that JDQL doesn't have subqueries. For full JPQL it might be somewhat harder, but of course the spec wouldn't require it for full JPQL.

otaviojava: @gavinking @njr-11 
I've created this change:

https://github.com/jakartaee/data/pull/1140/commits/d233ac4aa2ad4be02522f9243edaa8619397082d

njr-11: > I've created this change:
> 
> [d233ac4](https://github.com/jakartaee/data/commit/d233ac4aa2ad4be02522f9243edaa8619397082d)

I don't think it is necessary to include this sort of language.  It sounds like we are borderline declaring optional features. It is already assumed that vendors can do whatever they want outside of the spec, and they will document that themselves in their own products rather than the spec needing to.

otaviojava: Thank you. Done.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -829,6 +830,114 @@ When a repository method is called with a null value as an argument to one of it
 
 NOTE: The behavior of a query method when the method is called with a null argument is not defined by this specification, and is not portable between Jakarta Data providers.
 
+==== Restriction parameters
+
+The `Restriction<T>` type is a special parameter used to dynamically construct queries in a fluent and type-safe manner. Valid as parameters to methods that perform queries, such as methods annotated `@Find` or `@Delete`, restrictions allow the application to define additional conditions programmatically, often using a generated metamodel class (see <<metamodel-java-processor>>).
+
+When using `Restriction<T>` in repository methods:
+
+- The use of `Restriction<T>` is only considered valid in portable applications when used as a parameter to methods annotated with `@Find`, `@Delete`, or `@Query`. Methods using `Restriction<T>` without one of these annotations are not covered by the specification. Jakarta Data providers may reject such methods by throwing an `UnsupportedOperationException`, or they may choose to support them as vendor-specific extensions.
```
njr-11: Avoid declaring an optional feature. Vendors are free to add their own capability that involves `Restriction` regardless of whether we state this.
```suggestion
- The use of `Restriction<T>` is only considered valid in portable applications when used as a parameter to methods annotated with `@Find`, `@Delete`, or `@Query`. Methods using `Restriction<T>` without one of these annotations are not covered by the specification. Jakarta Data providers may reject such methods by throwing an `UnsupportedOperationException`.
```

njr-11 (approved):

Issue #1141, opened 2025-06-12
[clarification]: Shall We Introduce an Annotation to Mark Projection Records, or Leave It to Jakarta Data Providers?


otaviojava:
### Specification

https://github.com/jakartaee/data/blob/main/api/src/main/java/jakarta/data/repository/Find.java

### I need clarification on ...

Jakarta Data supports projections using Java records with `@Select` to map entity attributes:

```java
public record Name(String firstName,
                   String middleName,
                   @Select("lastName") String surname) {}
```

Currently, there is no annotation to explicitly declare a record as a projection target. This creates ambiguity for annotation processors and tooling that aim to generate metadata or validate mappings at compile time.

Should Jakarta Data define a dedicated annotation (e.g., @Projection) to mark these records as intended for projections, or should it be left to individual Jakarta Data providers to infer and handle this as needed?

This would not impact runtime behavior but would enhance compile-time clarity and tooling support.

gavinking:
I recently ran into a reason why we might need this: for execution on Graal VM, classes which should be exposed for reflection need to be explicitly marked as such. For entities and embeddables we have `@Entity` and `@Embeddable`,  but we have no such thing for these little projection objects.

@Sanne brought this up.

njr-11:
One scenario to think about here will be when a user wants to use a third-party record class that they don't control and to which they cannot add annotations.  We currently allow for that because `@Select` can be declared on the repository method instead.  Preserving that ability will need to be weighed against the needs/advantages brought up in this issue.

njr-11:
> So if we're going to introduce this thing

It would be fine with me if we didn't.  I liked what we had before adding it, with only the `Select` annotation for the user to be concerned with, and even that being optional when inferring names from record components.

gavinking:
> It would be fine with me if we didn't. 

I agree that we have not made any decision to go down this path, which is why I was a little bit surprised to see a pull request already. 

[Don't get me wrong, it's sometimes very useful to have a pull request to demonstrate a straw man proposal when the proposal is complicated and confusing, but I don't think that's the case here.]

beikov:
For reference, we already had some related discussions in the past about having a dedicated projection spec:

* https://github.com/jakartaee/data/issues/539
* https://github.com/jakartaee/data/issues/9

I think I'd personally prefer if projections were a standalone specification that JPA and JNoSQL providers can implement/extend/integrate. Jakarta Data wouldn't necessarily have to define all the things.

[...]

PR #1142, opened 2025-06-13
Generate the Projection annotation

otaviojava:
Ref: https://github.com/jakartaee/data/issues/1141#issuecomment-2969685905

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Projection.java
+++ b/api/src/main/java/jakarta/data/repository/Projection.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package jakarta.data.repository;
+
+/**
+ * Indicates that a {@code record} type is intended to serve as a projection
+ * for mapping selected attributes from an entity in a query result.
+ *
+ * <p>This annotation allows Jakarta Data providers to understand that the
+ * annotated record is designed for partial mapping of entity data, especially
+ * in queries using {@code @Query} or {@code @Find} annotations.</p>
+ *
+ * <p>All Jakarta Data providers <strong>must</strong> support the use of records
+ * annotated with {@code @Projection}. Providers <strong>may</strong> also choose
+ * to support projection mapping on records that are not explicitly annotated
+ * with {@code @Projection}.</p>
+ *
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/data/repository/Projection.java
```diff
@@ -0,0 +1,52 @@
...
+ * with {@code @Projection}.</p>
+ *
+ * <p>Example usage:</p>
+ *
+ * <pre>{@code
+ * @Projection
+ * public record ProductSummary(String name, @Select("price") BigDecimal value) {}
+ *
+ * @Repository
+ * public interface ProductRepository {
```
njr-11, line 40: Invalid repository because it does not define any entity type.
```suggestion
 * public interface Products extends BasicRepository<Product, Long> {
```

api/src/main/java/jakarta/data/repository/Projection.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ *     @Find
+ *     @Select("name")
+ *     @Select("price")
```
njr-11, line 44: Duplicate definition of `@Select("price")` on both the record and the repository method is valid, but unnecessary, and I would say borderline bad practice, so let's have our example choose one or the other, 
```suggestion
 *     @Select("name")
 *     @Select("price")
```

api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -109,6 +109,7 @@
      * <pre>
      * &#64;Repository
      * public interface Cars extends BasicRepository&lt;Car, String&gt; {
+     *      &#64;Projection
```
njr-11, line 112: For matching indentation,
```suggestion
     *     &#64;Projection
```

api/src/main/java/jakarta/data/repository/Projection.java
```diff
@@ -0,0 +1,52 @@
...
+ * for mapping selected attributes from an entity in a query result.
+ *
+ * <p>This annotation allows Jakarta Data providers to understand that the
+ * annotated record is designed for partial mapping of entity data, especially
+ * in queries using {@code @Query} or {@code @Find} annotations.</p>
+ *
+ * <p>All Jakarta Data providers <strong>must</strong> support the use of records
+ * annotated with {@code @Projection}. Providers <strong>may</strong> also choose
+ * to support projection mapping on records that are not explicitly annotated
+ * with {@code @Projection}.</p>
```
njr-11, line 31: Although the second sentence is true, it sounds like we are defining an optional feature. It is probably better to omit it and let the Jakarta Data provider declare their own support for it, if it wants to do so. That will make it an optional feature of the Data provider rather than an optional feature of the Data spec.
```suggestion
 * <p>All Jakarta Data providers <strong>must</strong> support the use of records
 * annotated with {@code @Projection}.</p>
```

gavinking (changes requested):
OK, look, I think we're perhaps getting a bit ahead of ourselves here, since we have not really discussed where we're going with this thing. Pull requests are generally a bad place to have high-level design discussions, so I think we should take the following questions back to the issue which was opened two days ago:

1. Since we've called its friend `@Select`, this annotation should be called `@Selection`, or the existing annotated should be named `@Project`.
2. The annotation should be allowed to specify which entity it is a projection for, allowing the use of projections for non-primary entities, e.g. `@Projection(Book.class) record BookSummary(String isbn, String title) {}`.
3. It should probably be allowed to annotate a repository method as an alternative to annoying the `record`.

Also, more trivially, this is missing `@Target` and `@Retention`.

Can we please continue the discussion on the issue?

PR #1143, opened 2025-06-13
Document where Restriction is a valid parameter

njr-11:
While reviewing another PR, I realized that the Javadoc hasn't been updated in a couple places to say that Restriction is a valid repository method parameter type.


api/src/main/java/module-info.java
```diff
@@ -805,98 +805,21 @@
  * Vehicle[] searchFor(String make, String model, int year, Sort&lt;?&gt;... sorts);
  * </pre>
  *
- * <h2>Restrictions</h2>
```
njr-11, line 808: Note to reviewers: This section is not removed. It is only moved to be under _Special parameters_ instead, which will hopefully make it clearer that `Restriction` ought to be treated like any other special parameter type when writing repository methods.

gavinking (approved):

PR #1146, opened 2025-06-14
start defining what a stateful repository is

gavinking:
Posting this here to get feedback.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -1003,9 +1003,11 @@ A _stateful repository_ is a repository which maintains an association with a se
 - A query method of a stateful repository which returns an entity type always returns managed instances belonging to the persistence context associated with the repository.
 
 - A <<Lifecycle methods, lifecycle method>> of a stateful repository must be annotated with a lifecycle annotation specific to the stateful model.
-  The annotations `@Save`, `@Insert`, `@Update`, `@Delete` may not be used to declare lifecycle methods of a stateful repository.
+  The annotations `@Save`, `@Insert`, `@Update`, `@Delete` are used to define stateless repositories and must not be used to declare lifecycle methods of a stateful repository.
   Instead, this specification defines the special lifecycle annotations `@Persist`, `@Merge`, `@Refresh`, `@Remove`, and `@Detach` for declaring lifecycle methods of stateful repositories.
-  Some lifecycle methods only accept managed entities associated with the persistence context underlying the repository.
+
+- A lifecycle method annotated `@Remove` or `@Refresh` only accepts managed entities associated with the persistence context underlying the repository.
+ On the other hand, a method annotated `@Persist`, `@Merge`, or `@Detach` also accepts unmanaged entities.
 
 - Any modification to the persistent state of a managed entity must be automatically detected by the Jakarta Data implementation and the corresponding database record must be automatically updated.
  The application program is not required to explicitly call a repository method after modifying a managed entity.
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -257,7 +257,7 @@ default void cleanup() {
 
 A repository may have at most one resource accessor method.
 
-A repository which declares a resource accessor method with return type `jakarta.persistence.EntityManager` is always a <<Stateful repositories,stateful repository>>.
+A resource accessor method with return type `jakarta.persistence.EntityManager` should only be declared by a <<Stateful repositories,stateful repository>>.
 
 === Conflicting repository method annotations
 
@@ -1035,7 +1035,7 @@ The annotations `@Save`, `@Insert`, `@Update`, `@Delete` are used to define stat
 Instead, this specification defines the special lifecycle annotations `@Persist`, `@Merge`, `@Refresh`, `@Remove`, and `@Detach` for declaring lifecycle methods of stateful repositories.
 
 A lifecycle method annotated `@Remove` or `@Refresh` only accepts managed entities associated with the persistence context underlying the repository.
- On the other hand, a method annotated `@Persist`, `@Merge`, or `@Detach` also accepts unmanaged entities.
+On the other hand, a method annotated `@Persist`, `@Merge`, or `@Detach` also accepts unmanaged entities.
 
 ==== Automatic change detection
 
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -998,31 +998,55 @@ That is, the repository does not maintain any relationship with any entity insta
 Any instance of an entity class may be passed to a method of a stateless repository which accepts the entity type.
 
 A _stateful repository_ is a repository which maintains an association with a set of _managed_ entity instances.
+Support for stateful repositories is defined in the dedicated module `jakarta.data.stateful`.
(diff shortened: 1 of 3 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -252,6 +252,8 @@ default void cleanup() {
 
 A repository may have at most one resource accessor method.
 
+A repository which declares a resource accessor method with return type `jakarta.persistence.EntityManager` is always a <<Stateful repositories,stateful repository>>.
+
```
njr-11: I would like to add this requirement exactly as you have written it, but unfortunately I think it is a change in behavior from Data 1.0, where the specification section on _Resource accessor methods_ says,

> if the Jakarta Data provider is backed by Jakarta Persistence, the return type might be `jakarta.persistence.EntityManager`.

(and EntityManager also appears in Javadoc on _Resource accessor methods_).

Data 1.0 didn't define the concept of stateful, and so it allows for `jakarta.persistence.EntityManager` as a resource accessor return type regardless of statement whether the repository is stateful or stateless.  Adding this requirement in Data 1.1 would be taking that away.  I'm tempted to just overlook this, because it certainly would be more consistent (and our implementation is still in beta waiting on the rest of EE 11, so we could still change our behavior here and not allow EntityManager for Data 1.0) but it needs to be pointed out as a change in behavior.

```suggestion
```

An alternative could be making a recommendation here that when using a Jakarta Data provider backed by Jakarta Persistence, that a resource accessor method for a stateful repository return EntityManager and a resource accessor method for a stateless repository return entity agent.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -875,3 +877,36 @@ The `@Asynchronous` annotation must not be used on repositories implemented usin
...
+- A stateful repository is backed by a _persistence context_, a set of managed entity instances in which at most one instance represents a given record in the database.
+ An entity instance never belongs to multiple persistence contexts.
+ The Jakarta Data implementation must ensure that a given persistence context never contains more than one entity instance representing the same record.
+ Multiple repositories might share a persistence context.
+ A persistence context is never shared across transactions.
+
+- A query method of a stateful repository which returns an entity type always returns managed instances belonging to the persistence context associated with the repository.
+
+- A <<Lifecycle methods, lifecycle method>> of a stateful repository must be annotated with a lifecycle annotation specific to the stateful model.
+  The annotations `@Save`, `@Insert`, `@Update`, `@Delete` may not be used to declare lifecycle methods of a stateful repository.
```
njr-11: Added some extra clarification,
```suggestion
  The annotations `@Save`, `@Insert`, `@Update`, `@Delete` are allowed only on stateless repositories and must not be used to declare lifecycle methods of a stateful repository.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -875,3 +877,36 @@ The `@Asynchronous` annotation must not be used on repositories implemented usin
...
+  Instead, this specification defines the special lifecycle annotations `@Persist`, `@Merge`, `@Refresh`, `@Remove`, and `@Detach` for declaring lifecycle methods of stateful repositories.
+  Some lifecycle methods only accept managed entities associated with the persistence context underlying the repository.
```
njr-11: The use of `Some lifecycle methods` in this sentence gives the impressions that maybe other lifecycle methods don't and could be used with either stateful or stateless. I'm pretty sure that's not what it is trying to say because that isn't correct, but it will confuse readers who don't already know the intent.  Here is my attempt at clarifying, 
```suggestion
  Of these lifecycle methods, some only accept managed entities associated with the persistence context underlying the repository, while others, such as `@Persist`, accept an unmanaged entity.
```

spec/src/main/asciidoc/repository.asciidoc
gavinking: Ugh. Oops. Well, we need to change that anyway, since `EntityAgent` is coming for certain.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -984,3 +991,62 @@ The `@Asynchronous` annotation must not be used on repositories implemented usin
...
+
+==== Stateful repository lifecycle operations
+
+A <<Lifecycle methods, lifecycle method>> of a stateful repository must be annotated with a lifecycle annotation specific to the stateful model.
+
+The annotations `@Save`, `@Insert`, `@Update`, `@Delete` are used to define stateless repositories and must not be used to declare lifecycle methods of a stateful repository.
+Instead, this specification defines the special lifecycle annotations `@Persist`, `@Merge`, `@Refresh`, `@Remove`, and `@Detach` for declaring lifecycle methods of stateful repositories.
+
+A lifecycle method annotated `@Remove` or `@Refresh` only accepts managed entities associated with the persistence context underlying the repository.
+ On the other hand, a method annotated `@Persist`, `@Merge`, or `@Detach` also accepts unmanaged entities.
```
```suggestion
On the other hand, a method annotated `@Persist`, `@Merge`, or `@Detach` also accepts unmanaged entities.
```

stateful/src/main/java/jakarta/data/repository/stateful/Merge.java
```diff
@@ -53,6 +53,25 @@
...
+ * the merged entity.
+ * <ul>
+ *     <li>If such a record exists, the merge operation schedules an update
+ *     of the existing record. The record is eventually updated to reflect the
+ *     merged changes.</li>
+ *     <li>Otherwise, if no such record exists, the merge operation schedules
+ *     the insertion of a new record with the state held by the merged entity.
+ * </ul>
+ * In either case, the method annotated {@code @Merge} returns a managed entity
+ * representing the inserted or updated record, and reflecting the merged changes,
```
njr-11:
```suggestion
 * representing the inserted or updated record and reflecting the merged changes,
```

njr-11 (commented):
The latest updates look good.  I only spotted 2 very trivial grammar updates.  There is also one review comment left over from the previous review around resource accessor methods and entity manager.  Maybe we can restate that one in terms of how repositories _should_ be written when referring to resource accessor methods returning EnityManager and how they _must_ be written when returning entity agent.  That way we don't break compatibility with Data 1.0 but mostly still get the point across.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -252,6 +257,8 @@ default void cleanup() {
 
 A repository may have at most one resource accessor method.
 
+A repository which declares a resource accessor method with return type `jakarta.persistence.EntityManager` is always a <<Stateful repositories,stateful repository>>.
```
njr-11: Here's an idea for turning this into a recommendation so that we maintain compatibility with Data 1.0,
```suggestion
A resource accessor method with return type `jakarta.persistence.EntityManager` should only be defined on a <<Stateful repositories,stateful repository>>.
```
Later, when we have entity agent, we could also write,
```
A resource accessor method with return type `jakarta.persistence.EntityAgent` must only be defined on a stateless repository.
```

gavinking:
The statement that "if the Jakarta Data provider is backed by Jakarta Persistence, the return type **might** be `jakarta.persistence.EntityManager`" does not actually contradict the statement that "A resource accessor method with return type `jakarta.persistence.EntityManager` should only be declared by a stateful repository."

So I think this is good enough for now, at least until we can start talking about `EntityManager` vs `EntityAgent`.

PR #1147, opened 2025-06-17
Restrict.all/any methods that accept a list of restrictions

njr-11:
A common use of Restriction will be when applications present their users with customizable filtering.  For these cases, the application will need to look through all of the various filtering options that might or might not be used and build up a composite restriction based on that.  When the application doesn't know in advance which filtering will be chosen, it will help if we offer forms of the `all` and `any` methods that accept a list rather than the varags array.  This will allow applications to write code such as,

```
if (maxPrice.isPresent())
    restrictionList.add(_Product.price.lessThanEqual(maxPrice.get()));
if (minPrice.isPresent())
    restrictionList.add(_Product.price.greaterThanEqual(minPrice.get()));
if ...

found = products.search(..., Restrict.all(restrictionList), Order.by(sortList));
```

### DIFF
--- a/api/src/main/java/jakarta/data/restrict/Restrict.java
+++ b/api/src/main/java/jakarta/data/restrict/Restrict.java
@@ -78,8 +78,10 @@ public class Restrict {
      *                                  includes a {@code null} value.
      */
     @SafeVarargs
-    public static <T> Restriction<T> all(Restriction<T>... restrictions) {
-        return new CompositeRestrictionRecord<>(CompositeRestriction.Type.ALL,
+    public static <T> Restriction<T> all(
+            Restriction<? super T>... restrictions) {
+        return new CompositeRestrictionRecord<T>(
+                CompositeRestriction.Type.ALL,
                 List.of(restrictions));
     }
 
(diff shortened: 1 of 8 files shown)


api/src/main/java/jakarta/data/restrict/Restrict.java
```diff
@@ -82,6 +83,43 @@ public static <T> Restriction<T> all(Restriction<T>... restrictions) {
...
+     * @param restrictions one or more restrictions obtained from a method of
+     *                     {@code Restrict} or from a static metamodel
+     *                     {@link Attribute} subtype.
+     * @return
+     * @throws IllegalArgumentException if the supplied restrictions list is
+     *                                  empty or {@code null}.
+     * @throws NullPointerException     if the supplied restrictions list
+     *                                  includes a {@code null} value.
+     */
+    public static <T> Restriction<T> all(List<Restriction<T>> restrictions) {
```
gavinking: We probably need make this:

```java
<T> Restriction<T> all(List<? extends Restriction<? super T>> restrictions)
```

njr-11: Regarding `? extends Restriction`,  I know we have `BasicRestriction` and `CompositeRestriction` as subtypes of `Restriction`, but users don't interact with those.  From the user's perspective, I think they would always have lists of just Restriction.

njr-11: It looks like `Order.by`, which is a very similar pattern, already has `? super T`, so that's good.

```
    public static <T> Order<T> by(List<Sort<? super T>> sorts) {
```

njr-11: Actually, it looks like we need `? extends Restriction` not for BasicRestriction/CompositeRestriction, but just to be able to use a `List<Restriction<Book>` rather than `List<Restriction<? super Book>>`.  And given that `Order.by(List)` would have the same pattern, I expect it will require the same.

njr-11: I confirmed that `Order.by(List<Sort<EntityClass>>)` does have the same bug.  I added a commit to fix this and include a test for it, as well as to similarly fix Restrict.any/all(List) per the recommendation of @gavinking  above.  This is ready for review again.

otaviojava (approved):

PR #1149, opened 2025-06-18
Consistent names for GTE and LTE

njr-11:
In some places (Restrictions, Expressions, Query by Method Name), we use names `GreaterThanEqual`, `LessThanEqual`, `greaterThanEqual(value)`, `lessThanEqual(value)`.
In Constraints, we have a slightly different form: `GreaterThanOrEqual` and `LessThanOrEqual`.
This PR makes all of the names consistent.

### DIFF
--- a/api/src/main/java/jakarta/data/constraint/Constraint.java
+++ b/api/src/main/java/jakarta/data/constraint/Constraint.java
@@ -89,12 +89,12 @@ public interface Constraint<V> {
         return LessThan.bound(bound);
     }
 
-    static <V extends Comparable<?>> GreaterThanOrEqual<V> greaterThanOrEqual(V bound) {
-        return GreaterThanOrEqual.min(bound);
+    static <V extends Comparable<?>> GreaterThanEqual<V> greaterThanEqual(V bound) {
+        return GreaterThanEqual.min(bound);
     }
 
-    static <V extends Comparable<?>> LessThanOrEqual<V> lessThanOrEqual(V bound) {
-        return LessThanOrEqual.max(bound);
+    static <V extends Comparable<?>> LessThanEqual<V> lessThanEqual(V bound) {
+        return LessThanEqual.max(bound);
     }
 
     static <V extends Comparable<?>> Between<V> between(V lowerBound, V upperBound) {
--- a/api/src/main/java/jakarta/data/constraint/GreaterThanEqual.java
+++ b/api/src/main/java/jakarta/data/constraint/GreaterThanEqual.java
@@ -20,16 +20,16 @@ package jakarta.data.constraint;
 import jakarta.data.expression.ComparableExpression;
 import jakarta.data.spi.expression.literal.ComparableLiteral;
 
-public interface GreaterThanOrEqual<V extends Comparable<?>> extends Constraint<V> {
+public interface GreaterThanEqual<V extends Comparable<?>> extends Constraint<V> {
 
-    static <V extends Comparable<?>> GreaterThanOrEqual<V> min(
+    static <V extends Comparable<?>> GreaterThanEqual<V> min(
             V minimum) {
-        return new GreaterThanOrEqualRecord<>(ComparableLiteral.of(minimum));
+        return new GreaterThanEqualRecord<>(ComparableLiteral.of(minimum));
     }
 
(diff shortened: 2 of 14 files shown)


gavinking:
An alternative solution, which I had considered and rejected, perhaps too fast, would be to call these `NotLessThan`, `NotGreaterThan`.

njr-11:
> An alternative solution, which I had considered and rejected, perhaps too fast, would be to call these `NotLessThan`, `NotGreaterThan`.

It's nice how those are more concise, and the tradeoff would be needing to think in terms of the negative (although that is already done for many of the other comparisons: NotLike, NotIn).
We can certainly consider some different options for the names, although probably as a separate issue if we do that.  Here, I was just trying to make all usage throughout the spec consistent.

otaviojava:
I mean, considering that we have the documentation like this: 

> evaluate to a value greater than or equal to the given value

Also, the method at the metamodel is still:

```java
 _Product.price.lessThanEqual(10);
```

And once we are using this `lessThanEqual` terminology all over the place, why should we rename the class to `atLeast` or `atMost` instead of keeping it super close to its proposal?

> I would argue that greaterThanOrEqual/lessThanOrEqual are the worst choice due to their length.

I prefer to keep it bigger if it brings a meaningful name.

njr-11:
> I prefer to keep it bigger if it brings a meaningful name.

I would say that `atMost`, even though shorter, is the more meaningful name.  The end user who inputs a maximum price they will pay into a filter is thinking they are willing to pay "at most" 50 dollars or "up to" 50 dollars.  You don't usually hear people say it as they are hoping to pay "less than or equal to" 50 dollars.

otaviojava:
how about the "contract form"?

I would be fine using:

```java
 _Product.price.lt(10);
```

That is what we have on Jakarta Persistence and Jakarta NoSQL.

Ref: 
https://jakarta.ee/specifications/persistence/2.2/apidocs/javax/persistence/criteria/criteriabuilder#lt(javax.persistence.criteria.Expression,java.lang.Number)

gavinking:
I gotta say that I find `AtLeast` and `AtMost` by far the must "natural" names here, that is, they're the names one would use in ordinary conversion.

otaviojava:
Gavin, could you clarify what you mean by "natural" here?

I searched across mainstream database DSLs and mappers
and found **zero usage** of `AtLeast` or `AtMost`.

Even in resources written in Portuguese, Spanish, French, and Italian, the standard terms remain:

- greaterThan, lessThanOrEqualTo
- or symbolic forms: GTE, LTE, etc.

It is not just an English convention—it's a global standard across the software industry.

My concern is that introducing `AtLeast/AtMost` creates inconsistency. 

It's like a Java ORM replaces JOIN in SQL with 'TOGETHER', because it sounds more intuitive—maybe, but it breaks alignment with developer expectations.


Why introduce new terms that conflict with decades of established developer vocabulary?

gavinking:
By "natural" I mean "what a typical native speaker would use in conversation" rather than "what someone who is trying to sound pretentiously pseudo-mathematical by repeating the phrase the learned in primary school math class".

Nobody says "greater than or equal to" when in regular speech.

otaviojava:
Thank you.

I’d be okay with adopting them as the primary methods. That said, since many developers—native speakers included—are already trained on DSLs like SQL, JPA, and Spring Data, I’d propose supporting both styles through method aliases:

Thus, we could support both:

```java
_Product.price.atLeast(10);         // natural language
_Product.price.greaterThanEqual(10); // DSL-aligned
```


Thus: the `AtLeast` with this alias method name.


Do you think it makes sense?

gavinking:
> I’d propose supporting both styles through method aliases:

That's completely fine by me. To be clear, my main problem was with calling the *types* `GreaterThanEqual` and `LessThanEqual`. I think `this.greaterThanEqual(that)` reads OK, since both subject and object are present.

otaviojava:
> > I’d propose supporting both styles through method aliases:
> 
> That's completely fine by me. To be clear, my main problem was with calling the _types_ `GreaterThanEqual` and `LessThanEqual`. I think `this.greaterThanEqual(that)` reads OK, since both subject and object are present.

For me, even better. Thus, having the class `AtLeast`, but keeping the `this.greaterThanEqual(that)` as a method.

Issue #1117, opened 2025-05-23
[Release]: 1.0.2 API Service Release


arjantijms:
### Proposal

API 1.0.1 jar is missing the OSGi manifest. This has already been committed into the repo, so only a release is needed still.

### Outstanding work

- [ ] 

#### Prepare
  - [X] Open this issue.
  - [ ] Complete outstanding work
  - [ ] [Update TCK Signatures to capture API changes](https://github.com/jakartaee/data/actions/workflows/update.yml)
#### Stage release
  - [ ] [Update the CHANGELOG.adoc file to reflect the changes put into the new release](https://github.com/jakartaee/data/blob/main/CHANGELOG.adoc)
  - [ ] [Build and stage artifacts to staging repository](https://ci.eclipse.org/data/view/Release%20Builds/job/jakarta-data-build-and-stage/)
  - [ ] [Build and stage the TCK distribution artifact to download.eclipse.org](https://ci.eclipse.org/data/view/Eclipse%20Builds/job/jakarta-data-tck-build-and-stage/)
  - [ ] Update this issue with a comment that links to the staged artifacts and the generated GitHub branch and tag
  - [ ] [Create a `draft` release on GitHub](https://github.com/jakartaee/data/releases/new)
#### Verify and modify
  - [ ] Ask for feedback from the community to verify the staged artifact has all the expected changes.
  - [ ] If anything needs to be added before publishing do that now and repeat the `Stage Release` section
#### Publish release
  - [ ] [Publish staged artifacts to the public repository](https://ci.eclipse.org/data/view/Release%20Builds/job/jakarta-data-stage-to-release/)
  - [ ] [Publish staged TCK distribution artifact to download.eclipse.org](https://ci.eclipse.org/data/view/Eclipse%20Builds/job/jakarta-data-tck-staged-to-promoted/)
  - [ ] Update this issue with a comment that links to the published artifacts and the generated GitHub branch and tag
  - [ ] [Modify the `draft` release on GitHub to be `pre-release` or `latest`](https://github.com/jakartaee/data/releases)
#### Follow up
  - [ ] [Email a Jakarta steering committee member to promote the TCK Distribution publically](https://ci.eclipse.org/jakartaee-spec-committee/job/promote-release/)
  - [ ] [Create pull request to https://github.com/jakartaee/specifications](https://github.com/jakartaee/specifications/blob/master/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md)
[...]

PR #1150, opened 2025-06-18
fix: add osgi manifest to api/orm jars

KyleAure:
Related to #1117

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -35,7 +35,7 @@
         <assertj.version>3.27.3</assertj.version>
         <!-- default is the same for backward compatibility reason
         Easy to override when building with a system property -->
-        <spec.version>1.0</spec.version>
+        <spec.version>1.1</spec.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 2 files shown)


api/pom.xml
```diff
@@ -141,15 +141,21 @@ Use is subject to <a href="{@docRoot}/doc-files/speclicense.html" target="_top">
...
+                        </Bundle-Description>
                         <Bundle-Name>Jakarta Data API jar</Bundle-Name>
                         <Bundle-SymbolicName>jakarta.data-api</Bundle-SymbolicName>
-                        <Bundle-Version>${spec.version}</Bundle-Version>
+                        <Bundle-Version>${project.version}</Bundle-Version>
                         <Extension-Name>jakarta.data</Extension-Name>
                         <Implementation-Version>${project.version}</Implementation-Version>
                         <Specification-Vendor>Eclipse Foundation</Specification-Vendor>
                         <Specification-Version>${spec.version}</Specification-Version>
+                        <Export-Package>jakarta.data.*; version=${project.version}</Export-Package>
```
njr-11: I wonder if we should avoid exporting internals (the jakarta.data.messages package) as we do in the module and what impact that would have on signature testing

KyleAure:
@otaviojava we will need both PRs one is for the service branch of release 1.0.1 and the other is for the main branch

api/pom.xml
KyleAure: Good catch.  I have updated the export's to exclude `jakarta.data.messaging`

PR #1159, opened 2025-06-25
Update spec to reflect addition of Constraint and @Is

gavinking:


api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -99,8 +99,8 @@
  *     of the query,</li>
  * <li>have exactly the same type and position within the parameter list of
  *     the method as a positional parameter of the query, or</li>
- * <li>be of type {@link Limit}, {@link Order}, {@link PageRequest},
- *     {@link Restriction}, or {@link Sort}.</li>
+ * <li>be of type {@link Limit}, {@link Restriction}, {@link Sort}, {@link Order},
+ *     or {@link PageRequest}.</li>
```
njr-11: I don't think this did anything except switch around the ordering, although I'm not sure why.  I suppose we could aim for the typical ordering we might expect/recommend a customer to use when writing a repository method and also pair up types next to each other that are most similar, such as `Order` near `Sort` and `PageRequest` near `Limit`.  If a user wanted all of the filtering related parameters together, they would place `Restriction` first, so maybe:
```suggestion
 * <li>be of type {@link Restriction}, {@link Limit}, {@link PageRequest},
 *     {@link Order}, or {@link Sort}.</li>
```

If you had a better reason for the prior ordering, just ignore this and the other comment on it. I don't think it matters very much.

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -64,8 +65,17 @@
...
+ *     and be of type {@code C<T>}, where {@code C} is {@link Constraint} or any
+ *     interface which extends {@link Constraint} and {@code T} is the type of
+ *     the persistent attribute,
+ * <li>have exactly the same name as a persistent attribute of the entity class,
+ *     be annotated {@link Is @Is(C.class)}, where {@code C} is {@link Constraint}
+ *     or any interface which extends {@link Constraint}, and be of the same type
+ *     as the parameter of a unary static method of {@code C} returning
+ *     {@code C<T>} where {@code T} is the type of the persistent attribute, or
+ * <li>be of type {@link Limit}, {@link Restriction}, {@link Sort}, {@link Order},
+ *     or {@link PageRequest}.</li>
```
njr-11:
```suggestion
 * <li>be of type {@link Restriction}, {@link Limit}, {@link PageRequest},
 *     {@link Order} or {@link Sort}.</li>
```

api/src/main/java/jakarta/data/repository/Find.java
```diff
@@ -64,8 +65,17 @@
...
  *     or</li>
- * <li>be of type {@link Limit}, {@link Order}, {@link PageRequest},
- *     {@link Restriction}, or {@link Sort}.</li>
+ * <li>have exactly the same name as a persistent attribute of the entity class
+ *     and be of type {@code C<T>}, where {@code C} is {@link Constraint} or any
+ *     interface which extends {@link Constraint} and {@code T} is the type of
+ *     the persistent attribute,
+ * <li>have exactly the same name as a persistent attribute of the entity class,
+ *     be annotated {@link Is @Is(C.class)}, where {@code C} is {@link Constraint}
+ *     or any interface which extends {@link Constraint}, and be of the same type
```
njr-11: `@Is(Constraint.class)` would not make sense,
```suggestion
 *     be annotated {@link Is @Is(C.class)}, where {@code C} is an interface
 *     which extends {@link Constraint}, and be of the same type
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -186,8 +186,10 @@ Each automatic query method must be assigned an entity type. The rules for infer
 
 Jakarta Data infers a query based on the parameters of the method. Each parameter must either:
 
-- have exactly the same type and name as a persistent attribute of the entity class, or
-- be of type `Limit`, `Order`, `PageRequest`, or `Sort`.
+- have exactly the same type and name as a persistent attribute of the entity class,
+- have exactly the same name as a persistent attribute of the entity class and be of type `C<T>`, where `C` is `Constraint` or any interface which extends `Constraint` and `T` is the type of the persistent attribute,
+- have exactly the same name as a persistent attribute of the entity class, be annotated `@Is(C.class)`, where `C` is `Constraint` or any interface which extends `Constraint`, and be of the same type as the parameter of a unary static method of `C` returning `C<T>` where `T` is the type of the persistent attribute, or
```
njr-11: Another place to update because `@Is(Constraint.class)` would not make sense,
```suggestion
- have exactly the same name as a persistent attribute of the entity class, be annotated `@Is(C.class)`, where `C` is an interface which extends `Constraint`, and be of the same type as the parameter of a unary static method of `C` returning `C<T>` where `T` is the type of the persistent attribute, or
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -267,7 +269,7 @@ reject such a method declaration at compile time.
 
 === Special parameters for limits, sorting, and pagination
```
njr-11: Because we are adding Restriction to this section,
```suggestion
=== Special parameters for additional filtering, limits, sorting, and pagination
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -267,7 +269,7 @@ reject such a method declaration at compile time.
...
 
-An <<Annotated query methods,annotated>>, <<Parameter-based automatic query methods,parameter-based>>, or Query by Method Name query method may have _special parameters_ of type `Limit`, `Order`, `Sort`, or `PageRequest` if the method return type indicates that the method may return multiple entities, that is, if the return type is:
+An <<Annotated query methods,annotated>>, <<Parameter-based automatic query methods,parameter-based>>, or Query by Method Name query method may have _special parameters_ of type `Limit`, `Restriction`, `Sort`, `Order`, or `PageRequest` if the method return type indicates that the method may return multiple entities, that is, if the return type is:
```
njr-11: Listing `Restriction` here will inadvertently cause support for it to be required for Query by Method Name (we didn't do that elsewhere).  My guess is that wasn't your intent.  I suppose it would be okay if we did support it for Query by Method Name, but we would use a separate PR to make it clear that's what we wanted and would update the other places as well.

The other problem with listing `Restriction` here is that it would require the method to have a signature that can return multiple entities.  I assume this was also not the intent.  `Restriction` will also be useful with methods that return 0 or 1 entity, as well as with delete queries, in which case entities are not returned at all.

Switching back to the previous text here:
```suggestion
An <<Annotated query methods,annotated>>, <<Parameter-based automatic query methods,parameter-based>>, or Query by Method Name query method may have _special parameters_ of type `Limit`, `Order`, `Sort`, or `PageRequest` if the method return type indicates that the method may return multiple entities, that is, if the return type is:
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -276,9 +278,21 @@ An <<Annotated query methods,annotated>>, <<Parameter-based automatic query meth
 A special parameter controls which query results are returned to the caller of a repository method, or in what order the results are returned:
```
njr-11, line 280: Insert language separately calling out `Restriction` as a special parameter because it can be used in a different set of places,
```suggestion
An <<Annotated query methods,annotated>> or <<Parameter-based automatic query methods,parameter-based>> query method may also have a _special parameter_ of type `Restriction`.

A special parameter controls which query results are returned to the caller of a repository method, or in what order the results are returned:
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -276,9 +278,21 @@ An <<Annotated query methods,annotated>>, <<Parameter-based automatic query meth
...
 
 - a `Limit` allows the query results to be limited to a given range defined in terms of an offset and maximum number of results,
+- a `Restriction` allows the query results to be restricted by applying constraints to entity attributes,
 - a `Sort` or `Order` allows the query results to be sorted by a given entity attribute or list of attributes, respectively, and
 - a `PageRequest` splits results into pages. A parameter of this type must be declared when the repository method returns a `Page` of results, as specified below in <<Offset-based pagination>>, or a `CursoredPage`, as specified in <<Cursor-based pagination>>.
 
+The types `Restriction`, `Sort`, and `Order` are parameterized by an entity type.
+Any special parameter declared with one of these types must be declared with type argument `E` or `? super E` where `E` is the queried entity type.
```
njr-11, line 288: Good idea adding this and the example!

njr-11 (commented):
Good idea to add this to the spec as well. I added a few comments. The ordering change seemed very minor although I couldn't figure out the reason for it, and I listed a possible ordering that I thought might make sense, but feel free to ignore that if you did have reasons for the other ordering.

api/src/main/java/jakarta/data/repository/Query.java
gavinking: I sorta wanted to keep `Restriction` next to `Order` and `Sort`. But I agree with you that it also makes sense to keep `Limit` and `PageRequest` together.

gavinking:
@njr-11 I've attempted to address your feedback. WDYT?

otaviojava:
@gavinking when you have time, could you rebase this one?

PR #1160, opened 2025-06-25
Clean up section on Restrictions

gavinking:
This pull request attempts to address problems introduced by #1140.

1. The new section about restrictions was added as a sub-subsection of "Null arguments to repository methods"
2. It was also in an unnatural  location with respect to the "natural" flow of topics
3. It smuggled in a requirement that the implementation throw `NullPointerException` for a null argument to a parameter of type `Restriction`, which was inconsistent with what we say about other special parameters
4. It also introduced a bunch of redundant, repetitive, and arguably *incorrect* statements about legal usage of `Restriction`

Folks, I'm sorry to have to be the bad guy again here. I had deliberately _not_ approved that pull request, and indeed I had explicitly asked that it not be merged.

@otaviojava 

- The Jakarta Data specification cannot explicitly enumerate every single thing that might not be portable between Jakarta Data providers, because that list is infinite, and the specification is constrained to be of finite length. 
- Furthermore, too  much "dicta" makes a specification much harder to read and understand. Any implementor of the specification needs to be able to clearly distinguish what is a positive requirement. This is especially the case when dicta crops up inconsistently on one part of the specification and not in others.
- When we say that something is **not** allowed, that means that this is a positive requirement that the implementor must detect and report an error in that case. This limits the ability of implementors to innovate and experiment with things which we do not require, but which may still be useful. We should therefore avoid saying that things are not allowed or not supported. At most we say that things are not portable between providers, and we only really need to say that when it's non-obvious.

Therefore, the specification should be written in terms of placing a finite list of requirements of things which an implementor must allow/support. And those things must be expressed in a consistent and non-redundant way. 

In this case, `Restriction` is just a special case of a generic class of things which we call "special parameters". The requirements relating to `Restriction` MUST be:

1.  consistent with, 
2. _expressed in the same language as_, and
3. whenever possible, _expressed in the same place_

as requirements we apply to other kinds of special parameter, for example `Sort` and `Order`.

If we don't say something about `Sort` and `Order`, then we almost certainly don't need to say it about `Restriction`.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -375,7 +375,7 @@ List<User> findByNamePrefix(String namePrefix, Sort<?>... sorts);
 
 === Restrictions
 
-A method annotated `@Find` or `@Delete` may have a <<special-parameters,special parameter>> of type `Restriction<T>`.
+A method annotated `@Find`, `@Delete`, or `@Query` may have a <<special-parameters,special parameter>> of type `Restriction<T>`.
 This parameter allows the caller of a repository method to programmatically express additional restrictions on the results returned by the repository method, often using a <<metamodel-java-processor,generated metamodel class>> for type safety.
 
 For example:
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -390,7 +390,7 @@ public interface ProductRepository {
 ----
 [source,java]
 ----
-List<Product> pencils = products.search(_Product.description.like("%pencil%");
+List<Product> pencils = products.search(_Product.description.like("%pencil%"));
 ----
 
 Programmatic restrictions passed to a parameter of type `Restriction` combine conjunctively with static restrictions:
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -264,7 +264,7 @@ If a method of a repository interface has more than one such annotation, the ann
 `UnsupportedOperationException` every time it is called. Alternatively, a Jakarta Data provider is permitted to
 reject such a method declaration at compile time.
 
-
+[[special-parameters]]
 === Special parameters for limits, sorting, and pagination
 
 An <<Annotated query methods,annotated>>, <<Parameter-based automatic query methods,parameter-based>>, or Query by Method Name query method may have _special parameters_ of type `Limit`, `Order`, `Sort`, or `PageRequest` if the method return type indicates that the method may return multiple entities, that is, if the return type is:
@@ -278,6 +278,8 @@ A special parameter controls which query results are returned to the caller of a
 - a `Limit` allows the query results to be limited to a given range defined in terms of an offset and maximum number of results,
 - a `Sort` or `Order` allows the query results to be sorted by a given entity attribute or list of attributes, respectively, and
 - a `PageRequest` splits results into pages. A parameter of this type must be declared when the repository method returns a `Page` of results, as specified below in <<Offset-based pagination>>, or a `CursoredPage`, as specified in <<Cursor-based pagination>>.
+
+A repository method throws `NullPointerException` if an argument to a special parameter of the method is null.
 
 A repository method must throw `UnsupportedOperationException` if it has:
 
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -371,6 +373,67 @@ Page<User> findByNamePrefix(String namePrefix,
...
+@Repository
+public interface ProductRepository {
+   @Find
+   List<Product> search(@By(_Product.NAME) name,
+                        Restriction<Product> restriction);
+
+   @Find
+   List<Product> search(Restriction<Product> restriction);
+}
+----
```
otaviojava, line 411: Once there is no description here, why not put it on the same block?

gavinking: I'm not at all a fan of long code fragments. Code samples should be short to focus attention on exactly the thing they are trying to illustrate, IMO.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -279,6 +279,8 @@ A special parameter controls which query results are returned to the caller of a
 - a `Sort` or `Order` allows the query results to be sorted by a given entity attribute or list of attributes, respectively, and
 - a `PageRequest` splits results into pages. A parameter of this type must be declared when the repository method returns a `Page` of results, as specified below in <<Offset-based pagination>>, or a `CursoredPage`, as specified in <<Cursor-based pagination>>.
 
+A repository method throws `NullPointerException` if an argument to a special parameter of the method is null.
+
```
njr-11, line 283: Unfortunately, if we didn't already state this anywhere during 1.0, then we cannot enforce it now.  If the behavior of `null` is undefined for 1.0, then an implementation could have chosen to make it work, maybe by ignoring the lack of a Limit/Order/Sort.  If we add this as a new requirement in 1.1, it could be a breaking change.

```suggestion
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -371,6 +373,67 @@ Page<User> findByNamePrefix(String namePrefix,
 List<User> findByNamePrefix(String namePrefix, Sort<?>... sorts);
 ----
 
+=== Restrictions
+
+A method annotated `@Find` or `@Delete` may have a <<special-parameters,special parameter>> of type `Restriction<T>`.
```
njr-11: The provided example includes `Restriction` on a method that is annotated `@Query`,
```suggestion
A method annotated `@Find`, `@Delete`, or `@Query` may have a <<special-parameters,special parameter>> of type `Restriction<T>`.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -371,6 +373,67 @@ Page<User> findByNamePrefix(String namePrefix,
...
+[source,java]
+----
+@Repository
+public interface ProductRepository {
+   @Find
+   List<Product> search(@By(_Product.NAME) name,
+                        Restriction<Product> restriction);
+
+   @Find
+   List<Product> search(Restriction<Product> restriction);
```
njr-11: The prior text claimed these to be example of combining with restrictions expressed by `@Query` or `@Find` parameters.  This second example method did neither of those.  No example was included for `@Query`, so I'll recommend switching it to that,

```suggestion
   @Query("WHERE (product.discount > 0)")
   List<Product> discounted(Restriction<Product> restriction);
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -826,117 +889,10 @@ A repository method that does not fit any of the above patterns and is not handl
 When a repository method is called with a null value as an argument to one of its parameters, the repository implementation might throw an exception:
 
 - when a <<_lifecycle_methods,lifecycle method>> is called with a null entity instance, the repository implementation must throw `NullPointerException`, or
-- when an <<_annotated_query_methods,annotated>> or <<_parameter_based_automatic_query_methods,parameter-based>> query method is called with a null argument, the repository implementation is permitted, but not required, to throw an appropriate exception type.
+- when an <<_annotated_query_methods,annotated>> or <<_parameter_based_automatic_query_methods,parameter-based>> query method is called with a null argument, the repository implementation is permitted, but not required, to throw an appropriate exception type, unless the null argument occurs as an argument to a <<special-parameters,special parameter>>, in which case the repository implementation is required to throw `NullPointerException`.
```
njr-11, line 895: It would have been nice if we had written it this way from the start, but changing it now could be a breaking change in behavior (see earlier comment)
```suggestion
- when an <<_annotated_query_methods,annotated>> or <<_parameter_based_automatic_query_methods,parameter-based>> query method is called with a null argument, the repository implementation is permitted, but not required, to throw an appropriate exception type, unless the null argument occurs as an argument to a <<special-parameters,special parameter>>.
```

spec/src/main/asciidoc/repository.asciidoc
gavinking: I don't care about this at all (I was just trying to preserve the statement which was being made about `Restriction`s).

That said, I don't believe it's a breaking change when we make something that was previously ambiguous/undefined into something that's now well-defined. 

I think it's generally the case that a "clarification" to behavior is not considered a breaking change.

njr-11: In most cases, it would be ambiguous because nothing is said about it.  But later on in the spec, we specifically say that the repository is not required to throw an exception and we would be reversing that statement.

> when an <<_annotated_query_methods,annotated>> or <<_parameter_based_automatic_query_methods,parameter-based>> query method is called with a null argument, the repository implementation is permitted, but not required, to throw an appropriate exception type

Now that I read this again, I think our intent there was really to state this about query parameters, not special parameters.

gavinking: > Now that I read this again, I think our intent there was really to state this about query parameters, not special parameters.

Right, the purpose of that statement was to stay agnostic about things like:

```java
@Find
List<Car> byMake(String make);

var carsWithUnknownMake = repo.byMake(null);
```

since `make = null` has semantics that vary between databases.

[In my implementation we actually interpret this to mean `make is null`, FTR.]

Again: I don't care what we decide to say about `null` for special parameters, and I wasn't really trying to smuggle anything new in here. I just think we should be say the same thing about `Restriction` and `Order`.

spec/src/main/asciidoc/repository.asciidoc
gavinking: How about we just simplify it instead?

njr-11: Sure, omitting this example method would also work.  That would leave the other method as an example that is consistent what it says it's an example of.

gavinking: @njr-11 @otaviojava I just pushed a simplification to the code examples that IMO does I better job of telling a "story". WDYT?

spec/src/main/asciidoc/repository.asciidoc
njr-11: I checked our implementation and we are always either raising NullPointerException, or in one case IllegalArgumentException, for null values for special parameters.  I'll just switch the one IllegalArgumentException over to NullPointerException and then we will already be compliant with this.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -371,6 +373,70 @@ Page<User> findByNamePrefix(String namePrefix,
...
+----
+@Repository
+public interface ProductRepository {
+    @Find
+    List<Product> search(Restriction<Product> restriction);
+}
+----
+[source,java]
+----
+List<Product> pencils = products.search(_Product.description.like("%pencil%");
```
```suggestion
List<Product> pencils = products.search(_Product.description.like("%pencil%"));
```

njr-11 (approved):
> @njr-11 @otaviojava I just pushed a simplification to the code examples that IMO does I better job of telling a "story". WDYT?

Looks good. I like how it gradually adds on more to the same example. That's a good way for users to learn the capability. One minor fix is needed to make it compile, but that's trivial and I'll approve it in advance.

otaviojava (approved):
it looks great to me as well.

Thank you, Gavin.

PR #1170, opened 2025-07-02
Update spec to reflect addition of Constraint and @Is

gavinking:
Rebase of #1159.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -286,7 +286,7 @@ A special parameter controls which query results are returned to the caller of a
 - a `PageRequest` splits results into pages. A parameter of this type must be declared when the repository method returns a `Page` of results, as specified below in <<Offset-based pagination>>, or a `CursoredPage`, as specified in <<Cursor-based pagination>>.
 
 The types `Restriction`, `Sort`, and `Order` are parameterized by an entity type.
-Any special parameter declared with one of these types must be declared with type argument `E` or `? super E` where `E` is the queried entity type.
+Any special parameter declared with one of these types must be declared with type argument `E` or `? super E` where `E` is the queried entity type. Usage of the generic type argument `?` in a special parameter of type `Sort<?>...` is deprecated.
 
 For example, this repository method allows a list of cars to be filtered and sorted by attributes of `Car` or of its supertype `Vehicle`:
 


spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -264,21 +266,36 @@ If a method of a repository interface has more than one such annotation, the ann
...
 
-- a `Limit` allows the query results to be limited to a given range defined in terms of an offset and maximum number of results,
-- a `Sort` or `Order` allows the query results to be sorted by a given entity attribute or list of attributes, respectively, and
+- a `Restriction` allows the query results to be restricted by applying constraints to entity attributes,
+- a `Sort` or `Order` allows the query results to be sorted by a given entity attribute or list of attributes, respectively,
+- a `Limit` allows the query results to be limited to a given range defined in terms of an offset and maximum number of results, and
 - a `PageRequest` splits results into pages. A parameter of this type must be declared when the repository method returns a `Page` of results, as specified below in <<Offset-based pagination>>, or a `CursoredPage`, as specified in <<Cursor-based pagination>>.
 
+The types `Restriction`, `Sort`, and `Order` are parameterized by an entity type.
+Any special parameter declared with one of these types must be declared with type argument `E` or `? super E` where `E` is the queried entity type.
```
njr-11: In Data 1.0, we allowed `?` for `Sort` and `Order`.  I think we need an allowance here for that usage to remain valid,
```suggestion
Any special parameter declared with one of these types must be declared with type argument `E` or `? super E` where `E` is the queried entity type. Usage of the generic type argument `?` in a special parameter of type `Sort<?>...` is deprecated.
```

gavinking: > In Data 1.0, we allowed `?` for `Sort` and `Order`.

Where precisely did we say that?

I found it in two code examples, but not in any text.

njr-11: It looks like we didn't say it anywhere for `Order`, only `Sort`, so I'll update the suggestion to only say it about `Sort`.  In addition to the code examples (I see the 2 in the spec doc and maybe 5 move in the javadoc), we also have this statement:

> Generic, untyped [Sort](https://jakarta.ee/specifications/data/1.0/apidocs/jakarta.data/jakarta/data/sort) criteria can be supplied directly to a repository method with a variable arguments Sort<?>... parameter.

gavinking: What is the exception if someone passes a `Sort<Author>` to a repository method that queries `Book` :-/

njr-11: > What is the exception if someone passes a `Sort<Author>` to a repository method that queries `Book` :-/

I don't think the spec says anything on this, but I would say that `IllegalArgumentException` seems appropriate if an implementation were to detect it.

njr-11: We probably should have removed `Sort<?>...` from the spec when `Order` was added, but we didn't.

gavinking: tsk.

otaviojava: @gavinking how are going to solve this?

gavinking: Well I agree with @njr-11 that we should have removed those references to `Sort<?>...`. It was an oversight to leave them there.

Perhaps we could say that this usage is deprecated?

njr-11: > Perhaps we could say that this usage is deprecated?

Good idea. That will put us on the path toward removing it completely when we move up major versions at some point.

njr-11: I created #1270 to deprecate it.

PR #1173, opened 2025-07-02
Javadoc for Constraint and Like

njr-11:
Write Javadoc for Constraint and one subtype of Constraint (Like).  If this looks acceptable, we can go on to do likewise for the other subtypes.

### DIFF
--- a/api/src/main/java/jakarta/data/constraint/Constraint.java
+++ b/api/src/main/java/jakarta/data/constraint/Constraint.java
@@ -36,10 +36,9 @@ import jakarta.data.restrict.Restriction;
  *
  * <h2>Method parameter</h2>
  *
- * <p>Constraint parameters can be subtypes of {@code Constraint}, such as
- * {@link Like} or {@link LessThan}. The type argument of the
- * {@code Constraint} subtype (or the type argument of {@code Constraint}
- * that the subtype defines) must match the entity attribute type.
+ * <p>A constraint parameters must be of type {@code C<T>} or {@code C<? super T>} 
+ * where {@code C} is {@code Constraint} or any subtype of {@code Constraint}, such 
+ * as {@link Like} or {@link LessThan}, and {@code T} is the entity attribute type.
  * For example,</p>
  *
  * <pre>
--- a/api/src/main/java/jakarta/data/constraint/Constraint.java
+++ b/api/src/main/java/jakarta/data/constraint/Constraint.java
@@ -63,7 +63,7 @@ import jakarta.data.restrict.Restriction;
  * <h2>Annotation value</h2>
  *
  * <p>Constraint parameters of repository methods can be annotated with the
- * {@link Is} annotation to indicate the subtype of {@code Constraint}.
+ * {@link Is @Is} annotation to indicate the subtype of {@code Constraint}.
  * The type of the method parameter must be the entity attribute type.
  * For example,</p>
  *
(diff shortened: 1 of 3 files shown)


api/src/main/java/jakarta/data/constraint/Constraint.java
```diff
@@ -19,88 +19,374 @@
...
+ * special parameters (such as {@link Restriction} and {@link Order}) in the
+ * method signature.</p>
+ *
+ * <h2>Method parameter</h2>
+ *
+ * <p>Constraint parameters can be subtypes of {@code Constraint}, such as
+ * {@link Like} or {@link LessThan}. The type argument of the
+ * {@code Constraint} subtype (or the type argument of {@code Constraint}
+ * that the subtype defines) must match the entity attribute type.
+ * For example,</p>
```
gavinking:
```suggestion
 * <p>A constraint parameters must be of type {@code C<T>} or {@code C<? super T>} 
 * where {@code C} is {@code Constraint} or any subtype of {@code Constraint}, such 
 * as {@link Like} or {@link LessThan}, and {@code T} is the entity attribute type.
 * For example,</p>
```

api/src/main/java/jakarta/data/constraint/Constraint.java
```diff
@@ -19,88 +19,374 @@
...
+ *                          Like.contains(modelSubstring),
+ *                          Order.by(_Car.year.desc(),
+ *                                   _Car.price.desc(),
+ *                                   _Car.vin.asc()));
+ * </pre>
+ *
+ * <h2>Annotation value</h2>
+ *
+ * <p>Constraint parameters of repository methods can be annotated with the
+ * {@link Is} annotation to indicate the subtype of {@code Constraint}.
```
gavinking:
```suggestion
 * <p>Constraint parameters of repository methods can be annotated with the
 * {@link Is @Is} annotation to indicate the subtype of {@code Constraint}.
```

api/src/main/java/jakarta/data/constraint/Constraint.java
```diff
@@ -19,88 +19,374 @@
...
+ * @param <V> type of the entity attribute.
+ * @since 1.1
+ */
 public interface Constraint<V> {
 
+    /**
+     * <p>Obtains the opposite {@code Constraint}. For example, the opposite of
+     * {@link Like} is {@link NotLike}, the opposite of {@link Null} is
+     * {@link NotNull}, and the opposite of {@link AtLeast} is {@link LessThan}.
+     * </p>
```
gavinking:
```suggestion
     * <p>Obtains the negation of the {@code Constraint}. For example, the negation 
     * of {@link Like} is {@link NotLike}, the negation of {@link Null} is {@link NotNull}, 
     * and the negation of {@link AtLeast} is {@link LessThan}. A value satisfies the
     * negation of a constraint if and only if it does not satisfy the constraint.
     * </p>
```

gavinking (approved):
LGTM, I left a couple of suggestions.

api/src/main/java/jakarta/data/constraint/Constraint.java
njr-11: Good idea to include `? super T`.  There's a grammar error in this ("A constraint parameters"), but I'll add it and then make the correction.

otaviojava (approved):

PR #1178, opened 2025-07-13
Javadoc for Constraint，new method for CrudRepository

luckygc:
Update the javadoc for jakarta.data.constraint.Constraint

### DIFF
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -17,9 +17,13 @@
  */
 package jakarta.data.repository;
 
+import jakarta.data.Order;
 import jakarta.data.exceptions.EntityExistsException;
 import jakarta.data.exceptions.OptimisticLockingFailureException;
 
+import jakarta.data.page.Page;
+import jakarta.data.page.PageRequest;
+import jakarta.data.restrict.Restriction;
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


graemerocher:
I don't think more methods should be added to `CrudRepository` for this. That interface should be kept simple.

api/src/main/java/jakarta/data/repository/CrudRepository.java
```diff
@@ -219,4 +223,32 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
...
+     * @param orderBy     sort criteria that must deterministically order the
+     *                    results; must not be {@code null}.
+     * @return a page of entities matching the restriction criteria; will
+     *         never be {@code null}.
+     * @throws NullPointerException when {@code restriction}, {@code pageRequest},
+     *                              or {@code orderBy} is {@code null}.
+     * @see PageRequest.Mode
+     */
+    @Find
+    Page<T> findAll(Restriction<T> restriction, PageRequest pageRequest, Order<T> orderBy);
```
otaviojava, line 253: We had this discussion on our meeting, at the end we agree to not include new methods at `CrudRepository`

luckygc:
I agree with your points: the built-in super-interfaces should stay minimal, and in many cases you don’t need to extend them at all—just declare the methods you actually need.
By the way, Jakarta Data’s parameter-based query derivation and its fluent query support provide a far better developer experience than Spring Data JPA.

PR #1190, opened 2025-07-30
new entity in TCK for testing restrictions, expressions, constraints

njr-11:
To test the new types of expressions and the restrictions and constraints that can be created from them, the TCK needs an entity with more types of attributes on it (temporal, navigable, and so forth), as well as attributes that can be meaningfully used for computations and comparisons with each other.

This PR adds such an entity and its metamodel, and includes a populator for it using some real world data, and writes a few tests to start out with, covering basic usage of constraints with the `Is` annotation, as well as a basic test of the `Between` constraint as a method parameter type.


tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
```diff
@@ -0,0 +1,110 @@
...
+            @By(_Country.AREA) @Is(LessThan.class) long squareKM);
+
+    @Find
+    @OrderBy(_Country.NAME)
+    List<Country> whereDaylightTimeEndsOn(
+            @Is(EqualTo.class) LocalDate daylightTimeEnds);
+
+    @Find
+    List<Country> withNameAfter(
+            @Is(GreaterThan.class) String name);
```
otaviojava, line 106: ⚠️ **Behavior of `GreaterThan (>)` on strings might differ between SQL and NoSQL databases.**

For example:


```sql
SELECT * FROM users WHERE username > 'john';
```

On MongoDB will be like this:

```java
db.users.find({ name: { $gt: "john" } });
```

Given this list:

1. alice
2. Bob
3. john
4. JOHN
5. zara
6. Zoë
7. José
8. João
9. juhn
10. jón

On Postgresql it returns **juhn**

On MonDB it returns: **JOHN**

On MongoDB there is the collation, but I am not sure at all the databases.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/City.java
```diff
@@ -0,0 +1,55 @@
...
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+package ee.jakarta.tck.data.framework.read.only;
+
+@jakarta.nosql.Embeddable
+@jakarta.persistence.Embeddable
+public class City {
```
otaviojava, line 20: I believe that both could be record right?
I mean, on Jakarta NoSQL we support it and on jakarta Persistence we have it since Jakarta EE 11.

it is up to you.

api/src/main/resources/checkstyle.xml
```diff
@@ -63,7 +63,8 @@
         <module name="UnusedImports"/>
 
         <module name="MethodLength">
-            <property name="max" value="250"/>
+            <!-- TCK prepopulator method -->
+            <property name="max" value="1500"/>
```
otaviojava: You can also put the generated classes, `**/_*.java` as exception.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Country.java
```diff
@@ -0,0 +1,171 @@
...
+        ASIA,
+        CARIBBEAN,
+        EUROPE,
+        NORTH_AMERICA,
+        OCEANIA,
+        SOUTH_AMERICA
+    }
+
+    @jakarta.persistence.Column(nullable = false)
+    private long area; // square kilometers
```
otaviojava, line 26: Remember to include the `jakarta.nosql.Column`as well.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Country.java
```diff
@@ -0,0 +1,171 @@
...
+ */
+package ee.jakarta.tck.data.framework.read.only;
+
+import java.time.LocalDate;
+
+@jakarta.nosql.Entity
+@jakarta.persistence.Entity
+public class Country {
+
+    public enum Region {
```
otaviojava: It is a taste and super subjective, I know.

But do you mind having Region as a class?

Just as suggestion as well.

api/src/main/resources/checkstyle.xml
otaviojava: On this line: https://github.com/jakartaee/data/blob/main/pom.xml#L99

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
njr-11: Thanks Otavio, that is excellent information to know!
In this particular dataset, the first letter is always capital, followed by lowercase, and only ASCII characters are used, so there should hopefully not be any ambiguity about ordering.  If you see a specific case in the data that will be a problem, point it out and we can adjust it (possibly by removal of a name or replacing it with something non-ambiguous).

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/City.java
njr-11: Yes, it would be valid to use a record for the embeddable in JPA.  Either way is fine with me, so I can switch it.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Country.java
njr-11: That's fine. I'll move the enumeration to its own class.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/City.java
njr-11: Actually, I'm going to wait on switching the embeddable to record because EclipseLink still has a known bug with records as embeddables that needs to be fixed and I'd like to still be able to run these tests to verify they are valid. I'll leave it as as normal class for now, but I'm willing to switch it to record in the future after the EclipseLink bug is fixed.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Country.java
njr-11: Thanks - I added it now, as well as addressed the comments to move the Region enumeration to a top level class and exclude only the populator class from the MethodLength rule.  It's ready for review again.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/City.java
otaviojava: It is up to you ok, I am fine with both.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Country.java
gavinking: Why does NoSQL requite an empty `@Column` annotation? Is it really the case that this implies a not null constraint?

gavinking: Or is it because fields on NoSQL are transient by default?

gavinking (approved):

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Country.java
otaviojava: Jakarta NoSQL by default is transient, thus, we need to put the annotation to define the attribute as persistable.

gavinking: OK, I see.

PR #1191, opened 2025-08-01
Internal review comment: Use present tense for arithmetic expressions

njr-11:
One of the comments from an internal review we had of what has been done thus far in the 1.1 spec is that present tense should be used for the arithmetic operators rather than past tense, so `subtractedFrom` would change to `subtractFrom` and `dividedInto` would change to `divideInto`.  Also, it was requested that we change `divide` to `divideBy` so that it reads better.  For example,

```
total.divideBy(amount);
```

instead of
```
total.divide(amount);
```

A few other things were brought up during the review as well, but should be looked at and considered separately from this.


gavinking (changes requested):
I actually disagree quite strongly with this.

Expressions are "pure", i.e. "functional" values.

They are not procedural "commands".

Present tense emphasizes a dynamic, procedural nature which simply doesn't exist in a mathematical expression. Past tense emphasizes that it's an immutable value.

I know this isn't the perspective that people with a background in Java and procedural OO are used to taking; it's a view that's more common in the FP community, and in mathematics.

Nobody reads the expression `5/2` as "divide five by 2", they read it as "five divided by two".

We don't need to go against centuries of mathematical convention here.

njr-11:
Given the discussion on this, I spoke with the reviewer who was requesting the change to present tense and we had a conversation about the need for consistency. I made the argument that always using the colloquial form of the expression -- how a person reads the expression -- is a way of having consistency.   In that case, only `divide` needs to change to `dividedBy`, so that we have:

`10 ÷ 5`: "ten divided by five"
`10 - 5`: "ten minus five"
`10 + 5`: "ten plus five"
`10 x 5`: "ten times five"
`-5 + 10`: "five subtracted from ten"
```
   ___
3 ) 15
```
: "3 divided into 15"

I changed the PR to make us consistent with the above.

api/src/main/java/jakarta/data/expression/NumericExpression.java
```diff
@@ -191,15 +191,15 @@ default NumericExpression<T, N> times(N factor) {
...
+     *         _Car.price.asDouble().dividedBy(1.0 + discountRate).lessThan(27000.0));
      * </pre>
      *
      * @param divisor the value by which to divide. Must not be {@code 0} or
      *                {@code null}.
      * @return an expression for the function that computes the quotient.
      * @throws NullPointerException if the supplied divisor value is null.
      */
-    default NumericExpression<T, N> divide(N divisor) {
+    default NumericExpression<T, N> dividedBy(N divisor) {
```
pzygielo, line 202: I'm confused. The opening comment in PR says "Also, it was requested that we change divide to divideBy so that it reads better" "that present tense should be used for the arithmetic operators" but here I see the past tense.

pzygielo: I understand https://github.com/jakartaee/data/pull/1191#issuecomment-3249723662 explains that.

pzygielo:
> I know this isn't the perspective that people with a background in Java and procedural OO 

Just out of curiosity - what is the target user of that API?

otaviojava:
> I actually disagree quite strongly with this.
> 
> Expressions are "pure", i.e. "functional" values.
> 
> They are not procedural "commands".
> 
> Present tense emphasizes a dynamic, procedural nature which simply doesn't exist in a mathematical expression. Past tense emphasizes that it's an immutable value.
> 
> I know this isn't the perspective that people with a background in Java and procedural OO are used to taking; it's a view that's more common in the FP community, and in mathematics.
> 
> Nobody reads the expression `5/2` as "divide five by 2", they read it as "five divided by two".
> 
> We don't need to go against centuries of mathematical convention here.

@gavinking I will assume that we agree on keep the divided by terminology, thus, we will merge this one.

Issue #1197, opened 2025-08-21
[Use Case]: Ordering projections


cunhazera:
### As a ...

- [ ] Application user/user of the configuration itself
- [x] API user (application developer)
- [ ] SPI user (container or runtime developer)
- [ ] Specification implementer

### I need to be able to ...

How can I combine `@OrderBy` or `Order.by` with methods that return a Java record representing a `JOIN` ?

Using the example from the documentation:

``` java
record AuthorBookSummary(
        String isbn,
        String ssn,
        String authorName,
        String title
) {}
```

``` java
    @Query("select isbn, ssn, name, title " +
            "from Author join books " +
            "where title like :pattern")
    List<AuthorBookSummary> summariesForTitle(@Pattern String pattern);
```

What can I do if I want to do something similar to the following:

``` java
    @Query("select isbn, ssn, name, title " +
            "from Author join books " +
            "where title like :pattern")
    @OrderBy("isbn")
    List<AuthorBookSummary> summariesForTitle(@Pattern String pattern);
```

Or even

``` java
    @Query("select isbn, ssn, name, title " +
            "from Author join books " +
            "where title like :pattern")
    List<AuthorBookSummary> summariesForTitle(@Pattern String pattern, Order<Author> order);
// ...
library.summariesForTitle("pattern", Order.by(_Author.name.desc()));
```

### Which enables me to ...

Make more use of the existing ordering resources, specially passing an `Order` object as parameter

gavinking:
> Do you have a link to the documentation where you found this example?

It looks a bit like some examples from the Hibernate docs, but not exactly like any of them.

> JDQL doesn't have a join, so I wouldn't expect the spec documentation or Javadoc to have examples with join. 

Sure, but implementations backed by by JPA are strongly encouraged by the spec to support all of JPQL, so the example is not unreasonable.

> that should have been `@Param("pattern")` there.

Actually that's not a requirement. Remember we said:

> The {@code Param} annotation is unnecessary when the method parameter name matches the query language named parameter name and the application is compiled with the {@code -parameters} compiler option making parameter names available at runtime.

All that said, I'm struggling to understand what is the new feature which is being requested here.

gavinking:
I'm pretty sure it was referring to this experimental annotation:

https://docs.jboss.org/hibernate/orm/6.6/javadocs/org/hibernate/annotations/processing/Pattern.html

which is marked `@Incubating` and which will be deprecated once Jakarta Data 1.1 is finished.

[...]

PR #1199, opened 2025-08-22
Statement allowing sorting on projection attributes

njr-11:
This PR won't cover everything that is asked for under #1197, but I assume we should at least say this much regardless of what else we decide to include, so I'll start out with this.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Select.java
+++ b/api/src/main/java/jakarta/data/repository/Select.java
@@ -52,7 +52,7 @@ import java.lang.annotation.Target;
  * <p>This annotation must not be used in other locations.</p>
  *
  * <p>Sort criteria for a repository method that returns a projection
- * can include any of the returned entity attributes that are sortable.
+ * can include any of the returned entity attributes that are of sortable type.
  * </p>
  *
  * @see Find


gavinking:
Yeah, that's a good clarification to start with.

Though I wonder if we should come up with something better than reciting "a repository method which returns a subset of entity attributes" in many places.

njr-11:
> Though I wonder if we should come up with something better than reciting "a repository method which returns a subset of entity attributes" in many places.

Good idea. We can just define the term in one place for the sake of new/inexperienced users and in every place thereafter be more concise.  I added a [commit](https://github.com/jakartaee/data/pull/1199/commits/124526cbfab351e01351cbeab00991f394e2ff29) to this PR to do that.  The commit also fixes up the formatting of few nearby places that I noticed.

api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -51,6 +51,10 @@
  *
  * <p>This annotation must not be used in other locations.</p>
  *
+ * <p>Sort criteria for a repository method that returns a projection
+ * can include any of the returned entity attributes that are sortable.
```
gavinking: Perhaps something like:

```suggestion
 * can include any of the returned entity attributes that are of sortable type.
```

api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -141,9 +143,8 @@
      * </pre>
      *
      * <p>For more concise code, the {@code Select} annotation can be omitted
-     * from
-     * record components that have the same name as the entity attribute name,
-     * such as {@code model} in the above example.</p>
+     * from record components that have the same name as the entity attribute
+     * name, such as {@code model} in the above example.</p>
```
gavinking:
```suggestion
     * from a record component that has the same name as the entity attribute,
     * such as {@code model} in the above example.</p>
```

gavinking (approved):

Issue #1168, opened 2025-06-27
Update Jenkins jobs to push to the new sonatype repository


KyleAure:
See original issue here: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5854

We need to: 

- [x] Update our builds to publish to `central.sonatype.org` instead of `jakarta.oss.sonatype.org`
- [x] #1235
  - [ ] Use the new [central-staging-plugin](https://github.com/eclipse-cbi/central-staging-plugins) to perform stage -> publish workload 
- [x] Update our builds to publish snapshots to `central.sonatype.com/repository/maven-snapshots/` instead of `jakarta.oss.sonatype.org`
  - [x] Use the new [central publishing plugin](https://central.sonatype.org/publish/publish-portal-maven/) to directly publish snapshots to `central.sonatype.org`

KyleAure:
This may actually need to wait until the EE4J parent project is updated.
See: https://github.com/eclipse-ee4j/ee4j/issues/98

Otherwise, we'll have to overwrite all the profiles.
I'll bring this up on the next platform meeting.

otaviojava:
@KyleAure 
We don't have control over this; when they release a new version of EE4J, it could take anywhere from one day to almost a year.

I would go to overwrite all the profiles, and when they update, we remove it.

Currently, Eclipse Payara requires version 1.0.2, and we need a snapshot that works.

njr-11:
Kyle just raised the issue this week with the Jakarta platform. I think we should give them a chance to resolve it before taking other actions of our own.

KyleAure:
@otaviojava 
The true blocker isn't our ability to push an artifact to maven central, but that we currently do not have a solution for snapshots/staged artifacts.  We need to have a solution for that so that Payara can verify the fix before we publish (publicly) to maven central.

otaviojava:
Sad, 
We haven't had a new update since June 6 on snapshot:

https://jakarta.oss.sonatype.org/content/repositories/snapshots/jakarta/data/jakarta.data-api/1.1.0-SNAPSHOT/

otaviojava:
> It looks like you have a solution for doing a release (GA and Snapshot), but not one for staging.

To keep the spec rolling, shall we fix at least one of them?
I know that implies not to release version 1.0.2, but at least we can release a snapshot and a milestone.

KyleAure:
> At Eclipse JNoSQL, we migrated the pom.xml and it is working.
[...]

PR #1204, opened 2025-09-05
fix: javadoc failures

KyleAure:
Related #1168 

Jenkins release build is catching javadoc errors when attempting to deploy that our current github actions build does not.
- add build step to github actions to catch errors put into javadoc during development
- fix errors that made it in while jenkins builds were broken. 
- skip javadoc for tck submodules (performance improvement and error prone)

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Select.java
+++ b/api/src/main/java/jakarta/data/repository/Select.java
@@ -71,7 +71,7 @@ public @interface Select {
      * <p>The {@code Select} annotation can be used with repository methods
      * that return a single entity attribute or multiple entity attributes.</p>
      *
-     * <h2>Method that returns Single Entity Attributes</h2>
+     * <h4>Method that returns Single Entity Attributes</h4>
      *
      * <p>Place the {@code Select} annotation on a repository find method and
      * assign the annotation value to be the name of a single entity attribute.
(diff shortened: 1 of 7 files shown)


KyleAure:
Looks like we still have some javadoc issues.  I'll pick this up again on Monday.

api/src/main/java/jakarta/data/repository/Select.java
```diff
@@ -71,7 +71,7 @@
      * <p>The {@code Select} annotation can be used with repository methods
      * that return a single entity attribute or multiple entity attributes.</p>
      *
-     * <h2>Method that returns Single Entity Attributes</h2>
+     * <h4>Method that returns Single Entity Attributes</h4>
```
otaviojava, line 74: @KyleAure why did you move it to h4?

KyleAure: Accessibility: when the javadoc is converted into HTML the element name `value` is at header `<h3>` so anything 
below that needs to be `<h4>` and higher. 

Screenshot 2025-09-08 at 14 35 07

KyleAure: And also the Javadoc plugin throws an error for it :D

njr-11: Nice catch. It looks like `<h2>` is fine elsewhere because we have it at class documentation level in the other cases, but here it is on the method documentation and doesn't render very well as you pointed out.

otaviojava:
Thank you, @KyleAure please let us know if we got our snapshot version soon :)

PR #1211, opened 2025-09-15
Pattern translation is broken when already-escaped pattern is supplied

njr-11:
While trying out `Like` constraints, I noticed that the API method for supplying an already escaped pattern is buggy in how it translates the pattern because the implementation has no awareness of whether or not the pattern is already escaped.  This PR adds that awareness and corrects the implementation. It could have been alternatively corrected in a less verbose form, but this seemed clearer.  Let me know if anyone prefers the other form and I can switch it.

This also adds test cases, including trying out the example usage of the API that we had stated in Javadoc, which in some cases was also broken by the same issue.

This also adds some clarity to the Javadoc when the pattern supplied to an API method does not include escape characters. Previously this was unstated, risking that a user might make wrong assumptions.

Also, I noticed some inconsistency between Like.pattern(pattern) and Like.pattern(pattern, charwildcard, stringwildcard) methods, where one of them ends up with NULL `escape()` and the other with `\`.  These really ought to be the same code path, except for the use of default vs custom wildcards.  This appears to be the only place where `LikeRecord` uses a NULL value, also requiring it specify `Character` type rather than `char`.  A commit added to this PR gets rid of the possibility of NULL here and makes both consistent, with Javadoc added to explain that the escape character returned by the `escape()` SPI method corresponds to the `pattern()` SPI method.

### DIFF
--- a/api/src/main/java/jakarta/data/constraint/Constraint.java
+++ b/api/src/main/java/jakarta/data/constraint/Constraint.java
@@ -195,7 +195,8 @@ public interface Constraint<V> {
 
     /**
      * <p>Requires that the constraint target match the given {@code pattern},
-     * in which {@code _} and {@code %} represent wildcards.</p>
+     * in which {@code _} and {@code %} represent wildcards, and no escape
+     * character is included.</p>
      *
      * @param pattern a pattern in which {@code _} matches a single character
      *                and {@code %} matches 0 or more characters.
@@ -209,7 +210,8 @@ public interface Constraint<V> {
 
     /**
      * <p>Requires that the constraint target match the given {@code pattern},
-     * in which the given characters represent wildcards.</p>
+     * in which the given characters represent wildcards, and no escape
+     * character is included.</p>
      *
      * @param pattern        a pattern that can include the given wildcard
      *                       characters.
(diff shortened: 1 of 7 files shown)


api/src/main/java/jakarta/data/constraint/Constraint.java
```diff
@@ -195,7 +195,8 @@ static <V> NotIn<V> notIn(Set<V> values) {
 
     /**
      * <p>Requires that the constraint target match the given {@code pattern},
-     * in which {@code _} and {@code %} represent wildcards.</p>
+     * in which {@code _} and {@code %} represent wildcards, and no escape
+     * character is included.</p>
```
gavinking, line 199: "no escape character is included" is very ambiguous to me. Does it mean:

- there is no escape character at all, or
- there is no escape character explicitly specified, so the escape character defaults to `\`?

njr-11: Good point. I'll give the wording another try.

njr-11: #1215 addresses this because the current PR is merged

Issue #1218, opened 2025-09-22
[Use Case]: disambiguate the query language


gavinking:
### As a ...

- [ ] Application user/user of the configuration itself
- [x] API user (application developer)
- [ ] SPI user (container or runtime developer)
- [ ] Specification implementer

### I need to be able to ...

explicitly specify the query language I'm using.

### Which enables me to ...

write queries in the native query language of my datastore (e.g. SQL, Cypher, etc).

### Additional information

`@Query` gives us a way to write queries in Jakarta Query.

But it's still very common for people to want to write native queries, for example, in SQL.

Currently we don't provide a standard way to do that, which is sort-of OK, because native queries aren't portable between providers anyway.

OTOH, every Persistence-backed implementation (and, one would assume, any implementation backed by a SQL database) is going to have the ability to query in SQL, and so it's annoying that such code would not be portable between at least those implementations.

I therefore propose that we introduce some sort of base-level support for native queries.

One obvious option would be to add a `language` element to `@Query`, but I dislike that for a couple of reasons.

First, the resulting annotation ends up a bit ugly:

```java
@Query(language = "SQL", 
       value = "select * from things for no key update")
```

Second, we would have to introduce a convention around `language` names: the spec would need to say that SQL is called `"SQL"` and not `"sql"`, Cypher is called `"Cypher"`, etc, etc. Yew.

On the other hand, while it's going to be very common that an implementation supports *two* query languages, it's not going to be common _at all_ for an implementation to support *more than two*. I therefore suggest a new annotation for just the *secondary* i.e. native query language:

 ```java
@NativeQuery("select * from things for no key update")
```

Now, without a way to specify fancy explicit result set mappings, this functionality is going to be very limited by nature, but it's better than _nothing_, it seems to me.

gavinking:
> I'm not sure that Jakarta Persistence actually guarantees a way to send in a SQL query, so this might not be implementable for some Persistence providers.

`EntityManager.createNativeQuery()`?

gavinking:
> For some reason, I had been looking for an API to obtain a `java.sql.Connection`

That is `runWithConnection()`/`callWithConnection()`, if you ever really need it (definitely not in this case).

gavinking:
In case it was unclear, the heart of this proposal is this bit at the end:

> On the other hand, while it's going to be very common that an implementation supports two query languages, it's not going to be common at all for an implementation to support more than two. I therefore suggest a new annotation for just the secondary i.e. "native" query language:
>
[...]

PR #1225, opened 2025-09-24
Add language and document limitations

njr-11:
This PR attempts to write up the proposal under #1218 to add a language attribute and document a bunch of limitations that would be needed for native queries.  I think we could go as far as to let the user supply native queries that we send directly to the database, but it is out of scope for a Jakarta Data provider to need to alter these queries to add sorting, restrictions, and so forth or infer parts of them that were not supplied.

Also, it should be noted that in new text, I referred to JQL / Jakarta Query Language, but I did not make any attempt to switch existing text over to JQL because we have not officially brought that in yet.  It seemed pointless to write new things in terms of JDQL/JPQL that we expect to be replaced.  It should be easy later on to scan the spec for occurrences of either when JQL is either voted in or out.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Query.java
+++ b/api/src/main/java/jakarta/data/repository/Query.java
@@ -20,6 +20,7 @@ package jakarta.data.repository;
 import jakarta.data.Limit;
 import jakarta.data.Order;
 import jakarta.data.Sort;
+import jakarta.data.page.CursoredPage;
 import jakarta.data.page.PageRequest;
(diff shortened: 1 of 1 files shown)


otaviojava (changes requested):
I prefer each language to have its own annotation, as I mentioned.

But I would be happier, at least, to use a different annotation as Gavin proposed.

However, before proposing new features, let's first complete what we have in our plan:

https://projects.eclipse.org/projects/ee4j.data/releases/1.1/plan

PR #986, opened 2025-02-24
fix build

gavinking:
Since https://github.com/jakartaee/data/pull/984 the build is broken.

PR #1232, opened 2025-09-30
Fix junit-bom type

pzygielo:
- https://github.com/jakartaee/data/pull/986


njr-11 (approved):
It looks like you and Kyle made the same fix at the same time. I'll approve both and let the two of you decide which to merge.

pzygielo:
- covered by #1231

PR #1234, opened 2025-10-01
refactor: update to project 2.0.0-SNAPSHOT

KyleAure:
I was able to publish a snapshot in the concurrency project with a similar configuration.
See: https://github.com/jakartaee/concurrency/pull/760

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/Attribute.java
+++ b/api/src/main/java/jakarta/data/metamodel/Attribute.java
@@ -127,7 +127,7 @@ public interface Attribute<T> {
      * @return the declaring class
      * @throws UnsupportedOperationException if the declaring type is not
      *                                       known.
-     * @apiNote This is only guaranteed to be known if a static <code>of</code>
+     * apiNote This is only guaranteed to be known if a static <code>of</code>
      *          method, such as {@link BasicAttribute#of(Class, String, Class)},
      *          was used to obtain the instance.
      * @since 1.1
@@ -143,7 +143,7 @@ public interface Attribute<T> {
      * @return the type of the entity attribute.
      * @throws UnsupportedOperationException if the entity attribute type is
      *                                       not known.
-     * @apiNote This is only guaranteed to be known if a static <code>of</code>
+     * apiNote This is only guaranteed to be known if a static <code>of</code>
      *          method, such as {@link BasicAttribute#of(Class, String, Class)},
      *          was used to obtain the instance.
      * @since 1.1
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
     <parent>
         <groupId>org.eclipse.ee4j</groupId>
         <artifactId>project</artifactId>
-        <version>1.0.9</version>
+        <version>2.0.0-SNAPSHOT</version>
     </parent>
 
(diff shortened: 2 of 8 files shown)


otaviojava (approved):

api/src/main/java/jakarta/data/metamodel/Attribute.java
```diff
@@ -127,7 +127,7 @@ public interface Attribute<T> {
      * @return the declaring class
      * @throws UnsupportedOperationException if the declaring type is not
      *                                       known.
-     * @apiNote This is only guaranteed to be known if a static <code>of</code>
+     * apiNote This is only guaranteed to be known if a static <code>of</code>
```
njr-11, line 130: Will this cause the text `apiNote` to appear in the Javadoc?  Same comment in 2 other places.

pom.xml
```diff
@@ -24,7 +24,7 @@
     <parent>
         <groupId>org.eclipse.ee4j</groupId>
         <artifactId>project</artifactId>
-        <version>1.0.9</version>
+        <version>2.0.0-SNAPSHOT</version>
```
njr-11, line 27: The next version of Jakarta Data is expected to be 1.1, not 2.0.

pzygielo, line 27: but it's parent, not the project, and this parent does not share lifecycle with children

---

_Apparently the comment I replied to was removed, which made my response to dangle with no context._

njr-11: Never mind. I was looking at the wrong place.  That is definitely not the version of Jakarta Data.

api/src/main/java/jakarta/data/metamodel/Attribute.java
KyleAure: Yes, the API note will appear in javadoc the way it is right now.
Since `@apiNote` is not a supported javadoc tag it was not clear to me what the purpose is. 
Do we want these notes in javadoc? Or are these notes only for implementations of Data and therefore should not be visible in javadoc?
Based on that answer I can fix these notes (as they are in the incorrect place the way they are now).

njr-11: If I remember correctly, the usage of `@apiNote` was asked for by a reviewer at one point.  I just checked and it is not showing in the Javadoc, but does appear in some IDE's,

image

gavinking: We use `@apiNote` all over the place and it's very nice. For example: https://docs.jboss.org/hibernate/orm/7.1/javadocs/org/hibernate/exception/LockTimeoutException.html

pzygielo: > Since `@apiNote` is not a supported javadoc tag it was not clear to me what the purpose is. 

https://nipafx.dev/javadoc-tags-apiNote-implSpec-implNote/#generating-javadoc

KyleAure: Okay, I'm inferring from the comments that the answer to my question is:
   yes, you want this visible in javadoc.
If that is the case I've fixed it here: https://github.com/jakartaee/data/pull/1238

PR #1257, opened 2025-10-17
TCK tests for basic constraint subtypes

njr-11:
Add TCK tests for constraint subtypes, such as Like, GreaterThan, NotIn, ... where basic values (not expressions) are supplied to the constraint.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
@@ -77,8 +77,18 @@ public interface Countries extends CrudRepository<Country, String> {
             @By(_Country.NAME) NotLike excludePattern);
 
     @Find
+    List<Country> highlyPopulousInRegion(
+            @By(_Country.REGION) EqualTo<Region> region,
+            @By(_Country.POPULATION) AtLeast<Long> minPopulation,
+            Order<Country> order);
+
+    @Find
     Stream<Country> inRegion(
             @By(_Country.REGION) @Is Region regionOfWorld);
+
+    @Find
+    List<Country> inSomeOtherRegionThan(
+            @By(_Country.REGION) NotEqualTo<Region> exclude);
 
     @Find
     List<Country> lessPopulousThan(
@@ -117,15 +127,6 @@ public interface Countries extends CrudRepository<Country, String> {
     @Find
     List<Country> outsideOfRegions(
             @By(_Country.REGION) @Is(NotIn.class) Collection<Region> excluded);
-
-    @Find
-    List<Country> outsideOfTheseRegions(
-            @By(_Country.REGION) NotEqualTo<Region> exclude1,
-            @By(_Country.REGION) NotEqualTo<Region> exclude2,
-            @By(_Country.REGION) NotEqualTo<Region> exclude3,
-            @By(_Country.REGION) NotEqualTo<Region> exclude4,
-            @By(_Country.REGION) NotEqualTo<Region> exclude5,
-            @By(_Country.REGION) NotEqualTo<Region> exclude6);
 
     @Find
     Stream<Country> populated(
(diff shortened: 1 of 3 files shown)


tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
```diff
@@ -88,6 +118,15 @@ List<Country> outsideOfRegion(
     List<Country> outsideOfRegions(
             @By(_Country.REGION) @Is(NotIn.class) Collection<Region> excluded);
 
+    @Find
+    List<Country> outsideOfTheseRegions(
+            @By(_Country.REGION) NotEqualTo<Region> exclude1,
```
otaviojava: Is this a valid?

Should we use: Restrict.all() instead?

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
```diff
@@ -58,10 +63,31 @@ List<Country> asLargeOrBigger(
     List<Country> byCountryCodes(
             @By(_Country.CODE) @Is(In.class) Collection<String> codes);
 
+    @Find
+    @Select(_Country.CODE)
+    List<String> countryCodesUpTo(
```
otaviojava, line 68: IMHO: this is also invalid, the goal of this is to have dynamic restriction, right? Where is the dynamic on this parameter?

The parameter should be `Restriction< Country>`

If we agree with this being valid, we should recondiser the: **Is** annotation.

IMHO: Why should I have two ways to do the samething in the API?

gavinking: Yes, we had said that this was going to be allowed. It’s perhaps not terribly useful for `AtMost` but for restriction types such as `Like` which have multiple constructor methods it’s useful.

otaviojava: Once we agreed that it is valid, should we have two ways to do the same thing?

I mean:

```java
@By(_Country.CODE) AtMost<String> maxCode
```

V.S:

```java
@Is(AtMost.class) String maxCode
```

Can we stay with only one?

gavinking: I agree they’re overlapping, but they’re not exactly the same. I think the extra flexibility is nice, given that neither feature adds much weight on its own. We’re not wasting much of our complexity budget on this.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
njr-11: Yes, having multiple constraints on the same attribute is valid.  I agree this test case uses that ability in a way that isn't very practical and that the more likely usage here would be to use Restriction and allow flexibility on how many.  But the point of these tests isn't to cover Restriction.  That will be a separate set of tests.  The point here is to cover the constraint types and what it is possible to do with them.  If you like, I can switch the NotEqualTo test to only have one parameter/constraint and then add a separate test for combining several constraints that might look more meaningful.

njr-11: @otaviojava is this better? I separated out the test for multiple constraints and made it more meaningful: https://github.com/jakartaee/data/pull/1257/commits/21db0eb9ef8c00ac6dbf83cee7a9b65501441353

otaviojava: Better, Thanks Nathan.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
otaviojava: My point is more worried with Frankinstain possibilities:

```java
List<Book> findBook(@By(_Country.CODE) AtMost<String> maxCode, @Is(AtLeast.class) String minCode, Restriction<Book> books, Order<Book> order);
```

gavinking: Well, I think that's a *very* strange way to write that method, and that nobody would do it that way, but I don't see why it's something that we need to explicitly *prevent*.

In general, it's not really possible to prevent people from doing things in a way that's inconsistent and inelegant. There's certain sorts of **bugs** which we can help prevent, and sometimes we can lead people down a path toward more elegant code, but ultimately some people are always going to find a way to make a mess.

PR #1270, opened 2025-11-05
Deprecate generic untyped Sort

njr-11:
Deprecation of untyped `Sort<?>...` as a repository method parameter in Data 1.1.

Due to the deprecation I have removed or switched all code examples and references to this pattern from the javadoc and specification, except for in the new Deprecations section which I added to document the pattern as deprecated.

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -886,7 +886,7 @@ import java.util.Set;
  * across invocations. One way to achieve this is to include the unique
  * identifier in the sort criteria.</p>
  *
- * <h3>Sorting at runtime</h3>
+ * <h3>Programmatic sorting</h3>
  *
  * <p>When a page is requested with a {@code PageRequest}, dynamic sorting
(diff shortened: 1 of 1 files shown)


api/src/main/java/module-info.java
```diff
@@ -873,22 +873,24 @@
...
+ * page1 = products.findByNameLike("%phone%",
+ *                                 PageRequest.ofSize(20));
  * </pre>
  *
  * <p>When using pagination, always ensure that the ordering is consistent
  * across invocations. One way to achieve this is to include the unique
  * identifier in the sort criteria.</p>
  *
- * <h3>Sorting</h3>
+ * <h3>Sorting at runtime</h3>
```
gavinking: All sorting happens at runtime. How about:

```suggestion
 * <h3>Programmatic sorting</h3>
```

gavinking (approved):

gavinking:
Question: if we're deprecating this, should we remove the TCK tests which test it, or leave them there?

PR #1303, opened 2025-12-05
attempt to respecify dataSource

gavinking:
Refinement of what was stated in #423 to reflect what we're actually doing with this in practice. Basically, say that the `dataSource` can just be the name of a persistence unit.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Repository.java
+++ b/api/src/main/java/jakarta/data/repository/Repository.java
@@ -66,22 +66,21 @@ import java.lang.annotation.Target;
 @Target(ElementType.TYPE)
 public @interface Repository {
     /**
-     * Value for the {@link Repository#provider()} attribute that allows the use
-     * of any available Jakarta Data provider that supports the type of entity
-     * annotation that is present on the repository's entity class.
+     * Value for the {@link #provider()} attribute that allows the use of
+     * any available Jakarta Data provider that supports the type of entity
+     * annotation that is present on the repository's entity classes.
      */
     String ANY_PROVIDER = "";
 
     /**
-     * <p>Value for the {@link Repository#dataStore()} attribute that indicates
-     * to use
+     * <p>Value for the {@link #dataStore()} attribute that indicates use of
      * a default data store.</p>
      *
-     * <p>When running in a Jakarta EE profile or platform and the entity
-     * annotations
-     * indicate a relational database, the default data store is the Jakarta EE
-     * default data source, {@code java:comp/DefaultDataSource}. Otherwise, the
-     * default data store is determined by the Jakarta Data provider.</p>
+     * <p>In a Jakarta EE profile or platform environment, if the repository
+     * implementation requires direct access to a {@code javax.sql.DataSource},
+     * the default data store is the Jakarta EE default data source with name
+     * {@code java:comp/DefaultDataSource}. Otherwise, the default data store
+     * is determined by the Jakarta Data provider.</p>
      *
      * <p>The default data store might require additional vendor-specific
      * configuration, depending on the vendor.</p>
@@ -100,14 +99,23 @@ public @interface Repository {
      * Interoperability with Jakarta Config is reserved for future versions
      * of Jakarta Data.
      * </li>
-     * <li>If running in a Jakarta EE profile or platform and the entity annotations
-     * indicate a relational database and the value begins with {@code java:} and
-     * matches the name of a {@code jakarta.annotation.sql.DataSourceDefinition},
-     * the JNDI name of a data source, or a resource reference to a data source,
-     * then the corresponding {@code javax.sql.DataSource} is used as the data store.
-     * If the same conditions are met but the value matches a persistence unit
-     * reference, then the corresponding {@code jakarta.persistence.PersistenceUnit}
-     * is used as the data store.
+     * <li>In a Jakarta EE profile or platform environment, if the repository
+     * implementation requires direct access to a {@code javax.sql.DataSource}
+     * and if the value begins with {@code java:} and matches the name of a
+     * {@code jakarta.annotation.sql.DataSourceDefinition}, the JNDI name of
+     * a {@code DataSource}, or a resource reference to a {@code DataSource},
+     * then the corresponding {@code DataSource} obtained from JNDI is used.
+     * </li>
+     * <li>In a Jakarta EE profile or platform environment, if the repository
+     * implementation requires direct access to a Jakarta Persistence
+     * {@code jakarta.persistence.EntityManagerFactory} and if the value does
+     * not begin with {@code java:} and matches the name of a persistence unit,
+     * then the container-managed {@code EntityManagerFactory} for that
+     * persistence unit is used, obtained as if it had been injected using the
+     * {@code PersistenceUnit} annotation. Or, if the value does begin with
+     * {@code java:} and matches the name of a persistence unit reference,
+     * then the corresponding {@code EntityManagerFactory} obtained from JNDI
+     * is used.
      * </li>
      * <li>Otherwise, the value serves as an identifier linking to vendor-specific
      * configuration for the Jakarta Data provider to interpret in a vendor-specific
--- a/api/src/main/java/jakarta/data/repository/Repository.java
+++ b/api/src/main/java/jakarta/data/repository/Repository.java
@@ -77,7 +77,7 @@ public @interface Repository {
      * a default data store.</p>
      *
      * <p>In a Jakarta EE profile or platform environment, if the repository
-     * implementation requires direct access to a {@code javax.sql.DataSource},
+     * requires a {@code javax.sql.DataSource},
      * the default data store is the Jakarta EE default data source with name
      * {@code java:comp/DefaultDataSource}. Otherwise, the default data store
      * is determined by the Jakarta Data provider.</p>
--- a/api/src/main/java/jakarta/data/repository/Repository.java
+++ b/api/src/main/java/jakarta/data/repository/Repository.java
@@ -77,8 +77,8 @@ public @interface Repository {
      * a default data store.</p>
      *
      * <p>In a Jakarta EE profile or platform environment, if the repository
-     * requires a {@code javax.sql.DataSource},
-     * the default data store is the Jakarta EE default data source with name
+     * requires a {@code javax.sql.DataSource}, the default data store is the
+     * Jakarta EE default data source with name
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/data/repository/Repository.java
```diff
@@ -100,14 +100,23 @@
      * Interoperability with Jakarta Config is reserved for future versions
      * of Jakarta Data.
      * </li>
-     * <li>If running in a Jakarta EE profile or platform and the entity annotations
-     * indicate a relational database and the value begins with {@code java:} and
-     * matches the name of a {@code jakarta.annotation.sql.DataSourceDefinition},
-     * the JNDI name of a data source, or a resource reference to a data source,
```
gavinking, line 106: @njr-11 One thing I'm not sure about with what you wrote here: with a resource reference (declared with `@Resource`), isn't it more common that the name does *not* belong to the `java:` namespace?

If that's correct, then this sentence looks wrong to me.

api/src/main/java/jakarta/data/repository/Repository.java
```diff
@@ -66,22 +66,21 @@
...
+     * <p>Value for the {@link #dataStore()} attribute that indicates use of
      * a default data store.</p>
      *
-     * <p>When running in a Jakarta EE profile or platform and the entity
-     * annotations
-     * indicate a relational database, the default data store is the Jakarta EE
-     * default data source, {@code java:comp/DefaultDataSource}. Otherwise, the
-     * default data store is determined by the Jakarta Data provider.</p>
+     * <p>In a Jakarta EE profile or platform environment, if the repository
+     * implementation requires direct access to a {@code javax.sql.DataSource},
```
njr-11: For Javadoc, use language that will be readily understood by the the application developer,
```suggestion
     * requires a {@code javax.sql.DataSource},
```

The prior wording "repository implementation requires direct access to" could be confusing to the application developer because they don't know the details of the implementation and whether the internals of the vendor implementation uses direct vs indirect access and what that means.

api/src/main/java/jakarta/data/repository/Repository.java
njr-11: > isn't it more common that the name does _not_ belong to the `java:` namespace?

No, resource reference JNDI names are in `java:comp`, `java:module`, `java:app`, or `java:global`, all of which begin with `java:`, so this is a very convenient way to identify when a name needs to be looked up in JNDI.

api/src/main/java/jakarta/data/repository/Repository.java
```diff
@@ -100,14 +99,23 @@
...
+     * <li>In a Jakarta EE profile or platform environment, if the repository
+     * implementation requires direct access to a Jakarta Persistence
+     * {@code jakarta.persistence.EntityManagerFactory} and if the value does
+     * not begin with {@code java:} and matches the name of a persistence unit,
+     * then the container-managed {@code EntityManagerFactory} for that
+     * persistence unit is used, obtained as if it had been injected using the
+     * {@code PersistenceUnit} annotation. Or, if the value does begin with
+     * {@code java:} and matches the name of a persistence unit reference,
+     * then the corresponding {@code EntityManagerFactory} obtained from JNDI
+     * is used.
```
njr-11, line 118: I remembered why I didn't write this with the PersistenceUnit name in the first place.
I'm not aware of an API or SPI that Jakarta Persistence provides to do this given the name of a persistence unit:
```
     * then the container-managed {@code EntityManagerFactory} for that
     * persistence unit is used, obtained as if it had been injected using the
     * {@code PersistenceUnit} annotation.
```
The closest I see is [Persistence.createEntityManagerFactory(persistenceUnitName)](https://javadoc.io/doc/jakarta.persistence/jakarta.persistence-api/latest/jakarta.persistence/jakarta/persistence/Persistence.html), but that is only for Java SE and also not container managed.  It's also possible that there is something in Jakarta Persistence that I am overlooking.

I would also be fine with including some language for Java SE usage here.

api/src/main/java/jakarta/data/repository/Repository.java
gavinking: But this wording is just incorrect. "Accesses a relational database" is not the same thing as "requires access to a `Datasource`" Those are completely different things.

For example, a reactive implementation based on Vert.x SQL client accesses a relational database but it certainly doesn't use anything in JDBC or JCA.

api/src/main/java/jakarta/data/repository/Repository.java
gavinking: `@PersistenceContext` and `@PersistenceUnit` both inject based on the unit name.

api/src/main/java/jakarta/data/repository/Repository.java
gavinking: Nope. There are two different things here. I checked the platform spec.

There is the `lookup` name, which is in the `java:` namespace. And there is the _reference_ name, the `name` member of `@Resource`, which is typically not. You can see that for sure, right here:

```
    /**
     * The JNDI name of the resource.  For field annotations,
     * the default is the field name.  For method annotations,
     * the default is the JavaBeans property name corresponding
     * to the method.  For class annotations, there is no default
     * and this must be specified.
     */
    String name() default "";
```

A field of method name does not have `java:` in it.

api/src/main/java/jakarta/data/repository/Repository.java
gavinking: I mean, the Javadoc is super-confusing:

```java
    /**
     * (Optional) The name of the persistence unit as defined
     * in the {@code persistence.xml} file. If specified, the
     * persistence unit for the entity manager factory that is
     * accessible in JNDI must have the same name.
     */
```

I don't know what that second sentence means. But I'm basing my assertion on the first sentence:

> The name of the persistence unit as defined in the `persistence.xml` file

api/src/main/java/jakarta/data/repository/Repository.java
njr-11: Resource reference documentation in Jakarta EE is awful and causes lots of confusion.  The `lookup` name is what is used by the container to obtain the reference resource from JNDI -- that is, if `lookup` (or `lookup-name` if in a deployment descriptor) is specified. It's optional. When lookup of a resource reference is done by application code, then the `name` value is used relative to the  `java:comp/env/` context.  So even if `name` is defaulted from a field name, such as `myFieldName`, because there is no `name`, the application uses `java:comp/env/myFieldName` as the JNDI name to look up the resource reference.

api/src/main/java/jakarta/data/repository/Repository.java
njr-11: > But this wording is just incorrect. "Accesses a relational database" is not the same thing as "requires access to a `Datasource`" Those are completely different things.
> 
> For example, a reactive implementation based on Vert.x SQL client accesses a relational database but it certainly doesn't use anything in JDBC or JCA.

Good point - my attempted improvement introduced a different problem.  I revised it with something closer to what you proposed but without the implementation parts.

api/src/main/java/jakarta/data/repository/Repository.java
njr-11: > `@PersistenceContext` and `@PersistenceUnit` both inject based on the unit name.

Right, but we have written the Data spec not to require a compile time step that generates code.  I think we would need an equivalent to Persistence.createEntityManagerFactory(persistenceUnitName) for Jakarta EE usage to ensure it is always implementable.

api/src/main/java/jakarta/data/repository/Repository.java
gavinking: Where are you getting that from? I couldn't find that in the platform spec.

njr-11: It's in section [5.7.1.3](https://jakarta.ee/specifications/platform/11/jakarta-platform-spec-11.0#a1183) where it gives additional information when discussing the `res-ref-name` deployment element that corresponds to the `name` field of `@Resource`,

> The res-ref-name element contains the name of the environment entry used in the application component’s code. The name of the environment entry is relative to the java:comp/env context (for example, the name should be jdbc/EmployeeAppDB rather than java:comp/env/jdbc/EmployeeAppDB )

It also hints at allowing `java:` names that are not relative to `java:comp/env` by explicitly including the full name in an example from section 5.7.1.1,

```java
@Resource(name="java:app/env/customerDB",
          type=javax.sql.DataSource.class)
public class AnApplicationClass {
```

java:app, java:module, and java:global were not originally in Java EE and were added at a later point, and the specification language to account for them is lacking in places, including here, where there ought to be an explicit exception to adding the java:comp/env when one of these others is explicitly stated.  The `name` field of `@Resource` also needs better documentation.

Assuming the above was the intent, an application will look up a resource reference either with a `java:comp/env/...` name or it explicitly specified a java:module, java:app, or java:global name and will use that name to look up the resource reference.

njr-11: This statement that I made earlier has an error in it,

> if `name` is defaulted from a field name, such as `myFieldName`, because there is no `name`, the application uses `java:comp/env/myFieldName` as the JNDI name to look up the resource reference.

Per the Jakarta Annotations spec, the defaulted name also includes the fully qualified class name, so assuming the class is `org.example.MyServlet`, then the defaulted name would be `org.example.MyServlet/myFieldName`. Because it is relative to the `java:comp/env` namespace, the application would then look it up as `java:comp/env/org.example.MyServlet/myFieldName`, not what I had stated earlier.

gavinking: OK, whatever, I'm just saying it seems like other injection annotations let us write stuff like:

```java
@Resource(name="myJMSCF")
ConnectionFactory myJMScf;
```

```java
@EJB(
  name = "ejb/shopping-cart",
  beanName = "cart1",
  beanInterface = ShoppingCart.class,
  description = "The shopping cart for this application"
)
private ShoppingCart myCart;
```

```java
@PersistenceContext(name="persistence/InventoryAppMgr")
```

```java
@Resource(name="myManagedExecutorService")
ManagedExecutorService myManagedExecutorService;
```

And not have to laboriously type out the `java:/comp/env` bit.

But I guess it's fine.

api/src/main/java/jakarta/data/repository/Repository.java
arjantijms: > I think we would need an equivalent to Persistence.createEntityManagerFactory(persistenceUnitName) for Jakarta EE usage to ensure it is always implementable.

+1

Issue #1309, opened 2025-12-24
[TCK Challenge]: Bug issue at the test 539 testReturnStreamOfRecordThatSelectsAttributesFindEntity


otaviojava:
### Specification

---

### Assertion

----

### TCK Version

1.1.0-M1

### Implementation being tested

Eclipse JNoSQL

### Challenge Scenario

Claims that a test assertion conflicts with the specification.

### Full Description

There is an issue with the test identified by method ID **539**, specifically
`testReturnStreamOfRecordThatSelectsAttributesFindEntity`.

The test invokes the following method:

```java
Stream<CardinalNumber> stream = characters.cardinalNumberStream(3L);
```

From the test context, I assume the underlying query to be equivalent to:

```sql
FROM NaturalNumber WHERE floorOfSquareRoot = 3
```

However, the test is currently failing.

When inspecting the generated data manually, the behavior appears to be inconsistent with the test expectations. Once the dataset is generated, **seven elements** are returned, which contradicts what the test seems to assume.

<img width="1085" height="211" alt="Image" src="https://github.com/user-attachments/assets/5c0d06c6-7330-48cd-8c3e-dff28d048763" />

### Is there an existing challenge for this?

- [x] I have searched the existing issues

PR #1310, opened 2025-12-24
Update test id: 539 [TCK]

otaviojava:
fixes: 
https://github.com/jakartaee/data/issues/1309

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
@@ -2610,11 +2610,14 @@ public class EntityTests {
 
         Stream<CardinalNumber> stream = characters.cardinalNumberStream(3L);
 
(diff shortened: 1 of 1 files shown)


njr-11 (approved):
Thanks for catching this! I think this is a second case that was impacted by me not including all of the updates when writing the test. Locally, I have the following, which passes,

```java
    public void testReturnStreamOfRecordThatSelectsAttributesFindEntity() {

        Stream<CardinalNumber> stream = characters.cardinalNumberStream(2L);

        assertEquals(List.of("4 COMPOSITE (3 bits)",
                             "5 PRIME (3 bits)",
                             "6 COMPOSITE (3 bits)",
                             "7 PRIME (3 bits)",
                             "8 COMPOSITE (4 bits)"),
                     stream
                                     .map(CardinalNumber::toString)
                                     .sorted()
                                     .collect(Collectors.toList()));
    }
```

with the difference being `2` rather than `3` supplied to the method.  What you have here also fixes it, by aligning the assertion with `3`.

Issue #1314, opened 2026-01-05
Contribute JDQL to Jakarta Query


njr-11:
Contribute Jakarta Data Query Language (JDQL) to Jakarta Query to become Jakarta Common Query Language (JCQL).

PR #1315, opened 2026-01-05
Replace JDQL with JCQL

njr-11:
First steps of #1314 to contribute Jakarta Data Query Language to Jakarta Query:
- remove the JDQL grammar from Jakarta Data.
- replace references to Jakarta Data Query Language with Jakarta Common Query Language
- replace JDQL acronym with JCQL
- go through the documentation sections on JDQL/JCQL in Jakarta Data and identify whether any statements need to remain because they aren't in Jakarta Query, or whether others can be removed as duplicated.

### DIFF
--- a/api/src/main/java/jakarta/data/page/CursoredPage.java
+++ b/api/src/main/java/jakarta/data/page/CursoredPage.java
@@ -18,6 +18,7 @@
 package jakarta.data.page;
 
 import jakarta.data.repository.OrderBy;
+import jakarta.data.repository.Query;
 import jakarta.data.Order;
 import jakarta.data.Sort;
 
@@ -108,6 +109,12 @@ import java.util.NoSuchElementException;
  * having previously been removed.</p>
  *
  * <h2>Cursor-based Pagination with {@code @Query}</h2>
+ *
+ * <p>The {@link Query} annotation from Jakarta Data and the similar
+ * {@code jakarta.persistence.StaticQuery} annotation from Jakarta Persistence
+ * allow the application to supply a Jakarta Common Query Language (JCQL) or
+ * Jakarta Persistence Query Language (JPQL) query for the repository method
+ * to perform.</p>
  *
  * <p>Cursor-based pagination involves generating and appending additional
  * restrictions involving the key elements to the {@code WHERE} clause of the
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -451,7 +451,7 @@ import java.util.Set;
  * Stream&lt;Person&gt; livingInCity(String address_city);
  * </pre>
  *
- * <h2>Query methods</h2>
+ * <h2>Methods annotated {@code @Query}</h2>
  *
  * <p>The {@link Query} annotation specifies that a method executes a query
  * written in Jakarta Common Query Language (JCQL) or Jakarta Persistence


api/src/main/java/jakarta/data/page/CursoredPage.java
```diff
@@ -111,7 +111,7 @@
  *
  * <p>Cursor-based pagination involves generating and appending additional
  * restrictions involving the key elements to the {@code WHERE} clause of the
- * query. For this to be possible, a user-provided JDQL or JPQL query must end
+ * query. For this to be possible, a user-provided JCQL or JPQL query must end
```
otaviojava, line 121: should we put this "JCQL" instead of the whole name.
I mean, if I jump in the javadoc, that might be confusing to the reader.

api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -30,16 +30,17 @@
...
  * <p>Annotates a repository method as a query method, specifying a query
- * written in Jakarta Data Query Language (JDQL) or in Jakarta Persistence
- * Query Language (JPQL). A Jakarta Data provider is not required to support
+ * written in Jakarta Common Query Language (JCQL) or in Jakarta Persistence
+ * Query Language (JPQL). JCQL and JPQL are defined by the Jakarta Query
+ * specification. A Jakarta Data provider is not required to support
  * the complete JPQL language, which targets relational data stores. However,
  * a given provider might offer features of JPQL which go beyond the subset
- * required by JDQL, or might even offer vendor-specific extensions to JDQL
+ * required by JCQL, or might even offer vendor-specific extensions to JCQL
```
otaviojava, line 38: this one makes more sense because we introduce it first, but I ma not sure in the CursoredPage.

api/src/main/java/module-info.java
```diff
@@ -451,10 +451,11 @@
  * Stream&lt;Person&gt; livingInCity(String address_city);
  * </pre>
  *
- * <h2>JDQL query methods</h2>
+ * <h2>Query methods</h2>
```
otaviojava: Should it be methods with  Query annotation instead?

api/src/main/java/jakarta/data/page/CursoredPage.java
njr-11: Added a few sentences to introduce it above. See https://github.com/jakartaee/data/pull/1315/commits/046634e722d186d5748fc87b7c18fd569ba23571

api/src/main/java/module-info.java
njr-11: I can update it if you think this isn't clear.  See https://github.com/jakartaee/data/pull/1315/commits/046634e722d186d5748fc87b7c18fd569ba23571

api/src/main/java/jakarta/data/repository/Query.java
```diff
@@ -177,7 +178,7 @@
 
     /**
      * <p>Specifies the query executed by the annotated repository method,
-     * in JDQL or JPQL.</p>
+     * in JCQL or JPQL.</p>
```
otaviojava, line 181: @njr-11 shall we use Jakarta Query here?

I mean, on this case it will run any of the languages of the query spec.

njr-11: If Jakarta Query adds additional language levels, we might or might not be able to support them in Jakarta Data. We would want to assess each on a case by case basis.

otaviojava: ahhh ok, good point.

otaviojava (approved):

PR #1318, opened 2026-01-06
Include first test to Jakarta Data Core [Part 1]

otaviojava:
This pull request introduces a set of new entity classes, repositories, and supporting utilities for standalone data-driven tests in the TCK. It also adds AssertJ as a test dependency and provides a custom annotation for parameterized assertions. The changes focus on enabling richer test scenarios with entities like `Fruit` and `Vehicle`, their suppliers, and repositories, as well as improving test metadata and assertion capabilities.

### Entity and Repository Additions

* Added new entity classes: `Fruit`, `Vehicle`, and supporting enum `Transmission`, each with fields, constructors, and standard methods for equality and string representation. [[1]](diffhunk://#diff-4ae1e01551fe28e89526ebcac4b9805f344ff9fa84723f5224a25050a4a7028aR1-R84) [[2]](diffhunk://#diff-88ff078be44d0441de763920d00e824226202d928d8ff738f29280ca3815a564R1-R120) [[3]](diffhunk://#diff-208fd0beaf3152428fdfc72a351867338497a1340367c30a8c4256e00a61145dR1-R20)
* Introduced repository interfaces: `FruitRepository` and `VehicleRepository`, providing query methods for CRUD and filtering operations on the respective entities. [[1]](diffhunk://#diff-3b54bdb1a7c563cbb7b203290749d3d33ca4c58bb2e4b01391b3e5c120c72ac4R1-R57) [[2]](diffhunk://#diff-9291c5f7cb01c7318951b2d4cff53eefdba88148e528e72e856fa475ec057de7R1-R42)
* Added projection record `VehicleSummary` for summarizing vehicle data in queries.

### Data Suppliers for Parameterized Tests

* Created supplier classes `FruitSupplier` and `VehicleSupplier` to provide lists of sample entities for parameterized tests, backed by static lists and utility methods for entity creation. [[1]](diffhunk://#diff-fbfef45928425eb763c4b54a1d23720949c5edbe9abde87f45e1f220fcf0d1f6R1-R47) [[2]](diffhunk://#diff-4bb373d5a451a0d4ff9b96e2cdc479beea41385e0478cf41651ea3b198eefbacR1-R56)
* Added an abstract base class `AbstractSupplier` implementing both `Supplier<List<T>>` and JUnit's `ArgumentsProvider` for seamless integration with parameterized testing.

### Test Framework and Metadata Enhancements

* Added a custom annotation `ParametizedAssertion` for tracking test metadata, such as GitHub issue/PR references and assertion strategies, to improve test documentation and traceability.
* Updated `tck/pom.xml` to include AssertJ as a dependency for fluent assertions in tests, and defined its version in the properties section. [[1]](diffhunk://#diff-c7973ae892437cdc63d039ee91cf6d93bc035ba2f77d5cfaf8c326924b6ffa2dL43-R45) [[2]](diffhunk://#diff-c7973ae892437cdc63d039ee91cf6d93bc035ba2f77d5cfaf8c326924b6ffa2dR57-R61)

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
@@ -104,7 +104,7 @@ public class JakartaQueryTests {
                     .hasSize(fruits.size())
                     .containsExactly(names.toArray(new String[0]));
         } catch (UnsupportedOperationException exp) {
-            if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
                 // Column and Key-Value databases might not be capable of sorting.
             } else {
                 throw exp;


otaviojava:
@njr-11 I've removed the natural numbers, and I will move it to another PR.

> I haven't looked at the other parts other than seeing that there is a Faker class. Hopefully that does not produce random data.

This part of the code uses **data-driven testing**. Faker is only used to generate input parameters; once those parameters are provided, test execution and assertions are fully **deterministic**. This approach is more efficient than fixed examples because it validates the same invariant across a broader input space without introducing non-determinism into the test logic. Any failure is reproducible, as it is directly tied to the concrete input values used in the test.

If this is a concern, I’m happy to change it, but when the focus is on validating behavior across data variations, a data-driven approach is, in my view, the most appropriate and effective choice.

tck/src/main/java/ee/jakarta/tck/data/framework/junit/anno/ParametizedAssertion.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Test metadata to track what assertion from GitHub is being tested, and the
+ * strategy used to test that assertion for data driven tests.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ParametizedAssertion {
+    /**
+     * The GitHub issue/PR number that lead to the test being created
```
njr-11:
```suggestion
     * The GitHub issue/PR number that led to the test being created
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/Fruit.java
```diff
@@ -0,0 +1,84 @@
...
+        this.quantity = quantity;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Fruit fruit = (Fruit) o;
+        return Objects.equals(id, fruit.id);
```
njr-11: I'm not sure if this is used in the tests, but it seems wrong that a new instance with same id that is created to update to a different quantity (or name) would be considered equal to the original that you are trying to change.

```suggestion
        return Objects.equals(id, fruit.id) &&
               Objects.equals(name, fruit.name) &&
               Objects.equals(quantity, fruit.quantity);
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/FruitRepository.java
```diff
@@ -0,0 +1,57 @@
...
+    List<Fruit> findQuantityLt(@Param("quantity") Long quantity);
+
+    @Query("FROM Fruit WHERE quantity <= :quantity")
+    List<Fruit> findQuantityLte(@Param("quantity") Long quantity);
+
+    @Query("FROM Fruit WHERE name IN (:name1, :name2)")
+    List<Fruit> findNameIn(@Param("name1") String name1, @Param("name2")  String name2);
+
+    @Query("FROM Fruit WHERE name = :name AND quantity = :quantity")
+    List<Fruit> findNameEqualsAndQuantitEquals(@Param("name") String name, @Param("quantity") Long quantity);
```
njr-11:
```suggestion
    List<Fruit> findNameEqualsAndQuantityEquals(@Param("name") String name, @Param("quantity") Long quantity);
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,414 @@
...
+public class JakartaQueryTests {
+
+    public static final Logger log = Logger.getLogger(EntityTests.class.getCanonicalName());
+
+    private final DatabaseType type = TestProperty.databaseType.getDatabaseType();
+
+    @Deployment
+    public static WebArchive createDeployment() {
+        return ShrinkWrap.create(WebArchive.class)
+                .addClasses(JakartaQueryTests.class, Vehicle.class, VehicleRepository.class);
```
njr-11: Both repositories and entities are used. Why only add one of them and not the other?
```suggestion
                .addClasses(JakartaQueryTests.class,
                            Vehicle.class, VehicleRepository.class,
                            Fruit.class, FruitRepository.class);
```

tck/src/main/java/ee/jakarta/tck/data/framework/junit/anno/ParametizedAssertion.java
```diff
@@ -0,0 +1,41 @@
...
+import java.lang.annotation.Target;
+
+
+/**
+ * Test metadata to track what assertion from GitHub is being tested, and the
+ * strategy used to test that assertion for data driven tests.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ParametizedAssertion {
```
njr-11: Why is this class any better than `ee.jakarta.tck.data.framework.junit.anno.Assertion` which existing tests are already using?

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,414 @@
...
+        fruitRepository.deleteAll();
+    }
+
+    @DisplayName("should find all entities as stream")
+    @ParameterizedTest
+    @ArgumentsSource(VehicleSupplier.class)
+    @ParametizedAssertion(id = "400",
+            strategy = "Persist a known collection of Vehicle entities and execute a repository query using the " +
+                    "'FROM Vehicle' clause that returns Stream<Vehicle>, asserting that the stream yields all " +
+                    "persisted Vehicle instances.")
```
njr-11: It is more useful to state the specification requirement that is being tested,
```suggestion
            strategy = "A query that is a FROM clause only can retrieve" +
                       " all persisted entities as a Stream.")
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,414 @@
...
+                    .isNotEmpty()
+                    .hasSize(vehicles.size())
+                    .containsAll(vehicles);
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH)) {
+                log.warning("database does not support keyword 'FROM' type: " + type);
+            } else {
+                throw exp;
+            }
+        }
```
njr-11, line 81: Is there somewhere in the spec that allows NoSQL databases to be incapable of `FROM [Entity]` ?  If not, we should not be ignoring failure,
```suggestion
        vehicleRepository.saveAll(vehicles);
        var result = vehicleRepository.findAllQuery();

        Assertions.assertThat(result)
                .isNotEmpty()
                .hasSize(vehicles.size())
                .containsAll(vehicles);
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,414 @@
...
+                    .map(Vehicle::getColor)
+                    .toList();
+
+            Assertions.assertThat(expectedColor)
+                    .isNotEmpty()
+                    .hasSize(vehicles.size())
+                    .containsExactly(colors.toArray(new String[0]));
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH)) {
+                log.warning("database does not support keyword 'FROM' type: " + type);
```
njr-11: This suggestion here remains valid. The only valid allowance in this is for Column and Key-Value that do not support sorting.
```suggestion
            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
                // Column and Key-Value databases might not be capable of sorting.
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,414 @@
...
+                    .sorted(Comparator.reverseOrder())
+                    .toList();
+
+            Assertions.assertThat(expectedColor)
+                    .isNotEmpty()
+                    .hasSize(vehicles.size())
+                    .containsExactly(colors.toArray(new String[0]));
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH)) {
+                log.warning("database does not support keyword 'FROM' type: " + type);
```
njr-11:
```suggestion
            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
                // Column and Key-Value databases might not be capable of sorting.
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,414 @@
...
+                    .isNotEmpty()
+                    .hasSize(vehicles.size())
+                    .containsAll(expected);
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH)) {
+                log.warning("database does not support keyword 'FROM' type: " + type);
+            } else {
+                throw exp;
+            }
+        }
```
njr-11, line 168:
```suggestion
        vehicleRepository.saveAll(vehicles);
        var result = vehicleRepository.findAllWithProjection();

        var expected = vehicles.stream()
                .map(VehicleSummary::of)
                .toList();

        Assertions.assertThat(result)
                .isNotEmpty()
                .hasSize(vehicles.size())
                .containsAll(expected);
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    @ParameterizedTest
+    @DisplayName("should test eq")
+    @ParametizedAssertion(id = "404",
+            strategy = "Persist Fruit entities and execute an equality comparison on the name attribute, asserting that all " +
+                    "returned entities have a name equal to the provided value, or accept UnsupportedOperationException if unsupported.")
```
njr-11:
```suggestion
            strategy = "A query consisting of a FROM clause and a WHERE clause" +
                       " that performs an equality comparison on a single attribute" +
                       " can retrieve a List of all entities having the attribute" +
                       " equal to the given value.")
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,414 @@
...
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .allMatch(fruit -> fruit.getName().equals(sample.getName()));
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH)) {
+                log.warning("database does not support keyword 'FROM' type: " + type);
+            } else {
+                throw exp;
+            }
+        }
```
njr-11, line 189: Existing tests do not allow for NoSQL databases to avoid supporting the equality comparison on a single attribute

```suggestion
        fruitRepository.saveAll(fruits);
        Fruit sample = fruits.getFirst();
        List<Fruit> result = fruitRepository.findNameEquals(sample.getName());

        Assertions.assertThat(result)
                .isNotEmpty()
                .allMatch(fruit -> fruit.getName().equals(sample.getName()));
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    @ParameterizedTest
+    @DisplayName("should test neq")
+    @ArgumentsSource(FruitSupplier.class)
+    @ParametizedAssertion(id = "405",
+            strategy = "Persist Fruit entities and execute a not-equal comparison on the name attribute, asserting that all " +
+                    "returned entities have a different name, or accept UnsupportedOperationException if unsupported.")
```
njr-11:
```suggestion
            strategy = "A query consisting of a FROM clause and a WHERE clause" +
                       " that performs an not-equal comparison on a single attribute" +
                       " can retrieve a List of all entities having a different" +
                       " value for the attribute than the given value.")
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,414 @@
...
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .allMatch(fruit -> !fruit.getName().equals(sample.getName()));
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH)) {
+                log.warning("database does not support keyword 'FROM' type: " + type);
+            } else {
+                throw exp;
+            }
+        }
```
njr-11, line 210: Existing tests do not make allowances for not-equals to be unsupported
```suggestion
        fruitRepository.saveAll(fruits);
        Fruit sample = fruits.getFirst();
        List<Fruit> result = fruitRepository.findNameNotEquals(sample.getName());

        Assertions.assertThat(result)
                .isNotEmpty()
                .allMatch(fruit -> !fruit.getName().equals(sample.getName()));
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/Vehicle.java
```diff
@@ -0,0 +1,120 @@
+/**
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
```
njr-11:
```suggestion
 * Copyright (c) 2026 Contributors to the Eclipse Foundation
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/Vehicle.java
```diff
@@ -0,0 +1,120 @@
...
+                ", transmission=" + transmission +
+                '}';
+    }
+
+    @Override
+    public boolean equals(Object object) {
+        if (!(object instanceof Vehicle vehicle)) {
+            return false;
+        }
+        return Objects.equals(id, vehicle.id);
```
njr-11: Same problem as the other entity class with comparing only the id.
```suggestion
        return Objects.equals(id, vehicle.id) &&
               Objects.equals(make, vehicle.make) &&
               Objects.equals(model, vehicle.model) &&
               Objects.equals(manufacturer, vehicle.manufacturur) &&
               Objects.equals(color, vehicle.color) &&
               Objects.equals(transmission, vehicle.transmission);
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/VehicleSupplier.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public List<Vehicle> get() {
+        return VEHICLES;
+    }
+
+    private static Vehicle vehicle(String id,
+                            String model,
+                            String make,
+                            String manufacturer,
```
njr-11: See the data that you supplied to this method above.  A vehicle's make does not mean the type of vehicle, such as Sedan, Hatchback, and so forth.  Instead, make refers to the brand of the vehicle (what you have for manufacturer in your data) such as Ford, Honda, Toyota, GMC, Cadillac, and so forth.  Keeping the same data is fine, but you should refer to former as the vehicle type and the latter as the make so that you don't cause confusion with everyone reading your tests.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,414 @@
...
+    @ParametizedAssertion(id = "411",
+            strategy = "Persist Fruit entities and execute a query combining two predicates with AND (name equals and quantity equals), " +
+                    "asserting every returned entity satisfies both predicates or accepting UnsupportedOperationException if unsupported.")
+    void shouldAnd(List<Fruit> fruits) {
+
+        try {
+            fruitRepository.saveAll(fruits);
+            Fruit sample = fruits.getFirst();
+
+            List<Fruit> result = fruitRepository.findNameEqualsAndQuantitEquals(sample.getName(), sample.getQuantity());
```
njr-11:
```suggestion
            List<Fruit> result = fruitRepository.findNameEqualsAndQuantityEquals(sample.getName(), sample.getQuantity());
```

njr-11 (changes requested):
I added a bunch of review comments on maybe half of the changes.  The PR tried to do too much at once and I find myself making very similar comments on each test method.  It would have been better to start out with only one or two tests and get those right before adding so many others.

It doesn't seem like the Fruit entity really adds anything that the Vehicle entity doesn't already cover, so I'm not sure why we have both.

None of these tests follow the precedent that we have for addressing eventual consistency.  You would need to add the delay in every time you add entities and remove entities.  And given that the same are continually added and removed, it seems like it would be better to follow the much more efficient model of existing read-only tests that have only the one-time setup cost.

tck/src/main/java/ee/jakarta/tck/data/framework/junit/anno/ParametizedAssertion.java
otaviojava: This one is to work on `ParametizedAnnotation`.

The assertion force `Assertion` to use `Test`

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
otaviojava: I cannot guarantee this on `key-value´

@njr-11

otaviojava:
> t seems like it would be better to follow the much more efficient model of existing read-only tests that have only the one-time setup cost.

Remember that is the part 1 of 5.

There are update and delete queries as the next step PRs.

njr-11:
> > t seems like it would be better to follow the much more efficient model of existing read-only tests that have only the one-time setup cost.
> 
> Remember that is the part 1 of 5.
> 
> There are update and delete queries as the next step PRs.

Maybe you should make one of the entities read-only and be used for all of the read operation tests, and the other entity allow for updates and only be used where update tests are needed.  That is the approach used by existing tests.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
njr-11: > I cannot guarantee this on `key-value´

I did not see anywhere that the spec makes that allowance (which you also made for all of NoSQL, not just Key-Value) and we have an existing TCK test that depends on being able to do it regardless of database type.  Please open a separate issue to get wording into the spec.  The place where it will go is currently in flux due to other modifications I am making there to remove what is duplicated in Jakarta Query.  We need to add these sorts of things to the spec before we can have TCK tests making allowances for them.  In the mean time, I recommend placing a TODO comment with the issue number on this test so that we can update it after the spec is made to allow it.  In the future, new tests that require new allowances ought to be their own PR, not grouped with lots of other new stuff where they will be overlooked.

njr-11:
> But, we eventually need to discuss how to move this code data design to a more modern approach, such as, data driven test

That is something to discuss in a separate issue and get agreement on before trying to implement.  I thought @KyleAure did an excellent job with the approach we currently have for the TCK.  Changes can be made, but I would want to be certain they are improvements and not making tests more fragile and non-deterministic, and that added complexity is justified.

otaviojava:
@njr-11, how about now?

We got 7 files.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+    protected List<Fruit> fruits = FruitPopulator.FRUITS;
+
+    @BeforeEach
+    public void setup() {
+        var populator = new FruitPopulator();
+        populator.populate(fruitRepository);
+    }
+
+    @DisplayName("should find all entities as stream")
+    @Assertion(id = "400",
```
KyleAure:
```suggestion
    @Assertion(id = "1318",
```
The point of having an ID was to relate the tests back to a specific Issue or PR in GitHub so years in the future we can find where tests came from and have context.  Especially helpful if tests move around or file names get changed and we cannot rely on commit history. So I'm not sure where the number 400 came from?

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/FruitPopulator.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+package ee.jakarta.tck.data.standalone.entity;
```
KyleAure:
```suggestion
package ee.jakarta.tck.data.framework.read.only;
```

This is where all the other read-only repositories are located.  It would be nice to keep these all together so that they can be re-used between tests suites.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+import org.junit.jupiter.api.DisplayName;
+import ee.jakarta.tck.data.framework.junit.anno.Assertion;
+
+import java.util.Comparator;
+import java.util.List;
+import java.util.logging.Logger;
+
+@Standalone
+@AnyEntity
+@ReadOnlyTest
```
KyleAure: Similar to previous review comment, you only need to add the `@ReadOnlyTest` annotation if you are using a repository from the `read.only` package.  See: https://github.com/jakartaee/data/blob/main/tck/src/main/java/ee/jakarta/tck/data/framework/arquillian/extensions/TCKArchiveProcessor.java#L67-L71

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+    @Inject
+    protected FruitRepository fruitRepository;
+
+    protected List<Fruit> fruits = FruitPopulator.FRUITS;
+
+    @BeforeEach
+    public void setup() {
+        var populator = new FruitPopulator();
+        populator.populate(fruitRepository);
+    }
```
njr-11, line 60: Only populate once:
```suggestion
    private boolean initialized = false;

    @BeforeEach
    public void setup() {
        if (!initialized) {
            var populator = new FruitPopulator();
            populator.populate(fruitRepository);
            initialized = true;
        }
    }
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+                    .isNotEmpty()
+                    .hasSize(fruits.size())
+                    .containsAll(fruits);
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of sorting.
+            } else {
+                throw exp;
+            }
+        }
```
njr-11, line 81: This repository method doesn't do any sorting,
```suggestion
        var result = fruitRepository.findAllQuery();

        Assertions.assertThat(result)
                .isNotEmpty()
                .hasSize(fruits.size())
                .containsAll(fruits);
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+                    .isNotEmpty()
+                    .hasSize(fruits.size())
+                    .containsAll(expected);
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of sorting.
+            } else {
+                throw exp;
+            }
+        }
```
njr-11, line 168: The repository method doesn't do an sorting,
```suggestion
        var result = fruitRepository.findAllWithProjection();

        var expected = fruits.stream()
                .map(FruitSummary::of)
                .toList();

        Assertions.assertThat(result)
                .isNotEmpty()
                .hasSize(fruits.size())
                .containsAll(expected);
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .allMatch(fruit -> fruit.getName().equals(sample.getName()));
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of sorting.
+            } else {
+                throw exp;
+            }
+        }
```
njr-11, line 189: Again, the repository method here doesn't do any sorting, so there should be no allowance for unsupported sorting
```suggestion
            strategy = "Execute an equality comparison on the name attribute, asserting that all " +
                    "returned entities have a name equal to the provided value.")
    void shouldEq() {
        Fruit sample = fruits.getFirst();
        List<Fruit> result = fruitRepository.findNameEquals(sample.getName());

        Assertions.assertThat(result)
                .isNotEmpty()
                .allMatch(fruit -> fruit.getName().equals(sample.getName()));
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .allMatch(fruit -> !fruit.getName().equals(sample.getName()));
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of sorting.
+            } else {
+                throw exp;
+            }
```
njr-11, line 209:
```suggestion
            List<Fruit> result = fruitRepository.findNameNotEquals(sample.getName());

            Assertions.assertThat(result)
                    .isNotEmpty()
                    .allMatch(fruit -> !fruit.getName().equals(sample.getName()));
        } catch (UnsupportedOperationException exp) {
            // TODO allow unsupported for key-value once language is added to the spec for that in a future PR
            //if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
                // Column and Key-Value databases might not be capable of sorting.
            //} else {
                throw exp;
            //}
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+        try {
+            Fruit sample = fruits.getFirst();
+            List<Fruit> result = fruitRepository.findQuantityGt(sample.getQuantity());
+
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .allMatch(fruit -> fruit.getQuantity() > sample.getQuantity());
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of sorting.
```
njr-11:
```suggestion
            if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
                // Key-Value databases might not be capable of >
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+        try {
+            Fruit sample = fruits.getFirst();
+            List<Fruit> result = fruitRepository.findQuantityGte(sample.getQuantity());
+
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .allMatch(fruit -> fruit.getQuantity() >= sample.getQuantity());
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of sorting.
```
njr-11:
```suggestion
            if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
                // Key-Value databases might not be capable of >=
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+        try {
+            Fruit sample = fruits.getFirst();
+            List<Fruit> result = fruitRepository.findQuantityLt(sample.getQuantity());
+
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .allMatch(fruit -> fruit.getQuantity() < sample.getQuantity());
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of sorting.
```
njr-11:
```suggestion
            if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
                // Key-Value databases might not be capable of <
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+    void shouldLte() {
+        try {
+            Fruit sample = fruits.getFirst();
+            List<Fruit> result = fruitRepository.findQuantityLte(sample.getQuantity());
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .allMatch(fruit -> fruit.getQuantity() <= sample.getQuantity());
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of sorting.
```
njr-11:
```suggestion
            if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
                // Key-Value databases might not be capable of <=
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+                    .allMatch(fruit -> fruit.getName().equals(sample1.getName())
+                            || fruit.getName().equals(sample2.getName()));
+
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of sorting.
+            } else {
+                throw exp;
+            }
+        }
```
njr-11, line 322: This repository method doesn't do any sorting,
```suggestion
            strategy = "Execute an IN comparison on the name attribute with multiple values, " +
                    "asserting membership in the provided set.")
    void shouldIn() {
        var sample1 = fruits.getFirst();
        var sample2 = fruits.get(1);
        List<Fruit> result = fruitRepository.findNameIn(sample1.getName(), sample2.getName());

        Assertions.assertThat(result)
                .isNotEmpty()
                .allMatch(fruit -> fruit.getName().equals(sample1.getName())
                        || fruit.getName().equals(sample2.getName()));
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+
+            List<Fruit> result = fruitRepository.findNameEqualsAndQuantityEquals(sample.getName(), sample.getQuantity());
+
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .allMatch(fruit -> fruit.getName().equals(sample.getName())
+                            && fruit.getQuantity().equals(sample.getQuantity()));
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of sorting.
```
njr-11:
```suggestion
                // Column and Key-Value databases might not be capable of AND.
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,373 @@
...
+            List<Fruit> result = fruitRepository.findNameEqualsORNameEquals(sample1.getName(), sample2.getName());
+
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .allMatch(fruit -> fruit.getName().equals(sample1.getName())
+                            || fruit.getName().equals(sample2.getName()));
+
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of sorting.
```
njr-11:
```suggestion
                // Column and Key-Value databases might not be capable of OR.
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/Transmission.java
```diff
@@ -0,0 +1,20 @@
...
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ * <p>
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+package ee.jakarta.tck.data.standalone.entity;
+
+public enum Transmission {
```
njr-11: Unused by tests - can delete this file.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
KyleAure: Actually, Nathan the Populator API already takes care to only initialize once.  Had to build it this way since we can't use Repositories inside the `@BeforeAll` method since it would run before injection.

njr-11: Nice - that's even better.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
otaviojava: This came from my head, I saw it in my mind, and it should be unique.
Sorry.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/FruitPopulator.java
otaviojava: moved, but we ended up with a circular dependency.

Once: `package ee.jakarta.tck.data.framework.read.only needs the package from the entity.
And the entity's test needs this package.

otaviojava:
@njr-11 I've updated all the comments

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
otaviojava: I've removed this annotation once, I will include operations that will change the data.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,374 @@
...
+import org.junit.jupiter.api.DisplayName;
+import ee.jakarta.tck.data.framework.junit.anno.Assertion;
+import ee.jakarta.tck.data.framework.read.only.FruitPopulator;
+
+import java.util.Comparator;
+import java.util.List;
+import java.util.logging.Logger;
+
+@Standalone
+@AnyEntity
```
KyleAure, line 37:
```suggestion
@ReadOnlyTest
@AnyEntity
```
I think something got mis-interpreted with my previous suggestions.  Having the `@ReadOnlyTest` annotation was a good thing, but the issue was that you didn't actually put the classes in the read only package. 

Add back in the annotation, but move the entity and repository to the read only package.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,374 @@
...
+
+    public static final Logger log = Logger.getLogger(JakartaQueryTests.class.getCanonicalName());
+
+    protected final DatabaseType type = TestProperty.databaseType.getDatabaseType();
+
+    @Deployment
+    public static WebArchive createDeployment() {
+        return ShrinkWrap.create(WebArchive.class)
+                .addClasses(JakartaQueryTests.class,
+                            Fruit.class, FruitRepository.class);
```
KyleAure, line 48:
```suggestion
                .addClasses(JakartaQueryTests.class);
```

If you follow the previous suggestion, then the Fruit entity, FruitRepository, and FruitPopulator will automatically be appended to the depoyment archive.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
otaviojava: But @KyleAure, the next commit will write, those are only the first 10 of 50 tests.

IMHO: it does not make sense to move to ReadOnly, and the next PR move it again.

KyleAure: That was not clear to me based off of how you were using the repository in this PR. (I cannot predict the future :D)

However, the more performant thing to do would be to have one repository (Fruits) that only does read operation testing and keep it read only, then have another repository (Vegetables?) that does read/write operations.

That way on a per-test basis you can ensure that whatever you wrote to the read/write repository is undone before moving onto the next test?

otaviojava: > I cannot predict the future :D


My bad :)

Can I use this Fruit entity for both writing and reading?

Or should I have one entity for reading operations and one for writing operations?

That is not clear to me.

otaviojava: @KyleAure

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,374 @@
...
+                    .map(Fruit::getName)
+                    .toList();
+
+            Assertions.assertThat(expectedColor)
+                    .isNotEmpty()
+                    .hasSize(fruits.size())
+                    .containsExactly(names.toArray(new String[0]));
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
+                // Column and Key-Value databases might not be capable of sorting.
```
njr-11:
```suggestion
            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
                // Column and Key-Value databases might not be capable of sorting.
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,374 @@
...
+        try {
+            var result = fruitRepository.findAllQuery();
+
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .hasSize(fruits.size())
+                    .containsAll(fruits);
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of find all elements.
```
njr-11, line 77: I don't see any language in the spec that makes this allowance

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,374 @@
...
+                    .map(FruitSummary::of)
+                    .toList();
+
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .hasSize(fruits.size())
+                    .containsAll(expected);
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable of find all elements.
```
njr-11, line 164: Again, I haven't been able to find any language in the spec allowing this to be ignored for key-value and column

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
```diff
@@ -0,0 +1,374 @@
...
+        try {
+            Fruit sample = fruits.getFirst();
+            List<Fruit> result = fruitRepository.findNameNotEquals(sample.getName());
+
+            Assertions.assertThat(result)
+                    .isNotEmpty()
+                    .allMatch(fruit -> !fruit.getName().equals(sample.getName()));
+        } catch (UnsupportedOperationException exp) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+                // Column and Key-Value databases might not be capable querying by attribute that is not a key.
```
njr-11, line 206: Comparing an attribute that is not the key is making a lot of these very basic tests not apply to Column and Key-Value.  I recommend we switch the repository method so that its JCQL query compares the key value instead.  Then we won't need to skip so many tests.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
otaviojava: If in both cases we can find by ID, and we are running a query without ID, then we won't be able to execute this query.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
otaviojava: @njr-11, do you prefer to update the query and put in (ids) instead?

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
otaviojava: I can do it, but from the user perspective, why not use the method that we have built in, such as `findById`?

I am fine to update it, but I would search non-ID attributes.

otaviojava: Please, let me know what you think, thus, I can update as you wish.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
njr-11: The point is that the specification doesn't allow for the exemption. You need spec language to cover that.  I just added that for you in https://github.com/jakartaee/data/pull/1319 , after which this one should now be okay.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
njr-11: See the discussion of my previous comment.  I proposed adding spec language in https://github.com/jakartaee/data/pull/1319 that would make the allowance here valid.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
njr-11: > I can do it, but from the user perspective, why not use the method that we have built in, such as `findById`?

Because this is a TCK test of query language, not a user example that limits itself to the easiest way to use the API.  I'm pretty sure we already have TCK tests for `findById` versus the valid query language `WHERE id = :id` which is probably not covered anywhere.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
otaviojava: ahhh ok, I got it.

Thank you.

I approved, and since Gavin approved as well, I already merged.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
otaviojava: @njr-11 in this case, I would go to the path of having both.

Thus, beyond those, include a query with `WHERE id = :id`

What do you think?

I can do this on this PR, or I can create a second one.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
KyleAure: If it was me, I would have one repository that only does read operation with static data that only needs to be written to the database once, and one repository that is used for read and write operations. 

The question of which to use for each test will depend on what you are testing.  If you are just testing some read operation, then use the read-only repository. If you want to test a write operation, then use the read-write repository. 

This just cuts down on expensive read operation (especially when eventual consistency is expected). But also helps to avoid the pitfall of writing tests with persistent data, which is, forgetting to clean up the database after the test is finished.

otaviojava: Let me move those tests; thus, we can refactor.

Thank you for your points Kyle.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaQueryTests.java
otaviojava: @njr-11 ^^

njr-11: Don't worry about it.  It was only a suggestion that would make one additional test valid for Key-value and Column.  It doesn't really matter.

otaviojava: I like the idea, and I will do it.

My question is: should I do it here or in another PR?

Once this PR is over one week, I would love to merge it and create the others.

I would love to include more tests until M2.

otaviojava: @njr-11 I guess we addressed all points to this PR, thus, can we approve to move it forward?

njr-11: I approved it.

PR #1319, opened 2026-01-08
Document WHERE clause restrictions for Key-value and Column databases

njr-11:
@otaviojava is trying to deliver TCK tests that exempt Key-value and Column databases from the requirement to support a query that lacks a `WHERE` clause.  Given that those database types might not be capable of a query that does not have an equality restriction on the key, an allowance for that needs to be written into the spec for it.

### DIFF
--- a/spec/src/main/asciidoc/query-language.asciidoc
+++ b/spec/src/main/asciidoc/query-language.asciidoc
@@ -46,7 +46,7 @@ Due to the diversity of NoSQL database types and their querying capabilities, th
 
 When using NoSQL databases, there are limitations to the support of equality and inequality operators:
 
-Key-Value Databases:: Support for the equality restriction on the key attribute is required. The key attribute is defined by the annotation `jakarta.nosql.Id`. Key-value databases are not required to support any other restrictions.
+Key-Value Databases:: Support for the equality restriction on the unique identifier attribute is required. The unique identifier attribute is defined by the annotation `jakarta.nosql.Id`. Key-value databases are not required to support any other restrictions.
 
 Wide-Column Databases:: Support for equality restriction and the inequality restriction on the `Id` attribute is required. Support for restrictions on other entity attributes is not required. These operations typically work only with the `Id` by default but might be compatible for other entity attributes if secondary indexes are configured in the database schema.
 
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -96,6 +96,14 @@ It is advisable to review your NoSQL provider's documentation to confirm the sup
 
 If the Jakarta Data provider does not support `count(this)`, it must throw `UnsupportedOperationException` when this aggregate expression occurs in a query.  Alternatively, the Jakarta Data provider is permitted to reject a repository method declaration at compilation time if its `@Query` annotation uses the unsupported aggregate expression.
 
+==== WHERE clause
+
+When working with NoSQL databases, the `WHERE` clause behavior may vary depending on the database structure and capabilities:
+
+Key-Value Databases:: Support for a `WHERE` clause that performs an equality restriction on the key attribute is required. Key-value databases might not be capable of queries without a `WHERE` clause or with a `WHERE` clause that does not include an equality restriction on the key attribute.
```
gavinking: I would unverb that phrase.

```suggestion
Key-Value Databases:: Support for a `WHERE` clause with an equality restriction on the key attribute is required. A key-value database might not be capable of processing  a query without a `WHERE` clause or with a `WHERE` clause that does not include an equality restriction on the key attribute.
```

Also, is "key attribute" a notion that is actually well-defined _internally_ to the spec? (I get that it's meaningful in context.)

spec/src/main/asciidoc/query-language.asciidoc
```diff
@@ -96,6 +96,14 @@ It is advisable to review your NoSQL provider's documentation to confirm the sup
...
+
+Wide-Column Databases:: Support for a `WHERE` clause that performs an equality restriction on the key attribute is required. Wide-column databases might not be capable of queries without a `WHERE` clause or with a `WHERE` clause that does not include an equality restriction on the key attribute.
```
gavinking:
```suggestion
Wide-Column Databases:: Support for a `WHERE` clause with an equality restriction on the key attribute is required. A wide-column database might not be capable of processing a query without a `WHERE` clause or with a `WHERE` clause that does not include an equality restriction on the key attribute.
```

gavinking (commented):
I am not competent to say whether this is right or wrong, so my approval would be meaningless, but I proposed a couple of wording changes. :)

spec/src/main/asciidoc/query-language.asciidoc
njr-11: That sounds better that way.  I'll update it.  And good point regarding the "key attribute".  I copied that text, but we should certainly refer to Id instead within Jakarta Data.

njr-11:
Updated to switch "key attribute" to "unique identifier attribute".
Regarding the other wording around key-value and wide-column, @otaviojava please confirm everything is accurate.  This is the PR that will allow your TCK tests to skip testing queries without a WHERE clause for those database types.

PR #1321, opened 2026-01-12
initial TCK tests for Restriction

njr-11:
Some of the most basic tests for Restriction.
These use a repository find method that has only a Restriction parameter.
Tests involve creating a Restriction that applies a single constraint.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
@@ -340,12 +340,13 @@ public class RestrictionTests {
 
     @Assertion(id = "829", strategy = """
             Supply an isNull Restriction to a repository
-            find method.
+            find method, where no entities in the database
+            match and are returned.
             """)
-    public void testIsNullRestriction() {
+    public void testIsNullRestrictionNoneFound() {
         try {
             List<Country> found = countries
-                    .filter(_Country.code.isNull());
+                            .filter(_Country.code.isNull());
 
             assertEquals(0,
                          found.size());
@@ -356,17 +357,24 @@ public class RestrictionTests {
                 throw x;
             }
         }
+    }
 
+    @Assertion(id = "829", strategy = """
+            Supply an isNull Restriction to a repository
+            find method, where some entities in the database
+            match and are returned.
+            """)
+    public void testIsNullRestrictionSomeFound() {
         try {
             List<String> found = countries
-                    .filter(_Country.daylightTimeEnds.isNull())
-                    .stream()
-                    .map(Country::getName)
-                    .collect(Collectors.toList());
-
+                            .filter(_Country.daylightTimeEnds.isNull())
+                            .stream()
+                            .map(Country::getName)
+                            .collect(Collectors.toList());
+        
             assertEquals(150,
                          found.size());
-
+        
             assertEquals(true,
                          found.contains("Burundi"));
             assertEquals(false,
@@ -588,9 +596,10 @@ public class RestrictionTests {
 
     @Assertion(id = "829", strategy = """
             Supply a notNull Restriction to a repository
-            find method.
+            find method, where all entities in the database
+            match and are returned.
             """)
-    public void testNotNullRestriction() {
+    public void testNotNullRestrictionAllFound() {
         try {
             List<String> found = countries
                     .filter(_Country.code.notNull())
@@ -610,6 +619,14 @@ public class RestrictionTests {
                 throw x;
             }
         }
+    }
+
+    @Assertion(id = "829", strategy = """
+            Supply a notNull Restriction to a repository
+            find method, where some entities in the database
+            match and are returned.
+            """)
+    public void testNotNullRestrictionSomeFound() {
 
         try {
             List<String> found = countries


tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
```diff
@@ -0,0 +1,696 @@
...
+                    .stream()
+                    .map(c -> c.getCode() + ": " + c.getName())
+                    .sorted()
+                    .collect(Collectors.toList());
+            assertEquals(List.of("GQ: Equatorial Guinea",
+                                 "IQ: Iraq",
+                                 "QA: Qatar"),
+                         found);
+            } catch (UnsupportedOperationException x) {
+                if (type.isKeywordSupportAtOrBelow(DatabaseType.GRAPH)) {
```
otaviojava, line 128: Just remember of this issue here:

https://github.com/jakartaee/query/issues/8

Not all like will work as SQL.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
```diff
@@ -0,0 +1,696 @@
...
+            assertEquals(List.of("LB: Lebanon",
+                                 "LC: Saint Lucia",
+                                 "LI: Liechtenstein",
+                                 "LK: Sri Lanka",
+                                 "LR: Liberia",
+                                 "LS: Lesotho",
+                                 "LT: Lithuania"),
+                         found);
+            } catch (UnsupportedOperationException x) {
+                if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
```
otaviojava, line 103: I believe code is not the **ID**, thus, column-family might not support it as well.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
```diff
@@ -0,0 +1,696 @@
...
+    }
+
+    @Assertion(id = "829", strategy = """
+            Supply a greaterThanEqual Restriction to a repository
+            find method.
+            """)
+    public void testGreaterThanEqualRestriction() {
+        try {
+            List<String> found = countries
+                    .filter(_Country.code.greaterThanEqual("ZA"))
```
otaviojava, line 205: I guess we got this discussion already about the greater to String might be different to/from database engines.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
```diff
@@ -0,0 +1,696 @@
...
+    }
+
+    @Assertion(id = "829", strategy = """
+                    Supply a greaterThan Restriction to a repository
+                    find method.
+                    """)
+    public void testGreaterThanRestriction() {
+        try {
+            List<String> found = countries
+                    .filter(_Country.code.greaterThan("ZL"))
```
otaviojava, line 230: We cannot guarantee this behavior, because string comparisons depend on database-specific collation rules, index usage, and query planner decisions, which vary between engines and configurations.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
```diff
@@ -0,0 +1,696 @@
...
+                            .stream()
+                            .map(c -> c.getCode() + ": " + c.getName())
+                            .sorted()
+                            .collect(Collectors.toList());
+            assertEquals(List.of("ZM: Zambia",
+                                 "ZW: Zimbabwe"),
+                         found);
+        } catch (UnsupportedOperationException x) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
```
otaviojava, line 239: remember if it is not an id...

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
```diff
@@ -0,0 +1,696 @@
...
+    }
+
+    @Assertion(id = "829", strategy = """
+            Supply a notNull Restriction to a repository
+            find method.
+            """)
+    public void testNotNullRestriction() {
+        try {
+            List<String> found = countries
+                    .filter(_Country.code.notNull())
```
otaviojava, line 605: We cannot guarantee this behavior, because null-handling semantics differ between schema-based and schemaless databases, and even among SQL engines, affecting how notNull() filters are interpreted and optimized.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
```diff
@@ -0,0 +1,696 @@
...
+        } catch (UnsupportedOperationException x) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.KEY_VALUE)) {
+                // Key-Value databases might not be capable of NOT NULL
+            } else {
+                throw x;
+            }
+        }
+
+        try {
+            List<String> found = countries
```
otaviojava, line 632: In this case, I would break it on two test, but, it is up to you.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
njr-11: No, `code` is the `@Id`.  I chose that attribute for these tests to ensure we can include that coverage.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
njr-11: I don't understand your concern here.  There is already a catch block allowing UnsupportedOperationException for all of the NoSQL database types with a comment stating, "NoSQL databases might not be capable of LIKE".  What more are you asking for?

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
njr-11: This test already allows UnsupportedOperationException or Key-Value.  I didn't see any other limitations documented.  If there are others, let me know exactly what they are.  Also note that in this case, all of the `code` values are exactly 2 capital letters.  There is no concern with case-sensitivity or characters outside of english letters `A` through `Z`.  Hopefully nothing would use a different ordering for that.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
njr-11: Again, this is excluded for Key-Value.  If there are other allowances needed, they need to be documented in the spec.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
njr-11: Not applicable.  The `code` attribute  _is_ the Id.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
njr-11: We have an exclusion for Key-Value, consistent with what we currently have written in the spec.  If others need to be excluded, please be specific about which so that I can update the spec language accordingly.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
njr-11: sure, I can split it

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
otaviojava: My point is that in NoSQL, `notNull` can be different.

In NoSQL, **notNull** is not a universally defined semantic concept, because null and absence are first-class and interpreted differently across databases.

Sample, given this insert when I insert as null and do not put this information explicitly:

```json
{ "id": "1", "code": "PT" }
{ "id": "2", "code": null }
{ "id": "3" }
```

## Result

| Database            | Returned documents                     |
|---------------------|----------------------------------------|
| MongoDB             | `{ "code": "PT" }`, `{ }`              |
| Couchbase           | `{ "code": "PT" }`                     |

otaviojava:
> @otaviojava I addressed your review comment about splitting the test that was doing 2 things into separate tests [1a9ff2a](https://github.com/jakartaee/data/commit/1a9ff2a58c661d7ac450a08e22545cb5de5540e6) For your other review comments about making allowances for NoSQL databases, please provide more precise detail that I can use to also update the Data specification with regard to the requirements it makes for different database types.

@njr-11 done, It just left this comment here:

https://github.com/jakartaee/data/pull/1321#discussion_r2685295320

But we can see how to solve it in the following PR.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
otaviojava: There are NoSQL databases that can vary in their collation settings.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
njr-11: For the data used by this test, neither of these are even possible:

> ```json
> { "id": "2", "code": null }
> { "id": "3" }
> ```

The `code` attribute (which happens to also be the `@Id`) is always set to a value that is two capital letters. It is never `null` and never absent. So a `notNull` restriction in this case will always return all entities, regardless of the above interpretation.  This particular test should be okay, although we do need to watch out elsewhere, and we ought to get some language into the spec.

PR #1331, opened 2026-01-16
Initial TCK tests for composite restrictions

njr-11:
TCK tests for simple scenarios with
- Restrict.any
- Restrict.all
- Restrict.unrestricted

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
@@ -81,6 +81,104 @@ public class RestrictionTests {
     private final DatabaseType type = TestProperty.databaseType.getDatabaseType();
 
     @Assertion(id = "829", strategy = """
+            Supply a composite ALL restriction to a repository
+            find method, where the ALL restriction includes only
+            a single restrictions that results must match.
+            """)
+    public void testAllOf1Restriction() {
+        List<Country> found =
+                countries.filter(Restrict.all(_Country.code.equalTo("MN")));
+
+        assertEquals(1, found.size());
+
+        Country country = found.get(0);
+
+        assertEquals("MN",
+                     country.getCode());
+        assertEquals("Mongolia",
+                     country.getName());
+        assertEquals(Region.ASIA,
+                     country.getRegion());
+        assertEquals(1553556L,
+                     country.getArea());
+        assertEquals(3543677L,
+                     country.getPopulation());
+        assertEquals(56474000000L,
+                     country.getGdp());
+        assertEquals(38176424000L,
+                     country.getDebt());
+        assertEquals(null,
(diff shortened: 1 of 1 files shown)


tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
```diff
@@ -710,4 +956,37 @@ public void testStartsWithRestriction() {
...
         }
     }
+
+    @Assertion(id = "829", strategy = """
+            Supply the negated unrestricted Restriction to a
+            repository find method to retrieve no entities.
+            """)
+    public void testUnmatchable() {
+        List<Country> found =
+                countries.filter(Restrict.not(Restrict.unrestricted()));
```
otaviojava, line 966: @njr-11 

It is fine we have this on API, if you agree, I am fine, but.


Is there any **real scenario** where I want to go to the database to return nothing?

IMHO: The API itself should throw an IllegalArgumentException exception in this case.

njr-11: > IMHO: The API itself should throw an IllegalArgumentException exception in this case.

It isn't as simple as having `Restrict.not(Restrict.unrestricted())` raise IllegalArgumentException.  `Restrict.not(Restrict.unrestricted())` is the most concise way for a user to attempt this, but they could also supply their own `CompositeRestriction` instance that has zero `restrictions()` and a `type()` of `ANY`, which would require at least 1 out of the 0 to be satisfied, effectively creating an unmatchable restriction.  And sure, we could write specification text somewhere that requires any repository method that accepts a restriction to detect this and raise `IllegalArgumentException`.  The spec doesn't currently have that.  Is it worth it to add these requirements?  It seems unlikely that very many users would care one way or the other.

> Is there any **real scenario** where I want to go to the database to return nothing?

A few that I can think of are:
- testing whether the connection is still valid is useful for connection pools, but they have better ways of doing that, like JDBC's Connection.isValid method
- checking whether the database tables exist. But again, I doubt this API was intended for that purpose.

That said, the above is really the wrong question to ask.  The `Restrict.unrestricted()` API (which I realize is not what you are challenging) was added to fill the need of a user wanting to invoke a repository method that requires a `Restriction` but they sometimes don't want to provide any, which is definitely a valid and useful scenario.  What we should be considering here is whether there are corresponding cases where a user might have code to which `N` restrictions must be supplied which are treated as `ANY`, but they only want to supply `N-1` of them. In that case, it would be useful to supply an unmatchable restriction along with the other `N-1`.  I have no idea whether any user will actually require that, but that is what I would guess that the unmatchable restriction is for.

otaviojava: Thank you Nathan. 
If it does make sense to SQL, it is fine.

PR #1336, opened 2026-01-21
TCK tests for composite restrictions

njr-11:
TCK tests for:

 - ANY restriction with multiple ALL restrictions under it
 - ALL restriction with multiple ANY restrictions under it
 - negated ANY restriction
 - negated ALL restriction

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
@@ -797,6 +797,99 @@ public class RestrictionTests {
     }
 
     @Assertion(id = "829", strategy = """
+            Supply a negated composite ALL restriction to a repository
+            find method, where the negated ALL restriction combines two
+            restrictions such that results must not satisfy both of the
+            two restrictions.
+            """)
+    public void testNotAllRestrictions() {
+        Restriction<Country> restriction =
+                Restrict.all(_Country.code.greaterThanEqual("B"),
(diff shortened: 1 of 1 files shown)


tck/src/main/java/ee/jakarta/tck/data/standalone/entity/RestrictionTests.java
```diff
@@ -325,6 +366,70 @@ public void testAnyOf4Restrictions() {
...
+    @Assertion(id = "829", strategy = """
+            Supply a composite ANY restriction to a repository
+            find method, where the ANY restriction combines three
+            ALL restrictions that are nested under it, such that
+            results must satisy both of the restrictions under ALL
+            for at least one of the restrictions under ANY.
+            """)
+    public void testAnyWithNestedAllRestrictions() {
+        Restriction<Country> restriction =
+                Restrict.any(Restrict.all(_Country.name.notNull(),
```
otaviojava, line 378: This is the discussion that I had. I guess you are assuming the relational database is the right way to understand SQL.

This might produce different results; thus, should we assume the SQL is correct?

https://github.com/jakartaee/query/issues/108#issuecomment-3763631361

njr-11: > This is the discussion that I had. I guess you are assuming the relational database is the right way to understand SQL.
> 
> This might produce different results; thus, should we assume the SQL is correct?

This will not produce different results.  Every `Country` entity in the dataset has a non-null name, so that particular condition is always satisfied, regardless of whether a relational database or non-relational database is used.

PR #1351, opened 2026-02-03
TCK tests of append, prepend, left, right with value and length, lower, upper

njr-11:
TCK tests of TextExpression methods:

- append(suffixExpression)
- append(suffixLiteral)
- left(numChars)
- length()
- lower()
- right(numChars)
- prepend(prefixExpression)
- prepend(prefixLiteral)
- upper()

I expect there will be some operations that some types of NoSQL databases will not be capable of which we have not yet documented in the spec.  When reviewing this, please point those out and specifically which categories of NoSQL databases need to be exempted, and I'll get the spec updated accordingly.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/ExpressionTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/ExpressionTests.java
@@ -81,9 +81,22 @@ public class ExpressionTests {
             """)
     public void testAppendExpression() {
         List<Country> found;
-        found = countries.filter(_Country.code.left(1)
-                                              .append(_Country.code.left(1))
-                                              .equalTo(_Country.code));
+        try {
+            found = countries.filter(_Country.code
+                                             .left(1)
+                                             .append(_Country.code.left(1))
+                                             .equalTo(_Country.code));
+        } catch (UnsupportedOperationException x) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.DOCUMENT)) {
+                // Document, Column, and Key-Value databases might not be
+                //  capable of ||.
+                // Document, Column, and Key-Value databases might not be
+                //  capable of LEFT.
+                return;
+            } else {
+                throw x;
+            }
+        }
 
         assertEquals(List.of("BB: Barbados",
                              "EE: Estonia",
@@ -104,8 +117,18 @@ public class ExpressionTests {
             """)
     public void testAppendValue() {
         List<Country> found;
-        found = countries.filter(_Country.code.append("STRALIA")
-                                              .equalTo("AUSTRALIA"));
+        try {
+            found = countries.filter(_Country.code.append("STRALIA")
+                                                  .equalTo("AUSTRALIA"));
+        } catch (UnsupportedOperationException x) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.DOCUMENT)) {
+                // Document, Column, and Key-Value databases might not be
+                //  capable of ||.
+                return;
+            } else {
+                throw x;
+            }
+        }
 
         assertEquals(List.of("AU: Australia"),
                      found.stream()
@@ -119,8 +142,18 @@ public class ExpressionTests {
             """)
     public void testLeft1() {
         List<Country> found;
-        found = countries.filter(_Country.code.left(1)
-                                              .equalTo("R"));
+        try {
+            found = countries.filter(_Country.code.left(1)
+                                                  .equalTo("R"));
+        } catch (UnsupportedOperationException x) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.DOCUMENT)) {
+                // Document, Column, and Key-Value databases might not be
+                //  capable of LEFT.
+                return;
+            } else {
+                throw x;
+            }
+        }
 
         assertEquals(List.of("RO: Romania",
                              "RS: Serbia",
@@ -143,7 +176,9 @@ public class ExpressionTests {
             found = countries.filter(_Country.name.left(3)
                                                   .equalTo("Gre"));
         } catch (UnsupportedOperationException x) {
-            if (type.isKeywordSupportAtOrBelow(DatabaseType.COLUMN)) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.DOCUMENT)) {
+                // Document, Column, and Key-Value databases might not be
+                //  capable of LEFT.
                 // Column and Key-Value databases might not be capable of
                 //  restrictions on attributes that are not the Id.
                 return;
@@ -168,8 +203,18 @@ public class ExpressionTests {
             """)
     public void testLength2() {
         List<Country> found;
-        found = countries.filter(_Country.code.length()
-                                              .equalTo(2));
+        try {
+            found = countries.filter(_Country.code.length()
+                                                  .equalTo(2));
+        } catch (UnsupportedOperationException x) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.DOCUMENT)) {
+                // Document, Column, and Key-Value databases might not be
+                //  capable of LENGTH.
+                return;
+            } else {
+                throw x;
+            }
+        }
 
         assertEquals(CountryPopulator.EXPECTED_TOTAL,
                      found.size());
@@ -181,8 +226,18 @@ public class ExpressionTests {
             """)
     public void testLength5() {
         List<Country> found;
-        found = countries.filter(_Country.name.length()
-                                              .equalTo(5));
+        try {
+            found = countries.filter(_Country.name.length()
+                                                  .equalTo(5));
+        } catch (UnsupportedOperationException x) {
+            if (type.isKeywordSupportAtOrBelow(DatabaseType.DOCUMENT)) {
+                // Document, Column, and Key-Value databases might not be
+                //  capable of LENGTH.
+                return;
+            } else {
+                throw x;
+            }
+        }
 
         assertEquals(List.of("AW: Aruba",
                              "BJ: Benin",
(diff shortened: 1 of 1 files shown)


tck/src/main/java/ee/jakarta/tck/data/standalone/entity/ExpressionTests.java
```diff
@@ -0,0 +1,373 @@
...
+            Use the left TextExpression to compare only the characters
+            at the beginning of a String attribute to a value.
+            """)
+    public void testLeft3() {
+        List<Country> found;
+
+        try {
+            found = countries.filter(_Country.name.left(3)
+                                                  .equalTo("Gre"));
+        } catch (UnsupportedOperationException x) {
```
otaviojava, line 178: I would remove those `UnsupportedOperationException` instances and move them to a relational package domain.

I would not assume those functions will work on NoSQL.

otaviojava: But I need more time to check; there are some functions that MongoDB can support, but I cannot assume that all databases follow the same approach.

njr-11: None of this is going to move to relational only.  If any NoSQL database decides to provide the capability (in this case, comparing only the first N characters of the value instead of the whole value) and the TCK runs with that database, then the TCK should test that the behavior follows the spec rather than doing something different.  NoSQL databases don't get a free pass to do whatever they feel like with these methods.  They either need to raise `UnsupportedOperationException` or implement them per spec.

> I would not assume those functions will work on NoSQL.

I'm not assuming anything.  See the PR description.  I'm counting on you to tell me exactly which categories of NoSQL databases will vs might not be capable of each of the functions, and then I'll add those requirements (or non-requirements) into the Data spec and align the TCK tests with it.

njr-11: > But I need more time to check; there are some functions that MongoDB can support, but I cannot assume that all databases follow the same approach.

Sorry for the out-of-sequence prior reply.  That's fine - take your time collecting the information.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/ExpressionTests.java
```diff
@@ -0,0 +1,373 @@
...
+
+    // Inject doesn't happen until after BeforeClass, so this is necessary
+    // before each test
+    @BeforeEach
+    public void beforeEach() {
+        assertNotNull(countries);
+        if (!initialized) {
+            CountryPopulator.get().populate(countries);
+            initialized = true;
+        }
```
KyleAure, line 66:
```suggestion
        CountryPopulator.get().populate(countries);
```
suggestion: the CountryPopulator does it's own test to verify it is run only once, and that includes between test classes. But needs to be called in the `@BeforeTest` because it needs to wait the countries repository to have been injected.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/ExpressionTests.java
```diff
@@ -0,0 +1,373 @@
...
+@Standalone
+public class ExpressionTests {
+
+    public static final Logger log =
+            Logger.getLogger(ExpressionTests.class.getCanonicalName());
+
+    @Inject
+    Countries countries;
+
+    private boolean initialized = false;
```
KyleAure:
```suggestion
```

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/ExpressionTests.java
KyleAure: I might work on making this better aligned with the JUnit lifecycle.

njr-11: This class was copied from ConstraintTests, so we would have the same pattern there, and likely in a few other classes, because I also copied ConstraintTests from another test class.

KyleAure (approved):

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/ExpressionTests.java
otaviojava: > Sorry for the out-of-sequence prior reply.

No problem, it looks like it was sent by two blocks instead of a single one. I hope to have all the information by Sunday. I will include a table to provide context and information so we can make a decision together.

njr-11: @otaviojava any progress on the above?

otaviojava: Yeah, sorry for the delay. 

I just got the internet yesterday.

##  Document Databases

| Function | MongoDB | Couchbase (N1QL) | CouchDB | Firestore | Oracle NoSQL|
|----------|----------|------------------|----------|------------|----------------|
| append(suffixExpression) | ✔⚠ | ✔ | ❌ | ❌ | ✔ |
| append(suffixLiteral) | ✔⚠ | ✔ | ❌ | ❌ | ✔ |
| prepend(prefixExpression) | ✔⚠ | ✔ | ❌ | ❌ | ✔ |
| prepend(prefixLiteral) | ✔⚠ | ✔ | ❌ | ❌ | ✔ |
| left(numChars) | ✔⚠ | ✔ | ❌ | ❌ | ✔ |
| right(numChars) | ✔⚠ | ✔ | ❌ | ❌ | ✔ |
| length() | ✔⚠ | ✔ | ❌ | ❌ | ✔ |
| lower() | ✔⚠ | ✔ | ❌ | ❌ | ✔ |
| upper() | ✔⚠ | ✔ | ❌ | ❌ | ✔ |

✔⚠ MongoDB supports these via the aggregation pipeline (`$concat`, `$substr`, `$toLower`, etc.), not as relational-style algebra in simple query predicates.

Document databases that aim for SQL compatibility (like Couchbase with N1QL) treat these as first-class query functions.  
More minimal document stores, such as CouchDB and Firestore, do not expose server-side string algebra.


## Wide-Column Databases

| Function | Cassandra | ScyllaDB | Bigtable |
|----------|------------|----------|----------|
| append(suffixExpression) | ❌ | ❌ | ❌ |
| append(suffixLiteral) | ❌ | ❌ | ❌ |
| prepend(prefixExpression) | ❌ | ❌ | ❌ |
| prepend(prefixLiteral) | ❌ | ❌ | ❌ |
| left(numChars) | ❌ | ❌ | ❌ |
| right(numChars) | ❌ | ❌ | ❌ |
| length() | ⚠ (UDF only) | ⚠ (UDF only) | ❌ |
| lower() | ❌ | ❌ | ❌ |
| upper() | ❌ | ❌ | ❌ |

Wide-column systems focus on partitioned storage and predictable access patterns. They intentionally avoid complex query-time algebra.


## Graph Databases

| Function | Neo4j (Cypher) | Neptune (openCypher) | Neptune (Gremlin) | JanusGraph (Gremlin) |
|----------|----------------|----------------------|-------------------|----------------------|
| append(suffixExpression) | ✔ | ✔ | ✔⚠ | ⚠ |
| append(suffixLiteral) | ✔ | ✔ | ✔⚠ | ⚠ |
| prepend(prefixExpression) | ✔ | ✔ | ✔⚠ | ⚠ |
| prepend(prefixLiteral) | ✔ | ✔ | ✔⚠ | ⚠ |
| left(numChars) | ✔ | ✔ | ✔⚠ | ⚠ |
| right(numChars) | ✔ | ✔ | ✔⚠ | ⚠ |
| length() | ✔ | ✔ | ✔⚠ | ⚠ |
| lower() | ✔ | ✔ | ✔⚠ | ⚠ |
| upper() | ✔ | ✔ | ✔⚠ | ⚠ |

Cypher provides built-in string functions similar to SQL (e.g., left, right, toLower, toUpper, size). However, Cypher is a pattern-matching pipeline language rather than a relational algebra language. While these functions are supported, the Neo4j documentation advises against applying heavy transformation functions to indexed properties, as they can disable index usage and impact performance.
 
Gremlin (Apache TinkerPop) is a traversal language, not an algebraic query language. Modern versions include native string steps, but these are traversal operations and may not be uniformly optimized across backends. Support exists, but string manipulation is not intrinsic to graph algebra itself.



## Key-Value Databases

| Function | Redis | DynamoDB | Riak |
|----------|--------|----------|------|
| append(suffixExpression) | ✔ | ❌ | ❌ |
| append(suffixLiteral) | ✔ | ❌ | ❌ |
| prepend(prefixExpression) | ❌ | ❌ | ❌ |
| prepend(prefixLiteral) | ❌ | ❌ | ❌ |
| left(numChars) | ❌ | ❌ | ❌ |
| right(numChars) | ❌ | ❌ | ❌ |
| length() | ✔ | ❌ | ❌ |
| lower() | ❌ | ❌ | ❌ |
| upper() | ❌ | ❌ | ❌ |

Key-value stores are designed for direct key lookups. 

## My opinion

While they are commonly available in SQL-based systems and other algebraic query languages, they are not intrinsic to many NoSQL models.

I am not opposed to supporting these capabilities. However, if we continue introducing SQL-style algebraic features into the core API, we risk shifting the abstraction toward a relational model.

If such capabilities are required, it may be more appropriate to place them in a relational or algebra-capable module rather than in the store-agnostic core.

njr-11: Otavio - thanks for providing the information on NoSQL database types!

> if we continue introducing SQL-style algebraic features into the core API, we risk shifting the abstraction toward a relational model.
> 
> If such capabilities are required, it may be more appropriate to place them in a relational or algebra-capable module rather than in the store-agnostic core.

Although the Jakarta Data 1.1 API representing the capability Expressions is new, the underlying capability is not anything new and was already required by JDQL from Jakarta Data 1.0.  Unfortunately, Data 1.0 neglected to indicate which types of NoSQL databases are allowed to not provide the capability.  I added that toleration under this PR (see added commits) - both in the spec and in the test cases.  Hopefully it is acceptable now with the toleration in place.

otaviojava: > Unfortunately, Data 1.0 neglected to indicate which types of NoSQL databases are allowed to not provide the capability.

I did not see it this way, I mean, what we did was the first step, the module creation and specialization go as the next step.

PR #1353, opened 2026-02-05
Prevent mutation of various records

njr-11:
@gavinking pointed out in https://github.com/jakartaee/persistence/pull/944 that records provided by the Jakarta Data API might be vulnerable to mutation.  I looked into this and found 3 places (fixed in this PR) plus one other place that I did not cover in this PR.

The one that is not covered is,
```java
Literal.of(anyValue)
```

A user could supply many different types of mutable objects to that method and later change them, or could invoke `value()` to retrieve its value and then mutate the value, impacting subsequent invocations of `.value()`.  However, it does not seem reasonable to try to identify all possible mutable types and provide code to copy them.

### DIFF
--- a/api/src/test/java/jakarta/data/mock/entity/Book.java
+++ b/api/src/test/java/jakarta/data/mock/entity/Book.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024,2025 Contributors to the Eclipse Foundation
+ * Copyright (c) 2024,2026 Contributors to the Eclipse Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,11 +19,12 @@ package jakarta.data.mock.entity;
 
 import java.time.Instant;
 import java.time.LocalDate;
+import java.util.Objects;
 
 /**
  * A mock entity class for tests
  */
-public class Book {
+public class Book implements Cloneable {
     String author;
     Instant copyrightDate;
     String id;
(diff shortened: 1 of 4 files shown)


api/src/main/java/jakarta/data/page/impl/CursoredPageRecord.java
```diff
@@ -49,6 +51,21 @@ public record CursoredPageRecord<T>
...
+                      long totalElements,
+                      PageRequest pageRequest,
+                      PageRequest nextPageRequest,
+                      PageRequest previousPageRequest) {
+        this.content = List.copyOf(content);
+        this.cursors = List.copyOf(cursors);
+        this.totalElements = totalElements;
+        this.pageRequest = copy(pageRequest);
+        this.nextPageRequest = copy(nextPageRequest);
+        this.previousPageRequest = copy(previousPageRequest);
```
gavinking: Why is it necessary to copy `PageRequest`s?

gavinking: `Pagination` is immutable, no?

njr-11: > `Pagination` is immutable, no?

Good point, if we assume the PageRequest implementation is always the immutable built-in one, then we don't need to copy it.

api/src/main/java/jakarta/data/page/PageRequestCursor.java
```diff
@@ -29,53 +29,52 @@ class PageRequestCursor implements PageRequest.Cursor {
     /**
      * Key values.
      */
-    private final Object[] key;
+    private final List<Object> key;
```
gavinking: Actually @njr-11 I don't think we need to make such a radical change here either. This is a private field, belonging to a package private class. As long as you just `clone()` the array in the constructor, you are fine. No need to change it to `List`. That's just an extra object instantiation.

njr-11: The reason I switched to a `List` was that PageRequestCursor's implementation of the `List<?> elements()` method was already converting from array to List (and doing so each time invoked), so I thought the List conversion would need to be done at least once regardless, making the overall usage less costly than including an extra clone.  However, I went back and checked, and there is an alternative to the `elements()` method, which is to invoke `cursor.get(index)` to retrieve each element individually.  If an implementation is doing that instead of using `cursor.elements()`, then cloning the array would be more efficient as you point out.  I'm not sure which approach implementations are actually going with.

gavinking: Aah ok, I see.

njr-11: I just noticed that the only API method that invokes the PageRequestCursor constructor is Cursor.forKey(Object ...)

```java
        static Cursor forKey(Object... key) {
            return new PageRequestCursor(key);
        }
```

Given that it's a varargs, it will almost always be an array instance created by Java that the application doesn't even have a reference to in order to mutate.  Technically, someone could pass in an `Object[]` to it, in which there is risk of the array being modified.  But I'm tempted to just put this code back how it was originally and do nothing about this.

njr-11: With my prior comment, I convinced myself that the original approach of no copying for Cursor.forKey is best.  Technically, it is just as vulnerable to someone mutating the individual values within the array as it is to someone altering the array itself.  A user doing that would be trying to make themselves fail.  I'd rather just have clarifying text indicating don't do that.

njr-11: I also improved the documentation of this method while I was there.

api/src/main/java/jakarta/data/page/impl/PageRecord.java
```diff
@@ -43,6 +43,17 @@ public record PageRecord<T>(PageRequest pageRequest, List<T> content,
                             long totalElements, boolean moreResults)
         implements Page<T> {
 
+    // Disallow mutation of PageRequest and List fields after creation
+    public PageRecord(PageRequest pageRequest,
```
otaviojava, line 47: Shall we have a test to prove this behavior?

gavinking (approved):

api/src/main/java/jakarta/data/page/impl/PageRecord.java
njr-11: > Shall we have a test to prove this behavior?

sure, I added a test for you,
https://github.com/jakartaee/data/pull/1353/changes/c41b0e2789461ae3b4cba4cbe6098bf8df4b6307

Issue #1280, opened 2025-11-12
Missing a way to statically reference a sort by an embedded or associated property


dstepanov:
I have an embedded entity `specs` and I want to sort by it's property:

```Sort.asc( "specs.weight")```

This works but I cannot reference it using the metamodel because `navigate` always returns an expression:

```NumericExpression<Train, Double> navigate = _Train.specs.navigate(_TrainSpecs.weight);```

Maybe there should be a way to create a path:

```_Train.specs.path(_TrainSpecs.weight).asc();```

gavinking:
> Ahhhh, this is probably on oversight in the new APIs.

Well, actually, not so fast ... in fact in JPQL we generally *don't* allow sorting by anything other than a value which already occurs in the `select` list. (There's actually a good reason for this prohibition, namely that it would interact a bit badly with `distinct` when translated to SQL.)

So the current design is deliberately restrictive. I agree it's *too* restrictive, but I'm not sure what a design with the correct level or restrictiveness would look like.

gavinking:
This was already fixed by #1455.

PR #1400, opened 2026-03-12
Sort for path attributes

njr-11:
The Sort methods for metamodel attributes are missing from the Path attributes for when the NavigableAttribute pattern is used.

The approach originally taken:
This PR looks into adding them by having the navigate(*Attribute) methods now return *Attribute.  To do that, the respective *Path classes are switched to be a *Attribute rather than only a *Expression.  With these changes, you can now get a `.name()` from the path classes like `NumericPath`, and because it now inherits `NumericAttribute` rather than just `NumericExpression`, we also get the default `asc()` and `desc()` methods that are based on the `name()`.  The same is true for the other sortable types, and with `TextPath`, we also inherit the default IgnoreCase variants of the asc/desc methods.

It was later decided that it would be better to allow sorts to be obtained from expressions, after which this PR was switched to add `asc` and `desc` sort to `ComparableExpression` and add `ascIgnoreCase` and `descIgnoreCase` sort to `TextExpression`.

I added a unit test with navigation to each of the path types, demonstrating that Sorts are now successfully obtained.

### DIFF
--- a/api/src/main/java/jakarta/data/Sort.java
+++ b/api/src/main/java/jakarta/data/Sort.java
@@ -86,11 +86,11 @@ import jakarta.data.repository.OrderBy;
  *
  * @param <T>          type of entity from which query results are obtained.
  * @param expression   an expression that computes a value by which to
- *                         order results. Alternatively, {@code null} if
- *                         {@code property} is supplied instead.
+ *                     order results. Alternatively, {@code null} if
+ *                     {@code property} is supplied instead.
  * @param property     name of an entity attribute to order by.
- *                         Alternatively, {@code null} if {@code expression}
- *                         is supplied instead.
+ *                     Alternatively, {@code null} if {@code expression}
+ *                     is supplied instead.
  * @param isAscending  whether ordering for this attribute is ascending
  *                     ({@code true}) or descending ({@code false}).
  * @param ignoreCase   whether or not to request case insensitive ordering
--- a/api/src/main/java/jakarta/data/Sort.java
+++ b/api/src/main/java/jakarta/data/Sort.java
@@ -18,9 +18,12 @@
 package jakarta.data;
 
 import jakarta.data.expression.ComparableExpression;
+import jakarta.data.expression.TextExpression;
 import jakarta.data.messages.Messages;
 import jakarta.data.metamodel.Attribute;
+import jakarta.data.metamodel.ComparableAttribute;
 import jakarta.data.metamodel.StaticMetamodel;
+import jakarta.data.metamodel.TextAttribute;
 import jakarta.data.repository.OrderBy;
 
 /**
@@ -337,6 +340,32 @@ public record Sort<T>(ComparableExpression<T, ? extends Comparable<?>> expressio
     }
 
     /**
+     * Create a {@link Sort} instance with {@link Direction#ASC ascending
+     * direction} that does not request case-insensitive ordering.
+     * <p>
+     * The {@code expression} can be obtained from the
+     * {@linkplain StaticMetamodel static metamodel}, which also allows a
+     * more concise way to sort on an attribute or expression,
+     *
+     * <pre>{@code
+     * _Person.ssn.asc()
+     * }</pre>
+     *
+     * @param <T>        entity class of the expression
+     * @param <V>        type of the expression
+     * @param expression the {@linkplain ComparableAttribute entity attribute}
+     *                   or {@linkplain ComparableExpression expression} by
+     *                   which to order
+     * @return a {@link Sort} instance. Never {@code null}
+     * @throws NullPointerException if the {@code expression} is {@code null}
+     * @since 1.1
+     */
+    public static <T, V extends Comparable<?>> Sort<T> asc(
+            ComparableExpression<T, V> expression) {
+        return new Sort<>(expression, null, true, false, Nulls.UNSPECIFIED);
+    }
+
+    /**
      * Create a {@link Sort} instance with
      * {@linkplain Direction#ASC ascending direction} that does not
      * request case-insensitive ordering.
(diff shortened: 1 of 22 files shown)


gavinking:
Coming back to this now. What exactly is the problem we're trying to solve here?

In the cases I can think of, we usually just want to do something like:

     Book_.publisher.navigate(Publisher_.name).asc()

I think the real problem we have here, which is definitely a problem, is that `asc()` is declared by `SortableAttribute`, which is not inherited by `TextExpression` or `NumericExpression`. We need to fix that.

gavinking:
Well, that's already issue https://github.com/jakartaee/data/issues/1280.

gavinking:
See also #1455.

njr-11:
Updated under https://github.com/jakartaee/data/pull/1400/changes/d1b2d333a486f934a8a1fff1c0fbc70ab9fba755 to correct a mistake that was noticed while reviewing on the Jakarta Data call where ComparableAttribute and TextAttribute's override of the asc/desc methods missed supplying themselves as the expression.

api/src/main/java/jakarta/data/metamodel/TextAttribute.java
```diff
@@ -35,8 +36,9 @@ public interface TextAttribute<T> extends ComparableAttribute<T, String>, TextEx
      * @return a request for an ascending, case-insensitive sort on the entity
      * attribute.
      */
+    @Override
     default Sort<T> ascIgnoreCase() {
-        return Sort.ascIgnoreCase(name());
+        return new Sort<T>(this, null, true, true, Nulls.UNSPECIFIED);
     }
```
gavinking: This implementation is identical to the supertype, and I believe it could be removed. (Unless we change the story with the construction of `Sort`, as suggested below.)

api/src/main/java/jakarta/data/Sort.java
```diff
@@ -88,7 +96,8 @@
  *                     {@link Nulls#FIRST FIRST}, {@link Nulls#LAST LAST}, or
  *                     {@linkplain Nulls#UNSPECIFIED by the data store}.
  */
-public record Sort<T>(String property,
+public record Sort<T>(ComparableExpression<T, ? extends Comparable<?>> expression,
+                      String property,
                       boolean isAscending,
                       boolean ignoreCase,
                       Nulls nullOrdering) {
```
gavinking, line 106: In my PR I decided to hide this constructor, so that `Sort` had better control over how it gets constructed.

So in my PR `Sort`s get constructed with something like `Sort.asc(this)` or `Sort.ascIgnoreCase(this)` instead of exposing the caller to the decision of whether to pass in a null `property`.

I do think that's better.

api/src/main/java/jakarta/data/Sort.java
```diff
@@ -181,18 +196,39 @@ public enum Nulls {
...
+     * sort {@link #expression()}.
+     *
+     * @return The attribute name to order by, or {@code null} if this
+     *         {@code Sort} instance pertains to an {@link #expression()}
      */
     public String property() {
-        return property;
+        return expression instanceof Attribute<?> attr
+                ? attr.name()
+                : property;
```
gavinking: Instead of this, my preference would be to initialize `property = attr.name()` in the constructor.

api/src/main/java/jakarta/data/metamodel/ComparableAttribute.java
```diff
@@ -86,5 +88,27 @@ static <T, V extends Comparable<?>> ComparableAttribute<T, V> of(
...
+    /**
+     * Obtain a request for a descending {@link Sort} based on the entity
+     * attribute.
+     *
+     * @return a request for a descending sort on the entity attribute.
+     */
+    @Override
+    default Sort<T> desc() {
+        return new Sort<T>(this, null, false, false, Nulls.UNSPECIFIED);
+    }
```
gavinking, line 111: Aren't these also identical to the supertype impl?

api/src/main/java/jakarta/data/Sort.java
njr-11: > In my PR I decided to hide this constructor, so that `Sort` had better control over how it gets constructed.

In order to achieve that, you made `Sort` not be a Java record anymore, which I suppose is fine, but I would have rather that been discussed as as separate issue of its own so I had opted not to change that part in my PR.  @otaviojava what do you think about the idea for `Sort` to stop being a Java record and be a normal Java class instead.  That is one of the differences between my PR and Gavin's and the decision on whether or not Sort ought to be a Java record should not be impacting which solution is chosen for having sortable expressions, so I'd like to reach a consensus on that and align the two PRs so we can make progress.

njr-11: > So in my PR `Sort`s get constructed with something like `Sort.asc(this)` or `Sort.ascIgnoreCase(this)` instead of exposing the caller to the decision of whether to pass in a null `property`.

I'm not necessarily opposed to `Sort.asc(expression)`, although the end user doesn't need it because they can already do `expression.asc()`.  If we add it, it seems that it would really be for our internal use within the spec.  That said, I suppose it would be a decent place to put Javadoc for making the user aware that they can obtain Sorts for/from expressions from the metamodel.

api/src/main/java/jakarta/data/metamodel/TextAttribute.java
njr-11: Oops. This was completely redundant. I'll get rid of it. Good catch.

api/src/main/java/jakarta/data/metamodel/ComparableAttribute.java
njr-11: > Aren't these also identical to the supertype impl?

They are identical to one of its supertypes (ComparableExpression) but not to its other supertype (SortableAttribute).  The duplication here is necessary to resolve the ambiguity and choose between the two supertypes.

api/src/main/java/jakarta/data/Sort.java
njr-11: Sure, that's fine.  I added commits to address this and your other review comments (except for switching away from a Java record pending Otavio's agreement).

api/src/main/java/jakarta/data/Sort.java
```diff
@@ -78,8 +84,13 @@
...
  *
- * @param <T>          entity class of the entity attribute upon which to sort.
- * @param property     name of the entity attribute to order by.
+ * @param <T>          type of entity from which query results are obtained.
+ * @param expression   an expression that computes a value by which to
+ *                         order results. Alternatively, {@code null} if
+ *                         {@code property} is supplied instead.
+ * @param property     name of an entity attribute to order by.
+ *                         Alternatively, {@code null} if {@code expression}
+ *                         is supplied instead.
```
gavinking:
```suggestion
 * @param expression   an expression that computes a value by which to
 *                     order results. Alternatively, {@code null} if
 *                     {@code property} is supplied instead.
 * @param property     name of an entity attribute to order by.
 *                     Alternatively, {@code null} if {@code expression}
 *                     is supplied instead.
```

gavinking (approved):
This looks good now.

PR #1406, opened 2026-03-18
Avoid ambiguity in dataStore name requirements

njr-11:
This was discussed on today's Jakarta Data call.  Javadoc for `@Repository(dataStore = ...)` has some ambiguity in the requirements for relational databases that could be interpreted to mean that Jakarta Data providers that access the database via JDBC vs Jakarta Persistence might need to support all of the possible types of names/JNDI names.  This PR splits the single list into 2 separate lists for better clarity.  It also elaborates on what is meant by the JNDI name of an EntityManagerFactory (recently added in Jakarta Persistence) vs a persistence unit reference JNDI name, which was also mentioned during the meeting.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/Repository.java
+++ b/api/src/main/java/jakarta/data/repository/Repository.java
@@ -107,17 +107,6 @@ public @interface Repository {
      * and the entities accessed by the {@code Repository}, then the resource
      * becomes the data store for the {@code Repository}.</p>
      *
-     * <p>Jakarta Data providers that access a relational database via JDBC
-     * must accept the following as the {@code dataStore} value:
-     *  <ul>
-     *   <li>the JNDI name of a {@link javax.sql.DataSource}
-     *   </li>
-     *   <li>the name of a {@code jakarta.annotation.sql.DataSourceDefinition}
-     *   </li>
-     *   <li>the JNDI name of a resource reference to a {@code DataSource}
-     *   </li>
-     *  </ul>
-     * </p>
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/data/repository/Repository.java
```diff
@@ -100,28 +100,39 @@
...
      * {@code dataStore} value has the prefix {@code java:}, the Jakarta Data
      * provider attempts to look up the value as a name in JNDI. If the
      * resource found in JNDI is compatible with the Jakarta Data provider
      * and the entities accessed by the {@code Repository}, then the resource
-     * becomes the data store for the {@code Repository}. For relational
-     * data access, the {@code dataStore} value can be:
+     * becomes the data store for the {@code Repository}.</p>
+     *
+     * <p>Jakarta Data providers that access a relational database via JDBC
+     * must accept the following as the {@code dataStore} value:
```
gavinking: I think I know what you're trying to say here, but since JPA implementations usually use JDBC to access the database, this is still slightly ambiguous.

You could maybe reverse the order of the cases and say "**Other** Jakarta Data providers that access a relational database via JDBC".

njr-11: Good idea.  And we would probably rather have the JPA information show up first regardless.  I added a [commit](https://github.com/jakartaee/data/pull/1406/changes/9ffd5601470c3658a5721a6f79779b7b612a4bca) to switch it.

KyleAure (approved):
Looks good to me.

PR #1409, opened 2026-03-24
Add `@NativeQuery` annotation

yrodiere:
Opening as a draft to gauge the appeal... I mostly focused on javadoc for now and have not looked in details at the spec. I will push an update to address that later.

---

Jakarta Data has `@Query`, Jakarta Persistence has `@StaticQuery`.

Jakarta Persistence also has `@StaticNativeQuery`, but Jakarta Data has... nothing?

This is an attempt at filling the gap, so that applications can rely on Jakarta Data to write (some) native queries, and more importantly (for me) can use an annotation with a simple, intuitive name to do so.

Note this is an entirely optional feature, and the syntax of native queries is implementation-dependent. As such, I do not know how to add TCK tests.

cc @gavinking

### DIFF
--- a/api/src/main/java/jakarta/data/exceptions/NonUniqueResultException.java
+++ b/api/src/main/java/jakarta/data/exceptions/NonUniqueResultException.java
@@ -19,6 +19,7 @@ package jakarta.data.exceptions;
 
 import jakarta.data.repository.Find;
 import jakarta.data.repository.First;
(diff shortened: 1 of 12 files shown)


yrodiere:
Closing, as `@NativeQuery` was added to Jakarta Persistence in https://github.com/jakartaee/persistence/pull/992. Thanks for the comments (and solution on Jakarta Persistence)!

PR #1418, opened 2026-04-03
how to sort nulls

njr-11:
Resolves #1416 which requested a way to indicate whether null values should be sorted first or last.  This is done with methods `.nullsFirst()` and `.nullsLast()` that can be invoked on a `Sort` instance to obtain an otherwise equivalent instance that specifies how to sort nulls.  This pattern aligns with designed `PageRequest` in Jakarta Data.  I have included language that allows NoSQL databases to reject a `Sort` instance that has `nullsFirst` or `nullsLast` if the NoSQL database is not capable of it.

### DIFF
--- a/api/src/main/java/jakarta/data/Sort.java
+++ b/api/src/main/java/jakarta/data/Sort.java
@@ -231,8 +231,13 @@ public record Sort<T>(String property,
     }
 
     /**
-     * Create a {@link Sort} instance, indicating how {@code null} values are
-     * ordered.
+     * <p>Create a {@link Sort} instance, indicating how {@code null} values
+     * are ordered.</p>
+     *
+     * <p>If the data store is a non-relational database that is not capable
+     * of ordering {@code null} values according to the given
+     * {@code nullsFirst} argument, then repository methods to which this
+     * instance is supplied must raise {@link IllegalArgumentException}.</p>
      *
      * @param <T>        entity class of the sortable entity attribute.
      * @param attribute  name of the entity attribute to order by
@@ -315,11 +320,8 @@ public record Sort<T>(String property,
 
     /**
      * <p>Returns an otherwise-equivalent sort that orders {@code null} values
-     * first. If the data store is not capable of ordering {@code null} values
-     * first, then repository methods to which this instance is supplied must
-     * raise {@link UnsupportedOperationException}.</p>
+     * first. For example,</p>
      *
-     * <p>For example,</p>
      * <pre>{@code
      * page1 = customers.livingIn(country,
      *                            PageRequest.ofSize(50),
@@ -327,6 +329,11 @@ public record Sort<T>(String property,
      *                                     _Customer.cityName.asc().nullsFirst(),
      *                                     _Customer.id.asc()));
      * }</pre>
+     *
+     * <p>If the data store is a non-relational database that is not capable
+     * of ordering {@code null} values first, then repository methods to which
+     * this instance is supplied must raise {@link IllegalArgumentException}.
+     * </p>
      *
      * @return a sort with {@link #orderNullsFirst} set to {@code TRUE}.
      */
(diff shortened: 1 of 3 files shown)


api/src/main/java/jakarta/data/Sort.java
```diff
@@ -78,27 +78,46 @@
...
+ *                        ({@code TRUE}), last ({@code FALSE}), or according
+ *                        to the capability and behavior of the underlying
+ *                        data store ({@code null}).
  */
-public record Sort<T>(String property, boolean isAscending,
-                      boolean ignoreCase) {
+public record Sort<T>(String property,
+                      boolean isAscending,
+                      boolean ignoreCase,
+                      Boolean orderNullsFirst) {
```
otaviojava: I guess you want to include this in Jakarta Persistence, right?

otaviojava (dismissed):
Null ordering is fundamentally **non-portable across databases**.

Given that Jakarta Data aims to be domain-centric, exposing null ordering as a first-class abstraction would promote an infrastructure concern into the domain model, which seems misaligned with its goals.

A better approach would be to treat null ordering as a provider-specific capability (e.g., via hints or extensions), rather than standardizing it in the core API.

This allows the specification to remain focused on domain concepts while still providing an escape hatch for datastore-specific behavior.

api/src/main/java/jakarta/data/Sort.java
njr-11: > I guess you want to include this in Jakarta Persistence, right?

Yes, Jakarta Persistence has it. According to an AI search, a few NoSQL databases have it as well, although you will know more the detail on that.

gavinking: Not sure I'm happy with `Boolean` here.

Because `orderNullsFirst = false` isn't completely obviously the same thing as "order nulls last".

I think an enum would be more appropriate, as in JPA.

njr-11: I was wondering about that, too, when I wrote this up, whether it would be better for clarity to use an enum.  I switched to an enum under https://github.com/jakartaee/data/pull/1418/changes/9af32530e93c399051335765f7a6620d5312bd5c and also included `@since 1.1` which I had missed in my previous changes.

otaviojava: I am not a huge fan of increasing this class, what do you think of having components instead?

njr-11: > I am not a huge fan of increasing this class, what do you think of having components instead?

@otaviojava I'm not sure what you mean by having components instead. From the other part of your comment, I am guess that you want the enum moved out of the `Sort` class.  I liked it there because you can write `Sort.Nulls.FIRST` which reads very well, but if you prefer it moved to the same level as `Sort`, I could do that.  I'm not sure if that is what you are asking for though.

otaviojava: I want to understand how big this Sort class can become, given that we want to put some Sort configurations in a single place.

What I would suggest is: instead of having it here, we could consider moving it to another class, such as `SortNullable`, and so on. 

I understand we as a team want this option, but I recommend creating classes or specializations for any new configuration type rather than increasing this one.

What do you think?

njr-11: The decision to represent `Sort` configuration as a Java record means that it cannot be subclassed with a `SortNullable`.  That means if we create a `SortNullable` then it would not be an instance of `Sort`, and so it would not be usable in `Order.by(Sort...)`, and then you need `Order.by(SortNullable...)` even though only one out of several items that the user supplies intends to sort a nullable entity attribute and the others should have been fine as just `Sort` but now need to be `SortNullable` since one of the other items is a `SortNullable`.  It gets worse.  Sorts are obtained from the metamodel and so then the metamodel needs a way to obtain a `SortNullable` for something you don't actually want to be nullable so that you can combine it with things that are nullable in `Order.by(SortNullable...)`.  The end user would need to do something like this:

```java
Order.by(_Product.price.desc().nullsLast(), // this is fine
         _Product.name.asc().asSortNullable(), // confusing and annoying to user
         _Product.id.asc().asSortNullable()); // confusing and annoying to user
```

That is just one example of where the specialization makes the API uglier and more burdensome for users instead of better.

I also want to point out that you are rightly referring to `Sort` as a configuration.  It is perfectly normal for configuration to include some options that not all users will take advantage of.  It is unnatural to subclass or copy a configuration when you introduce a new configuration property.

FroMage (approved):
This looks great, thanks.

api/src/main/java/jakarta/data/Sort.java
otaviojava: Good point, thank.

gavinking (approved):
There are a couple of `@param` javadocs which have periods (inconsistently).

Otherwise it's fine.

PR #1421, opened 2026-04-10
Some clarifications on stateful repositories around persistence context

njr-11:
The language we currently have implies in some places there is only one persistence context associated with a repository:

> A query method of a stateful repository which returns an entity type always returns managed instances belonging to **the persistence context** associated with the repository.

> A lifecycle method annotated `@Remove` or `@Refresh` only accepts managed entities associated with **the persistence context** underlying the repository.

> A stateful repository is backed by a _persistence context_

Methods of a repository can be invoked by multiple threads at the same time. Persistence context is not designed to be accessed by multiple threads or from multiple transactions, as stated in another line from the Data spec

> A persistence context is never shared across transactions.

The first 3 references cited above need to be adjusted, which is done in commit 1.

Commit 2 strengthens requirements and guarantees to the developer around using a stateful repository within a transaction.

Usage of a stateless repository outside of a transaction is another problem.  Considering the following pattern,

```java
person = people.find(id);
person.setStreetAddress(newAddress);
person.setCity(newCity);
person.setState(newState);
```

How does the Jakarta Data provider know that the application is done with the persistence context such that updates ought to be flushed to the database versus whether the application intends to make another update,

```java
person.setZipCode(newZipCode);
```

The only way I know of to indicate this currently is to use a resource accessor method on a JPA-backed Data provider to explicitly flush,

```java
EntityManager em = people.getEntityManager();
em.flush();
em.close(); // ?
```

Resource accessor methods are an advanced pattern. I also don't like the problem the user is confronted with here of whether it's a good idea or terrible idea to close the EntityManager obtained from the stateful repository.  Assuming usage outside of a transaction is intended to be supported -- the Data spec doesn't currently require a transaction -- I think we can make this easier on users by providing a built-in `flush()` operation.  But that will require an interface for `StatefulRepository`.  I haven't included that in this PR because we should probably discuss it first.

### DIFF
--- a/spec/src/main/asciidoc/repository.asciidoc
+++ b/spec/src/main/asciidoc/repository.asciidoc
@@ -999,17 +999,19 @@ Support for stateful repositories is defined in the dedicated module `jakarta.da
 
 ==== Persistence contexts
 
-A stateful repository is backed by a _persistence context_, a set of managed entity instances in which at most one instance represents a given record in the database.
+Each operation on a stateful repository is associated with a _persistence context_, a set of managed entity instances in which at most one instance represents a given record in the database.
 
 - An entity instance never belongs to multiple persistence contexts.
 
 - The Jakarta Data implementation must ensure that a given persistence context never contains more than one entity instance representing the same record.
 
-- Multiple repositories might share a persistence context, especially if they share a datastore.
+- Multiple operations on a single repository, in particular those running in the same transaction or thread, can share a persistence context.
+
+- Operations on multiple repositories might share a persistence context, especially if the repositories share a datastore.
 
 - A persistence context is never shared across transactions.
 
-A query method of a stateful repository which returns an entity type always returns managed instances belonging to the persistence context associated with the repository.
+A query method of a stateful repository which returns an entity type always returns managed instances belonging to a persistence context associated with the repository.
 
 [WARNING]
 ====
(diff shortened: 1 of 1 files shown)


gavinking:
I agree that this could be clarified by saying that the association between persistence context and repository might be scoped to a thread, or to a transaction, or whatever, but I'm not keen on these specific changes.

gavinking:
All you need to really say is that the association is contextual.

njr-11:
> All you need to really say is that the association is contextual.

Alright, I reduced the scope of this to only cover the first item, to at least avoid the misinterpretation and restate things along those lines.

We can discuss the other other items separately from this PR.

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -999,7 +999,7 @@ Support for stateful repositories is defined in the dedicated module `jakarta.da
 
 ==== Persistence contexts
 
-A stateful repository is backed by a _persistence context_, a set of managed entity instances in which at most one instance represents a given record in the database.
+A stateful repository operation applies to a _persistence context_, a set of managed entity instances in which at most one instance represents a given record in the database. Persistence context can span multiple repository operations, such as all operations performed within a transaction.
```
```suggestion
A stateful repository operation is always performed in a _persistence context_, operating on a set of managed entity instances in which at most one instance represents a given record in the database.
A given persistence context might span multiple repository operations. 
For example, a persistence context often spans the operations performed within a given transaction.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -1009,7 +1009,7 @@ A stateful repository is backed by a _persistence context_, a set of managed ent
 
 - A persistence context is never shared across transactions.
 
-A query method of a stateful repository which returns an entity type always returns managed instances belonging to the persistence context associated with the repository.
+A query method of a stateful repository which returns an entity type always returns managed instances belonging to the persistence context in which the repository operation is running.
```
```suggestion
A query method of a stateful repository which returns an entity type always returns managed instances belonging to the persistence context in which the repository operation is performed.
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -1031,7 +1031,7 @@ A <<Lifecycle methods, lifecycle method>> of a stateful repository must be annot
 The annotations `@Save`, `@Insert`, `@Update`, `@Delete` are used to define stateless repositories and must not be used to declare lifecycle methods of a stateful repository.
 Instead, this specification defines the special lifecycle annotations `@Persist`, `@Merge`, `@Refresh`, `@Remove`, and `@Detach` for declaring lifecycle methods of stateful repositories.
 
-A lifecycle method annotated `@Remove` or `@Refresh` only accepts managed entities associated with the persistence context underlying the repository.
+A lifecycle method annotated `@Remove` or `@Refresh` only accepts managed entities associated with the persistence context under which the repository method is running.
```
```suggestion
A lifecycle method annotated `@Remove` or `@Refresh` only accepts managed entities associated with the persistence context in which the repository method is performed.
```

njr-11:
Sorry I missed this one last week.  All of your suggestions look good and have been applied now.

otaviojava (approved):

PR #1424, opened 2026-04-14
StatefulRepository with flush method?

njr-11:
For discussion at the Jakarta Data meeting:
Do we need something like this for applications that use a stateful repository to modify a persistence context outside of a transaction?

```java
Person person = people.find(ssn);
person.setHomeAddress(newAddress);
person.setZipCode(newZipCode);
```

Without a flush operation, how can the application know that its changes will be persisted vs lost?
The same applies to our TCK tests.  In order to make assertions about what went in to the database, we need guarantees that operations we perform will be persisted after a certain point.

### DIFF
--- a/stateful/src/main/java/jakarta/data/repository/stateful/StatefulRepository.java
+++ b/stateful/src/main/java/jakarta/data/repository/stateful/StatefulRepository.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


gavinking:
> For discussion at the Jakarta Data meeting:

Which meeting? Isn't the meeting later today?

> Do we need something like this for applications that use a stateful repository to modify a persistence context outside of a transaction? Without a flush operation, how can the application know that its changes will be persisted vs lost?

I don't know what you're trying to say here, but I'm pretty sure the answer is "no".

gavinking:
>  Isn't the meeting later today?

Oh, you wrote "for", not "from". Sorry, misread.

njr-11:
Canceling this PR after agreement on today's Jakarta Data call that stateless repositories are not intended to make updates or have lifecycle methods run outside of a transaction.  PR #1429 adds documentation of that restriction.

PR #1425, opened 2026-04-15
Update spec.version to 1.1 in pom.xml

otaviojava:
This pull request updates the Maven build configuration in `spec/pom.xml` to improve version management by dynamically setting the specification version and introducing a new plugin for parsing version information.

**Build and versioning improvements:**

* The `spec.version` property is now set dynamically using the parsed major and minor version numbers from the project version, instead of being hardcoded.
* Added the `build-helper-maven-plugin` (version 3.6.1) with the `parse-version` goal to the build plugins, enabling automatic extraction of version components for use in properties.

### DIFF
--- a/spec/pom.xml
+++ b/spec/pom.xml
@@ -42,7 +42,7 @@
 
         <!-- default is the same for backward compatibility reason
         Easy to override when building with a system property -->
-        <spec.version>1.0</spec.version>
+        <spec.version>1.1</spec.version>
     </properties>
 
     <build>
--- a/spec/pom.xml
+++ b/spec/pom.xml
@@ -134,6 +134,19 @@
                     </attributes>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <version>3.6.1</version>
+                <executions>
+                    <execution>
+                        <id>parse-version</id>
+                        <goals>
+                            <goal>parse-version</goal>
+                        </goals>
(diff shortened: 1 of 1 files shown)


spec/pom.xml
```diff
@@ -42,7 +42,7 @@
 
         <!-- default is the same for backward compatibility reason
         Easy to override when building with a system property -->
-        <spec.version>1.0</spec.version>
+        <spec.version>1.1</spec.version>
```
pzygielo: Could it be derived from `project.version` to not need to be updated manually?

otaviojava: Can we use the semantic version, e.g.: 1.1.0, instead of the one required to Jakarta EE spec, e.g.: 1.1?

The last time we did that, we had to roll back.

otaviojava: @ivargrimstad

pzygielo: I've not asked to use `project.version` but to extract `major.minor` from it.

ivargrimstad: I usually use the spec version separately from the project version to be able to build snapshots and miletones. But it is probably because of how the spec version plugin is set up. How you do it is totally up to you. The result should be the same with major.minor in the spec documents

otaviojava: I got it @pzygielo how about now?

Thank you @ivargrimstad

pzygielo: >  how about now?

For me - perfect :fireworks:

otaviojava:
@njr-11 @KyleAure could you review this one?

njr-11 (approved):

PR #1427, opened 2026-04-15
Address review comment to use consistent punctuation in param javadoc

njr-11:
Fixes this review comment from #1418 about inconsistent punctuation of `@param` javadoc in the `Sort` class,

https://github.com/jakartaee/data/pull/1418#pullrequestreview-4114034349

The majority of `@param` javadoc in this class ends with `.`, so I updated the places where it was lacking.

### DIFF
--- a/api/src/main/java/jakarta/data/Sort.java
+++ b/api/src/main/java/jakarta/data/Sort.java
@@ -248,11 +248,11 @@ public record Sort<T>(String property,
      * Create a {@link Sort} instance.
      *
      * @param <T>        entity class of the sortable entity attribute.
-     * @param attribute  name of the entity attribute to order by
+     * @param attribute  name of the entity attribute to order by.
      * @param direction  the direction in which to order.
      * @param ignoreCase whether to request a case insensitive ordering.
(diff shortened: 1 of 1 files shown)


pzygielo:
> The majority of `@param` javadoc in this class ends with `.`, so I updated the places where it was lacking.

But https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html says:

> End the phrase with a period only if another phrase or sentence follows it.

njr-11:
> > The majority of `@param` javadoc in this class ends with `.`, so I updated the places where it was lacking.
> 
> But https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html says:
> 
> > End the phrase with a period only if another phrase or sentence follows it.

That's unfortunate. It means nearly all of the Javadoc I've contributed to specifications is out of compliance with Oracle's style guide.  It will be a disruptive change to correct all of it - probably best to wait until a point when there are not other PRs pending to change that sort of thing due to the large number of lines it will impact and the merge conflicts that would cause.

gavinking (approved):

PR #1429, opened 2026-04-15
Document the requirement for a transaction for stateful repositories in Javadoc

njr-11:
I decided to start with the Javadoc for explicitly stating the requirement to use stateful repositories within a transaction (except when using find to read and not update entities) because that is primarly what applications which need to know this information will be looking at.

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -310,9 +310,15 @@ import java.util.Set;
  * <h2>Lifecycle methods</h2>
  *
  * <p>A lifecycle method makes changes to persistent data in the data store.
- * A lifecycle method must be annotated with a lifecycle annotation such as
- * {@link Insert}, {@link Update}, {@link Save}, or {@link Delete}. The
- * method must accept a single parameter, whose type is either:</p>
+ * The lifecycle methods of a stateless repository must be annotated with a
+ * stateless lifecycle annotation such as {@link Insert}, {@link Update},
+ * {@link Save}, or {@link Delete}. The lifecycle methods of a
+ * <a href="../jakarta.data.stateful/">stateful repository</a>, which apply
+ * to a persistence context, must be annotated with one of the stateful
+ * lifecycle annotations: {@code Detach}, {@code Merge}, {@code Persist}.
+ * {@code Refresh}, or {@code Remove}. For both stateful as well as stateless
+ * lifecycle methods, the lifecycle method must accept a single parameter,
+ * whose type is either:</p>
  *
  * <ul>
  * <li>the class of the entity, or</li>
@@ -321,11 +327,15 @@ import java.util.Set;
  * </ul>
  *
  * <p>The annotated method must be declared {@code void}, or, except in the
- * case of {@code @Delete}, have a return type that is the same as the type
+ * case of {@code @Delete} and the stateful lifecycle annotations other than
+ * {@code @Merge}, have a return type that is the same as the type
  * of its parameter.</p>
  *
+ * <p>A repository without lifecycle methods is considered to be a
+ * stateless repository.</p>
+ *
  * <table style="width: 100%">
- * <caption><b>Lifecycle Annotations</b></caption>
+ * <caption><b>Stateless Lifecycle Annotations</b></caption>
  * <tr style="background-color:#ccc">
  * <td style="vertical-align: top; width: 10%"><b>Annotation</b></td>
  * <td style="vertical-align: top; width: 25%"><b>Description</b></td>
@@ -334,7 +344,7 @@ import java.util.Set;
  *
  * <tr style="vertical-align: top; background-color:#eee"><td>{@link Delete}</td>
  * <td>deletes entities</td>
- * <td>{@code @Delete}<br>{@code public void remove(person);}</td></tr>
+ * <td>{@code @Delete}<br>{@code public void delete(person);}</td></tr>
  *
  * <tr style="vertical-align: top"><td>{@link Insert}</td>
  * <td>creates new entities</td>
@@ -349,8 +359,13 @@ import java.util.Set;
  * <td>{@code @Update}<br>{@code public boolean modify(Product modifiedProduct);}</td></tr>
  * </table>
  *
- * <p>Refer to the API documentation for {@link Insert}, {@link Update}, {@link Delete},
- * and {@link Save} for further information about these annotations.</p>
+ * <p>Refer to the API documentation for {@link Insert}, {@link Update},
+ * {@link Delete}, and {@link Save} for further information about the
+ * stateless lifecycle annotations.</p>
+ *
+ * <p>Refer to the API documentation of the Jakarta Data
+ * <a href="../jakarta.data.stateful/">Stateful Repositories module</a>
+ * for additional information about the stateful lifecycle annotations.</p>
  *
  * <h2>Parameter-based {@code Find} and {@code Delete} methods</h2>
  *
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -310,15 +310,13 @@ import java.util.Set;
  * <h2>Lifecycle methods</h2>
  *
  * <p>A lifecycle method makes changes to persistent data in the data store.
- * The lifecycle methods of a stateless repository must be annotated with a
- * stateless lifecycle annotation such as {@link Insert}, {@link Update},
- * {@link Save}, or {@link Delete}. The lifecycle methods of a
- * <a href="../jakarta.data.stateful/">stateful repository</a>, which apply
- * to a persistence context, must be annotated with one of the stateful
- * lifecycle annotations: {@code Detach}, {@code Merge}, {@code Persist}.
- * {@code Refresh}, or {@code Remove}. For both stateful as well as stateless
- * lifecycle methods, the lifecycle method must accept a single parameter,
- * whose type is either:</p>
+ * Each lifecycle method of a stateless repository must be annotated with
+ * a stateless lifecycle annotation, such as {@link Insert}, {@link Update},
+ * {@link Save}, or {@link Delete}. Each lifecycle method of a
+ * <a href="../jakarta.data.stateful/">stateful repository</a>
+ * must be annotated with a stateful lifecycle annotation.
+ * A lifecycle method, whether stateful or stateless,
+ * must accept a single parameter, whose type is either:</p>
  *
  * <ul>
  * <li>the class of the entity, or</li>
@@ -326,10 +324,9 @@ import java.util.Set;
  *     entities.</li>
  * </ul>
  *
- * <p>The annotated method must be declared {@code void}, or, except in the
- * case of {@code @Delete} and the stateful lifecycle annotations other than
- * {@code @Merge}, have a return type that is the same as the type
- * of its parameter.</p>
+ * <p>The annotated method must be declared {@code void}, or, in cases where
+ * allowed by the lifecycle annotation, have a return type that is the same
+ * as the type of its parameter.</p>
  *
  * <p>A repository without lifecycle methods is considered to be a
  * stateless repository.</p>
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -328,8 +328,8 @@ import java.util.Set;
  * allowed by the lifecycle annotation, have a return type that is the same
  * as the type of its parameter.</p>
  *
- * <p>A repository without lifecycle methods is considered to be a
- * stateless repository.</p>
+ * <p>A repository with neither lifecycle methods nor a resource accessor method is 
+ * considered to be a stateless repository.</p>
  *
  * <table style="width: 100%">
  * <caption><b>Stateless Lifecycle Annotations</b></caption>
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -316,7 +316,7 @@ import java.util.Set;
  * <a href="../jakarta.data.stateful/">stateful repository</a>
  * must be annotated with a stateful lifecycle annotation.
  * A lifecycle method, whether stateful or stateless,
- * must accept a single parameter, whose type is either:</p>
+ * must declare a single parameter, whose type is either:</p>
  *
  * <ul>
  * <li>the class of the entity, or</li>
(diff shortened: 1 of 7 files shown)


api/src/main/java/module-info.java
```diff
@@ -310,9 +310,15 @@
...
- * method must accept a single parameter, whose type is either:</p>
+ * The lifecycle methods of a stateless repository must be annotated with a
+ * stateless lifecycle annotation such as {@link Insert}, {@link Update},
+ * {@link Save}, or {@link Delete}. The lifecycle methods of a
+ * <a href="../jakarta.data.stateful/">stateful repository</a>, which apply
+ * to a persistence context, must be annotated with one of the stateful
+ * lifecycle annotations: {@code Detach}, {@code Merge}, {@code Persist}.
+ * {@code Refresh}, or {@code Remove}. For both stateful as well as stateless
+ * lifecycle methods, the lifecycle method must accept a single parameter,
+ * whose type is either:</p>
```
gavinking: Pluralization does not agree here.

> lifecycle methods ... a lifecycle annotation

Please revert back to the singular form which makes clear that each (singular) lifecycle method has one (singular) lifecycle annotation.

gavinking: I also dislike this change and think it should simply be rolled back. It was perfectly fine as it was before. This new language makes it look like you've enumerated all possible lifecycle annotations, rather than them being an open-ended set.

api/src/main/java/module-info.java
```diff
@@ -321,11 +327,15 @@
  * </ul>
  *
  * <p>The annotated method must be declared {@code void}, or, except in the
- * case of {@code @Delete}, have a return type that is the same as the type
+ * case of {@code @Delete} and the stateful lifecycle annotations other than
+ * {@code @Merge}, have a return type that is the same as the type
```
gavinking: Again, this is getting worse. Lifecycle annotation are an open-ended set. We should not be explicitly enumerating them like this.

stateful/src/main/java/jakarta/data/repository/stateful/Merge.java
```diff
@@ -36,6 +36,9 @@
  * database. This typically occurs later, when the persistence context is
  * flushed.
  * </p>
+ * <p>A transaction must be active on the thread from which a repository
+ * method annotated {@code @Merge} is invoked.
+ * </p>
```
gavinking: This is not correct. JDBC transactions are not associated with threads, and so when a repository is backed by a non-JTA datasource, it's the persistence context which determines the transaction. I mean, sure, you can argue that, well, usually there's some affinity between the thread and the persistence context, but even that need not be true.

This is what I was trying to tell you on the call. The thing you want to say here is a very thing to formulate in the abstract.

stateful/src/main/java/module-info.java
```diff
@@ -16,8 +16,128 @@
...
+ * </p>
+ *
+ * <h2>Transactions</h2>
+ *
+ * <p>When using a stateful repository, all lifecycle methods and all updates
+ * performed on entities obtained from query methods must run within a
+ * transaction. A persistence context remains active for the duration of the
+ * transaction, after which all updates must be committed or rolled back in the
+ * data store, per the outcome of the transaction. In Jakarta EE environments,
+ * a Jakarta Transactions (JTA) transaction is used.</p>
```
gavinking: No, wrong. It's just not the case that every persistence context is a JTA transaction-scoped persistence context. This is exactly the problem I was trying to explain yesterday.

gavinking (changes requested):
Yesterday on the call I tried to explain the reasons this is difficult to specify. You've fallen right into the trap I mentioned.

gavinking:
It's important to realize that, in full generality:

1. transactions need not be associated with threads (resource local transactions)
2. persistence contexts need not be associated with transactions
3. a persistence context might span multiple transactions

stateful/src/main/java/jakarta/data/repository/stateful/Merge.java
njr-11: > This is not correct. JDBC transactions are not associated with threads

JDBC transactions are associated with a java.sql.Connection object.  Although it is true the JDBC specification doesn't explicitly forbid JDBC drivers from supporting concurrent access to a java.sql.Connection, it is certainly not a requirement of the JDBC specification, nor is it a practice I have ever seen offered or encouraged by a JDBC vendor.  That part is not a concern.

The problem with JDBC, that I just thought of, is that it has no explicit begin operation for a transaction, lazily starting (with autocommit=false) when a statement needs to be sent to the database. That won't always have happened before `merge(entity)` is called.  So even though the transaction is tied to the thread for any normal usage, it won't yet be active in some cases.  So we can't use the wording I came up with and we need something else here.

gavinking: What I'm saying is that the whole concept of a "transaction being active on [a] thread" is specific to JTA transactions. It doesn't even make sense to speak in those terms for non-JTA datasources, for reactive repositories, and so on.

njr-11:
>     2. persistence contexts need not be associated with transactions

True, but lacking a transaction, there is no boundary that Jakarta Data offers after which the application can expect their changes to be persisted to the database vs rolled back or left in pending state.  So I though we decided on the call that Jakarta Data would require a transaction.

>     3. a persistence context might span multiple transactions

We already wrote into the Jakarta Data spec a requirement that, "A persistence context is never shared across transactions."  Maybe that was intended to say that it is never shared across transactions that are active at the same time and make an allowance to span transactions serially?  But we currently don't have that allowance in there.

gavinking:
> True, but lacking a transaction, there is no boundary that Jakarta Data offers after which the application can expect their changes to be persisted to the database vs rolled back or left in pending state. 

Saying that there is a transaction is not the same thing as saying that there is a direct affinity between the persistence content and the transaction.

> So I though we decided on the call that Jakarta Data would require a transaction.

For _write_ operations, sure. We were talking about write operations. 

But that tx might be controlled by, for example, the application directly starting/ending a resource-local transaction via `EntityTransaction` or a JDBC transaction via a `Connection` obtained from a non-JTA `DataSource`, or by some other means we're entirely ignorant of.

> Maybe that was intended to say that it is never shared across transactions that are active at the same time and make an allowance to span transactions serially?

Yes.

api/src/main/java/module-info.java
njr-11: I removed the list of stateful methods and switched back to singular.
The intended purpose of my changes in this section was:
* indicate that we also have stateful lifecycle annotations, and link to the stateful module where they can be found
* require that stateful and stateless not be intermixed on a repository.  The singular wording "A lifecycle method must..." did not achieve this. That was my reason for switching to plural, but thanks for pointing out it causes other confusion. I like your suggestion of "each" for singular, which solves that without adding confusion.

api/src/main/java/module-info.java
njr-11: I didn't like what I added here either, but with the introduction of the stateless lifecycle annotations, the old statement here is no longer true and cannot be left as is.  In my followup commit, I have switched the statement to be more generic and defer to the lifecycle method instead.

stateful/src/main/java/jakarta/data/repository/stateful/Merge.java
njr-11: I made an attempt at improving the language to allow for these other patterns.

stateful/src/main/java/module-info.java
njr-11: I rewrote this section in the same way I attempted to correct the Merge/Remove/Persist documentation.

api/src/main/java/module-info.java
```diff
@@ -310,22 +310,29 @@
...
- * A lifecycle method must be annotated with a lifecycle annotation such as
- * {@link Insert}, {@link Update}, {@link Save}, or {@link Delete}. The
- * method must accept a single parameter, whose type is either:</p>
+ * Each lifecycle method of a stateless repository must be annotated with
+ * a stateless lifecycle annotation, such as {@link Insert}, {@link Update},
+ * {@link Save}, or {@link Delete}. Each lifecycle method of a
+ * <a href="../jakarta.data.stateful/">stateful repository</a>
+ * must be annotated with a stateful lifecycle annotation.
+ * A lifecycle method, whether stateful or stateless,
+ * must accept a single parameter, whose type is either:</p>
```
gavinking:
```suggestion
 * must declare a single parameter, whose type is either:</p>
```

A method "declares" parameters; it "accepts" arguments.

api/src/main/java/module-info.java
```diff
@@ -310,22 +310,29 @@
...
  *
- * <p>The annotated method must be declared {@code void}, or, except in the
- * case of {@code @Delete}, have a return type that is the same as the type
- * of its parameter.</p>
+ * <p>The annotated method must be declared {@code void}, or, in cases where
+ * allowed by the lifecycle annotation, have a return type that is the same
+ * as the type of its parameter.</p>
+ *
+ * <p>A repository without lifecycle methods is considered to be a
+ * stateless repository.</p>
```
gavinking:
```suggestion
 * <p>A repository with neither lifecycle methods nor a resource accessor method is 
 * considered to be a stateless repository.</p>
```

spec/src/main/asciidoc/repository.asciidoc
```diff
@@ -1007,7 +1007,7 @@ A stateful repository is backed by a _persistence context_, a set of managed ent
 
 - Multiple repositories might share a persistence context, especially if they share a datastore.
 
-- A persistence context is never shared across transactions.
+- A persistence context is never shared across transactions that are active at the same time.
```
```suggestion
- A persistence context is never shared across concurrent transactions.
```

stateful/src/main/java/jakarta/data/repository/stateful/Merge.java
```diff
@@ -36,6 +36,12 @@
  * database. This typically occurs later, when the persistence context is
  * flushed.
  * </p>
+ * <p>To ensure a boundary at which updates are persisted to the data store,
+ * the application should arrange to have repository methods annotated
+ * {@code @Merge}, as well as updates made directly to entities, participate
+ * in a transaction, the duration of which the same persistence context
+ * applies.
```
gavinking:
```suggestion
 * <p>The application program may ensure that pending changes held in entities 
 * associated with a persistence context are synchronized with the database by
 * committing the transaction associated with the persistence context.
```

stateful/src/main/java/jakarta/data/repository/stateful/Persist.java
```diff
@@ -35,6 +35,11 @@
  * when the annotated repository method is invoked, or it might occur later,
  * when the persistence context is flushed.
  * </p>
+ * <p>To ensure a boundary at which updates are persisted to the data store,
+ * the application should arrange to have repository methods annotated
+ * {@code @Persist} participate in a transaction, the duration of which
+ * the same persistence context applies.
```
gavinking:
```suggestion
 * <p>The application program may ensure that pending insertions of entities 
 * associated with a persistence context are synchronized with the database by
 * committing the transaction associated with the persistence context.
```

stateful/src/main/java/jakarta/data/repository/stateful/Remove.java
```diff
@@ -33,6 +33,11 @@
  * Deletion might occur immediately, when the annotated repository method is
  * invoked, or it might occur later, when the persistence context is flushed.
  * </p>
+ * <p>To ensure a boundary at which updates are persisted to the data store,
+ * the application should arrange to have repository methods annotated
+ * {@code @Remove} participate in a transaction, the duration of which
+ * the same persistence context applies.
```
gavinking:
```suggestion
 * <p>The application program may ensure that pending removals of entities 
 * associated with a persistence context are synchronized with the database by
 * committing the transaction associated with the persistence context.
```

stateful/src/main/java/module-info.java
```diff
@@ -16,8 +16,129 @@
...
+import jakarta.data.repository.Update;
+import jakarta.data.repository.stateful.Detach;
+import jakarta.data.repository.stateful.Merge;
+import jakarta.data.repository.stateful.Persist;
+import jakarta.data.repository.stateful.Refresh;
+import jakarta.data.repository.stateful.Remove;
+
 /**
- * API for Stateful Repositories in Jakarta Data
+ * <p>API for Stateful Repositories in Jakarta Data</p>
```
gavinking, line 32: Not important at all but note that putting `<p>` tags around the first line of a Javadoc comment is contrary to the conventions: https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html

stateful/src/main/java/module-info.java
```diff
@@ -16,8 +16,129 @@
...
+ * <li>{@link Detach @Detach}</li>
+ * <li>{@link Merge @Merge}</li>
+ * <li>{@link Persist @Persist}</li>
+ * <li>{@link Refresh @Refresh}</li>
+ * <li>{@link Remove @Remove}</li>
+ * </ul>
+ *
+ * <p>Stateful repository interfaces can optionally inherit from the built-in
+ * {@link DataRepository} interface to define a primary entity class and the
+ * type of its Id attribute. Otherwise, if all life cycle methods specify the
```
gavinking:
```suggestion
 * type of its Id attribute. Otherwise, if all lifecycle methods specify the
```

stateful/src/main/java/jakarta/data/repository/stateful/Remove.java
gavinking: I think it's *probably* okay to say:

> transaction associated with the persistence context

since that doesn't really commit us to any sort of affinity running in either direction. All it says is that there's some (transient) association between a tx and a pc. Which is correct.

api/src/main/java/module-info.java
njr-11: Good catch - I forgot about resource accessor methods here

stateful/src/main/java/module-info.java
njr-11: Wow- all these years I never realized that and have been writing old style html into Javadoc everywhere with the begin and end tags on paragraphs.

njr-11:
Thanks for all the improvements and clarifications to wording - Your suggestions looked good as you wrote them and I accepted/added all of them.

stateful/src/main/java/module-info.java
gavinking: Most people don't know about that page.

gavinking (approved):

PR #1432, opened 2026-04-20
TCK tests for Sort.nullsFirst() and nullsLast()

njr-11:
TCK tests covering Sort methods `nullsFirst()` and `nullsLast()`.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/SortNullableTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/SortNullableTests.java
@@ -1,0 +1,556 @@
+/*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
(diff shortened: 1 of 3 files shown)


tck/src/main/java/ee/jakarta/tck/data/standalone/entity/SortNullableTests.java
```diff
@@ -0,0 +1,618 @@
...
+                    50_000_000L,
+                    Order.by(_Country.daylightTimeBegins.asc().nullsFirst()));
+        } catch (IllegalArgumentException x) {
+            if (type.capableOfSortingNulls()) {
+                throw x;
+            } else {
+                return;
+            }
+        } catch (UnsupportedOperationException x) {
+            if (type.capableOfConstraintsOnNonIdAttributes() &&
```
otaviojava, line 136: I would say while we don't define how to handle it, I would roll back this and keep the classic based on the types.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/SortNullableTests.java
```diff
@@ -0,0 +1,618 @@
...
+            non-null values are in ascending order.
+            """)
+    public void testAscendingNullsLast() {
+        List<Country> found;
+        try {
+            found = countries.findLargest(
+                    50_000_000L,
+                    Order.by(_Country.daylightTimeBegins.asc().nullsLast()));
+        } catch (IllegalArgumentException x) {
+            if (type.capableOfSortingNulls()) {
```
otaviojava: The same here, mainly, when we talk about `SortingNulls` 
For now, I would say, let's keep the previous strategy for type.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/SortNullableTests.java
njr-11: I don't understand your comment here, but I did notice that the checking for sorting of nulls is not sufficiently granular enough. Is that what you are commenting on?  I added a commit to address that.

KyleAure (approved):

PR #1441, opened 2026-05-04
 TCK tests for the First anno

### LINKED ISSUE #532: [Use Case]: Statically specify First/Limit at dev time so that singular return type can be valid on methods
### As a ...

- [X] Application user/user of the configuration itself
- [X] API user (application developer)

### I need to be able to ...

statically specify a limit at development time

### Which enables me to ...

write a repository method with a singular return type (`Optional<MyEntity>` or `MyEntity`).

### Additional information

This is currently only possible (and awkwardly so) with Query by Method Name.

To allow it with `@Query` and parameter based queries (`@Find`), we can add `First` (proposed by Gavin in https://github.com/jakartaee/data/pull/530#discussion_r1518094185)

```
@Find @First
@OrderBy(value = _Employee.SALARY, descending = true)
@OrderBy(value = _Employee.ID)
Optional<Employee> highestPaid(String jobTitle);
```

The value would default to `1`.

Reminder to update recommendations in `NonUniqueResultException` once we have this.

This is being proposed for after Jakarta Data version 1.0.

njr-11:
Adds TCK tests for the `@First` annotation, covering usage with `@Find` and `@Query`, obtaining a single result or multiple results, including cases where there are less results available than the requested First amount.

Resolves #532

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
@@ -23,6 +23,7 @@ import java.util.stream.Stream;
 
 import jakarta.data.Limit;
 import jakarta.data.Order;
+import jakarta.data.Sort;
 import jakarta.data.Sort.Nulls;
 import jakarta.data.constraint.AtLeast;
 import jakarta.data.constraint.AtMost;
@@ -43,8 +44,10 @@ import jakarta.data.page.PageRequest;
 import jakarta.data.repository.By;
 import jakarta.data.repository.CrudRepository;
 import jakarta.data.repository.Find;
+import jakarta.data.repository.First;
 import jakarta.data.repository.Is;
 import jakarta.data.repository.OrderBy;
+import jakarta.data.repository.Query;
 import jakarta.data.repository.Repository;
 import jakarta.data.repository.Select;
 import jakarta.data.restrict.Restriction;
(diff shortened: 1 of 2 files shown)


KyleAure (commented):
Looks good.  While I was reviewing this I thought of a couple of missing negative tests: 

- Verify a repository method with the `@First` annotation and a `Limit` parameter throws an `UnsupportedOperationException`
- Verify a repository method with the `@First` annotation and the `First` keyword (query by method name) throws an `UnsupportedOperationException`

I think it would also be useful to have one of the repository methods return an array instead of a `List` since there is an example of that in javadoc. 

I did not find this scenario discussed, but should the following repository method work, or be rejected?

```java
    @Find
    @First
    List<Country> leastBy(Sort<Country> sort);
```
We left the default value `1` but still want to return a list. Not something I'd expect a user to WANT to do, but we might want to document what behavior we expect.

njr-11:
Thanks Kyle - those are some good comments.

>     * Verify a repository method with the `@First` annotation and a `Limit` parameter throws an `UnsupportedOperationException`
> 
>     * Verify a repository method with the `@First` annotation and the `First` keyword (query by method name) throws an `UnsupportedOperationException`

According to the spec, UnsupportedOperationException is one option, but the Data provider alternatively has the option to fail these at build time, so while the above are good ideas to have coverage for, it will be something for Data providers to test on their own.

> I think it would also be useful to have one of the repository methods return an array instead of a `List` since there is an example of that in javadoc.

added in the latest commit
 
> I did not find this scenario discussed, but should the following repository method work, or be rejected?
> 
> ```java
>     @Find
>     @First
>     List<Country> leastBy(Sort<Country> sort);
> ```
> 
> We left the default value `1` but still want to return a list. Not something I'd expect a user to WANT to do, but we might want to document what behavior we expect.

Yes, the spec allows this usage. Although it is not often likely to be useful, I don't see a good reason to go out of our way to add extra language in the spec saying that it must be rejected. I added a test for it.

The latest commit also rebased to resolve a merge conflict.

KyleAure (approved):

PR #1449, opened 2026-05-18
TCK tests for stateful repository method annotations

njr-11:
Some initial TCK tests for Detach, Merge, Persist, Refresh, and Remove methods of stateful repositories.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/Inventory.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/Inventory.java
@@ -35,9 +35,7 @@ import jakarta.transaction.Transactional;
 @Repository
 public interface Inventory {
 
-    @Query("""
-            DELETE FROM Product
-            """)
+    @Query("DELETE FROM Product")
     void erase();
 
     @Find
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/Inventory.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/Inventory.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+package ee.jakarta.tck.data.standalone.persistence;
+
+
+import java.util.List;
+import java.util.stream.Stream;
+
+import jakarta.data.Order;
+import jakarta.data.repository.Find;
+import jakarta.data.repository.Query;
+import jakarta.data.repository.Repository;
+import jakarta.data.repository.stateful.Merge;
+import jakarta.data.repository.stateful.Persist;
+import jakarta.data.repository.stateful.Remove;
+import jakarta.data.restrict.Restriction;
+import jakarta.transaction.Transactional;
+
+/**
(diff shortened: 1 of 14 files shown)


otaviojava:
@njr-11 could you update this branch?

njr-11:
> @njr-11 could you update this branch?

Thanks for pointing that out.  I rebased onto the latest and resolved the merged conflict, so it is ready for review now.

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/Inventory.java
```diff
@@ -0,0 +1,65 @@
...
+import jakarta.transaction.Transactional;
+
+/**
+ * A stateful repository for the Product entity.
+ */
+@Repository
+public interface Inventory {
+
+    @Query("""
+            DELETE FROM Product
```
otaviojava: Nothing critical, but once it has a single line, what are the benefits of this approach?

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/Inventory.java
```diff
@@ -0,0 +1,65 @@
...
+    @Persist
+    @Transactional
+    void persist(Product product);
+
+    @Remove
+    @Transactional
+    void remove(Product product);
+
+    @Query("""
+            FROM Product
```
otaviojava, line 57: This one makes sense because it has three lines.

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/StatefulPersistenceEntityTests.java
```diff
@@ -0,0 +1,474 @@
...
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+package ee.jakarta.tck.data.standalone.persistence;
```
otaviojava: Given this is a specialization of persistence, 

I believe it makes sense to move it to:

```
package ee.jakarta.tck.data.standalone.persistence.stateful;
```

What do you think?

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/Inventory.java
njr-11: > Nothing critical, but once it has a single line, what are the benefits of this approach?

There aren't any benefits.  I used the same style for all of the `@Query` methods, and for this one, the multi-line string isn't needed. I added a commit to switch it to a single line as you suggested.

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/StatefulPersistenceEntityTests.java
njr-11: These tests use the same entity `ee.jakarta.tck.data.standalone.persistence.Product` that is already used by `PersistenceEntityTests` in the `ee.jakarta.tck.data.standalone.persistence` package.  So this also covers that you can define both stateless and stateful repositories for the same entity, and if the stateless repository is in the same package as the entity, then it makes sense for the stateful repository to also be in the same package as the entity.

njr-11: Alternatively, the existing stateless repository and tests from Data 1.0 TCK could be moved to a stateless subpackage, in which case it would then make sense to put the stateful repository and tests into a stateful subpackage.  You can let me know which you prefer.

otaviojava: I prefer moving those to the stateful subpackage, if you are ok with it.

njr-11: The latest commit does what you asked for.

njr-11:
I only opened this PR 2 days ago -- using what was at the time the latest level -- and it has already been hit with merge conflicts to resolve twice.  The latest round of merge conflicts is resolved now.

otaviojava (approved):

PR #1455, opened 2026-05-20
make ComparableExpressions sortable

### LINKED ISSUE #1280: Missing a way to statically reference a sort by an embedded or associated property

gavinking:
This had a bigger impact than I expected.

Question: do we need `SortableExpression`?

### DIFF
--- a/api/src/main/java/jakarta/data/Sort.java
+++ b/api/src/main/java/jakarta/data/Sort.java
@@ -18,8 +18,12 @@
 package jakarta.data;
 
 import jakarta.data.messages.Messages;
+import jakarta.data.metamodel.Attribute;
 import jakarta.data.metamodel.StaticMetamodel;
+import jakarta.data.metamodel.TextAttribute;
 import jakarta.data.repository.OrderBy;
+
+import java.util.Objects;
 
 /**
  * <p>Requests sorting on a given entity attribute.</p>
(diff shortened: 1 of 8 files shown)


api/src/main/java/jakarta/data/Sort.java
```diff
@@ -168,28 +175,75 @@ public enum Nulls {
...
+     * <li>{@link #descIgnoreCase(String) Sort.descIgnoreCase(attributeName)}
+     *     for case-insensitive descending sort on an entity attribute.</li>
+     * <li>{@link #descIgnoreCase(String) Sort.descIgnoreCase(attributeName)}.{@link #nullsLast()}
+     *     for case-insensitive descending sort on an entity attribute,
+     *     where {@code null} values are ordered last.</li>
+     * </ul>
+     * This constructor will be removed in a future release.
      */
+    @Deprecated(since = "1.1", forRemoval = true)
     public Sort(String property, boolean isAscending, boolean ignoreCase) {
```
gavinking, line 224: 👀

api/src/main/java/jakarta/data/Sort.java
```diff
@@ -78,20 +82,14 @@
...
- *                     from a database with case sensitive collation.
- * @param nullOrdering whether {@code null} values are ordered
- *                     {@link Nulls#FIRST FIRST}, {@link Nulls#LAST LAST}, or
- *                     {@linkplain Nulls#UNSPECIFIED by the data store}.
  */
-public record Sort<T>(String property,
-                      boolean isAscending,
-                      boolean ignoreCase,
-                      Nulls nullOrdering) {
+public final class Sort<T> {
```
gavinking, line 86: 👀

njr-11 (commented):
I'd rather not introduce `Sortable`.  It doesn't do anything useful for the end user and I don't think we need it.  I wanted to try out adding the asc/desc methods to ComparableExpression and the ascIgnoreCase/descIgnoreCase methods to TextExpression.  This seems to work fine in a [commit of 1400](https://github.com/jakartaee/data/pull/1400/changes/1da4b4905cfe355f0ad0882d60333375a2b47d6f) where I tried it out.  I created a couple of [new unit tests](https://github.com/jakartaee/data/pull/1400/changes/1d4ab0239398bb27cbb7677c56320e20178c0ebb) to verify it as well as updating the test that were previously under 1400 so they test navigable expressions rather than navigable attributes.  That all seemed to work.

api/src/main/java/jakarta/data/Sortable.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Obtain an ascending {@linkplain Sort sorting criterion} based on
+     * this entity attribute or expression.
+     *
+     * @return the sorting criterion.
+     */
+    default Sort<T> asc() {
+        return Sort.asc(this);
+    }
```
gavinking, line 36: So @njr-11 the thing I would still defend here after our last conversation is I think this is a much more elegant way to construct a `Sort`.

njr-11: I agree `Sort.asc(expression)` is more elegant than using a constructor, but this is internal usage and the end user already has a better option in directly using the metamodel.  Nonetheless, I added it to the other PR given that I don't see anything objectionable about it.

gavinking:
Closing this, since Nathan's version of it was merged.

PR #1456, opened 2026-05-20
TCK tests of Asynchronous repository method

njr-11:
Adds a couple of TCK tests to cover Asynchronous on a repository method.
The tests include logic to check whether Jakarta Concurrency is present in the environment and if not present, the tests skip themselves.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/web/async/Account.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/web/async/Account.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+package ee.jakarta.tck.data.web.async;
+
+import java.time.LocalDateTime;
+
+@jakarta.nosql.Entity
+@jakarta.persistence.Entity
+public class Account {
+    @jakarta.nosql.Id
+    @jakarta.persistence.Id
+    int accountId;
+
+    @jakarta.nosql.Column
+    @jakarta.persistence.Column(nullable = false)
+    boolean active;
+
+    @jakarta.nosql.Column
+    @jakarta.persistence.Column(nullable = false)
+    float balance;
+
+    @jakarta.nosql.Column
+    @jakarta.persistence.Column(nullable = false)
+    LocalDateTime created;
+
+    @jakarta.nosql.Column
+    String email;
+
+    public static Account of(int accountId,
+                             boolean active,
+                             float balance,
+                             LocalDateTime created,
+                             String email) {
+        Account account = new Account();
+        account.setAccountId(accountId);
+        account.setActive(active);
+        account.setBalance(balance);
+        account.setCreatedOn(created);
+        account.setEmail(email);
+        return account;
+    }
+
+    public int getAccountId() {
+        return accountId;
+    }
+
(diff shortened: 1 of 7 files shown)


tck/src/main/java/ee/jakarta/tck/data/web/async/Account.java
```diff
@@ -0,0 +1,105 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+package ee.jakarta.tck.data.web.async;
+
+import java.time.LocalDateTime;
+
+@jakarta.nosql.Entity
+@jakarta.persistence.Entity
+public class Account {
+    @jakarta.nosql.Id
```
otaviojava, line 24: Do you mind putting a line here?

njr-11: > Do you mind putting a line here?

I don't understand your question. My best guess is that you want a new line prior to the `Id` annotation.  I added a commit that does that:

https://github.com/jakartaee/data/pull/1456/changes/d882dfcfa6c0f91f1f970c90babb9a4e50aba0ab

gavinking:
@njr-11 I would like to better understand your intention with this one. As the spec is written today, I believe async repositories are not required:

> NOTE: Every Jakarta Data provider is encouraged, but not required, to support asynchronous repository methods returning `java.util.concurrent.CompletionStage`.

How is a provider which does not support `CompletionStage` going to indicate that to the TCK?

njr-11:
> How is a provider which does not support `CompletionStage` going to indicate that to the TCK?

Thanks for spotting that - I included logic to allow for Concurrency spec not being available, but forgot to include the catch block that tolerates the former like we have elsewhere throughout the TCK.  This falls under the general category of unsupported repository method signatures, but I'll also add a note to the specification pointing out how to reject it.

Corrections are included in:
https://github.com/jakartaee/data/pull/1456/changes/c9cd9cb779703c0420f770a9f4f8a3a222270fe8

gavinking:
OK thanks.

tck/src/main/java/ee/jakarta/tck/data/web/async/Accounts.java
```diff
@@ -0,0 +1,47 @@
...
+ * annotation need not be available at run time, because the TCK will
+ * detect the presence or absence of the Jakarta Concurrency API and
+ * skip the tests when not present.
+ */
+@Repository
+public interface Accounts extends BasicRepository<Account, Integer> {
+
+    @Asynchronous
+    @Insert
+    void add(Account... accounts);
```
gavinking:
```suggestion
    CompletionStage<Void> add(Account... accounts);
```

njr-11: I was wondering why you proposed switching this one to `CompletionStage` given that the other repository method already covers a return type of `CompletionStage`.   Then I went back and looked at the documentation.  The specification does doesn't mention `void` as a valid return type, which is probably what you are looking at here.  This pattern does appear in the Javadoc under its section on integration with Jakarta Concurrency,

> The return type must be `void` or `CompletionStage<R>` where `R` is a type that would be a valid return type for a non-asynchronous repository method.

The pattern with `void` return is the one that was already inadvertently working as of Data 1.0 due to requirements around interceptors and Concurrency's Asynchronous being an interceptor.

Here is a commit that adds mention of it in the spec,
https://github.com/jakartaee/data/pull/1456/changes/ba38774d4c87bc841de7fefa52683d31dd2ebba2

gavinking: So the question is whether we're allowing fire-and-forget.

I don't have a problem with fire-and-forget for `@Asynchronous`, but I don't think the TCK should test it. The reason I say that is that only *some* models for asynchrony support that. In particular reactive definitely *doesn't* support it.

I think the basic most vanilla case is that you get a notification when your `void` operation has completed.

njr-11: Sure, that's fine to omit fire-and-forget from the TCK.  I'll accept your change and then update the test accordingly.

njr-11: Updated in https://github.com/jakartaee/data/pull/1456/changes/efc917f539d7598a638a9ce43ef24cd03ba26604 .  It turns out switching from fire-and-forget to `CompletionStage<Void>` also makes for a much cleaner test.

gavinking (approved):
Cool, thanks!

PR #1461, opened 2026-05-22
respecify EntityExistsException

gavinking:
This is needed because we cannot always determine the precise cause of a unique key violation. This change aligns the Data spec with a change already made in Persistence.

### DIFF
--- a/api/src/main/java/jakarta/data/repository/CrudRepository.java
+++ b/api/src/main/java/jakarta/data/repository/CrudRepository.java
@@ -97,12 +97,14 @@ import java.util.List;
 public interface CrudRepository<T, K> extends BasicRepository<T, K> {
 
     /**
-     * <p>Inserts an entity into the database. If an entity of this type with
-     * the same
-     * unique identifier already exists in the database and the database
-     * supports ACID transactions, then this method raises
-     * {@link EntityExistsException}. In databases that follow the BASE model or
-     * use an append model to write data, this exception is not thrown.</p>
+     * <p>Inserts an entity into the database.<p>
+     *
+     * <p>If the database supports ACID transactions and insertion of the given
+     * entity would violate a uniqueness constraint, for example, if an entity
+     * of this type with the same unique identifier already exists in the database,
+     * then this method raises {@link EntityExistsException}. In databases that
+     * follow the BASE model or use an append model to write data, this exception
+     * is not thrown.</p>
      *
      * <p>The entity instance returned as a result of this method must include
      * all values that were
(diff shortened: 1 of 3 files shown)


api/src/main/java/jakarta/data/repository/CrudRepository.java
```diff
@@ -117,21 +119,24 @@ public interface CrudRepository<T, K> extends BasicRepository<T, K> {
      * @return the inserted entity, which may or may not be a different instance
      * depending on whether the insert caused values to be generated or
      * automatically incremented.
-     * @throws EntityExistsException if the entity is already present in the
-     *                               database (in ACID-supported databases).
+     * @throws EntityExistsException (in ACID-supported databases) if the entity
```
otaviojava, line 122: Gavin, we have a problem here with NoSQL databases. Because:

 In many NoSQL databases, this assumption is not always true because insert and update operations are often unified as an upsert operation. E.g.: Apache Cassandra treats inserts as append-only writes, where existing data may simply be replaced or versioned rather than violating a uniqueness constraint.

gavinking: @otaviojava I didn't change anything in this respect. The "in ACID-supported databases" was the pre-existing language (which this commit retains).

api/src/main/java/jakarta/data/repository/Insert.java
```diff
@@ -64,10 +64,11 @@
  * of entities in the argument. After the annotated method returns, an original entity instance supplied as an argument
  * might not accurately reflect the inserted state.
  * </p>
- * <p>If an entity of the given type, and with the same unique identifier already exists in the database when the
- * annotated method is called, and if the databases uses ACID (atomic, consistent, isolated, durable) transactions,
- * then the annotated method must raise {@link jakarta.data.exceptions.EntityExistsException}.
- * If the database follows the BASE model, or uses an append model to write data, this exception is not thrown.
+ * <p>If the databases uses ACID (atomic, consistent, isolated, durable) transactions, and insertion of an entity would
```
njr-11:
```suggestion
 * <p>If the database uses ACID (atomic, consistent, isolated, durable) transactions, and insertion of an entity would
```

njr-11 (approved):
I noticed one minor typo but otherwise looks good and is consistent with recent updates to Jakarta Persistence. Approving in advance.

gavinking:
Thanks Nathan, applied your fix.

api/src/main/java/jakarta/data/repository/CrudRepository.java
otaviojava: ahhh ok, I will propose an update in next PR, thanks Gavin.

PR #1466, opened 2026-05-27
Various TCK updates

marko-bekhta:
Hey 👋🏻 🙂 

Here are a few things I've found so far while going through the new tests and running them against Hibernate ORM.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
@@ -242,7 +242,11 @@ public interface Countries extends CrudRepository<Country, String> {
     @Find
     @OrderBy(_Country.CAPITAL_POPULATION)
     List<Country> withCapitalBiggerThan(
-            GreaterThan<Integer> capitalPopulation);
+            // CHECKSTYLE:OFF
+            // using an underscore in the parameter name on purpose
+            // to test the specification feature:
+            GreaterThan<Integer> capital_population);
+            // CHECKSTYLE:ON
 
     @Find
     List<Country> withCapitalNamed(
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/NaturalNumbers.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/NaturalNumbers.java
@@ -50,7 +50,7 @@ public interface NaturalNumbers extends BasicRepository<NaturalNumber, Long>, Id
     CardinalNumber cardinalNumberOf(long id);
 
     @Find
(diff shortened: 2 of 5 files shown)


gavinking (approved):
We still don't require that implementations support implicit numeric conversions, so +1.

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/Countries.java
```diff
@@ -242,7 +242,7 @@ List<Country> whereDaylightTimeStartsOnOrAfter(
     @Find
     @OrderBy(_Country.CAPITAL_POPULATION)
     List<Country> withCapitalBiggerThan(
-            GreaterThan<Integer> capitalPopulation);
+           @By(_Country.CAPITAL_POPULATION) GreaterThan<Integer> capitalPopulation);
```
gavinking: Alternatively, `capitalPopulation` -> `capital_population`.

njr-11: Given that this method is used in a test that explicitly states it is testing the `GreaterThan` parameter type, we should go with Gavin's suggested alternative that preserves this

>     @Assertion(id = "965", strategy = """
>                          Use a repository find method that has a constraint
>                          parameter of type GreaterThan.
>                     """)
>     public void testGreaterThanConstraint() {
> 
>         List<String> found;
>         try {
>             found = countries
>                             .withCapitalBiggerThan(GreaterThan.bound(5000000))


```suggestion
            GreaterThan<Integer> capital_population);
```

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/NaturalNumbers.java
```diff
@@ -164,13 +164,13 @@ List<WholeNumber> wholeNumberList(@By(_NaturalNumber.NUMTYPEORDINAL) int numType
     @Select(_NaturalNumber.NUMTYPEORDINAL)
     @Select(_NaturalNumber.FLOOROFSQUAREROOT)
     @Select(_NaturalNumber.ID)
-    Optional<WholeNumber> wholeNumberOf(@By(ID) int id);
+    Optional<WholeNumber> wholeNumberOf(@By(ID) long id);
```
njr-11, line 167: Sorry about mistakenly using `int` in all of these places.  I wasn't paying close enough attention to the correct type when adding these.

marko-bekhta: no worries 🙂

njr-11 (approved):

PR #1467, opened 2026-05-28
Always return same instance of CurrentDate, CurrentTime, or CurrentDateTime

njr-11:
Currently we are creating a new instance of CurrentDate (and likewise for CurrentTime/CurrentDateTime) every time the TemporalExpression.localDate or CurrentDate.now method is invoked, such that CurrentDate.now() is unequal to CurrentDate.now() even though both represent the same datetime expression, `LOCAL DATE`.

### DIFF
--- a/api/src/main/java/jakarta/data/spi/expression/function/CurrentDate.java
+++ b/api/src/main/java/jakarta/data/spi/expression/function/CurrentDate.java
@@ -47,6 +47,9 @@ public interface CurrentDate<T> extends TemporalExpression<T, LocalDate> {
 class CurrentDateInstance implements CurrentDate<Object> {
     static final CurrentDate<?> instance = new CurrentDateInstance();
 
+    private CurrentDateInstance() {
+    }
+
     @Override
     public Class<LocalDate> type() {
         return LocalDate.class;
--- a/api/src/main/java/jakarta/data/spi/expression/function/CurrentDate.java
+++ b/api/src/main/java/jakarta/data/spi/expression/function/CurrentDate.java
@@ -37,17 +37,23 @@ public interface CurrentDate<T> extends TemporalExpression<T, LocalDate> {
      *
      * @return an expression representing the current date.
      */
+    @SuppressWarnings("unchecked")
     static <T> CurrentDate<T> now() {
-        return new CurrentDate<>() {
-            @Override
-            public Class<LocalDate> type() {
-                return LocalDate.class;
-            }
+        return (CurrentDate<T>) CurrentDateInstance.instance;
+    }
+}
 
-            @Override
-            public String toString() {
-                return "LOCAL DATE";
-            }
-        };
+// Internal implementation of single instance obtained from CurrentDate.now()
+class CurrentDateInstance implements CurrentDate<Object> {
+    static final CurrentDate<?> instance = new CurrentDateInstance();
+
+    @Override
+    public Class<LocalDate> type() {
+        return LocalDate.class;
+    }
+
+    @Override
+    public String toString() {
+        return "LOCAL DATE";
     }
 }
--- a/api/src/test/java/jakarta/data/expression/TemporalExpressionTest.java
+++ b/api/src/test/java/jakarta/data/expression/TemporalExpressionTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ * Copyright (c) 2025,2026 Contributors to the Eclipse Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


api/src/main/java/jakarta/data/spi/expression/function/CurrentDate.java
```diff
@@ -37,17 +37,23 @@ public interface CurrentDate<T> extends TemporalExpression<T, LocalDate> {
...
-            @Override
-            public Class<LocalDate> type() {
-                return LocalDate.class;
-            }
+        return (CurrentDate<T>) CurrentDateInstance.instance;
+    }
+}
+
+// Internal implementation of single instance obtained from CurrentDate.now()
+class CurrentDateInstance implements CurrentDate<Object> {
```
otaviojava, line 47: make it private?

api/src/test/java/jakarta/data/expression/TemporalExpressionTest.java
```diff
@@ -98,4 +102,22 @@ void shouldRestrictLocalDateBetween() {
                 .isEqualTo(LocalDate.of(2025, Month.APRIL, 30));
         });
     }
+
+    @Test
+    void testCurrentDateEqualsCurrentDate() {
+        assertEquals(CurrentDate.now(),
```
otaviojava: I would use the assertj on it.

api/src/main/java/jakarta/data/spi/expression/function/CurrentDate.java
njr-11: No - for two reasons.  Java does not allow `private` on top level classes.  Even if it did allow that, the `CurrentDate` class needs access to it, so the default access level of package-private is what we want.

pzygielo: make it inner, static, private then

I understand CS might not be used here, but there also is https://checkstyle.org/checks/design/onetoplevelclass.html#Description to consider.

api/src/test/java/jakarta/data/expression/TemporalExpressionTest.java
njr-11: That's fine, I switched it for you.

api/src/main/java/jakarta/data/spi/expression/function/CurrentDate.java
njr-11: > make it inner, static, private then

Java does not allow `private` inner classes on interfaces.  That does not compile.

> I understand CS might not be used here, but there also is https://checkstyle.org/checks/design/onetoplevelclass.html#Description to consider.

This project does use checkstyle, but the OneTopLevelClass check is not enabled.  I looked over that check (thanks for posting the direct link) and it does seem like a nice guideline to follow in a lot of cases but I do not see how it would add value here.  Given that the `CurrentDateInstance` is only used by `CurrentDate` (which is why it would have made sense as a private static inner class if Java allowed that on interfaces) and there is no value in any other classes using it, it seems appropriate to isolate it to the same file as `CurrentDate` in this case.

pzygielo: > > make it inner, static, private then
> 
> Java does not allow `private` inner classes on interfaces. That does not compile.

Missed that it's interface 😞

pzygielo: > Given that the `CurrentDateInstance` is only used by `CurrentDate` (which is why it would have made sense as a private static inner class if Java allowed that on interfaces) and there is no value in any other classes using it, it seems appropriate to isolate it to the same file as `CurrentDate` in this case.

But this will not prevent using it by other classes (in the package I mean), will it?

~Having this as separate class smells funny.~
On the second look, it looks quite nice.

njr-11: > But this will not prevent using it by other classes (in the package I mean), will it?

I agree, it doesn't prevent other classes in the package from using it. Mostly, it just keeps it out of their way and isolates it to the only place that cares about it.

njr-11:
I did think of one improvement to add: making the constructor private to enforce that only the single static instance can ever be obtained.  This is covered in commit https://github.com/jakartaee/data/pull/1467/changes/0a0b0f007c1bfe983dfcd2d22869f854aa2545fb

api/src/main/java/jakarta/data/spi/expression/function/CurrentDate.java
```diff
@@ -37,17 +37,26 @@ public interface CurrentDate<T> extends TemporalExpression<T, LocalDate> {
...
-            public Class<LocalDate> type() {
-                return LocalDate.class;
-            }
-
-            @Override
-            public String toString() {
-                return "LOCAL DATE";
-            }
-        };
+        return (CurrentDate<T>) CurrentDateInstance.instance;
```
gavinking, line 42: I'm sure I'm in the minority here, but honestly I kinda prefer an unnecessary instantiation to an unchecked cast. Yes, I get that it's perfectly sound here due to type argument erasure but even so ...

otaviojava (approved):

PR #1482, opened 2026-06-03
minimal API changes for 0-based offsets

njr-11:
This PR starts with some of Gavin's updates from #1479 and pares down to a minimum the amount of API updates needed to accommodate a developer that wants to think about making PageRequests as a 0-based offset from the first page and defining Limits in terms of a 0-based offset from first result.

To achieve this, we really only need 2 methods.

`pageOffset(offset)` for `PageRequest`:
```java
page0Request = PageRequest.ofSize(10).pageOffset(0);
```

`limit.of(maxResults, offset)` for `Limit`:
```java
limitToFifthGroupOf10 = Limit.of(10, 40);
```

### DIFF
--- a/api/src/main/java/jakarta/data/Limit.java
+++ b/api/src/main/java/jakarta/data/Limit.java
@@ -17,14 +17,22 @@
  */
 package jakarta.data;
 
+import jakarta.data.messages.Messages;
 import jakarta.data.page.PageRequest;
+import jakarta.data.repository.Find;
+import jakarta.data.repository.Query;
 
 /**
  * <p>Specifies a limit on the number of results retrieved by a repository
  * method. The results of a single invocation of a repository method may be
- * limited to a given {@linkplain #of(int) maximum number of results} or to a
- * given {@linkplain #range(long, long) positioned range} defined in terms of an
- * offset and maximum number of results.</p>
+ * limited to
+ * <ul>
+ * <li>a {@linkplain #of(int) maximum number of results},</li>
+ * <li>a {@linkplain #of(int, long) maximum number of results relative to an
+ *     offset}, or</li>
+ * <li>a {@linkplain #range(long, long) positioned range} defined in terms of
+ * a starting position and maximum number of results.</li>
+ * </ul></p>
  *
  * <p>A query method of a repository may have a parameter of type
  * {@code Limit} if its return type indicates that it may return multiple
@@ -105,14 +113,14 @@ public record Limit(int maxResults, long startAt) {
 
     /**
      * <p>The position at which to start when returning query results.</p>
-     * <p>The first query result is at position one. If the start position
+     * <p>The first query result is at position {@code 1}. If the start position
      * is greater than one, some results at the beginning of the result set
      * are skipped.</p>
      *
      * @return position of the first result.
      *
      * @apiNote Positions are indexed from one;
-     *          {@linkplain #startOffset offsets} are indexed from zero.
+     *          offsets are indexed from zero.
      */
     public long startAt() {
         return startAt;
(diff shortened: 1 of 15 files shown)


api/src/main/java/jakarta/data/Limit.java
```diff
@@ -17,14 +17,22 @@
...
- * limited to a given {@linkplain #of(int) maximum number of results} or to a
- * given {@linkplain #range(long, long) positioned range} defined in terms of an
- * offset and maximum number of results.</p>
+ * limited to
+ * <ul>
+ * <li>a {@linkplain #of(int) maximum number of results},</li>
+ * <li>a {@linkplain #of(int, long) maximum number of results relative to an
+ *     offset}, or</li>
+ * <li>a {@linkplain #range(long, long) positioned range} defined in terms of
+ * a starting position and maximum number of results.</li>
```
FroMage:
```suggestion
 * a starting position and an ending position.</li>
```

api/src/main/java/jakarta/data/page/PageRequest.java
```diff
@@ -259,16 +262,55 @@ static PageRequest beforeCursor(Cursor cursor, long pageNumber, int maxPageSize,
...
+            throw new IllegalStateException(
+                    Messages.get("014.mode.disallows.offset", mode()));
+        }
+
+        if (offset < 0) {
+            throw new IllegalArgumentException(
+                    Messages.get("004.arg.negative", "offset"));
+        }
+
+        if (offset == Long.MAX_VALUE) {
```
otaviojava, line 317: I am not sure about this one here.

1) the Long.MAX_VALUE still a valid number right?

Why is this not supported?

njr-11: >     1. the Long.MAX_VALUE still a valid number right?
> 
> Why is this not supported?

Of course it's a valid number.  But an offset of Long.MAX_VALUE from the first result has a page number (Long.MAX_VALUE + 1) that is not representable as a positive `long` value.  If we allowed it, what would you propose to return when a user/Data provider asks the API for the page number?

gavinking: 10^19 is an implausible number of rows.

FroMage: You know, that's precisely how many rows I have. Make it work.

njr-11:
Includes [additional changes](https://github.com/jakartaee/data/pull/1482/changes/8c47327c67aed182cf78ca30a8b3d3d5a0c3fdfa) that were agreed to on the Jakarta Data meeting:

- `.page()` is deprecated in favor of a new `.pageNumber()`
- `.page(num)` is renamed `.pageNumber(num)` which is possible because the method is new in 1.1

Usage for page requests is now:

```java
PageRequest.ofSize(10).pageNumber(1);
```
or

```java
PageRequest.ofSize(10).pageOffset(0);
```

depending on whether you want 1-based or 0-based.

otaviojava (approved):

api/src/main/java/jakarta/data/page/PageRequest.java
```diff
@@ -259,15 +273,54 @@ static PageRequest beforeCursor(Cursor cursor, long pageNumber, int maxPageSize,
...
+            throw new IllegalArgumentException(
+                    Messages.get("013.arg.invalid", "offset", offset));
```
otaviojava, line 319: Need to update to 15.

njr-11:
@otaviojava I resolved the soft merge conflict with commit https://github.com/jakartaee/data/pull/1482/changes/ef45af271850e14320b5a28984c16bb59f5c6b94

PR #1494, opened 2026-06-19
add nullability annotations

### LINKED ISSUE #1493: Nullability annotations
I'm now very happy with how the use of `jakarta.annotation.Nullable` and `jakarta.annotation.Nonnull` has turned out in the persistence spec, and I think we should do something similar here. That is:

- annotate methods and method parameters of Jakarta Data APIs with `@Nullable` or `@Nonnull` as appropriate, and even
- encourage the developer to annotate repository methods and their parameters with these annotations, allowing the repository implementation to make use of the additional information (we already support this in Hibernate).

There's no real downside to any of this that I can see.

gavinking:
These were mostly robotically generated. I have done some review, but perhaps not quite as thorough as I should.


api/src/main/java/jakarta/data/repository/Is.java
```diff
@@ -117,6 +117,5 @@
      *
      * @return the type of comparison operation.
      */
-    @SuppressWarnings("rawtypes")
     Class<? extends Constraint> value() default EqualTo.class;
```
njr-11, line 121: Was SuppressWarnings removed by mistake?  Without it, I get the warning:
"Constraint is a raw type. References to generic type Constraint<V> should be 
parameterized"

```suggestion
    @SuppressWarnings("rawtypes")
    Class<? extends Constraint> value() default EqualTo.class;
```

api/src/main/java/jakarta/data/repository/OrderBy.java
```diff
@@ -73,9 +73,9 @@
  * <ul>
  * <li>the <em>Query by Method Name</em> {@code OrderBy} keyword in its
  *     name,</li>
- * <li>a {@link Query} or {@code jakarta.persistence.query.JakartaQuery}
+ * <li>a {@link Query} or {@link jakarta.persistence.query.JakartaQuery}
  *     annotation specifying a query with an {@code ORDER BY} clause, nor</li>
- * <li>a {@code jakarta.persistence.query.NativeQuery} annotation.</li>
+ * <li>a {@link jakarta.persistence.query.NativeQuery} annotation.</li>
```
njr-11, line 78: We definitely want to change this to `link`, but I noticed that afterward the build generates the Javadoc without any link and without any package, as,

Image

which is worse than just `{@code` which at least includes the package name when there is no link.  Are there more build steps that are needed to get it to find Persistence 4.0?

njr-11 (commented):
The changes will Nonnull and Nullable look fine. I noted a couple of minor things with some other changes in the PR.
Given how many places this modifies and the merge conflicts it is likely to cause (as well as soft merge conflicts), we should consider delaying merging it until after the current backlog of PRs is merged

api/src/main/java/jakarta/data/repository/Is.java
gavinking: God damn it, sorry. This keeps happening. 

I just don't see that warning,neither in the Maven build nor in IntelliJ. Instead I get a warning from IntelliJ that the suppression is redundant. But of course this is not the first time that's happened and I should have been more careful.

gavinking: Screenshot 2026-06-19 at 7 07 46 PM

I have _no idea_ what we could have set up differently that would cause these inconsistencies.

api/src/main/java/jakarta/data/repository/OrderBy.java
gavinking: @njr-11 Fixed now, I believe, but please test it.

gavinking:
>  the merge conflicts it is likely to cause

I strongly suspect that the merge conflicts will be pretty trivial, but also FYI Codex is fantastic at resolving merge conflicts, even highly nontrivial ones.

api/src/main/java/jakarta/data/repository/Is.java
gavinking: Anyway I have reverted the change.

api/src/main/java/jakarta/data/repository/OrderBy.java
njr-11: Nice - it is working now.

njr-11 (approved):
Looks good now

> I have _no idea_ what we could have set up differently that would cause these inconsistencies.

I don't know why we are getting inconsistent warnings either.  Probably the best thing to do in the hopefully rare cases where this occurs will be to align with the maven build output and then put a comment in the code where the SuppressWarnings would be indicating that the warning suppression should either be left there or not added, so that we don't go back and forth on adding and removing the suppression.  In this case, it looks like Maven is okay with removing SuppressWarnings, so we can remove it and add the note so that I don't put it back in later when I see it.

> > the merge conflicts it is likely to cause
> 
> I strongly suspect that the merge conflicts will be pretty trivial, but also FYI Codex is fantastic at resolving merge conflicts, even highly nontrivial ones.

Every outstanding PR that adds a method will have a soft merge conflict with this.  Nothing will show up as a merge conflict, but those PRs will be lacking the Nonnull/Nullable

gavinking:
> Every outstanding PR that adds a method will have a soft merge conflict with this. Nothing will show up as a merge conflict, but those PRs will be lacking the Nonnull/Nullable

Sure, but nothing a robotic review won't quickly catch.

api/src/main/java/jakarta/data/Sort.java
```diff
@@ -213,25 +215,29 @@ public Sort(String property, boolean isAscending, boolean ignoreCase) {
...
      * @return The attribute name to order by, or {@code null} if this
-     *         {@code Sort} instance pertains to a {@link #property()}
+     *         {@code Sort} instance was constructed without an expression.
      */
+    @Nonnull
     public ComparableExpression<T, ? extends Comparable<?>> expression() {
+        if ( expression == null ) {
+            throw new IllegalStateException(
+                    Messages.get("013.no-expression"));
+        }
```
gavinking, line 229: 👀

gavinking:
I have updated to resolve the conflicts.

otaviojava:
@gavinking could you solve this conflict, please?

otaviojava (approved):

gavinking:
> @gavinking could you solve this conflict, please?

Done.

otaviojava:
Given the huge amount of code, I will merge to avoid any new conflicts.

PR #1497, opened 2026-06-24
TCK tests for sorting based on expressions

njr-11:
TCK tests for sorting based on expressions:

- sorting based on a navigable expression that allows access to an attribute of an embeddable
- sorting based on a single expression that performs an arithmetic computation involving two entity attributes
- sorting based on multiple sort expressions


tck/src/main/java/ee/jakarta/tck/data/standalone/entity/ExpressionTests.java
```diff
@@ -443,6 +450,167 @@ public void testRight3() {
...
+                             "AZ: Azerbaijan"),
+                found.stream()
+                     .map(c -> c.getCode() + ": " + c.getName())
+                     .toList());
+    }
+
+    @Assertion(id = "829", strategy = """
+            Sort based on a navigable expression.
+            """)
+    public void testSortByNavigableExpression() {
```
otaviojava: how about use: shouldBe...

otaviojava:
How about we move it in either as Nested classes?

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/ExpressionTests.java
njr-11: > how about use: shouldBe...

I'd rather not change just this one test. Hundreds of other tests in the TCK all start with `test`, including all of the other test methods within this class.

njr-11: If we decide to rename tests from `test*` to `shouldBe*`, let's do that in a separate PR that can cover all occurrences.

njr-11:
> How about we move it in either as Nested classes?

I don't understand what your question means.  Are you asking me to move the 3 new test methods that cover sorting by expressions into their own class, and you are suggesting that to be a nested class?

njr-11:
I added commit https://github.com/jakartaee/data/pull/1497/changes/e37b1164ff51f19677574a64c5c17620ff328631 to move the new tests to a class of their own.  Is that better?

otaviojava (approved):

PR #1498, opened 2026-06-25
Ignore expression when asserting that sorts are equal

marko-bekhta:
```java
  Expected :Sort[expression=null, property=hexadecimal, isAscending=true, ignoreCase=true, nullOrdering=UNSPECIFIED]
  Actual   :Sort[expression=TextAttributeRecord[name=hexadecimal], property=hexadecimal, isAscending=true, ignoreCase=true, nullOrdering=UNSPECIFIED]
```

Since string-based sorts don't get an expression ... but the ones created from attributes do ... let's just not expect them to match ?

cc: @njr-11

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/EntityTests.java
@@ -126,6 +126,13 @@ public class EntityTests {
 
     private final DatabaseType type = TestProperty.databaseType.getDatabaseType();
 
+    private static void assertSortEqualsIgnoreExpression(Sort<?> expected, Sort<?> actual) {
+        assertEquals(expected.property(), actual.property(), "Sort property");
+        assertEquals(expected.isAscending(), actual.isAscending(), "Sort isAscending");
+        assertEquals(expected.ignoreCase(), actual.ignoreCase(), "Sort ignoreCase");
+        assertEquals(expected.nullOrdering(), actual.nullOrdering(), "Sort nullOrdering");
+    }
+
     @Assertion(id = "136", strategy = "Ensures that the prepopulation step for readonly entities was successful")
     public void ensureNaturalNumberPrepopulation() {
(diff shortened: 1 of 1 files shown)


njr-11 (changes requested):
> Since string-based sorts don't get an expression ... but the ones created from attributes do ... let's just not expect them to match ?

It looks like you are recommending to have the TCK overlook the change in behavior by altering the test case.  Even if the test case is made to conveniently not notice, Data 1.1 would still be introducing a behavior change vs Data 1.0.  We cannot do that.

marko-bekhta:
I mean ... 
https://github.com/jakartaee/data/blob/60413f550ee4675977f62c36a3d90515c6d856a8/api/src/main/java/jakarta/data/expression/TextExpression.java#L56-L58

https://github.com/jakartaee/data/blob/f37ba8e32875c56009f6e508e75e65fa46d5d81c/api/src/main/java/jakarta/data/Sort.java#L393-L395

it's Data itself that broke it ?

njr-11:
> it's Data itself that broke it ?

Yes, and I was the one who broke it due to a mistake in one of my changes.  I have a correction to what I broke in #1496 which has not been merged yet.  We should be fixing the break rather than altering the TCK to overlook it.

marko-bekhta:
ah 🙂 so you already were working on a fix! Let's close this one then.
(note: we are currently running the snapshot version of this TCK as part of the Hibernate ORM build, hence we now notice such kind of failures in the build. Btw let me know if you'd want running that TCK integration as part of verifying your changes 🙂)

njr-11:
> (note: we are currently running the snapshot version of this TCK as part of the Hibernate ORM build, hence we now notice such kind of failures in the build. Btw let me know if you'd want running that TCK integration as part of verifying your changes 🙂)

I did already test #1496 by verifying the TCK passes again, but you are certainly welcome to test it with Hibernate.  The only reason the fix wasn't merged is that it still needs a review by a Jakarta Data committer.

PR #1506, opened 2026-07-14
Increase Test Coverage

otaviojava:
This pull request adds comprehensive unit tests for the current date, time, and date-time expression classes in the `jakarta.data.spi.expression.function` package. These tests verify that the expression objects are correctly created, expose the expected types, have the right string representations, and ensure that each invocation returns a new instance.

**New unit tests for expression classes:**

*CurrentDate expression:*
- Added `CurrentDateTest` to verify creation, type exposure as `LocalDate`, string representation as `"LOCAL DATE"`, and that each call to `CurrentDate.now()` returns a new instance.

*CurrentDateTime expression:*
- Added `CurrentDateTimeTest` to check creation, type exposure as `LocalDateTime`, string representation as `"LOCAL DATETIME"`, and that each call to `CurrentDateTime.now()` returns a new instance.

*CurrentTime expression:*
- Added `CurrentTimeTest` to test creation, type exposure as `LocalTime`, string representation as `"LOCAL TIME"`, and that each call to `CurrentTime.now()` returns a new instance.


api/src/test/java/jakarta/data/spi/expression/function/CurrentTimeTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+        @Test
+        @DisplayName("should create a new expression for each invocation")
+        void shouldCreateNewExpressionForEachInvocation() {
+            CurrentTime<TestEntity> firstExpression = CurrentTime.now();
+            CurrentTime<TestEntity> secondExpression = CurrentTime.now();
+            assertThat(firstExpression).isNotSameAs(secondExpression);
+        }
+    }
+
```
njr-11: That is how the implementation I put there originally behaved, but it is currently being considered whether a single instance should be reused. See #1467

```suggestion
```

api/src/test/java/jakarta/data/spi/expression/function/CurrentDateTimeTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+        @Test
+        @DisplayName("should create a new expression for each invocation")
+        void shouldCreateNewExpressionForEachInvocation() {
+            var firstExpression = CurrentDateTime.now();
+            var secondExpression = CurrentDateTime.now();
+            assertThat(firstExpression).isNotSameAs(secondExpression);
+        }
+    }
+
```
njr-11: That is how the implementation I put there originally behaved, but it is currently being considered whether a single instance should be reused. See #1467 

```suggestion
```

api/src/test/java/jakarta/data/spi/expression/function/CurrentDateTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+        @Test
+        @DisplayName("should create an independent expression for each invocation")
+        void shouldCreateIndependentExpressionForEachInvocation() {
+            CurrentDate<TestEntity> firstExpression = CurrentDate.now();
+            CurrentDate<TestEntity> secondExpression = CurrentDate.now();
+            assertThat(firstExpression).isNotSameAs(secondExpression);
+        }
+    }
+
```
njr-11, line 70: That is how the implementation I put there originally behaved, but it is currently being considered whether a single instance should be reused. See #1467

```suggestion
```

njr-11 (approved):

PR #1507, opened 2026-07-15
Add Jakarta Annotations security annotation support to spec and TCK

### LINKED ISSUE #1293: [Use Case]: Inheritance of annotations beyond Jakarta interceptor bindings
### As a ...

### I need to be able to ...

... apply annotations to my repository interfaces, and have them taken into account in repository implementations even if the annotations are not CDI interceptor bindings.

### Which enables me to ...

* secure access to Jakarta Data repositories according to roles/permissions defined by `jakarta.security` annotations
* \<put the effect of any method-level annotation you can think of here>

### Additional information

Section 7.2 of the spec states:

> A repository interface or method of a repository interface may be annotated with an interceptor binding annotation. In the Jakarta EE environment—​or in any other environment where Jakarta Interceptors [[6](https://jakarta.ee/specifications/data/1.1/jakarta-data-1.1.0-m1#_footnotedef_6)] is available and integrated with Jakarta CDI—​if the repository implementation is instantiated by the CDI bean container then the interceptor binding annotation is inherited by the repository implementation. That is, the interceptor binding annotation must be treated as if it were placed directly on the repository implementation bean. The interceptors bound to the annotation are applied automatically by the implementation of Jakarta Interceptors.

This is a very useful feature, but unfortunately limited to interceptor binding annotations.

There are more annotations that could be applied to repositories, and that would make sense to be "inherited" -- copied, really -- by the Jakarta Repository implementation.

One example, and I want to stress that even if it's arguable, it's just an example among others, is Jakarta Security annotations. Frameworks would likely comply with `jakarta.annotation.security.RolesAllowed` through some kind of interceptors, perhaps even CDI interceptors, but the annotation itself is not an interceptor binding -- it's not meta-annotated with `@jakarta.interceptor.InterceptorBinding`

We could say that "security annotation should be annotated with `@Inherited`", but we would be wrong -- from https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/annotation/Inherited.html: "this meta-annotation only causes annotations to be inherited from superclasses; annotations on implemented interfaces have no effect". So, not a good match for Jakarta Data repositories.

---

Could we extend the definition of which annotations should be copied from the interface to the implementation?
Perhaps list Jakarta Security annotations?
Maybe decide that `@Inherited` annotations will be copied to implementations, thereby fixing the weird "official" definition of `@Inherited` in the JDK?

Alternatively -- or perhaps additionally -- could you confirm that the processor may chose to copy more annotations than what's mentioned in the spec? In which case this problem could be shifted to implementations.

yrodiere:
Specify that some security annotations from the `jakarta.annotation.security` package (`@DenyAll`,` @PermitAll`, `@RolesAllowed`) on repository interfaces must be treated as if placed directly on the repository implementation, analogously to interceptor bindings.

Add tests to the "web" TCK -- since that's the kind of environment where security annotations must be handled.
Note this involves a CDI bean implementing a custom `TestSecurityContext` interface, to let tests configure caller roles; I'm not entirely sure if we want to add that kind of requirement to the TCK, but I'm sure someone will let me know :)

* Fixes #1293

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -1430,8 +1430,9 @@ import java.util.Set;
  *
  * <h2>Jakarta Annotations</h2>
  *
- * <p>A repository interface or method of a repository interface may be annotated with a
- * security annotation defined in the {@code jakarta.annotation.security} package:
+ * <p>A repository interface or method of a repository interface may be
+ * annotated with one of the following annotations defined in the
+ * {@code jakarta.annotation.security} package:
  * {@code @DeclareRoles}, {@code @DenyAll}, {@code @PermitAll}, {@code @RolesAllowed},
  * or {@code @RunAs}.
  * In the Jakarta EE environment, or in any other environment where Jakarta Annotations
@@ -1440,12 +1441,6 @@ import java.util.Set;
  * as if it were placed directly on the repository implementation bean. The security
  * constraints declared by the annotation are enforced automatically by the Jakarta EE
  * container.</p>
- *
- * <p>The {@code jakarta.annotation.security} annotations are defined by the Jakarta
- * Annotations specification, but their processing semantics are defined by other
- * specifications such as Jakarta Enterprise Beans and Jakarta Servlet. Depending on
- * the Jakarta EE container, some of these annotations—in particular
- * {@code @DeclareRoles} and {@code @RunAs}—may not be supported on CDI beans.</p>
  */
 //TODO switch @code to @link for jakarta.persistence.query.* once it is determined
 // why Javadoc links to API from other Jakarta spec modules are not working
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -1432,7 +1432,8 @@ import java.util.Set;
  *
  * <p>A repository interface or method of a repository interface may be annotated with a
  * security annotation defined in the {@code jakarta.annotation.security} package:
- * {@code @DenyAll}, {@code @PermitAll}, or {@code @RolesAllowed}.
+ * {@code @DeclareRoles}, {@code @DenyAll}, {@code @PermitAll}, {@code @RolesAllowed},
+ * or {@code @RunAs}.
  * In the Jakarta EE environment, or in any other environment where Jakarta Annotations
  * is available and integrated with Jakarta CDI, the security annotation is inherited
  * by the repository implementation. That is, the security annotation must be treated
@@ -1440,9 +1441,11 @@ import java.util.Set;
  * constraints declared by the annotation are enforced automatically by the Jakarta EE
  * container.</p>
  *
- * <p>The behavior of {@code @DeclareRoles} and {@code @RunAs} on a repository interface
- * is not defined by this specification, and is not portable between Jakarta Data
- * providers.</p>
+ * <p>The {@code jakarta.annotation.security} annotations are defined by the Jakarta
+ * Annotations specification, but their processing semantics are defined by other
+ * specifications such as Jakarta Enterprise Beans and Jakarta Servlet. Depending on
+ * the Jakarta EE container, some of these annotations—in particular
+ * {@code @DeclareRoles} and {@code @RunAs}—may not be supported on CDI beans.</p>
  */
 //TODO switch @code to @link for jakarta.persistence.query.* once it is determined
 // why Javadoc links to API from other Jakarta spec modules are not working
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -1427,6 +1427,22 @@ import java.util.Set;
  *     CompletionStage<Integer> setPriceAsync(float newPrice, Long productId);
  * }
  * }</pre>
+ *
+ * <h2>Jakarta Annotations</h2>
+ *
+ * <p>A repository interface or method of a repository interface may be annotated with a
+ * security annotation defined in the {@code jakarta.annotation.security} package:
+ * {@code @DenyAll}, {@code @PermitAll}, or {@code @RolesAllowed}.
+ * In the Jakarta EE environment, or in any other environment where Jakarta Annotations
+ * is available and integrated with Jakarta CDI, the security annotation is inherited
+ * by the repository implementation. That is, the security annotation must be treated
+ * as if it were placed directly on the repository implementation bean. The security
+ * constraints declared by the annotation are enforced automatically by the Jakarta EE
+ * container.</p>
+ *
+ * <p>The behavior of {@code @DeclareRoles} and {@code @RunAs} on a repository interface
+ * is not defined by this specification, and is not portable between Jakarta Data
+ * providers.</p>
  */
 //TODO switch @code to @link for jakarta.persistence.query.* once it is determined
 // why Javadoc links to API from other Jakarta spec modules are not working
--- a/spec/src/main/asciidoc/jakarta-ee.adoc
+++ b/spec/src/main/asciidoc/jakarta-ee.adoc
@@ -221,10 +221,10 @@ public interface School extends DataRepository<Student, String> {
 
 === Jakarta Annotations
 
-A repository interface or method of a repository interface may be annotated with a security annotation defined in the `jakarta.annotation.security` package: `@DenyAll`, `@PermitAll`, or `@RolesAllowed`.footnote:[Jakarta Annotations 3.0, https://jakarta.ee/specifications/annotations/3.0/]
+A repository interface or method of a repository interface may be annotated with a security annotation defined in the `jakarta.annotation.security` package: `@DeclareRoles`, `@DenyAll`, `@PermitAll`, `@RolesAllowed`, or `@RunAs`.footnote:[Jakarta Annotations 3.0, https://jakarta.ee/specifications/annotations/3.0/]
 In the Jakarta EE environment--or in any other environment where Jakarta Annotations is available and integrated with Jakarta CDI--the security annotation is inherited by the repository implementation. That is, the security annotation must be treated as if it were placed directly on the repository implementation bean. The security constraints declared by the annotation are enforced automatically by the Jakarta EE container.
 
-NOTE: The behavior of `@DeclareRoles` and `@RunAs` on a repository interface is not defined by this specification, and is not portable between Jakarta Data providers.
+NOTE: The `jakarta.annotation.security` annotations are defined by the Jakarta Annotations specification, but their processing semantics are defined by other specifications such as Jakarta Enterprise Beans and Jakarta Servlet.footnote:[See also https://github.com/jakartaee/security/issues/295[Jakarta Security issue #295\].] Depending on the Jakarta EE container, some of these annotations--in particular `@DeclareRoles` and `@RunAs`--may not be supported on CDI beans.
 
 The following example illustrates how security annotations might be applied to a repository:
 
--- a/spec/src/main/asciidoc/jakarta-ee.adoc
+++ b/spec/src/main/asciidoc/jakarta-ee.adoc
@@ -221,7 +221,7 @@ public interface School extends DataRepository<Student, String> {
 
 === Jakarta Annotations
 
-A repository interface or method of a repository interface may be annotated with a security annotation defined in the `jakarta.annotation.security` package: `@DeclareRoles`, `@DenyAll`, `@PermitAll`, `@RolesAllowed`, or `@RunAs`.footnote:[Jakarta Annotations 3.0, https://jakarta.ee/specifications/annotations/3.0/]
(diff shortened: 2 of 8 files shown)


yrodiere:
Implementation for Hibernate: https://github.com/hibernate/hibernate-orm/pull/13072

api/src/main/java/module-info.java
```diff
@@ -1427,6 +1427,22 @@
  *     CompletionStage<Integer> setPriceAsync(float newPrice, Long productId);
  * }
  * }</pre>
+ *
+ * <h2>Jakarta Annotations</h2>
+ *
+ * <p>A repository interface or method of a repository interface may be annotated with a
+ * security annotation defined in the {@code jakarta.annotation.security} package:
```
njr-11: Clarifies that we are only speaking of these 3 annotations, not the other 2 in the package
```suggestion
 * <p>A repository interface or method of a repository interface may be
 * annotated with one of the following annotations defined in the
 * {@code jakarta.annotation.security} package:
```

api/src/main/java/module-info.java
```diff
@@ -1427,6 +1427,22 @@
...
+ * In the Jakarta EE environment, or in any other environment where Jakarta Annotations
+ * is available and integrated with Jakarta CDI, the security annotation is inherited
+ * by the repository implementation. That is, the security annotation must be treated
+ * as if it were placed directly on the repository implementation bean. The security
+ * constraints declared by the annotation are enforced automatically by the Jakarta EE
+ * container.</p>
+ *
+ * <p>The behavior of {@code @DeclareRoles} and {@code @RunAs} on a repository interface
+ * is not defined by this specification, and is not portable between Jakarta Data
+ * providers.</p>
```
njr-11: This is now unnecessary to say and adds clutter to the Javadoc. We should omit it.
```suggestion
```

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -218,3 +218,28 @@ public interface School extends DataRepository<Student, String> {
     List<Student> findByAgeLessThanEqual(@Min(18) int age);
 }
 ----
+
+=== Jakarta Annotations
+
+A repository interface or method of a repository interface may be annotated with a security annotation defined in the `jakarta.annotation.security` package: `@DenyAll`, `@PermitAll`, or `@RolesAllowed`.footnote:[Jakarta Annotations 3.0, https://jakarta.ee/specifications/annotations/3.0/]
```
njr-11: Similar to prior comment and update to EE 12 spec version,
```suggestion
A repository interface or method of a repository interface may be annotated one of the following annotations defined in the `jakarta.annotation.security` package: `@DenyAll`, `@PermitAll`, or `@RolesAllowed`.footnote:[Jakarta Annotations 3.1, https://jakarta.ee/specifications/annotations/3.1/]
```

spec/src/main/asciidoc/jakarta-ee.adoc
```diff
@@ -218,3 +218,28 @@ public interface School extends DataRepository<Student, String> {
...
+In the Jakarta EE environment--or in any other environment where Jakarta Annotations is available and integrated with Jakarta CDI--the security annotation is inherited by the repository implementation. That is, the security annotation must be treated as if it were placed directly on the repository implementation bean. The security constraints declared by the annotation are enforced automatically by the Jakarta EE container.
+
+NOTE: The behavior of `@DeclareRoles` and `@RunAs` on a repository interface is not defined by this specification, and is not portable between Jakarta Data providers.
```
njr-11: If you want to keep this text in the spec doc, that's fine. I would also be okay with removing it.  If it is kept here, there is one grammar fix to make,
```suggestion
NOTE: The behavior of `@DeclareRoles` and `@RunAs` on a repository interface is not defined by this specification and is not portable between Jakarta Data providers.
```

tck/src/main/java/ee/jakarta/tck/data/web/annotations/SecuredProducts.java
```diff
@@ -0,0 +1,41 @@
...
+public interface SecuredProducts {
+
+    @Find
+    @DenyAll
+    SecuredProduct findByProductId(String productId);
+
+    @Find
+    @PermitAll
+    List<SecuredProduct> findAll();
+
```
njr-11, line 37: In order to run tests, I think you will need a way to populate entities into the database so that you can try to find and remove them.  Individual tests will need to be updated as well.

```suggestion
    @Insert
    @RolesAllowed({"manager", "admin"})
    void add(SecuredProduct product);

```

njr-11 (commented):
Overall this looks good. I added a few suggested updates.  I don't see anywhere that this is adding entities to the database that the TCK tests that find/remove will be able to operate on.  There will also need to be some instructions in the TCK about the requirements for the TCK runner to set up roles and so forth.  This might be something to talk with @KyleAure about once he returns to the office

api/src/main/java/module-info.java
gavinking: Just out of curiosity, why precisely are we excluding `@RunAs`? Does it make no sense to have `@RunAs("DatabaseClient")` on a repository?

tck/src/main/java/ee/jakarta/tck/data/web/annotations/SecuredProducts.java
yrodiere: I changed the tests to use `@Insert` instead of `@Delete`, which makes things simpler.

spec/src/main/asciidoc/jakarta-ee.adoc
yrodiere: I had to change it for other reasons... will add a comment below.

yrodiere:
So. I was convinced only `@RolesAllowed`/`@DenyAll`/`@PermitAll` were supported on CDI bean for some reason, which is why I had excluded the other two annotations.

But the sad truth is that outside of Quarkus... none of these annotations are supported on CDI beans -- only on EJBs. Supporting them on CDI beans is planned in the Security spec, though; see https://github.com/jakartaee/security/issues/295

So, I changed the approach: we simply state the annotations must behave the same way they would behave on CDI beans. If the container doesn't support that, so be it. If it does, even better. In either case: using these annotations on repositories won't lead to any surprise, which was the goal.

WDYT?

tck/src/main/java/ee/jakarta/tck/data/web/annotations/AnnotationsSecurityTests.java
```diff
@@ -0,0 +1,119 @@
...
+            that the product was added.
+            """)
+    public void testRolesAllowedAllowsWithMatchingRole() {
+        int before = products.findAll().size();
+
+        securityContext.setCallerRoles("admin");
+        products.add(new SecuredProduct("P888", "Gizmo"));
+
+        securityContext.clearCallerRoles();
+        assertTrue(products.findAll().size() > before);
```
njr-11: A more precise assertion would be:
```suggestion
        assertEquals(before + 1, products.findAll().size());
```

tck/src/main/java/ee/jakarta/tck/data/web/annotations/AnnotationsSecurityTests.java
```diff
@@ -0,0 +1,119 @@
...
+            Verify that a repository method annotated with @RolesAllowed
+            allows access when the caller has one of the required roles.
+            The insert must succeed, and findAll must confirm
+            that the product was added.
+            """)
+    public void testRolesAllowedAllowsWithMatchingRole() {
+        int before = products.findAll().size();
+
+        securityContext.setCallerRoles("admin");
+        products.add(new SecuredProduct("P888", "Gizmo"));
```
njr-11, line 117: Need to account for NoSQL databases that use eventual consistency,

```suggestion
        products.add(new SecuredProduct("P888", "Gizmo"));

        TestPropertyUtility.waitForEventualConsistency();
```

njr-11 (commented):
I added a couple of review comments on the tests. My prior comment about needing to document requirements for the TCK test runner also still applies.

yrodiere:
Thanks, I addressed your comments.

Also added instructions regarding TCK runners, but... considering that:

> But the sad truth is that outside of Quarkus... none of these annotations are supported on CDI beans -- only on EJBs. Supporting them on CDI beans is planned in the Security spec, though; see [jakartaee/security#295](https://github.com/jakartaee/security/issues/295)
> 
> So, I changed the approach: we simply state the annotations must behave the same way they would behave on CDI beans. If the container doesn't support that, so be it. If it does, even better. In either case: using these annotations on repositories won't lead to any surprise, which was the goal.

I'm not sure there is anything we can test. The requirement is behavior consistent with a CDI bean -- not necessarily enforcement.

njr-11:
> I'm not sure there is anything we can test. The requirement is behavior consistent with a CDI bean -- not necessarily enforcement.

We have been experimenting with this in Open Liberty.  All we are able to test for now is to assert that the annotation is copied to the respective bean method for method level annotations. For DenyAll, PermitAll, and RolesAllowed annotations at class level we can assert that the annotation is either copied to the class or method, although we haven't actually been able to get Weld to copy annotations to the repository bean class.  If that remains a problem, I don't see how we can include RunAs and DeclareRoles, which are permitted at TYPE level only.

yrodiere:
> > I'm not sure there is anything we can test. The requirement is behavior consistent with a CDI bean -- not necessarily enforcement.
> 
> We have been experimenting with this in Open Liberty. All we are able to test for now is to assert that the annotation is copied to the respective bean method for method level annotations. For DenyAll, PermitAll, and RolesAllowed annotations at class level we can assert that the annotation is either copied to the class or method, although we haven't actually been able to get Weld to copy annotations to the repository bean class. If that remains a problem, I don't see how we can include RunAs and DeclareRoles, which are permitted at TYPE level only.

Fair enough -- IMO those are the least useful annotations anyway. Please let me know.

Regarding testing though, I want to point out that the requirement is not to copy annotations, it's to have the container behave 
"as if" the implementation was annotated, because the interface is annotated. It's entirely possible for the annotations to not actually exist on the implementation. In which case (back to my original statement) we can't test for annotation presence in the TCK.

njr-11:
Right, I don't mean to say that the TCK can take the same approach as a provider's own tests which can test aspects of its own implementation.  Given that Jakarta Security doesn't yet have a requirement for these annotations for CDI beans, the TCK cannot expect to enforce anything regarding the annotations.  Prior to that, we should really have the TCK tests for this disabled or commented out.  Also, re-reading the issue in the Jakarta Security spec, it seems like we are getting too far ahead of it in predicting what another specification is going to do.  Although the security annotations in the common annotations spec seems to be a preference, there is also  some discussion in that issue about possibly using different annotations or a combination.  Unless Jakarta Security commits to including something specific in Jakarta EE 12, we should probably wait on this altogether until we know what the outcome will be.

yrodiere:
I'm worried that this "wait" will take years, during which people will continue to add security annotations to their repositories and discovering rather late that they are being ignored.

But I agree there's only so much we can do on the Data side -- if this is not seen as a pressing issue in the Security spec, even if it applies to *every single CDI bean out there*... so be it.

---

That said, this problem is worked around in Quarkus, so I guess my main *personal* concern is addressed and this would mostly have been a way to move the hack to a proper implementation in Hibernate ORM.

Going by [this message](https://github.com/jakartaee/data/issues/1293#issuecomment-3564620436) (quoted below), I suppose I could just go ahead and merge hibernate/hibernate-orm#13072 without waiting for the spec?

> > Alternatively -- or perhaps additionally -- could you confirm that the processor may chose to copy more annotations than what's mentioned in the spec? In which case this problem could be shifted to implementations.
> 
> Implementations are always free to add their own vendor-specific behaviors as long as they don't conflict with the specification. And even if there were a conflict, an implementation is still free to add the behavior as non-default with configuration to enable it. That said, it would be better to standardize a solution in Jakarta EE then rely on vendor behavior.

njr-11:
> Going by [this message](https://github.com/jakartaee/data/issues/1293#issuecomment-3564620436) (quoted below), I suppose I could just go ahead and merge [hibernate/hibernate-orm#13072](https://github.com/hibernate/hibernate-orm/pull/13072) without waiting for the spec?

I don't see anything here that is contrary to the specification, so yes, you could go ahead and add it as vendor-specific function for now.

After experimentation that we did, I'm not comfortable with defining this in the Jakarta Data spec in advance of Jakarta Security saying how it will work.  We have run into issues with Weld bean proxies/interceptor proxies not copying over the class level annotation.  On the specification side, I think we should wait and see whether Jakarta Security makes these annotations into interceptor bindings (which would help) or takes another approach like defining different annotations.

yrodiere:
Okay. Thanks for looking into this.

I will proceed in Hibernate ORM.

Marking this PR as draft for clarity.

KyleAure:
Closing based on the discussion on the parent issue.

PR #1513, opened 2026-07-16
Fix cursors check when no results are present

### LINKED ISSUE #1512: TCK testCursoredPageOfNothing fails after CursoredPageRecord API change
The TCK test `EntityTests.testCursoredPageOfNothing()` fails after the API change introduced in https://github.com/jakartaee/data/commit/66f5a7efdd66.

The test still expects a `NoSuchElementException`:

```java
try {
    page.nextPageRequest();
    fail("nextPageRequest must raise NoSuchElementException when current page is empty.");
} catch (NoSuchElementException x) {
    // expected
}
```

Whereas the new implementation throws `UnsupportedOperationException` when the `cursors.isEmpty()` condition is true.

mbellade:
When https://github.com/jakartaee/data/commit/66f5a7efdd66 introduced the new `UnsupportedOperationException`, it only checked that `cursors.isEmpty()` as a signal of uncomputable sort-criterias. Queries returning no results though will also produce no cursors in the first place, so this fix should restore the expected exception order (and the failing test reported in the issue).

* Fixes https://github.com/jakartaee/data/issues/1512

### DIFF
--- a/api/src/main/java/jakarta/data/page/impl/CursoredPageRecord.java
+++ b/api/src/main/java/jakarta/data/page/impl/CursoredPageRecord.java
@@ -123,21 +123,21 @@ public record CursoredPageRecord<T>
 
     @Override
     public PageRequest nextPageRequest() {
+        if (nextPageRequest == null)
+            throw new NoSuchElementException();
         if (cursors.isEmpty())
             throw new UnsupportedOperationException(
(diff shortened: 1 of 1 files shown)


njr-11 (approved):
Thanks for fixing!

njr-11:
Merging right away because this corrects a breaking change that I accidentally made

PR #1515, opened 2026-07-16
Review comments from tech writer on recently added messages

njr-11:
When we pulled in the latest spec API to Open Liberty, one of our tech writers reviewed the recently added (013, 014, and 015) messages and recommended changes which I have copied to this PR.

### DIFF
--- a/api/src/main/java/jakarta/data/Sort.java
+++ b/api/src/main/java/jakarta/data/Sort.java
@@ -218,15 +218,11 @@ public record Sort<T>(@Nullable ComparableExpression<T, ? extends Comparable<?>>
      * An expression to order by, or {@code null} if this {@code Sort}
      * instance was constructed without an expression.
      *
-     * @return The attribute name to order by, or {@code null} if this
+     * @return The expression to order by, or {@code null} if this
      *         {@code Sort} instance was constructed without an expression.
      */
-    @Nonnull
+    @Nullable
     public ComparableExpression<T, ? extends Comparable<?>> expression() {
-        if ( expression == null ) {
-            throw new IllegalStateException(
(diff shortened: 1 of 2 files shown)


gavinking (approved):

otaviojava (approved):

otaviojava:
@njr-11 could you rebase it, please?

njr-11:
> @njr-11 could you rebase it, please?

Yes, I'll rebase it, although the real problem is that another PR added a soft merge conflict adding a second message also numbered 013.  That soft conflict was with the existing codebase, not this PR.  I'll plan to correct this as well.

njr-11:
> > @njr-11 could you rebase it, please?
> 
> Yes, I'll rebase it, although the real problem is that another PR added a soft merge conflict adding a second message also numbered 013. That soft conflict was with the existing codebase, not this PR. I'll plan to correct this as well.

I found where it happened. It's in https://github.com/jakartaee/data/commit/a84450fe8fb8ddb85eb7841ada3fc8f943df0c6d#diff-b8648338e503d9bb49fd01e556069418661c6ee7f0e816c71e0364481530c432
which added the following exception to Sort.expression(),
```java
        if ( expression == null ) {
            throw new IllegalStateException(
                    Messages.get("013.no-expression"));
        }
```
despite Javadoc which says that Sort.expression() returns `{@code null} if this {@code Sort} instance was constructed without an expression.`

It seems like a Data provider might want to invoke Sort.expression() on main path code to find out whether or not there is an expression (vs an entity attribute), so what the Javadoc says seems preferable.  In that case, we don't need the message at all, and I'll remove it after updating the expression() method accordingly.

PR #1526, opened 2026-08-05
add missing @Nonnull annotations

### LINKED ISSUE #1493: Nullability annotations

gavinking:
For #1493.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/_AsciiCharacter.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/framework/read/only/_AsciiCharacter.java
@@ -59,11 +59,11 @@ public class _AsciiCharacter {
 
 
 
-    private record TextAttr(String name,
+    private record TextAttr(@Nonnull String name,
                             @Nonnull Sort<AsciiCharacter> asc,
-                            Sort<AsciiCharacter> ascIgnoreCase,
+                            @Nonnull Sort<AsciiCharacter> ascIgnoreCase,
                             @Nonnull Sort<AsciiCharacter> desc,
-                            Sort<AsciiCharacter> descIgnoreCase)
+                            @Nonnull Sort<AsciiCharacter> descIgnoreCase)
             implements TextAttribute<AsciiCharacter> {
         private TextAttr(String name) {
             this(name, Sort.asc(name), Sort.ascIgnoreCase(name), Sort.desc(name), Sort.descIgnoreCase(name));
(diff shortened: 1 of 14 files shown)


tck/src/main/java/ee/jakarta/tck/data/framework/read/only/_AsciiCharacter.java
```diff
@@ -58,10 +59,12 @@ private NumericAttr(String name) {
...
 
-    private record TextAttr(String name, Sort<AsciiCharacter> asc,
-                                   Sort<AsciiCharacter> ascIgnoreCase,
-                                   Sort<AsciiCharacter> desc,
-                                   Sort<AsciiCharacter> descIgnoreCase) implements TextAttribute<AsciiCharacter> {
+    private record TextAttr(String name,
+                            @Nonnull Sort<AsciiCharacter> asc,
+                            Sort<AsciiCharacter> ascIgnoreCase,
+                            @Nonnull Sort<AsciiCharacter> desc,
+                            Sort<AsciiCharacter> descIgnoreCase)
```
njr-11: I'm not sure if you intended to update this TCK class, but if we do, let's make it consistent across record components given that none of them can be null.
```suggestion
    private record TextAttr(@Nonnull String name,
                            @Nonnull Sort<AsciiCharacter> asc,
                            @Nonnull Sort<AsciiCharacter> ascIgnoreCase,
                            @Nonnull Sort<AsciiCharacter> desc,
                            @Nonnull Sort<AsciiCharacter> descIgnoreCase)
```

otaviojava (approved):

tck/src/main/java/ee/jakarta/tck/data/framework/read/only/_AsciiCharacter.java
gavinking: I did intend it, since without those annotations, IntelliJ shows me warnings on the interface it implements. I have applied your suggestion.

njr-11 (approved):

PR #1531, opened 2026-08-18
Life cycle events do not enforce nonnull

njr-11:
LifeCycleEvent is accepting a null entity value and returns the null value from `entity()` despite the NonNull annotations on both.  This PR adds enforcement to prevent a null entity value from being accepted.


gavinking (approved):
LGTM

PR #1534, opened 2026-08-19
Double insert in TCK test results in EntityExistsException

njr-11:
Found a TCK bug where repeated insert of the same entities results in EntityExistsException.  It looks like the insert of the entities was only intended to be done once.


otaviojava (approved):

otaviojava:
Thanks @njr-11

PR #1535, opened 2026-08-19
add missing long pageOffset() getter

gavinking:
We are missing a getter for the page offset.

### DIFF
--- a/api/src/main/java/jakarta/data/page/PageRequest.java
+++ b/api/src/main/java/jakarta/data/page/PageRequest.java
@@ -251,6 +251,7 @@ public interface PageRequest {
      * @return the requested page number
      * @apiNote Page <em>numbers</em> are indexed from one;
      *          page {@linkplain #pageOffset() offsets} are indexed from zero.
+     * @since 1.1
      * @see #pageNumber(long)
      */
     long pageNumber();
@@ -261,6 +262,7 @@ public interface PageRequest {
      * @return the requested page offset
      * @apiNote Page <em>offsets</em> are indexed from zero;
      *          page {@linkplain #pageNumber() numbers} are indexed from one.
+     * @since 1.1
      * @see #pageOffset(long)
      */
     default long pageOffset() {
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/data/page/PageRequest.java
```diff
@@ -250,10 +250,21 @@ default long page() {
...
      * @return the requested page number
      * @apiNote Page <em>numbers</em> are indexed from one;
-     *          page {@linkplain #pageOffset offsets} are indexed from zero.
+     *          page {@linkplain #pageOffset() offsets} are indexed from zero.
+     * @see #pageNumber(long)
      */
     long pageNumber();
 
+    /**
+     * Returns the requested page offset. Page numbers begin with {@code 0}.
```
njr-11: The first sentence has a copy/paste error which likely came from the pageNumber Javadoc.  The minimum offset is 0, not 1, because it is relative to the first page.
```suggestion
     * Returns the offset of the requested page relative to the first page.
     * Page offsets begin with {@code 0}.
```

api/src/main/java/jakarta/data/page/PageRequest.java
```diff
@@ -250,10 +250,21 @@ default long page() {
...
+     *
+     * @return the requested page offset
+     * @apiNote Page <em>offsets</em> are indexed from zero;
```
njr-11, line 264:
```suggestion
     * @since 1.1
     * @apiNote Page <em>offsets</em> are indexed from zero;
```

njr-11 (commented):
I actually left this out intentionally, thinking that it would only be needed by the Data provider, and the Data provider can easily subtract one from the pageNumber().  If we expect this to be helpful for users as well, here are a few review comments,

gavinking:
> If we expect this to be helpful for users as well

I think it's certainly useful to clients. It's also more symmetric.

api/src/main/java/jakarta/data/page/PageRequest.java
gavinking: Yeah, this was also missing from `pageNumber()`, which is also new.

gavinking:
@njr-11 fixed those two things

otaviojava (approved):

PR #1536, opened 2026-08-20
doc: update spec license to EFSL-2.0

KyleAure:
As directed in Platform mailing list: https://www.eclipse.org/lists/jakartaee-platform-dev/msg04872.html


otaviojava:
I believe we need to update the header of the Java file, right?

otaviojava (approved):

PR #1541, opened 2026-08-21
Javadoc for NavigableExpression and Path

njr-11:
Add Javadoc for NavigableExpression, Path, and one of the Path subclasses.  After getting that correct, the same can be applied to the other Path subclasses

### DIFF
--- a/api/src/main/java/jakarta/data/expression/NavigableExpression.java
+++ b/api/src/main/java/jakarta/data/expression/NavigableExpression.java
@@ -148,7 +148,7 @@ import jakarta.annotation.Nonnull;
 public interface NavigableExpression<T, U> {
 
     /**
-     * Represents navigation to an attribute that can be further navigated.
+     * Allows navigation to an attribute that can be further navigated.
      *
      * <p>The resulting {@link NavigableExpression} represents the path formed
      * by navigating from this expression to the given
--- a/api/src/main/java/jakarta/data/expression/NavigableExpression.java
+++ b/api/src/main/java/jakarta/data/expression/NavigableExpression.java
@@ -56,8 +56,8 @@ import jakarta.annotation.Nonnull;
  * }</pre>
  *
  * <p>
- * The following entity and embeddables are used in examples throughout the
- * documentation of this class.
+ * The following Jakarta Persistence entity and embeddable types are used in examples 
+ * throughout the documentation of this class.
  *
  * <pre>{@code
  * @Entity
--- a/api/src/main/java/jakarta/data/expression/NavigableExpression.java
+++ b/api/src/main/java/jakarta/data/expression/NavigableExpression.java
@@ -33,36 +33,279 @@ import jakarta.data.spi.expression.path.TemporalPath;
 import jakarta.data.spi.expression.path.TextPath;
 import jakarta.annotation.Nonnull;
 
+/**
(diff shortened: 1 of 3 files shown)


api/src/main/java/jakarta/data/expression/NavigableExpression.java
```diff
@@ -33,36 +33,279 @@
 import jakarta.data.spi.expression.path.TextPath;
 import jakarta.annotation.Nonnull;
 
+/**
+ * An {@linkplain Expression expression} that navigates to an attribute of an
+ * embeddable or association.
```
gavinking: Not sure this is the right way to say it.

- I don't think we define the term "embeddable" anywhere in Jakarta Data
- it's not the expression itself that navigates an attribute or association; the expression represents a type which allows such navigation

api/src/main/java/jakarta/data/expression/NavigableExpression.java
```diff
@@ -33,36 +33,279 @@
...
+ *
+ * <p>
+ * The {@code navigate} methods accept a metamodel attribute of an intermediate
+ * type {@code U} and return a new expression of a more specific subtype,
```
gavinking: I don't think "more specific subtype" is quite correct.

api/src/main/java/jakarta/data/expression/NavigableExpression.java
gavinking: > it's not the expression itself that navigates an attribute or association; the expression represents a type which allows such navigation

well, actually scratch what I said ... I suppose this is actually correct but somehow I still find it confusing....

njr-11: Updated to remove the term embeddable and hopefully be less confusing.

api/src/main/java/jakarta/data/expression/NavigableExpression.java
```diff
@@ -33,36 +33,279 @@
...
+ * <pre>{@code
+ * Restrict.all(_Customer.shippingAddress.navigate(_Address.city)
+ *                                       .equalTo("Rochester"),
+ *              _Customer.shippingAddress.navigate(_Address.state)
+ *                                       .equalTo(StateCode.MN));
+ * }</pre>
+ *
+ * <p>
+ * The following entity and embeddables are used in examples throughout the
+ * documentation of this class.
```
gavinking:
```suggestion
 * The following Jakarta Persistence entity and embeddable types are used in examples 
 * throughout the documentation of this class.
```

api/src/main/java/jakarta/data/expression/NavigableExpression.java
```diff
@@ -33,36 +33,279 @@
...
+ *
+ * @param <T> entity type
+ * @param <U> type of the intermediate object whose attributes can be
+ *            navigated to
+ * @since 1.1
+ */
 public interface NavigableExpression<T, U> {
 
+    /**
+     * Represents navigation to an attribute that can be further navigated.
```
gavinking:
```suggestion
     * Allows navigation to an attribute that can be further navigated.
```

This is merely a suggestion. I personally prefer to reserve the word "represents" for types.

gavinking (approved):
Couple of very minor suggestions. Ignore if you prefer.

PR #1544, opened 2026-08-22
allow navigation to fields of supertypes

gavinking:
Assuming I'm not experiencing a massive brain fart (quite a strong assumption), the `navigate()` method needs to be modified to accept attributes declared by supertypes of the attribute type.


njr-11 (commented):
The changes look good, but I'd like to see an example of what problem it solves.  I realized we didn't sufficiently define how the static metamodel is generated when it comes to inheritance.  If one entity or embeddable inherits from another, does the static metamodel for the first entity include inherited attributes or only the ones it directly declares, with separate static metamodel classes being generated for super types and their respective attributes?  Or do these changes have a different purpose in mind?

gavinking:
>  I realized we didn't sufficiently define how the static metamodel is generated when it comes to inheritance. If one entity or embeddable inherits from another, does the static metamodel for the first entity include inherited attributes or only the ones it directly declares, with separate static metamodel classes being generated for super types and their respective attributes?

Yeah OK, I agree, we should pin this down. My preferred implementation is that the metamodel class for a subclass inherits the metamodel class for the superclass. But I guess we don't say that.

> Or do these changes have a different purpose in mind?

Nope, you nailed it.

njr-11 (approved):

otaviojava:
@gavinking could you rebase this one?

PR #1548, opened 2026-08-25
Define how inherited attributes work in static metamodel

njr-11:
The spec should indicate how static metamodel classes are generated when some attributes are inherited from a superclass rather than being declared directly on the entity/embeddable class.

### DIFF
--- a/api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
+++ b/api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
@@ -32,9 +32,9 @@ import jakarta.data.restrict.Restriction;
  * Annotates a static metamodel class.
  *
  * <p>
- * A static metamodel class represents the declared attributes of an entity
- * class, an association (such as a Jakarta Persistence Embeddable class),
- * or a superclass of an entity or association. The static metamodel enables
+ * A static metamodel class holds a representation of the declared attributes of an entity
+ * class, an associated class (such as a Jakarta Persistence Embeddable class),
+ * or a superclass of an entity or associated class. The static metamodel enables
  * type-safe access to attribute names as well as an {@code *Attribute} object
  * from which is obtained {@link Expression}s, {@link Restriction}s, and
  * {@link Sort}s on the attribute. A metamodel class contains one or more
--- a/api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
+++ b/api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
@@ -25,48 +25,71 @@ import java.lang.annotation.Target;
 import java.util.UUID;
 
 import jakarta.data.Sort;
+import jakarta.data.expression.Expression;
+import jakarta.data.restrict.Restriction;
 
 /**
- * <p>Annotates a class which serves as a static metamodel for an entity,
- * enabling
- * type-safe access to entity attribute names and related objects such as
- * instances of {@link Sort}s for an attribute. A metamodel class contains one
- * or more {@code public static} fields corresponding to attributes of the
- * entity class. The type of each of these fields must be either {@link String},
- * {@link Attribute}, or a subinterface of {@code Attribute} defined in this
- * package.</p>
+ * Annotates a static metamodel class.
  *
- * <p>The following subinterfaces of {@code Attribute} are recommended to
- * obtain
- * the full benefit of the static metamodel:</p>
+ * <p>
+ * A static metamodel class represents the declared attributes of an entity
+ * class, an association (such as a Jakarta Persistence Embeddable class),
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
```diff
@@ -25,48 +25,71 @@
...
- * <p>The following subinterfaces of {@code Attribute} are recommended to
- * obtain
- * the full benefit of the static metamodel:</p>
+ * Annotates a static metamodel class.
+ *
+ * <p>
+ * A static metamodel class represents the declared attributes of an entity
+ * class, an association (such as a Jakarta Persistence Embeddable class),
+ * or a superclass of an entity or association. The static metamodel enables
+ * type-safe access to attribute names as well as an {@code *Attribute} object
```
gavinking: What does `{@code *Attribute}` do? Is that something I don't know about?

api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
```diff
@@ -25,48 +25,71 @@
...
  *
- * <p>Jakarta Data defines the following conventions for static metamodel classes:</p>
+ * <p>Jakarta Data defines the following conventions for static metamodel
+ * classes:
  * <ul>
  * <li>The metamodel class can be an interface or concrete class.</li>
- * <li>The name of the static metamodel class should consist of underscore ({@code _})
- *     followed by the entity class name.</li>
+ * <li>The name of the static metamodel class should consist of underscore
+ *     ({@code _}) followed by the name of the modeled class.</li>
```
gavinking:
```suggestion
 * <li>The name of the static metamodel class is formed by prefixing
 *     the name of the modeled class with an underscore ({@code _}).</li>
```

spec/src/main/asciidoc/entity.asciidoc
```diff
@@ -382,12 +382,14 @@ NOTE: Application programmers are strongly encouraged to follow Java's camel cas
 
 Jakarta Data provides a static metamodel that allows entity attributes to be accessed by applications in a type-safe manner.
 
-For each entity class, the application developer or a compile-time annotation processor can define a corresponding metamodel class following a prescribed set of conventions.
+A static metamodel class represents the declared attributes of an entity class, an association (such as a Jakarta Persistence embeddable class), or a superclass of an entity or association.
```
gavinking:
```suggestion
A static metamodel class holds a representation of the declared attributes of an entity class, an associated class (such as a Jakarta Persistence embeddable class), or a superclass of an entity or associated class.
```

api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
```diff
@@ -25,48 +25,71 @@
...
- *
- * <p>The following subinterfaces of {@code Attribute} are recommended to
- * obtain
- * the full benefit of the static metamodel:</p>
+ * Annotates a static metamodel class.
+ *
+ * <p>
+ * A static metamodel class represents the declared attributes of an entity
+ * class, an association (such as a Jakarta Persistence Embeddable class),
+ * or a superclass of an entity or association. The static metamodel enables
```
gavinking: Suggest something like:

```suggestion
 * A static metamodel class holds a representation of the declared attributes of an entity
 * class, an associated class (such as a Jakarta Persistence Embeddable class),
 * or a superclass of an entity or associated class. The static metamodel enables
```

spec/src/main/asciidoc/entity.asciidoc
```diff
@@ -416,7 +418,8 @@ The fields may be statically initialized, or they may be initialized by the prov
 
 The following are conventions for static metamodel classes:
 
-- The name of the static metamodel class should consist of underscore (`_`) followed by the entity class name.
+- The name of the static metamodel class should consist of underscore (`_`) followed by the name of the modeled class.
```
gavinking:
```suggestion
- The name of the static metamodel class is formed by prefising the name of the modeled class with an underscore (`_`).
```

api/src/main/java/jakarta/data/metamodel/StaticMetamodel.java
njr-11: Thanks for spotting that!  It isn't anything special, just a badly written way of trying to communicate "subclass of Attribute" which happen to all have class names that end with `Attribute`.  I updated it to explicitly state subclass and removed the `*`.

gavinking: Ahah yeah, OK, now I understand.  :)

gavinking (approved):

PR #1556, opened 2026-09-03
Add TCK coverage for exceptional async completion

mhnuk2007:
## Description

Adds TCK coverage for exceptional completion of an asynchronous Jakarta Data repository method.

The new test verifies that when an asynchronous `@Insert` operation fails because an entity with the same identifier already exists, the returned `CompletionStage` completes exceptionally with `EntityExistsException` as the underlying cause.

### Testing

* `mvn clean verify` — successful
* `AsyncTests` against Open Liberty — 3 tests passed
* `testAsynchronousInsertExceptionalCompletion` against Open Liberty — passed
* `git diff --check` — clean

The change is limited to `tck/src/main/java/ee/jakarta/tck/data/web/async/AsyncTests.java`.

Related to #19.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/web/async/AsyncTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/web/async/AsyncTests.java
@@ -207,13 +207,13 @@ public class AsyncTests {
                 LocalDateTime.of(2026, 7, 28, 10, 30, 00),
                 "asyncUser108@eclipse.org");
 
+        CompletionStage<Void> insert;
         try {
-            CompletionStage<Void> insert;
-            try {
-                insert = accounts.add(account);
-            } catch (UnsupportedOperationException x) {
-                return; // Data provider is not capable of CompletionStage return type
-            }
+            insert = accounts.add(account);
+        } catch (UnsupportedOperationException x) {
+            return; // Data provider is not capable of CompletionStage return type
+        }
+        try {
 
             insert.toCompletableFuture()
                     .get(TIMEOUT_SECONDS, TimeUnit.SECONDS);
--- a/tck/src/main/java/ee/jakarta/tck/data/web/async/AsyncTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/web/async/AsyncTests.java
@@ -32,6 +32,8 @@ import ee.jakarta.tck.data.framework.junit.anno.AnyEntity;
 import ee.jakarta.tck.data.framework.junit.anno.Assertion;
 import ee.jakarta.tck.data.framework.junit.anno.Web;
 import ee.jakarta.tck.data.framework.utilities.TestPropertyUtility;
+
+import jakarta.data.exceptions.EntityExistsException;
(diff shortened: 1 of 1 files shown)


tck/src/main/java/ee/jakarta/tck/data/web/async/AsyncTests.java
```diff
@@ -187,4 +189,49 @@ public void testAsynchronousInsert() throws Exception {
...
+                LocalDateTime.of(2026, 7, 28, 10, 30, 00),
+                "asyncUser108@eclipse.org");
+
+        try {
+            CompletionStage<Void> insert;
+            try {
+                insert = accounts.add(account);
+            } catch (UnsupportedOperationException x) {
+                return; // Data provider is not capable of CompletionStage return type
+            }
```
njr-11: This is good - it correctly allows for Jakarta Data providers that are not capable of Asynchronous repository methods that return CompletionStage and catches the correct exception for it.  A very subtle issue is that the delete(account) operation in the finally block will fail if the insert didn't happen.  We can get around that by moving the beginning of the `try` block to after the attempted insert,
```suggestion
        CompletionStage<Void> insert;
        try {
            insert = accounts.add(account);
        } catch (UnsupportedOperationException x) {
            return; // Data provider is not capable of CompletionStage return type
        }
        try {
```

tck/src/main/java/ee/jakarta/tck/data/web/async/AsyncTests.java
```diff
@@ -187,4 +189,49 @@ public void testAsynchronousInsert() throws Exception {
...
+
+            insert.toCompletableFuture()
+                    .get(TIMEOUT_SECONDS, TimeUnit.SECONDS);
```
njr-11, line 219: To allow for eventual consistency databases (applies to some NoSQL databases), every time an update is made that subsequent test code will rely upon, the tests need to apply a delay to be sure the change has gone into effect.

```suggestion
            insert.toCompletableFuture()
                    .get(TIMEOUT_SECONDS, TimeUnit.SECONDS);

            TestPropertyUtility.waitForEventualConsistency();
```

njr-11 (commented):
Nice work on your first contribution to the TCK tests!  The test looks good and I ran it as well and also saw it passing.  I added two comments and suggested fixes for some more obscure aspects that need to be accounted for.

mhnuk2007:
@njr-11 
Thanks, Nathan. I’ve applied both suggestions and revalidated the updated test.

* `mvn clean verify` — successful
* `testAsynchronousInsertExceptionalCompletion` against Open Liberty — 1/1 passed
* `git diff --check` — clean
* Working tree is clean and synchronized with the PR branch.

The test is ready for further review.

njr-11 (approved):
Thanks!  Looks good.  I added an approval.  I'll leave this open for a couple more days in case anyone else wants to review and offer feedback, and then we'll merge it.

gavinking (approved):
No objections here, thanks.

Issue #965, opened 2025-02-06
TCK scenarios for Data 1.1


njr-11:
List of TCK scenarios to cover for Jakarta Data 1.1

- [ ] #964
- [ ] ... add more items to this list

mhnuk2007:
@njr-11 
I’ve opened PR #1556 to address the **Asynchronous repository methods** TCK scenario listed here.

The PR adds coverage for exceptional completion of an asynchronous repository method, verifying that an asynchronous `@Insert` failure completes the returned `CompletionStage` exceptionally with `EntityExistsException` as the underlying cause.

PR: #1556

njr-11:
> I’ve opened PR [#1556](https://github.com/jakartaee/data/pull/1556) to address the **Asynchronous repository methods** TCK scenario listed here.

Thanks!  I added a checkbox for #1556 to the list of TCK testing.  I'll plan to review your new tests and once we get it merged, then we can check off the list item for it.

mhnuk2007:
TCK implementation is complete for `@QueryOptions` on a `@JakartaQuery` repository method (`Inventory.averagePrice`, `StatefulPersistenceEntityTests#testJakartaQueryWithQueryOptions`).

Local verification is currently blocked by an independently reproduced Open Liberty runtime failure ([[OpenLiberty/open-liberty#35697](https://github.com/OpenLiberty/open-liberty/issues/35697)](https://github.com/OpenLiberty/open-liberty/issues/35697)). Persistence unit initialization fails with an NPE in `InMemoryUrlStreamHandler$PsURLConnection` when running existing, unmodified Jakarta Data TCK tests, before the test method executes.

njr-11:
> TCK implementation is complete for `@QueryOptions` on a `@JakartaQuery` repository method (`Inventory.averagePrice`, `StatefulPersistenceEntityTests#testJakartaQueryWithQueryOptions`).

Go ahead and create a PR to share your TCK test, and I'll link to it in the list of TCK tests.  One of the rest of us may be able to run it successfully for you.

njr-11:
Thanks for opening the PR. I added a link to it from the corresponding entry of the TCK tests list.

Now that I can see the source, I know one important reason it wouldn't be working in Open Liberty yet

> Local verification is currently blocked by an independently reproduced Open Liberty runtime failure

[...]

PR #1561, opened 2026-09-04
TCK tests for QueryOptions on JakartaQuery

mhnuk2007:
## Summary

Adds TCK coverage for `@QueryOptions` on a repository method annotated with `@JakartaQuery`.

### Changes

* Added a `@JakartaQuery` repository method using `@QueryOptions(flush = QueryFlushMode.FLUSH)`.
* Added a stateful persistence TCK test covering the `@JakartaQuery` + `@QueryOptions` combination.
* Uses a pending entity modification before executing the query to exercise the query option.

### Verification

* `mvn clean verify` — **BUILD SUCCESS**
* API tests: **321 passed**
* TCK framework tests: **8 passed**
* Checkstyle: **0 violations**
* `git diff --check` — **clean**

### Runtime testing

The new test was also attempted against the local Open Liberty Jakarta Data 1.1 TCK server.

The test currently cannot reach the test body because persistence-unit bootstrap fails in Open Liberty with an `InMemoryUrlStreamHandler$PsURLConnection` `NullPointerException`.

The same persistence bootstrap failure was reproduced with existing TCK tests, including:

* `StatefulPersistenceEntityTests#testModifyPersistedEntity`
* `JakartaQueryTests#testJPQLSelectAggregate`

Therefore, this runtime failure is independent of the changes in this PR.

Issue: #965

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/StatefulPersistenceEntityTests.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/StatefulPersistenceEntityTests.java
@@ -34,6 +34,9 @@ import ee.jakarta.tck.data.standalone.persistence.Product.Department;
 
 import jakarta.data.Order;
 import jakarta.inject.Inject;
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.FlushModeType;
+import jakarta.persistence.PersistenceContext;
 import jakarta.transaction.UserTransaction;
 
 /**
@@ -61,6 +64,9 @@ public class StatefulPersistenceEntityTests {
 
     @Inject
     UserTransaction tran;
+
+    @PersistenceContext
+    EntityManager entityManager;
 
     @Assertion(id = "471", strategy = """
             Use the Detach annotation to detach entities from the persistence context.
@@ -472,5 +478,36 @@ public class StatefulPersistenceEntityTests {
 
         assertEquals(false,
                      products.byNumber("TEST-PROD-1015").isPresent());
+    }
+
+    @Assertion(id = "965", strategy = """
+    Use a repository method annotated with JakartaQuery and QueryOptions
+    with QueryFlushMode.FLUSH. Set the persistence context flush mode to
+    EXPLICIT, modify a managed entity without explicitly flushing, and
+    verify that the query sees the modification.
+    """)
+    public void testJakartaQueryWithQueryOptions() throws Exception {
+        inventory.erase();
+
+        Product product = Product.of("tennis racket",
+                82.99,
+                "TEST-PROD-1017",
+                Department.SPORTING_GOODS);
+
+        tran.begin();
+        inventory.persist(product);
+        product.setPrice(84.98);
+
+        entityManager.setFlushMode(FlushModeType.EXPLICIT);
+
+        assertEquals(84.98,
+                inventory.averagePrice("TEST-PROD-1017"),
+                0.01);
+
+        entityManager.setFlushMode(FlushModeType.AUTO);
+
+        tran.commit();
+
+        inventory.remove(product);
     }
 }
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/Inventory.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/Inventory.java
@@ -27,6 +27,9 @@ import jakarta.data.repository.stateful.Merge;
 import jakarta.data.repository.stateful.Persist;
 import jakarta.data.repository.stateful.Remove;
 import jakarta.data.restrict.Restriction;
+import jakarta.persistence.QueryFlushMode;
+import jakarta.persistence.query.JakartaQuery;
+import jakarta.persistence.query.QueryOptions;
 import jakarta.transaction.Transactional;
 
 /**
@@ -60,4 +63,12 @@ public interface Inventory {
             """)
     List<Product> withDiscountedPriceUpTo(double max,
                                           double percentOff);
+
+    @JakartaQuery("""
+        SELECT AVG(p.price)
+          FROM Product p
+         WHERE p.productNum = ?1
+        """)
+    @QueryOptions(flush = QueryFlushMode.FLUSH)
+    double averagePrice(String productNumber);
 }


njr-11:
FYI - I posted a [comment in the TCK tests issue](https://github.com/jakartaee/data/issues/965#issuecomment-5547008129) pointing out that this would not be able to run in Open Liberty yet because Open Liberty only has beta capability for Data 1.1 and not currently for Persistence 4.0, which this test relies on.  Maybe another provider will be able to run it.

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/StatefulPersistenceEntityTests.java
```diff
@@ -473,4 +479,35 @@ public void testRemoveMultiple() throws Exception {
...
+        inventory.persist(product);
+        product.setPrice(84.98);
+
+        entityManager.setFlushMode(FlushModeType.EXPLICIT);
+
+        assertEquals(84.98,
+                inventory.averagePrice("TEST-PROD-1017"),
+                0.01);
+
+        entityManager.setFlushMode(FlushModeType.AUTO);
```
njr-11: There is no guarantee that updates made to entityManager will have any effect on the Jakarta Data repository operation. The Jakarta Data provider might be using a different EntityManager, or it might not be using an EntityManager at all and might be using an EntityAgent instead.

mhnuk2007: Thanks, you're right. The test currently assumes that the `@PersistenceContext`-injected `EntityManager` is the same persistence context used by the Jakarta Data repository operation, which Jakarta Data does not guarantee.

I’m going to revisit the test setup so that it does not rely on that assumption while still exercising the `QueryFlushMode.FLUSH` behavior required by the scenario.

mhnuk2007: @njr-11 
For this TCK scenario, what is the intended portable way to put the persistence context used by the stateful repository into `FlushModeType.EXPLICIT`?

The current `@PersistenceContext`-injected `EntityManager` isn't guaranteed to be the persistence context used by the repository operation, while the Jakarta Data spec leaves persistence-context propagation unspecified.

njr-11: As far as I am aware, the only portable way to specify a particular FlushModeType in Jakarta Data is via the QueryOptions, which applies only to the query performed by the annotated repository method.

mhnuk2007: Thanks, that makes sense.

Given that there is no portable way in Jakarta Data to set the repository's persistence context to `FlushModeType.EXPLICIT`, should assertion 965's test strategy itself be revised? Its current wording requires setting the persistence context to `EXPLICIT` before verifying that `QueryOptions(flush = QueryFlushMode.FLUSH)` makes the pending modification visible.

I don't want to remove that setup and leave a test that passes due to the normal `AUTO` behavior, since that would no longer specifically exercise the `FLUSH` option.

One option I considered is adding a second repository method identical to `averagePrice` but without `QueryOptions` (or with the default flush behavior), and comparing the results immediately after the pending change. However, that still doesn't reliably distinguish `FLUSH` from `AUTO`, since `AUTO` may already guarantee visibility.

Is there a portable pattern you'd recommend for constructing a "would fail without FLUSH" precondition, or should assertion 965 be scoped down to verify the `FLUSH` behavior without claiming to prove that it is load-bearing versus `AUTO`?

I'll hold off on further code changes until the intended assertion strategy is clear.

njr-11: > iven that there is no portable way in Jakarta Data to set the repository's persistence context to `FlushModeType.EXPLICIT`, should assertion 965's test strategy itself be revised? Its current wording requires setting the persistence context to `EXPLICIT` before verifying that `QueryOptions(flush = QueryFlushMode.FLUSH)` makes the pending modification visible.

Given that there is no portable way to set the persistence context to `EXPLICIT`, maybe change the test strategy to something like: "Within an active transaction, persist an entity and modify it without explicitly flushing, then verify that a repository method annotated `@QueryOptions(flush = QueryFlushMode.FLUSH)` returns a result that reflects the pending modification" ?

mhnuk2007: Yes, that makes sense to me. This removes the non-portable assumption about setting the persistence context to `EXPLICIT` while still exercising the `FLUSH` option on the `@JakartaQuery` method.

I'll update the assertion strategy along those lines and remove the `EntityManager` flush-mode setup.

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/StatefulPersistenceEntityTests.java
```diff
@@ -474,4 +474,31 @@ public void testRemoveMultiple() throws Exception {
...
+    public void testJakartaQueryWithQueryOptions() throws Exception {
+        inventory.erase();
+
+        Product product = Product.of("tennis racket",
+                82.99,
+                "TEST-PROD-1017",
+                Department.SPORTING_GOODS);
+
+        tran.begin();
+        inventory.persist(product);
```
njr-11:
```suggestion
        inventory.persist(product);
        tran.begin();
```

njr-11 (approved):
This looks good now, and passes on a development copy of Open Liberty that isn't beta yet.  I made one suggestion: to move persisting the entity to happen before the transaction that updates the entity and tests for flush of the update.

otaviojava (approved):

mhnuk2007:
For tracking: the remaining `@QueryOptions` TCK coverage for `@Find`, `@Query`, `@Delete`, and `@NativeQuery` is being added in **PR #1567**: https://github.com/jakartaee/data/pull/1567

PR #1563, opened 2026-09-09
fix: TCK test deployment issues

KyleAure:
- make sure deployments have the right set of classes to be able to run
- use `@Assertion` instead of `@DisplayName/@Test` we scan for `@Assertion` to automate sections of the TCK Distribution documentation.
- other cleanup of TCK tests.

### DIFF
--- a/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaEventBuiltInRepositoryTest.java
+++ b/tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaEventBuiltInRepositoryTest.java
@@ -15,23 +15,20 @@
  */
 package ee.jakarta.tck.data.standalone.entity;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.tuple;
+
 import ee.jakarta.tck.data.framework.junit.anno.AnyEntity;
+import ee.jakarta.tck.data.framework.junit.anno.Assertion;
 import ee.jakarta.tck.data.framework.junit.anno.Standalone;
-import ee.jakarta.tck.data.framework.utilities.DatabaseType;
-import ee.jakarta.tck.data.framework.utilities.TestProperty;
 import ee.jakarta.tck.data.framework.utilities.TestPropertyUtility;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.assertj.core.groups.Tuple;
 import org.junit.jupiter.api.DisplayName;
-import org.junit.jupiter.api.Nested;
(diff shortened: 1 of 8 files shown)


tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaEventBuiltInRepositoryTest.java
```diff
@@ -15,172 +15,147 @@
...
     @BeforeEach
-    void setUp() {
+    public void setUp() {
         this.repository.deleteAll();
         this.observer.reset();
         TestPropertyUtility.waitForEventualConsistency();
     }
 
-    @Nested
-    @DisplayName("When inserting an entity")
```
otaviojava, line 66: @KyleAure, any reason for removing the nested class? 
Any framework limitations?

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaEventCustomRepositoryTest.java
```diff
@@ -15,45 +15,49 @@
...
+                    MusicRecordRepository.class,
+                    MusicRecord.class,
+                    MusicStore.class,
+                    ObservedEvent.class
+                );
     }
 
     @Inject
-    private MusicRecordLifecycleObserver observer;
+    MusicRecordLifecycleObserver observer;
```
otaviojava, line 57: The same question here: why remove the private attribute?

KyleAure: No they could be private, but I'm just trying to keep consistency between the test classes.

tck/src/main/java/ee/jakarta/tck/data/standalone/entity/JakartaEventBuiltInRepositoryTest.java
KyleAure: Yes, Arquillian does not natively support the `@Nested` Junit annotation. To make this work you would need to annotate the nested class with the following: 

```java
@Standalone
@AnyEntity
@Nested
@DisplayName("When inserting an entity")
class MyNestedTestClass {
```

I did not test it but I think you might need to also inject the resources for each nested test class.

njr-11 (approved):
Looks good to me. I'll add my approval and Otavio can continue the discussion with Kyle about the questions he asked.

otaviojava:
@KyleAure @njr-11 I will merge, so you can test on Archilian, but we need to pay more attention to this quality in our TCK.

PR #1566, opened 2026-09-11
Adjust tests to use matching types and add missing ordering

marko-bekhta:
From Hibernate ORM builds against the recent TCK snapshots:

- Since `AsciiCharacter` uses a long id having those its for the filters does not play very nicely 😕 
- missing sorting/oredering results in:
```
error: method with return type 'jakarta.data.page.CursoredPage' has no parameter of type 'Order' or 'Sort' and no '@OrderBy' annotation for inherited member 'withTruncatedSquareRoot'
interface NaturalNumbers$ extends NaturalNumbers {
```

cc: @gavinking @njr-11


njr-11 (approved):
Thanks for spotting these mistakes.

Tests that use withTruncatedSquareRoot are consistent with ordering by ID as well as requiring a single sort (because they have `type.capableOfSingleSort()`).  Most likely the way this was intended to be done was with an `@OrderBy` on the method that was accidentally omitted, which matches the fix you made.

Also I confirmed that the type of `AsciiCharacter` is `long`, not `int`, which confirms your other update as well.

PR #1567, opened 2026-09-12
Add QueryOptions TCK tests for Find, Query, Delete, and NativeQuery

mhnuk2007:
Adds TCK coverage for three QueryOptions scenarios from #965:

* `@QueryOptions(flush = QueryFlushMode.FLUSH)` on a `@Find` repository method
* `@QueryOptions(flush = QueryFlushMode.FLUSH)` on a `@Query` repository method
* `@QueryOptions(flush = QueryFlushMode.FLUSH)` on a `@NativeQuery` repository method

The tests verify that a pending entity modification is visible to the repository operation when invoked within an active transaction.

The `@Delete` scenario was initially included but has been removed from this PR following review. The Delete case requires further investigation because the stateless repository method and stateful repository used by the test involve separate persistence contexts.

Local compilation succeeds with:

`mvn -pl tck -am -DskipTests compile`

Runtime verification with Open Liberty remains limited by the existing Jakarta Persistence 4.0 / `@JakartaQuery` implementation limitation discussed in the related work.


tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/StatefulPersistenceEntityTests.java
```diff
@@ -263,6 +263,121 @@ public void testModifyPersistedEntity() throws Exception {
         inventory.remove(found.get(0));
     }
 
+    @Assertion(id = "965", strategy = """
```
njr-11:
```suggestion
    @Assertion(id = "1312", strategy = """
```

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/StatefulPersistenceEntityTests.java
```diff
@@ -263,6 +263,121 @@ public void testModifyPersistedEntity() throws Exception {
...
+        assertEquals(84.98,
+                found.getPrice(),
+                0.01);
+
+        tran.commit();
+
+        inventory.remove(found);
+    }
+
+    @Assertion(id = "965", strategy = """
```
njr-11:
```suggestion
    @Assertion(id = "1312", strategy = """
```

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/StatefulPersistenceEntityTests.java
```diff
@@ -263,6 +263,121 @@ public void testModifyPersistedEntity() throws Exception {
...
+        tran.begin();
+        product.setPrice(84.98);
+
+        assertEquals(1L,
+                inventory.deleteByPrice(84.98));
+
+        tran.commit();
+    }
+
+    @Assertion(id = "965", strategy = """
```
njr-11:
```suggestion
    @Assertion(id = "1312", strategy = """
```

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/StatefulPersistenceEntityTests.java
```diff
@@ -263,6 +263,121 @@ public void testModifyPersistedEntity() throws Exception {
...
+        assertEquals(84.98,
+                inventory.priceOf("TEST-PROD-1018"),
+                0.01);
+
+        tran.commit();
+
+        inventory.remove(product);
+    }
+
+    @Assertion(id = "965", strategy = """
```
njr-11:
```suggestion
    @Assertion(id = "1312", strategy = """
```

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/Inventory.java
```diff
@@ -64,11 +68,31 @@ Stream<Product> filter(Restriction<Product> restriction,
...
         SELECT AVG(p.price)
           FROM Product p
          WHERE p.productNum = ?1
         """)
     @QueryOptions(flush = QueryFlushMode.FLUSH)
     double averagePrice(String productNumber);
+
+    @Delete
+    @QueryOptions(flush = QueryFlushMode.FLUSH)
+    long deleteByPrice(@By("price") double price);
```
njr-11: Violates a specification/API requirement that stateless repository methods (such as `@Delete` methods) cannot be intermixed with stateful repository methods (such as `@Merge`, `@Persist`, and `@Remove` methods) on the same repository. This method will need to be placed on a stateless repository instead.
```suggestion
```

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/StatefulPersistenceEntityTests.java
```diff
@@ -263,6 +263,121 @@ public void testModifyPersistedEntity() throws Exception {
...
+
+        tran.begin();
+        product.setPrice(84.98);
+
+        assertEquals(1L,
+                inventory.deleteByPrice(84.98));
+
+        tran.commit();
+    }
+
```
njr-11: Due to the deleteByPrice method not being valid on the stateful repository

```suggestion
```

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/StatefulPersistenceEntityTests.java
```diff
@@ -263,6 +263,121 @@ public void testModifyPersistedEntity() throws Exception {
...
+        Product product = Product.of("tennis racket",
+                82.99,
+                "TEST-PROD-1019",
+                Department.SPORTING_GOODS);
+
+        inventory.persist(product);
+
+        tran.begin();
+        product.setPrice(84.98);
```
njr-11, line 284: ensure that the entity is managed within the transaction rather than detached
```suggestion
        product = inventory.merge(product);
        product.setPrice(84.98);
```

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/StatefulPersistenceEntityTests.java
```diff
@@ -263,6 +263,121 @@ public void testModifyPersistedEntity() throws Exception {
...
+
+        Product product = Product.of("tennis racket",
+                82.99,
+                "TEST-PROD-1021",
+                Department.SPORTING_GOODS);
+
+        inventory.persist(product);
+
+        tran.begin();
+        product.setPrice(84.98);
```
njr-11, line 348: ensure that the entity is managed within the transaction rather than detached
```suggestion
        product = inventory.merge(product);
        product.setPrice(84.98);
```

tck/src/main/java/ee/jakarta/tck/data/standalone/persistence/stateful/StatefulPersistenceEntityTests.java
```diff
@@ -263,6 +263,121 @@ public void testModifyPersistedEntity() throws Exception {
...
+
+        Product product = Product.of("tennis racket",
+                82.99,
+                "TEST-PROD-1018",
+                Department.SPORTING_GOODS);
+
+        inventory.persist(product);
+
+        tran.begin();
+        product.setPrice(84.98);
```
njr-11, line 319: ensure that the entity is managed within the transaction rather than detached
```suggestion
        product = inventory.merge(product);
        product.setPrice(84.98);
```

njr-11 (commented):
Thanks for adding these tests. Most of them look good, although the one for Delete relied on mixing stateful and stateless methods on the same repository, which is not allowed in Jakarta Data.  When I ran them, this was detected and failed, but after removing it (see suggestions) and making a few updates to Open Liberty locally to a development copy of Open Liberty to support the tested capability, the tests pass.

It should be possible to reintroduce the removed test for Delete/QueryOptions after moving the Delete method to a stateless repository.

mhnuk2007:
Thanks @njr-11 for the review. I’ve addressed the feedback and pushed the updates in `0f5cee5b`.

* Reintroduced the `@Delete` / `@QueryOptions(flush = FLUSH)` test in this PR.
* Moved `deleteByPrice(...)` from the stateful `Inventory` repository to the stateless `Catalog` repository, avoiding the stateful/stateless repository-method mixing issue.
* Updated `testDeleteWithQueryOptions()` to call `catalog.deleteByPrice(...)`.
* Applied the suggested `merge()` pattern before modifying the entity so that the entity is managed within the active transaction, consistently across the Find, Delete, Query, and NativeQuery tests.
* Corrected the duplicate assertion ID: `testDeleteWithQueryOptions()` now uses `1312`, matching the other QueryOptions tests.

The changes compile cleanly, and the working-tree checks are clean. Runtime verification is still limited by the currently available Jakarta Persistence/Open Liberty implementation, as discussed earlier.

Please take another look when convenient, and let me know if you would prefer the Delete test to be split into a separate PR.

njr-11 (commented):
Thanks for adding the change and updating the `Delete` scenario.
A couple of things to note.  It looks like 3 of the suggestions adding `inventory.merge(product)` still need to be applied.
Regarding the `@Delete` scenario -- I just tried it out and it failed the assertion because the deletion count was 0 rather than 1.  There are two reasons for this that I can think of.  One reason is that floating point numbers aren't exact, so an equals comparison on them won't always work and so you might need to do something like `catalog.deleteByPrice(Between.bounds(84.97, 84.99))` instead.  The second reason is that with the `@Delete` method now on a stateless repository, and the testDeleteWithQueryOptions method now involving 2 different repositories, we end up with different persistence contexts. After the transaction begins, the `product` is merged to the persistence context of the stateful repository, where it is subsequently updated.  After that, the `deleteByPrice` method with `FLUSH` runs, but that is on a stateless repository independent of the persistence context associated with the stateful repository.  So the flush does not apply to the former, and the database still has the original price, resulting in no deletion.
Due to this finding, I would recommend moving the Delete scenario to a separate PR.  It needs more thought as to whether there is a way to write a valid test for it that actually covers the behavior.  We might also want to think about whether we should relax the prohibition on `@Delete` on a stateful repository to possibly allow non-lifecycle method use.

mhnuk2007:
Thanks @njr-11 for testing this and for explaining the persistence-context issue. That makes sense and also explains why the deletion count remained 0 even after applying the `merge()` pattern.

I’ll move the Delete scenario to a separate PR for further investigation and keep #1567 focused on the Find, Query, and NativeQuery cases. I’ll also verify that all three remaining `inventory.merge(product)` suggestions are applied.

njr-11 (approved):
I added the 3 remaining review comments to ensure the entity is managed within the transaction, ran the tests to confirm they pass, and am approving this PR so we can merge the new tests. Thanks for the contribution!

PR #1574, opened 2026-09-16
Align TemporalLiteral toString with Jakarta Query grammar

### LINKED ISSUE #1102: Align TemporalLiteral toString with the outcome of Jakarta Persistence 453
> This will probably need to be revisited in light of https://github.com/jakartaee/persistence/issues/453.  (Currently an issue in the Persistence project, but really belongs in Query.)

_Originally posted by @gavinking in https://github.com/jakartaee/data/pull/1101#discussion_r2083098285_

njr-11:
Resolves #1102 by aligning toString output with the Jakarta Query grammar where defined.  Year on its own is not defined in the grammar so I chose something for toString that looked similar.  Also, the grammar mentions nothing about padding with leading zeros, but does not prohibit it either, so I reused the toString implementations of LocalDate and LocalTime which include padding.


api/src/main/java/jakarta/data/spi/expression/literal/TemporalLiteralRecord.java
```diff
@@ -45,18 +45,21 @@ public String toString() {
...
+            return "DATETIME " + dateString + ' ' + d.toLocalTime().toString();
+        } else if (temporal instanceof LocalDate d) {
+            return "DATE " + (d.getYear() >= 10000
+                                ? d.toString().substring(1) // omit leading +
+                                : d.toString());
         } else if (temporal instanceof LocalTime) {
-            return "{t '" + value + "'}";
+            return "TIME " + temporal.toString();
+        } else if (temporal instanceof Year y) {
+            return "YEAR " + y.getValue();
```
gavinking, line 71: Interesting. Perhaps Jakarta Query should allow this syntax.

api/src/main/java/jakarta/data/spi/expression/literal/TemporalLiteralRecord.java
```diff
@@ -45,18 +45,21 @@ public String toString() {
...
-        if (temporal instanceof Year y) {
-            return "{d '" + y.getValue() + "'}";
-        } else if (temporal instanceof LocalDate) {
-            return "{d '" + value + "'}";
-        } else if (temporal instanceof LocalDateTime d) {
-            return "{ts '" + d.toLocalDate() + ' ' + d.toLocalTime() + "'}";
+        if (temporal instanceof LocalDateTime d) {
+            String dateString = d.getYear() >= 10000
+                    ? d.toLocalDate().toString().substring(1) // omit leading +
+                    : d.toLocalDate().toString();
```
gavinking: Since this is repeated in three places, better to extract a little function.

njr-11: Good idea to have it only in one place. I switched it under https://github.com/jakartaee/data/pull/1574/changes/3ef3d3add296dfe272818c52d7f94071648e3424

gavinking (approved):
LGTM

Issue #1185, opened 2025-07-22
[Use Case]: Fetching


gavinking:
### As a ...

- [ ] Application user/user of the configuration itself
- [x] API user (application developer)
- [ ] SPI user (container or runtime developer)
- [ ] Specification implementer

### I need to be able to ...

... specify that certain fields of an entity should be fetched by a repository method.

### Which enables me to ...

... control the number of queries executed against the datastore and the amount of data which is fetched.

### Additional information

By analogy with `@OrderBy`, I propose a `@Fetch` or `@Fetching` annotation:

```java
@Find 
@Fetching(_Book.AUTHORS)
@OrderBy(_Book.ISBN)
Book bookWithAuthors();
```

For implementations based on Jakarta Persistence, this might be backed by construction of an `EntityGraph`.

gavinking:
Relevant: https://github.com/jakartaee/persistence/issues/735

gavinking:
> I understand the desire to have this, but it seems like Jakarta Data does not specify lazy loading, so it feels a bit awkward to have this annotation, don't you think?

It's quite easy to provide well-defined semantics for this operation without needing to talk explicitly about lazy fetching.

> Let the user define interfaces that represent the subset of data they want to expose and fetch eagerly at the same time. When the entity type implements that interface, we can derive load graphs automatically from the methods of the interface.

I'm not quite sure I have understood what you're envisioning here, but perhaps something like the following.

I think you're suggesting the user would write something like:

```java
@FetchProfile
interface BookWithAuthors {
    String getIsbn();
    String getTitle();
    Set<Author> getAuthors();
}
```

(Not sure if the annotation is really necessary, so ignore if for now.)

And then the entity class would actually implement this interface, as a witness that the entity really has the required fields:

```java
@Entity
class Book implements BookWithAuthors {
    @Id String isbn;
    String title;
    @OneToMany Set<Author> authors;

    // getters, setters
    ...
}
```

And then the repository would indicate this fetch profile in its return type:

```java
@Find
BookWithAuthors bookWithAuthors(String isbn);
``` 

This is a nice idea and frankly quite elegant. Definitely appealing in some ways.

I believe it can even in principle be extended to handle "deeper" fetch graphs by writing stuff like:

```java
Set<? extends AuthorWithThing> getAuthors();
```

But it does have some distinct downsides:

1. There's a need to declare not only associations, but even basic attributes of `Book` on the interface. 
[...]

PR #1576, opened 2026-09-18
add @Fetching annotation

gavinking:
Rough solution to #1185.

Note that I picked `@Fetching` instead of `@Fetch` because JPA already has an annotation and an interface named `Fetch`.


api/src/main/java/jakarta/data/repository/Fetching.java
```diff
@@ -0,0 +1,62 @@
...
+public @interface Fetching {
+    /**
+     * An attribute of the entity returned by the parameter-based
+     * automatic query method which will be accessed by the
+     * application program.
+     */
+    String value();
+
+    /**
+     * Enables multiple {@code Fetch} annotations on the method.
```
njr-11:
```suggestion
     * Enables multiple {@code Fetching} annotations on the method.
```

api/src/main/java/jakarta/data/repository/Fetching.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (c) 2022,2026 Contributors to the Eclipse Foundation
```
njr-11:
```suggestion
 * Copyright (c) 2026 Contributors to the Eclipse Foundation
```

api/src/main/java/jakarta/data/repository/Fetching.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.lang.annotation.*;
+
+/**
+ * A hint to the Jakarta Data implementation specifying that
+ * an attribute of the entity returned by a parameter-based
+ * automatic query method will be accessed by the application
+ * program. The repository implementation should ensure that
+ * access to the value of the specified attribute is efficient.
+ * <pre>{@code
```
otaviojava, line 28: maybe include a note about:

```
An implementation may use this information to influence how it obtains the specified attribute when that behavior is meaningful for the underlying data model.
```

api/src/main/java/jakarta/data/repository/Fetching.java
```diff
@@ -0,0 +1,62 @@
...
+package jakarta.data.repository;
+
+import java.lang.annotation.*;
+
+/**
+ * A hint to the Jakarta Data implementation specifying that
+ * an attribute of the entity returned by a parameter-based
+ * automatic query method will be accessed by the application
+ * program. The repository implementation should ensure that
+ * access to the value of the specified attribute is efficient.
```
njr-11, line 27: There are a couple of aspects of this that applications will care about. One is efficient access, which is covered above, although does have the risk of informed users thinking it will beneficial to declare every entity attribute they use as `@Fetching`. Another use case is an application using a stateful repository needing an entity attribute to be accessible after the entity is detached from the persistence context. The description here does not offer the latter.

api/src/main/java/jakarta/data/repository/Fetching.java
```diff
@@ -0,0 +1,62 @@
...
+ * Book bookWithAuthorsAndPublisher(String isbn);
+ * }</pre>
+ *
+ * @see Find
+ * @since 1.1
+ */
+@Repeatable(Fetching.List.class)
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface Fetching {
```
otaviojava, line 41: 🤔 
Maybe use the 3.2. from Jakarta Query as a reference here?

https://jakarta.ee/specifications/query/1.0/jakarta-query-1.0-m2#joins

I mean, as a Java developer to know more

api/src/main/java/jakarta/data/repository/Fetching.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * An attribute of the entity returned by the parameter-based
+     * automatic query method which will be accessed by the
+     * application program.
```
njr-11, line 45: Grammar nitpick here - this means the "method ... will be accessed by the application program", which is true, but the intention is that the attribute will be accessed.
```suggestion
     * Name of an entity attribute that will be accessed by the application program.
     * The entity defining the attribute must be returned by the {@link Find} method
     * that is annotated {@code @Fetching}.
```

njr-11 (commented):
Added a few minor suggestions and something to think about with respect to the how we present this annotation to users.

api/src/main/java/jakarta/data/repository/Fetching.java
gavinking: I'm OK with adding that.

api/src/main/java/jakarta/data/repository/Fetching.java
gavinking: Not sure I understand?

